PR #1, opened 2019-06-11
EVG-6226 Create patch and variant components and add login modal

dominoweir:
The main things I have done here are: created a component for a patch, created a component for a build variant, created a login modal with some handlers, and played around with colors and styling on the main app page.


src/components/login/login.tsx
```diff
@@ -0,0 +1,111 @@
...
+    this.setState({
+      open : false,
+      submitted : true,
+    });
+  }
+
+  public render() {
+    return (
+      <div>
+        <Button onClick={this.handleClickOpen}>
```
john-m-liu: I think this button belongs in the menu in App.tsx. That way the UI will be consistent and also the contents of that are in a single place (we may want to split that off into its own component later too)

src/components/login/login.tsx
```diff
@@ -0,0 +1,111 @@
+import { Button, Dialog, DialogActions, DialogContent, DialogContentText, DialogTitle, TextField } from '@material-ui/core';
+import * as React from 'react';
+import * as rest from "../../rest/interface";
+import '../../styles.css';
+
+interface State {
+  APIClient: rest.Evergreen;
```
john-m-liu: This should be a part of Props because it's data that should be set by the main app and passed to all its child components

src/components/login/login.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+export class Login extends React.Component<Props, State> {
+
+  constructor(props: Props) {
+    super(props);
+    this.handleClickOpen = this.handleClickOpen.bind(this);
+    this.handleClickClose = this.handleClickClose.bind(this);
+    this.handleClickSubmit = this.handleClickSubmit.bind(this);
+    this.state = {
+      APIClient: rest.EvergreenClient("admin", "e4f2c40463dcade5248d36434cb93bac", "http://localhost:8080/api"),
```
john-m-liu: Similarly, the REST client should be passed in by the parent component

src/components/login/login.tsx
```diff
@@ -0,0 +1,111 @@
...
+      open : false,
+      submitted : false,
+      username : "",
+      password : "",
+    };
+  }
+
+  public handleClickOpen() : any{
```
john-m-liu: These functions here don't return anything, so they should be of type void (delete the :any return signature). Also in general functions shouldn't return "any" because we know the type they return

src/components/login/login.tsx
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  public handleClickClose(): any {
+    this.setState({
+      open : false,
+    });
+  }
+
+  public handleClickSubmit(): any {
+    console.log(this.state.username);
```
john-m-liu: I would take out this debug code unless you're still using it for investigating

src/components/login/login.tsx
```diff
@@ -0,0 +1,111 @@
...
+    });
+  }
+
+  public render() {
+    return (
+      <div>
+        <Button onClick={this.handleClickOpen}>
+          Log In
+        </Button>
+        <Dialog open={this.state.open} onClose={this.handleClickClose} aria-labelledby="form-dialog-title">
```
john-m-liu: This isn't a big deal, but aria- properties aren't necessary for Evergreen code. They are used to annotate elements for accessibility software such as screen readers, but accessibility is not one of our goals

src/components/patch/patch.tsx
```diff
@@ -0,0 +1,62 @@
...
+import * as rest from "../../rest/interface";
+import '../../styles.css';
+import Variant from '../variant/variant';
+
+interface State {
+  APIClient: rest.Evergreen;
+}
+
+class Props {
+  public name : string;
```
john-m-liu: Since these fields are specific to this component and not passed in by the parent component, they should be state, not props

src/components/patch/patch.tsx
```diff
@@ -0,0 +1,62 @@
+import { ExpansionPanel, ExpansionPanelDetails, ExpansionPanelSummary, Grid, Typography } from '@material-ui/core';
+import ExpandMoreIcon from '@material-ui/icons/ExpandMore';
+import * as React from 'react';
+import * as rest from "../../rest/interface";
+import '../../styles.css';
+import Variant from '../variant/variant';
+
+interface State {
+  APIClient: rest.Evergreen;
```
john-m-liu: This should be passed in by the parent, so it should be props

package.json
```diff
@@ -28,7 +29,7 @@
     "eslint-plugin-import": "2.14.0",
     "eslint-plugin-jsx-a11y": "6.1.1",
     "eslint-plugin-react": "7.11.1",
-    "evergreen.js": "^1.0.6",
+    "evergreen.js": "^1.0.8",
```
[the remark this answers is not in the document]
dominoweir: Fixed

src/components/login/login.tsx
dominoweir: Fixed- switched return type for all three handlers to void.

src/components/login/login.tsx
dominoweir: Totally forgot about this, have now removed it.

src/components/login/login.tsx
dominoweir: Good to know. The Dialog example I found in the material-ui documentation included these properties, and I wasn't totally sure what they did so I just left them alone. I have now removed them.

src/components/login/login.tsx
dominoweir: Fixed

src/components/login/login.tsx
dominoweir: Fixed

src/components/app/App.tsx
```diff
@@ -19,54 +22,70 @@ export class Evergreen extends React.Component<Props, State> {
...
+                </MenuItem>
               </Menu>
             </Toolbar>
           </AppBar>
           <div className="App-intro">
             <Route path="/admin" render={admin} />
           </div>
         </HashRouter>
+        <div className="searchBar">Insert search bar here</div>
+        <Grid className="patchContainer" container={true} spacing={24}>
```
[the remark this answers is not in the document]
dominoweir: Fixed by adding a new PatchContainer component

src/components/login/login.tsx
dominoweir: Moved to the app bar, still working on getting the positioning correct though.

src/components/patch/patch.tsx
dominoweir: Fixed

src/components/login/login.tsx
john-m-liu: Just to clarify, I meant the dropdown menu with this comment, not the toolbar. Though if you think it makes more sense there, that works too

src/components/login/Login.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+class Props {
+  public client: rest.Evergreen;
+}
+
+export class Login extends React.Component<Props, State> {
+
+  constructor(props: Props) {
+    super(props);
+    this.handleClickOpen = this.handleClickOpen.bind(this);
```
john-m-liu: This works, but it would be cleaner to avoid having to tell the function what "this" refers to by using an arrow function, which will pass along the scoping information: https://stackoverflow.com/a/16158550

src/components/login/Login.tsx
```diff
@@ -0,0 +1,109 @@
...
+    this.handleClickSubmit = this.handleClickSubmit.bind(this);
+    this.state = {
+      open : false,
+      submitted : false,
+      username : "",
+      password : "",
+    };
+  }
+
+  public handleClickOpen() : void {
```
john-m-liu: It doesn't make sense for any other component to call these functions, so they should be private

dominoweir: Fixed

src/components/login/Login.tsx
dominoweir: Fixed

src/components/app/App.less
```diff
@@ -23,9 +22,21 @@
 .appBar {
     width: 100%;
     padding-left: 140px;
+    background-color: #3E4347 !important;
+}
+
+.title {
+  color: white !important;
```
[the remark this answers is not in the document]
dominoweir: Fixed

src/components/app/App.tsx
```diff
@@ -4,11 +4,14 @@ import * as React from 'react';
...
 import '../../styles.css';
-import {Admin} from "../admin/admin";
+import { Admin } from "../admin/Admin";
+import { Login } from "../login/Login";
+import { PatchContainer } from '../patch/PatchContainer';
 
 interface State {
   APIClient: rest.Evergreen;
   MenuAnchor?: HTMLElement;
+  Patches: JSON;
```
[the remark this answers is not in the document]
dominoweir: You're totally right, I have now removed this variable from the state.

src/components/login/Login.tsx
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  private handleClickSubmit = () => {
+    this.setState({
+      open : false,
+      submitted : true,
+    });
+  }
+ 
+  private updateSingleField = (fieldName:string) => {
```
john-m-liu: I didn't know about this before, but we can enforce that the field names passed into here have to be valid properties on the state object by using the "keyof" keyword in the function signature here. Could you try changing this function to only allow valid property names?

src/components/app/App.tsx
```diff
@@ -19,54 +22,61 @@ export class Evergreen extends React.Component<Props, State> {
...
               </Menu>
+              <Login client={this.state.APIClient} />
             </Toolbar>
           </AppBar>
           <div className="App-intro">
             <Route path="/admin" render={admin} />
           </div>
         </HashRouter>
+        <div className="searchBar">Insert search bar here</div>
+        <PatchContainer client={this.state.APIClient} />
```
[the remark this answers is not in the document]
dominoweir: Fixed

src/components/app/App.tsx
```diff
@@ -19,54 +21,56 @@ export class Evergreen extends React.Component<Props, State> {
...
-                      <NavLink to="/admin"> Admin page</NavLink>
-                    </MenuItem>
+              <Menu id="mainAppMenu" open={menuOpen} anchorEl={this.state.MenuAnchor} 
+                anchorOrigin={{vertical: 'top', horizontal: 'right',}}
+                transformOrigin={{vertical: 'top', horizontal: 'right',}}
+                onClose={this.closeMenu}>
+                <MenuItem onClick={this.closeMenu}>
+                  <NavLink to="/admin"> Admin page</NavLink>
+                </MenuItem>
+                <MenuItem>
```
john-m-liu: add an onClick here to close the menu

dominoweir: fixed!

PR #3, opened 2019-06-21
EVG-6306: Add config file dropzone

dominoweir:
This set of changes includes: creating a new drop zone component to upload and read in a local config file, adding that component to App.tsx, and updating functions to match the most recent version of evergreen.js.


src/components/app/App.tsx
```diff
@@ -60,6 +61,7 @@ export class Evergreen extends React.Component<Props, State> {
             <Route path="/admin" render={admin} />
             <Route path="/patches" render={patches} />
           </div>
+          <ConfigDrop updateClientConfig={this.updateConfig}/>
```
john-m-liu: Embedding this here will make it show up on all pages won't it? I think we want a specific place the user goes to set the config, or possibly hiding this once we have a config

src/components/admin/ConfigDrop.tsx
```diff
@@ -0,0 +1,57 @@
...
+class Props {
+  public updateClientConfig: (configObj: ClientConfig) => void;
+}
+
+interface State {
+  processing: boolean,
+  error: string
+}
+
+export class ConfigDrop extends React.Component<Props, State> {
```
john-m-liu: Could you add a couple tests for this? It would be neat to have an integration test where we simulate dropping a sample config (that we've committed to this repo) and parsing that config file to make sure everything is set correctly

src/components/app/App.tsx
dominoweir: That is correct. When we talked about this initially, we talked about putting it on the admin page (which is why it's currently in the admin folder), but we ran in to that issue with the routing to that page so we decided to just throw it on the main app page instead. Where do you think it should go now, since I'm still having that problem locally?

john-m-liu: One idea is to make it a modal that you can get to from the menu (but the menu option only shows if we have no config). The main thing is to not show it on every page

src/components/admin/ConfigDrop.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+
+  private upload = (file: File) => {
+    const reader = new FileReader();
+    reader.onload = () => {
+      const raw = reader.result.toString();
+      const configObj = ConvertToClientConfig(raw);
+      this.props.updateClientConfig(configObj);
+    };
+    reader.readAsText(file);
```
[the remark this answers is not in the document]
john-m-liu: In general it means displaying what went wrong to the user, in a clear way if possible (ie. not directly printing the error without any translation). In this case since it's an admin interface and we aren't sure of the potential errors, displaying it verbatim is fine

src/components/app/App.tsx
dominoweir: Just discovered that the admin page does load for me locally now (not sure what changed), so I moved the config drop zone to that page.

src/components/admin/ConfigDrop.tsx
```diff
@@ -0,0 +1,57 @@
...
+      this.upload(fileArray[0]);
+    }
+  }
+
+
+  private upload = (file: File) => {
+    const reader = new FileReader();
+    reader.onload = () => {
+      const raw = reader.result.toString();
+      const configObj = ConvertToClientConfig(raw);
```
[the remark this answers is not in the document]
dominoweir: That makes sense. I added validation that checks all required properties are present in the new config object before doing anything with it.

src/components/admin/ConfigDrop.tsx
dominoweir: I added 3 new tests: a snapshot, simulating a drag event with an empty json file object, and simulating a drag event with a valid json file object.

src/components/admin/ConfigDrop.tsx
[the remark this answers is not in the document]
dominoweir: That makes sense. I added error notifications using the material-ui snackbar component.

src/components/app/App.tsx
```diff
@@ -19,15 +21,15 @@ class Props {
 export class Evergreen extends React.Component<Props, State> {
   constructor(props: Props) {
     super(props);
-    // TODO: get the API client from some config
     this.state = {
-      APIClient: rest.EvergreenClient("admin", "e4f2c40463dcade5248d36434cb93bac", "http://localhost:8080/api"),
+      APIClient: rest.EvergreenClient("", "", "", ""),
```
john-m-liu: We should initialize this to be null rather than a blank client

src/components/configdrop/ConfigDrop.test.tsx
```diff
@@ -0,0 +1,53 @@
...
+    const wrapper = enzyme.mount(<ConfigDrop updateClientConfig={verifyConfig}/>);
+
+    it("matches snapshot", () => { 
+      // note: using enzyme.mount here (instead of enzyme.shallow) causes the test suite to hang, so this test needs its own shallow wrapper
+      const shallowWrapper = enzyme.shallow(<ConfigDrop updateClientConfig={verifyConfig}/>);
+      expect(shallowWrapper).toMatchSnapshot();
+    })
+
+    it("check config state object is not updated with an empty file", () => {
+      // check that the config object in state is null initially
```
john-m-liu: There are differing opinions on this, but I personally think comments should only be used to explain things that may not be obvious. Could you remove the comments that explain the same thing that reading the code would?

src/components/configdrop/ConfigDrop.test.tsx
```diff
@@ -0,0 +1,53 @@
+import * as enzyme from "enzyme";
+import { DropzoneArea } from 'material-ui-dropzone';
+import * as React from "react";
+import { ClientConfig } from '../../models/client_config';
+import { ConfigDrop } from "./ConfigDrop";
+
+describe("ConfigDrop", () => {
+    const verifyConfig = (configObj: ClientConfig) => {
```
john-m-liu: There is a potential problem here where the test will pass if this function isn't called. Take a look at Banner.test.tsx for how you can ensure that it's called

src/components/configdrop/ConfigDrop.test.tsx
```diff
@@ -0,0 +1,53 @@
...
+        "ui_url": "https://evergreen.mongodb.com"
+      };
+      const jsonString = JSON.stringify(validJSON);
+      const validFile = new File ([jsonString], "config.json");
+      // find the dropzone and simulate dropping our new file in to that drop zone
+      const drop = wrapper.find(DropzoneArea);
+      // we will verify that the config object contains what we expect in the verifyConfig function passed in to the wrapper
+      drop.simulate("change", {target: {files: [validFile]}});
+    })
+})
```
john-m-liu: Could you add 1 more test with an invalid JSON file, and assert that the element contains a visible snackbar?

src/components/app/App.tsx
dominoweir: I originally had this as null in my code earlier today, but I noticed that if you initialize it to null and do not upload a config file first, the admin page will not load. It fails because you can't call ConvertToAdminSettings with a null API client, but you can with an empty one. Should I still change it for now?

john-m-liu: Strange, that shouldn't happen since the function doesn't use the client at all. Let's initialize it to null and fix it in evergreen.js if needed

src/components/configdrop/ConfigDrop.test.tsx
```diff
@@ -0,0 +1,71 @@
+import * as enzyme from "enzyme";
+import { DropzoneArea } from 'material-ui-dropzone';
+import * as React from "react";
+import { ClientConfig } from '../../models/client_config';
+import { ConfigDrop } from "./ConfigDrop";
+
+describe("ConfigDrop", () => {
+
+  const verifyValidConfig = jest.fn((configObj: ClientConfig) => {
+    if (configObj !== undefined) {
```
john-m-liu: Add an expect(configObj !== undefined) as well

PR #4, opened 2019-06-27
EVG-6227: Create a default description for patches

dominoweir:
This set of changes includes using the evergreen API to get a user's patches, then populating the patch container using that data. If a patch does not have a description, we create one using other information about the patch (author, date/time, project, githash, etc.)


src/components/app/App.tsx
```diff
@@ -22,7 +22,7 @@ export class Evergreen extends React.Component<Props, State> {
   constructor(props: Props) {
     super(props);
     this.state = {
-      APIClient: null,
+      APIClient: rest.EvergreenClient("", "", "", "", true),
```
john-m-liu: We can leave this as null can't we? Assuming that you upload a config file before going to a page that needs to use the client

dominoweir: Yes you can, the only problem is that the patches page depends on that client existing and being not null in order to load all the patches. I switched it back to an empty client because it was really annoying to have to upload a config file every time I wanted to test something on the patches page. I can make it null again if you think that's better, it was just inconvenient when testing.

john-m-liu: Fair, let's leave it like this. I was annoyed by having to upload it each time as well

PR #5, opened 2019-06-28
EVG-6337: implement expanded view for a patch

dominoweir:
This set of changes includes: adding more mock data (which is why it says so many lines were added), creating a variant component for each build variant in a patch labeled with its name, populating that variant with cards containing a count of tasks for each status type (success, failure, etc.), and organizing those cards by priority (failures and system errors are displayed first).


src/components/patch/Patch.tsx
```diff
@@ -34,34 +35,35 @@ export class Patch extends React.Component<Props, State> {
...
   public render() {
 
+    const Variants = () => (
+      <Grid container={true} spacing={24}>
+        <Grid item={true} xs={12}>
+          <Typography>
+            Applied at {this.state.datetime.format("D MMM YYYY, h:mm a")} on {this.state.project}
+          </Typography>
+        </Grid>
+        {Object.keys(this.state.builds).map(index => (
```
john-m-liu: Since this.state.builds is an array, what's the reason for enumerating its keys? It should return the arrays of the index. I think you can get the same behavior by doing this.state.builds.map and passing the object itself rather than the indices

src/components/variant/Variant.tsx
```diff
@@ -1,37 +1,93 @@
...
 
-export class Variant extends React.Component<Props> {
+export class Variant extends React.Component<Props, State> {
+
   constructor(props: Props) {
     super(props);
+    const statusCount = {};
+    for (const task of this.props.build.Build.tasks) {
+      if (statusCount[task.status]) {
+        statusCount[task.status] = statusCount[task.status] + 1;
```
john-m-liu: nit: I would use ++

src/components/variant/Variant.tsx
```diff
@@ -1,37 +1,93 @@
...
+      "inactive": 9,
+    }
+    return displayPriority[a] > displayPriority[b] ? 1 : -1;
+  }
+
+  private orderByPriority() {
+    const sortedStatus = [];
+    const asArray = [];
+    for (const key of Object.keys(this.state.statusCount)) {
+      asArray[asArray.length] = key
```
john-m-liu: I would use asArray.push(key) to make it clearer what this is doing

src/components/variant/Variant.tsx
```diff
@@ -1,37 +1,93 @@
...
         <Typography variant="body1" className="variant-title">
-          {this.props.name}
-        </Typography>      
+          {this.state.name}
+        </Typography>
       </Card>
     );
   }
+
+  private compareByPriority(a: any, b: any) {
```
john-m-liu: the types of the params should be strings

src/components/variant/Variant.tsx
```diff
@@ -1,37 +1,93 @@
...
+      "system-failed": 2,
+      "system-unresponsive": 3,
+      "setup-failed": 4,
+      "success": 5,
+      "started": 6,
+      "dispatched": 7,
+      "undispatched": 8,
+      "inactive": 9,
+    }
+    return displayPriority[a] > displayPriority[b] ? 1 : -1;
```
john-m-liu: nit: a slick way to do this is to just return displayPriority[a] - displayPriority[b]. I think that might be less clear though so feel free to resolve this if you prefer what you have now

src/components/variant/Variant.tsx
dominoweir: you're right- when I wrote this code originally, typescript yelled at me saying the type had to be any, but it appears that the string type is fine now.

src/components/variant/Variant.tsx
dominoweir: huh, I didn't know that was a thing you could do in Javascript, I assumed it wouldn't actually update anything in the object. fixed

src/components/variant/Variant.tsx
dominoweir: Fixed

src/components/variant/Variant.tsx
```diff
@@ -1,37 +1,93 @@
...
+  public build: UIBuild;
 }
 
-export class Variant extends React.Component<Props> {
+export class Variant extends React.Component<Props, State> {
+
   constructor(props: Props) {
     super(props);
+    const statusCount = {};
+    for (const task of this.props.build.Build.tasks) {
```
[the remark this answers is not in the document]
dominoweir: Fixed

src/components/patch/Patch.tsx
dominoweir: I have no particular reason for doing it this way besides a lack of familiarity with how the map function works. I tried just passing the object itself and you're right- it behaves the same.

src/components/patch/Patch.tsx
```diff
@@ -34,34 +35,35 @@ export class Patch extends React.Component<Props, State> {
...
     };
   }
 
   public render() {
 
+    const Variants = () => (
+      <Grid container={true} spacing={24}>
+        <Grid item={true} xs={12}>
+          <Typography>
+            Applied at {this.state.datetime.format("D MMM YYYY, h:mm a")} on {this.state.project}
```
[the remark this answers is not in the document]
dominoweir: The user's browser locale is not something I had thought about before but I definitely agree. I changed the places that use a datetime to format based on locale. I also swapped out "applied" for "created" (I didn't like that wording either, I just kept it because I wasn't sure if the word applied was an intentional choice with strong reasoning behind it)

src/components/variant/Variant.tsx
```diff
@@ -72,17 +77,17 @@ export class Variant extends React.Component<Props, State> {
     return displayPriority[a] > displayPriority[b] ? 1 : -1;
   }
 
-  private orderByPriority() {
+  private orderByPriority(statusCount: {}) {
```
john-m-liu: Since you use the statusCount type declaration in 2 places now, can you create a private type for it in this file and declare this param as well as this.state.statusCount as that type?

package-lock.json
```diff
@@ -5803,9 +5803,9 @@
       }
     },
     "evergreen.js": {
-      "version": "1.0.12",
-      "resolved": "https://registry.npmjs.org/evergreen.js/-/evergreen.js-1.0.12.tgz",
-      "integrity": "sha512-fANM+lV6h7GuOq2VNlu6zI/31CD0d1ibwruGVp2FkNc/Hjlof56iCWYbgu2FjCVWFQyhV9CbhtL9f5WsmHFeRQ==",
+      "version": "1.0.13",
```
john-m-liu: Also doesn't this need to be 14?

src/components/variant/Variant.tsx
dominoweir: yup, declared a new type: 
interface StatusCount { [id: string]: number };

package-lock.json
dominoweir: It does- it looks like the version is correct in package.json but not package-lock.json because I forgot to run npm update before pushing. Fixed now.

PR #6, opened 2019-07-02
EVG-6308: Add searching of patches

dominoweir:
This set of changes includes the implementation of search by patch description, some element styling, and a test to confirm that search works as expected.

### DIFF
--- a/src/components/patch/Patch.less
+++ b/src/components/patch/Patch.less
@@ -7,8 +7,13 @@
   padding-right: 100px;
 }
 
-.patch-search-bar {
-  text-align: center;
-  padding-top: 50px;
-  padding-bottom: 50px;
+.search-container {
+  padding-top: 40px;
+  padding-bottom: 40px;
+  padding-left: 300px;
+  padding-right: 300px;
+}
+
+.search-input {
+  min-width: 400;
 }
--- a/src/components/patch/Patch.less
+++ b/src/components/patch/Patch.less
@@ -10,13 +10,13 @@
 .search-container {
   padding-top: 40px;
   padding-bottom: 40px;
-  min-width: 400px;
-  max-width: 700px;
-  text-align: center;
+  padding-left: 300px;
+  padding-right: 300px;
 }
 
 .search-input {
   padding-left: 15px;
   padding-top: 5px;
   padding-bottom: 5px;
+  min-width: 400px;
 }
--- a/src/components/patch/Patch.less
+++ b/src/components/patch/Patch.less
@@ -10,10 +10,13 @@
 .search-container {
   padding-top: 40px;
   padding-bottom: 40px;
-  padding-left: 300px;
-  padding-right: 300px;
+  min-width: 400px;
+  max-width: 700px;
+  text-align: center;
 }
 
 .search-input {
-  min-width: 400;
+  padding-left: 15px;
+  padding-top: 5px;
+  padding-bottom: 5px;
 }
--- a/src/components/patch/PatchContainer.test.tsx
+++ b/src/components/patch/PatchContainer.test.tsx
@@ -47,7 +47,6 @@ describe("PatchContainer", () => {
   })
 
   it("check that search returns correct results", () => {
-    // note: this test uses a shallow wrapper because simulating events on a wrapper instantiated with enzyme.mount doesn't really work
     const event = {currentTarget: { value: "pull request #2428" }};
     const expectedResults = ["5d138b6b61837d77f9dda2a1", "5d1391b63e8e860e458573a5"];
     const notInResults = ["5d1385720305b932b1d50d01", "5d126fa93627e070b33dbbc0"];
--- a/src/components/patch/PatchContainer.test.tsx
+++ b/src/components/patch/PatchContainer.test.tsx
@@ -52,7 +52,7 @@ describe("PatchContainer", () => {
     const notInResults = ["5d1385720305b932b1d50d01", "5d126fa93627e070b33dbbc0"];
     const input = wrapper.find(InputBase);
     expect(input).toHaveLength(1);
-    input.prop("onChange")(event);
+    input.prop("onChange")(event as React.ChangeEvent<HTMLInputElement>);
     const visibleIds = Object.keys(wrapper.state("visible"));
     for(const versionId of expectedResults) {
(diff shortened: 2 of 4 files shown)


src/components/patch/PatchContainer.test.tsx
```diff
@@ -31,7 +31,7 @@ describe("PatchContainer", () => {
     const patch = wrapper.findWhere(node => node.key() === "5d126fa93627e070b33dbbc0").find(Patch);
     expect(patch).toHaveLength(1);
     expect(patch.state("description")).toBeDefined();
-    expect(patch.state("description")).toBe("Patch from domino.weir at Tuesday, June 25, 2019 3:02 PM on project spruce");
```
dominoweir: This was causing my patch to fail in Evergreen because I made an assumption about the locale it would be running from- (expected time was 3:02 PM, actual time was 7:02 PM). The main point of this test is to make sure a description is defined and nonempty for every patch, so instead of looking for an exact match I'm just checking that the patch description contains the text that it should start with.

src/components/patch/PatchContainer.tsx
```diff
@@ -48,9 +60,40 @@ export class PatchContainer extends React.Component<Props, State> {
...
+    const query = event.currentTarget.value;
+    const filteredPatches = this.filterItems(query);
+    this.setState({
+      visible: filteredPatches,
+    });
+  }
+
+  private filterItems(query: string) {
+    const filtered = {};
+    if(query === "") {
```
john-m-liu: Space after the if

src/components/patch/PatchContainer.tsx
```diff
@@ -48,9 +60,40 @@ export class PatchContainer extends React.Component<Props, State> {
...
+      return this.state.versions;
+    }
+    Object.keys(this.state.versions).map(versionId => {
+      const patch = this.state.versions[versionId];
+      let description = patch.Version.message;
+      if (description === undefined) {
```
john-m-liu: There shouldn't be a need to default a description here. But if there is, the right way would be to create a separate function for the default description rather than copying the code from the patch component

src/components/patch/PatchContainer.tsx
```diff
@@ -39,7 +43,15 @@ export class PatchContainer extends React.Component<Props, State> {
 
     return (
       <div>
-        <div className="patch-search-bar">Insert patch search bar here</div>
+        <div className="search-container">
+          <Paper>
+            <InputBase startAdornment={<IconButton><SearchIcon /></IconButton>}
+              className="search-input" fullWidth={true}
+              placeholder="Search Patch Descriptions..."
```
john-m-liu: Nit: I think the label shouldn't have the ellipses

src/components/patch/PatchContainer.tsx
```diff
@@ -39,7 +43,15 @@ export class PatchContainer extends React.Component<Props, State> {
 
     return (
       <div>
-        <div className="patch-search-bar">Insert patch search bar here</div>
+        <div className="search-container">
+          <Paper>
+            <InputBase startAdornment={<IconButton><SearchIcon /></IconButton>}
```
john-m-liu: Since this searches as you type, we should just have an icon here rather than a button

src/components/patch/Patch.less
```diff
@@ -7,8 +7,13 @@
   padding-right: 100px;
 }
 
-.patch-search-bar {
-  text-align: center;
-  padding-top: 50px;
-  padding-bottom: 50px;
+.search-container {
```
john-m-liu: Was the reason for changing the alignment approach so that you could left-justify the label text? If so I think a better approach would be to keep text-align:center on the container, but override it for the input element. That way you have a more consistent styling hierarchy for all screen sizes. It also lets you set a max-width for the input (which I think we should do since it's huge on my 26 inch monitor)

src/components/patch/PatchContainer.tsx
dominoweir: Fixed

src/components/patch/PatchContainer.tsx
dominoweir: Fixed

src/components/patch/PatchContainer.tsx
dominoweir: Fixed (I thought the icon button was required for an icon for some reason)

src/components/patch/Patch.less
dominoweir: I set a max width, which definitely does look better, but I'm having trouble figuring out how to center align the search bar now.

src/components/patch/PatchContainer.tsx
dominoweir: It is necessary here since the description is what we're searching on. I moved this code in to a function.

src/components/patch/Patch.less
dominoweir: Leaving this alone for now since we discovered the CSS is mysterious

src/components/patch/PatchContainer.tsx
john-m-liu: Isn't it already set by the patch itself?

dominoweir: It wasn't before, but I just changed that (I thought initially that the change wouldn't be reflected in the parent PatchContainer, only the Patch). The code looks cleaner/shorter/all around better now.

src/components/patch/PatchContainer.test.tsx
```diff
@@ -44,4 +45,20 @@ describe("PatchContainer", () => {
...
+
+  it("check that search returns correct results", () => {
+    const event = {currentTarget: { value: "pull request #2428" }};
+    const expectedResults = ["5d138b6b61837d77f9dda2a1", "5d1391b63e8e860e458573a5"];
+    const notInResults = ["5d1385720305b932b1d50d01", "5d126fa93627e070b33dbbc0"];
+    const input = wrapper.find(InputBase);
+    expect(input).toHaveLength(1);
+    input.prop("onChange")(event as React.ChangeEvent<HTMLInputElement>);
+    const visibleIds = Object.keys(wrapper.state("visible"));
+    for(const versionId of expectedResults) {
```
john-m-liu: nit: spacing

PR #7, opened 2019-07-02
add a deployment script, attempt to load config automatically

john-m-liu:
Initial setup:
- Have the AWS CLI configured
- Create a S3 bucket
- Give the bucket a policy that allows public read of objects
- Enable the bucket to host static websites
- Create a config.json file at the root

Deploying:
- npm run build if needed (creates build folder)
- BUCKET=yourbucket npm run deploy (syncs the build folder to the bucket)

http://john.liu2.s3-website-us-east-1.amazonaws.com/#/


src/models/client_config.ts
```diff
@@ -10,4 +10,9 @@ export class ClientConfig {
   public api_key?: string;
   public api_url?: string;
   public ui_url?: string;
+}
+
+export function IsValidConfig(toCheck: object): boolean {
+  const config = toCheck as ClientConfig;
+  return config.api_url !== undefined && config.ui_url !== undefined;
```
dominoweir: Should this also be checking to make sure the username and api key are defined?

john-m-liu: My approach was to spot check key fields, since it would be annoying to check everything as we add more config options. Also the credentials will be going away soon

src/components/app/App.tsx
```diff
@@ -84,6 +87,20 @@ export class Evergreen extends React.Component<Props, State> {
       APIClient: rest.EvergreenClient(configObj.user, configObj.api_key, configObj.api_url, configObj.ui_url),
     });
   }
+
+  private tryLoadConfig = () => {
+    fetch(configPath).then((resp: Response) => {
```
bsamek: nit: dunno if we have a convention or care, but there's inconsistent spacing after the parens after the `then`

src/components/app/App.tsx
```diff
@@ -2,7 +2,7 @@ import { AppBar, IconButton, Menu, MenuItem, Toolbar, Typography } from '@materi
 import * as MenuIcon from '@material-ui/icons/Menu';
 import * as React from 'react';
 import { HashRouter, NavLink, Route } from 'react-router-dom';
-import { ClientConfig } from '../../models/client_config';
+import { ClientConfig, IsValidConfig } from '../../models/client_config';
```
tychoish: is relative pathing really the only way to do imports?

john-m-liu: I think so to avoid ambiguity with imports of node_modules

PR #8, opened 2019-07-03
EVG-6307: Plumb real evergreen API to patch page

dominoweir:
This set of changes includes: ordering patches by most recent first, infinite scroll of patches using the real API, fixing duplicate keys in Patch components, and reworking a couple of small things for testing purposes.


src/components/patch/PatchContainer.tsx
```diff
@@ -51,41 +57,64 @@ export class PatchContainer extends React.Component<Props, State> {
...
+    if (this.state.hasMore) {
+      this.props.client.getPatches((err, resp, body) => {
+        const newPatches = ConvertToPatches(resp.body).VersionsMap;
+        if (Object.keys(newPatches).length === 0) {
+          this.setState({ 
+            hasMore: false 
+          });
+          return;
+        }
+        const sortedPatches = this.state.allPatches;
```
john-m-liu: Do we need to do this? I would think that the API returns patches in reverse chronological order, so that we only need to append the new results

dominoweir: The API does return them in order, but with the oldest patch first (we want the newest patch first). I already needed a compare by date function elsewhere in the code so I just used it here as well to order everything.

src/components/patch/PatchContainer.tsx
```diff
@@ -67,22 +67,19 @@ export class PatchContainer extends React.Component<Props, State> {
...
             hasMore: false 
           });
           return;
         }
-        const sortedPatches = this.state.allPatches;
-        Object.keys(newPatches).map(versionId => {
-          sortedPatches.push(newPatches[versionId]);
-        });
-        sortedPatches.sort(this.compareByDate);
+        // reorder the patches so that the newest element is at the front of the array
```
dominoweir: I know you don't like comments in code, but reversing the array seems cryptic enough without context to warrant an explanation.

src/components/patch/PatchContainer.tsx
```diff
@@ -51,41 +57,61 @@ export class PatchContainer extends React.Component<Props, State> {
...
-      const patch = this.state.versions[versionId];
+    Object.keys(this.state.allPatches).map(versionId => {
+      const patch = this.state.allPatches[versionId];
       const description = patch.Version.message;
       if (description.toLowerCase().indexOf(query.toLowerCase()) !== -1) {
-        filtered[versionId] = patch;
+        filtered.push(patch);
       }
     });
+    filtered.sort(this.compareByDate);
```
john-m-liu: I don't think you need to do this. Javascript array methods like map iterate from the lowest index in the array, so what you have at this point should already be sorted

dominoweir: Good to know- I did this originally because I wasn't totally sure that map would iterate from the lowest index. I have now removed that code.

PR #9, opened 2019-07-10
EVG-6411: Patch expand panels collapse when a new page of patches is loaded

dominoweir:
This set of changes includes adding logic to determine whether a patch should be expanded on rendering after changes occur somewhere on the page.

### DIFF
--- a/src/components/patch/PatchContainer.test.tsx
+++ b/src/components/patch/PatchContainer.test.tsx
@@ -1,8 +1,9 @@
-import { InputBase } from '@material-ui/core';
+import { ExpansionPanel, InputBase } from '@material-ui/core';
 import * as enzyme from "enzyme";
 import { UIVersion } from 'evergreen.js/lib/models';
 import * as moment from 'moment';
 import * as React from "react";
+import * as InfiniteScroll from 'react-infinite-scroller';
 import * as rest from "../../rest/interface";
 import { Variant } from '../variant/Variant';
 import { Patch } from './Patch';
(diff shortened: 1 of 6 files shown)


src/components/patch/PatchContainer.tsx
```diff
@@ -39,7 +41,7 @@ export class PatchContainer extends React.Component<Props, State> {
       <Grid className="patch-container" container={true} spacing={24}>
         {this.state.visiblePatches.map(patchObj => (
           <Grid item={true} xs={12} key={patchObj.Version.id}>
-            <Patch Patch={patchObj} />
+            <Patch Patch={patchObj} updateOpenPatches={this.updateOpenPatches} expanded={this.isExpanded(patchObj)} />
```
john-m-liu: This check is O(n^2) since indexOf needs to potentially scan the entire patches array. Can you instead track the expanded patches by whatever unique key you're using to repeat them?

dominoweir: Good call, I totally forgot indexOf would be O(n^2). I'm now representing those with a dict using the patch ID as a key.

john-m-liu:
For seeing why this feels slower than before - you can use the react profiler in the react dev tools to record clicking an expansion panel before and after this change. The profile will display as a flame graph, which shows how long each function called during that window took. If you see something obviously taking longer before and after, let's try to fix that

dominoweir:
Based on the React Profiler, it doesn't seem like anything is obviously taking longer. I think removing that O(n^2) code helped.

src/components/patch/PatchContainer.test.tsx
```diff
@@ -10,7 +10,19 @@ import { PatchContainer } from "./PatchContainer";
 
 describe("PatchContainer", () => {
 
-  const wrapper = enzyme.mount(<PatchContainer client={rest.EvergreenClient("", "", "", "", true)} />);
+  const checkState = jest.fn(() => {
```
john-m-liu: Since this is only used for 1 test, let's scope the variable to just the test

src/components/patch/PatchContainer.test.tsx
```diff
@@ -10,7 +10,19 @@ import { PatchContainer } from "./PatchContainer";
...
+    wrapper.update();
+    expect(wrapper.state("expandedPatches")).toEqual({ "5d1385720305b932b1d50d01": 1 });
+    const patch = wrapper.findWhere(node => node.key() === "5d1385720305b932b1d50d01");
+    expect(patch).toHaveLength(1);
+    // tslint:disable-next-line: no-string-literal
```
john-m-liu: If this is scoped to the test, you can avoid this code duplication and just set the patch variable once

src/components/patch/PatchContainer.test.tsx
```diff
@@ -44,17 +56,30 @@ describe("PatchContainer", () => {
     expect(failedBox).toHaveLength(1);
     expect(variant.state("name")).toBe("Ubuntu 16.04");
     expect(variant.state("statusCount")).toEqual({ success: 2, failed: 1 });
-    expect(variant.state("sortedStatus")).toEqual([{"count": 1, "status": "failed"}, {"count": 2, "status": "success"}]);
+    expect(variant.state("sortedStatus")).toEqual([{ "count": 1, "status": "failed" }, { "count": 2, "status": "success" }]);
+  })
+
+  it("check that expanded state persists on re-render", () => {
+    expect(wrapper.state("expandedPatches")).toEqual({});
+    const patch = wrapper.findWhere(node => node.key() === "5d1385720305b932b1d50d01");
```
john-m-liu: This actually finds the wrapper around the patch component. Would it be easier to find the patch component itself so you don't have to look at its children?

src/components/patch/PatchContainer.test.tsx
```diff
@@ -44,17 +56,30 @@ describe("PatchContainer", () => {
...
+    expect(patch).toHaveLength(1);
+    // tslint:disable-next-line: no-string-literal
```
john-m-liu: Let's avoid having to do this by casting to a patch component

src/components/patch/PatchContainer.test.tsx
dominoweir: Fixed

src/components/patch/PatchContainer.test.tsx
dominoweir: Fixed (sort of)

PR #13, opened 2019-07-23
EVG-6385: create task picker component

dominoweir:
This set of changes includes the implementation of the BuildView sidebar, fixing patch searching (broke when I changed how patches are loaded), and some other changes from running npm update.


src/components/log/LogContainer.tsx
```diff
@@ -41,25 +43,25 @@ export class LogContainer extends React.Component<Props, State> {
...
 
   public render() {
 
-    // TODO: use Evergreen API to determine the severity of each log line (see example below)
-    // (https://evergreen.mongodb.com/json/task_log/spruce_ubuntu1604_lint_37ec5816d14e05bd02535788d114da8b97ed0231_19_07_11_18_09_18/0?type=T)
     const Logs = () => (
       <InfiniteScroll loadMore={this.dummyLoadMore} className="log-scrollable">
         {this.state.logText.split("\n").map(textLine => (
-          <Typography className={this.lineContainsError(textLine.toLowerCase()) ? "log-text-error" : "log-text-normal"}>
+          <Typography className={this.lineContainsError(textLine.toLowerCase()) ? "log-text-error" : "log-text-normal"} key={new Date().getTime().toString()}>
```
dominoweir: I was getting lots of console warnings about the log lines not having unique IDs. I decided to use datetime as the basis for a unique key at the suggestion of a StackOverflow post (using an incrementing number as you iterate through the list is a bad idea apparently). Let me know if you have a better/less strange solution.

src/components/build/BuildSidebar.tsx
```diff
@@ -38,15 +43,8 @@ export class BuildSidebar extends React.Component<Props, State> {
...
-      <Grid item={true} xs={12} key={taskObj.task_id}>
-        <ExpansionPanel className="task-panel">
-          <ExpansionPanelSummary expandIcon={<ExpandMoreIcon />}>
-            <Typography>{taskObj.display_name}</Typography>
-          </ExpansionPanelSummary>
-          <ExpansionPanelDetails>
-            <Typography>No tests to show</Typography>
-          </ExpansionPanelDetails>
-        </ExpansionPanel>
+      <Grid item={true} xs={12} key={taskObj.display_name}>
```
john-m-liu: display_name isn't guaranteed to be unique. How come you switch from the task ID?

src/components/task/TaskPanel.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+export const StyledExpansionPanel = withStyles({
+  root: {
+    border: '1px solid rgba(0, 0, 0, .125)',
+    boxShadow: "none"
+  }
+})(MuiExpansionPanel);
+
+interface State {
+  hasFailingTests: boolean
```
john-m-liu: Do we need these 2 boolean properties? Would checking the length of the arrays work instead?

src/components/task/TaskPanel.tsx
```diff
@@ -0,0 +1,186 @@
...
+      isShowingOtherTests: false
+    }
+  }
+
+  public componentDidMount() {
+    this.props.client.getTestsForTask((err, resp, body) => {
+      const tests = ConvertToAPITests(body) as unknown as APITest[];
+      let failingTests: APITest[] = [];
+      let otherTests: APITest[] = [];
+      Array.from(tests).forEach(element => {
```
john-m-liu: While the vast majority of tests will either be pass or fail, there are potentially 2 others states: https://github.com/evergreen-ci/evergreen/blob/master/globals.go#L76. I think we need a way for someone to discover the actual status of a test

src/components/task/TaskPanel.tsx
```diff
@@ -0,0 +1,186 @@
...
+import MuiExpansionPanel from '@material-ui/core/ExpansionPanel';
+import ExpandLessIcon from '@material-ui/icons/ExpandLess';
+import ExpandMoreIcon from '@material-ui/icons/ExpandMore';
+import { withStyles } from '@material-ui/styles';
+import { APITask, APITest, ConvertToAPITests } from 'evergreen.js/lib/models';
+import * as React from 'react';
+import * as rest from "../../rest/interface";
+import '../../styles.css';
+
+export const StyledExpansionPanel = withStyles({
```
john-m-liu: I'm not 100% sure what I would change, but I feel like the sidebar could use some color or labels to help differentiate the different pieces of data. Things like task status, time taken, and display task hierarchy aren't really apparent. What do you think?

src/components/task/TaskPanel.tsx
```diff
@@ -0,0 +1,186 @@
...
+            Other Tests
+          </Typography>
+        </CardActions>
+        <Collapse in={this.state.isShowingOtherTests} timeout="auto" unmountOnExit={true}>
+          <CardContent className="other-tests-list">
+            <Grid container={true} spacing={1}>
+              {this.state.otherTests.map(test => (
+                <Grid item={true} xs={12} key={test.test_file}>
+                  <Typography>
+                    {test.test_file.split("/").pop()}
```
john-m-liu: I sort of expected clicking a test would show me the test logs in the panel. Are you planning on implementing that later?

src/components/log/LogContainer.tsx
```diff
@@ -41,25 +43,25 @@ export class LogContainer extends React.Component<Props, State> {
   }
 
   public componentDidUpdate() {
-    if (this.props.task.logs !== undefined && this.state.logText === "") {
+    if ((this.props.task.logs !== undefined && this.state.logText === "") || (this.props.task.task_id !== this.state.taskId)) {
```
john-m-liu: I'm not sure if it was this change or if it was like this before, but having a task selected goes into an infinite loop of requesting task logs

src/components/log/LogContainer.tsx
john-m-liu: I'd prefer to not have any key since I see little need to do so. Do you know why a typography component needs a key?

src/components/build/BuildSidebar.tsx
dominoweir: The task ID is also not guaranteed to be unique. In fact, when I use my patch data, the task ID is the same for every element. Is there something else that would be better to use than the display name?

src/components/log/LogContainer.tsx
dominoweir: It's not actually the typography component that needs it. This is a general requirement of using the map() function to generate elements in React. Things still work without unique keys, it just fills the console with messages warning that elements should have a unique key. You can read more about that [here](https://reactjs.org/docs/lists-and-keys.html).

src/components/task/TaskPanel.tsx
dominoweir: No, we don't really need them, I just thought it was cleaner. I can swap these properties out for length checks in my code.

src/components/task/TaskPanel.tsx
dominoweir: I agree with that, at least for including the status. The time taken I'm not so sure about- none of the people I interviewed mentioned that as being a helpful piece of information, but maybe I'm just missing a use case. The reason I didn't add the display task hierarchy is because it is really not apparent to me how that works just looking at data returned from the API, and it's not displayed well enough in the current UI for me to understand it.

src/components/task/TaskPanel.tsx
dominoweir: I agree with that. Maybe it can be displayed when you hover over the test name, or displayed in a different color? Also, do we consider a skip or a silent failure an actual failure, or would those go in the "other tests" dropdown?

src/components/task/TaskPanel.tsx
dominoweir: Yes, I was. There's kind of a lingering design question there- because the test logs are linked through LogKeeper, I'm not totally sure how to retrieve or display them.

src/components/log/LogContainer.tsx
dominoweir: I'm not sure if that was happening before either, but I'll look in to it.

src/components/task/TaskPanel.tsx
dominoweir: Fixed

src/components/task/TaskPanel.tsx
john-m-liu: They shouldn't be considered failures. A color highlight seems a simple way to denote the status, but I think it would be useful to also say it somewhere. How do you feel about taking either the bottom (or top) of the sidebar to display some metadata about the task?

src/components/task/TaskPanel.tsx
john-m-liu: You can retrieve them the same way. The link property of the task will just point to logkeeper, which should be able to serve up the log in plain text format as well

src/components/build/BuildSidebar.tsx
dominoweir: Here's an example of what I get back as a response currently, where all task IDs are the same: https://evergreen.mongodb.com/api/rest/v2/tasks/mongodb_mongo_master_enterprise_rhel_62_64_bit_aggregation_auth_patch_aa9f6a202e0709adf14046cb27504864adaf732b_5d28d45fd1fe0716daa6b903_19_07_12_18_41_36/tests

dominoweir: Oh wait... I see what you mean. I think this was a mistaken change, I thought the warning about unique IDs was pointing to this mapping of elements, but I think it was actually warning me about the tests in each task (it is very hard to determine the source for that warning). You are correct that I should be using the task IDs here and that they should be unique. Sorry for the misunderstanding.

src/components/task/TaskPanel.tsx
dominoweir: Additional test colors have been added. I can definitely add metadata to the top of the task card. What do you think we should display?

src/components/log/LogContainer.tsx
dominoweir: Are you sure it goes in to an infinite loop? I don't see any evidence of that when I test it on my machine.

dominoweir: Fixed by adding a componentDidMount method

src/components/task/TaskPanel.tsx
dominoweir: Done

PR #14, opened 2019-07-25
EVG-6478: hide authentication related UI

dominoweir:
This set of changes includes hiding login related UI components (the code is still there, just commented out), and adding the username as a URL parameter when viewing patches.


src/components/app/App.tsx
```diff
@@ -49,7 +49,7 @@ export class Evergreen extends React.Component<Props, State> {
   constructor(props: Props) {
     super(props);
     this.state = {
-      APIClient: rest.EvergreenClient("", "", "", "", true),
+      APIClient: rest.EvergreenClient("https://evergreen.mongodb.com/api", "https://evergreen.mongodb.com")
```
john-m-liu: This should come from the config

dominoweir: You're right- I misunderstood this ticket initially and thought all the config stuff should be gone too. I've fixed this now.

dominoweir:
I have now removed that menu. The only thing with that is now I don't have a way to upload a config file for testing purposes, although I don't know how much that matters at this point.

src/components/patch/PatchContainer.tsx
```diff
@@ -67,6 +67,14 @@ export class PatchContainer extends React.Component<Props, State> {
 
   private loadPatches = () => {
     if (this.state.hasMore && !this.state.isSearching) {
+      let username = this.props.username;
+      const urlParams = window.location.href.split("?");
```
john-m-liu: I would use URLSearchParams to parse this: https://davidwalsh.name/query-string-javascript

PR #16, opened 2019-07-29
MCI-3066: add banner to initial load of the patches page

dominoweir:


src/components/patch/PatchContainer.tsx
```diff
@@ -25,13 +27,18 @@ class Props {
...
+    }
     this.state = {
       hasMore: true,
       pageNum: 0,
       allPatches: [],
       visiblePatches: [],
       expandedPatches: {},
-      isSearching: false
+      isSearching: false,
+      bannerIsHidden: shouldHideBanner && false
```
john-m-liu: Was this debug code?

src/components/patch/PatchContainer.tsx
```diff
@@ -49,6 +56,24 @@ export class PatchContainer extends React.Component<Props, State> {
 
     return (
       <div>
+        <div className="banner-container" hidden={this.state.bannerIsHidden}>
```
john-m-liu: I'd like to see this in a separate component so that it can be reused on future converted pages. The message should be configurable, and whether or not it's been acknowledged should be tracked separately. There should also be tests of it

PR #17, opened 2019-07-29
EVG-6447: create a banner linking to user settings

dominoweir:
I renamed this branch to line up with the ticket it actually belongs to, and in the process Github automatically closed my other PR under the old branch name.


dominoweir:
From John on the old PR: "I'd like to see this in a separate component so that it can be reused on future converted pages. The message should be configurable, and whether or not it's been acknowledged should be tracked separately. There should also be tests for it."

dominoweir:
It definitely makes sense for the message to be configurable, but I'm unclear on how to pass that message through to the Banner component. Specifically, what if I want to pass something besides just a string? In the case of the patches page, the banner message also contains a link to the user settings page that cannot be passed as a string.

package.json
```diff
@@ -148,7 +148,8 @@
       "jsx",
       "json",
       "node"
-    ]
+    ],
+    "setupTestFrameworkScriptFile": "<rootDir>/src/components/banner/MockLocalStorage.tsx"
```
john-m-liu: According to jest documentation we should be using setupFilesAfterEnv in new code

src/components/banner/Banner.tsx
```diff
@@ -0,0 +1,57 @@
...
+import CloseIcon from '@material-ui/icons/Close';
+import * as React from 'react';
+import '../../styles.css';
+
+interface State {
+  bannerIsHidden: boolean
+}
+
+class Props {
+  public message: any
```
john-m-liu: This isn't used at all

src/components/banner/MockLocalStorage.tsx
```diff
@@ -0,0 +1,18 @@
+const localStorageMock = (() => {
```
john-m-liu: This definitely works but it's a little Javascript-y in that we're creating and using arbitrary objects. Can you do it this way: https://stackoverflow.com/a/41434763

src/components/banner/Banner.tsx
```diff
@@ -0,0 +1,57 @@
...
+  public render() {
+
+    return (
+      <div className="banner-container" hidden={this.state.bannerIsHidden}>
+        <Paper className="banner">
+          <Grid container={true}>
+            <Grid item={true} xs={11}>
+              <p className="vertical-center">
+                {"Welcome to the new patches page! You can opt out of this page or report bugs "}
+                <Link href={"https://evergreen.mongodb.com/settings"}>here</Link>.
```
john-m-liu: Use the base URL from the config here

src/components/banner/Banner.tsx
dominoweir: I know, it's a placeholder while I figure out the message passing thing I left a comment about above.

package.json
dominoweir: Link? When I use that variable name, I get the following warning: 
"Validation Warning: Unknown option "setupFilesAfterEnv" with value <rootDir>/src/components/banner/MockLocalStorage.tsx was found. This is probably a typing mistake. Fixing it will remove this message."

src/components/banner/MockLocalStorage.tsx
dominoweir: Fixed

src/components/banner/Banner.tsx
dominoweir: Fixed

package.json
john-m-liu: https://jestjs.io/docs/en/configuration#setupfilesafterenv-array

dominoweir: Hmm, the tests do still pass despite that warning so I guess it's ok?

john-m-liu: Ah the problem is that we're using jest v23, but that field is added in v24. You can do npm install jest@latest --save to update it, but it looks like this is actually not needed because node.js actually has some implementation of local storage - you can log localStorage.getItem.toString() in a test to see what that is. So I think this can just be deleted

dominoweir:
Okay, I have now updated the banner component to use the message and an optional link. Let me know what you think.

src/components/banner/Banner.test.tsx
```diff
@@ -0,0 +1,35 @@
+import CloseIcon from '@material-ui/icons/Close';
+import * as enzyme from "enzyme";
+import * as React from "react";
+import * as rest from "../../rest/interface";
+import Banner from './Banner'
+
+describe("Banner", () => {
+
+  const wrapper = enzyme.mount(<Banner client={rest.EvergreenClient("https://evergreen.mongodb.com/api", "https://evergreen.mongodb.com")} message={""} showOptOut={true} onFinishStateUpdate={null} storageKey={"shouldHideBanner"}/>);
```
john-m-liu: Could you also add a test showing that setting showOptOut=false doesn't show the link?

src/components/banner/Banner.tsx
```diff
@@ -0,0 +1,62 @@
...
+import * as rest from "../../rest/interface";
+import '../../styles.css';
+
+interface State {
+  bannerIsHidden: boolean
+}
+
+class Props {
+  public client: rest.Evergreen
+  public message: any
```
john-m-liu: Shouldn't this be type string?

dominoweir: Fixed

src/components/banner/Banner.test.tsx
dominoweir: Done

PR #18, opened 2019-07-30
EVG-6407: add searching by patch attribute

dominoweir:
The basic search functionality is done here, although I'm still working on how to style the drop down for selecting a property to search on. I'm worried that my implementation of search wouldn't be very fast with lots of patches- let me know what you think.


src/components/app/App.tsx
```diff
@@ -49,7 +49,7 @@ export class Evergreen extends React.Component<Props, State> {
   constructor(props: Props) {
     super(props);
     this.state = {
-      APIClient: rest.EvergreenClient("", "", "", "", true)
+      APIClient: rest.EvergreenClient("https://evergreen.mongodb.com/api", "https://evergreen.mongodb.com", "", "")
```
john-m-liu: This shouldn't be populated in the constructor (and if it did it should come from the config)

src/components/patch/PatchContainer.tsx
```diff
@@ -1,20 +1,28 @@
...
-import { ConvertToPatches, UIVersion } from 'evergreen.js/lib/models';
+import { ConvertToPatches, UIPatch, UIVersion } from 'evergreen.js/lib/models';
 import * as React from 'react';
 import * as InfiniteScroll from 'react-infinite-scroller';
 import * as rest from "../../rest/interface";
 import '../../styles.css';
 import Banner from '../banner/Banner';
 import Patch from './Patch';
 
+enum SearchType {
```
john-m-liu: My intent was to have the project and status be discrete filters in addition to searching patch descriptions. I think the use cases will want an intersection of the filters

src/components/patch/PatchContainer.tsx
```diff
@@ -108,37 +128,56 @@ export class PatchContainer extends React.Component<Props, State> {
...
-      if (description.toLowerCase().indexOf(query.toLowerCase()) !== -1) {
+      const patch = this.state.allPatches[versionId] as UIPatch;
+      let queryField = "";
+      switch (this.state.searchType) {
+        case SearchType.description:
+          queryField = patch.Patch.Description
+          break;
+        case SearchType.project:
+          queryField = patch.Patch.Project
+          console.log(queryField);
```
john-m-liu: Take out these statements if you don't need them anymore

src/components/app/App.tsx
dominoweir: You're right- I just had this in for convenience's sake while I was testing. I'll go ahead and remove it.

src/components/patch/PatchContainer.tsx
dominoweir: Oh, I totally misunderstood the intent of this ticket. That's my bad, I'll go back now and start reworking my code.

src/components/patch/PatchContainer.tsx
dominoweir: Thanks, I totally missed these before I committed. I'll go ahead and remove any other stray logging statements I find.

dominoweir:
I figured out why some of the patches have empty messages- it was the difference between checked for an undefined field or an empty string. Fixed now.

src/components/patch/PatchContainer.tsx
```diff
@@ -41,24 +62,68 @@ export class PatchContainer extends React.Component<Props, State> {
...
+                fullWidth={true}
+                placeholder="Search Patch Descriptions"
+                onChange={this.search}
+              />
+            </ExpansionPanelSummary>
+            <ExpansionPanelDetails className="advanced-search">
+              <div>
+                <FormControl className="advanced-select">
+                  <InputLabel>Status</InputLabel>
+                  <Select
```
john-m-liu: I was thinking that these filters would be independent of the search (ie. you can filter based on status without typing in the search bar). Right now changing this doesn't call the search. What do you think? Also the placement of these suggests that they work in combination with the search, but I kind of think they shouldn't

src/components/patch/PatchContainer.tsx
```diff
@@ -108,37 +192,59 @@ export class PatchContainer extends React.Component<Props, State> {
...
   }
 
-  private isExpanded = (patchObj: UIVersion) => {
-    return patchObj.Version.id in this.state.expandedPatches;
+  private isExpanded = (patchObj: UIPatch) => {
+    return patchObj.Patch.Id in this.state.expandedPatches;
+  }
+
+  private onStatusSelectChange = (event: React.ChangeEvent<{ name?: string; value: string }>) => {
+    const selectedValues = event.target.value as unknown as string[];
```
john-m-liu: Do you need to case twice for this to work?

dominoweir: Yes- if not, you get the following error: "Conversion of type 'string' to type 'string[]' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first."

src/components/patch/PatchContainer.tsx
dominoweir: I think you and I thought this ticket meant different things. I was under the impression that the intent here was for these filters to work as a kind of "advanced search," such that changing them would only affect what is displayed while searching. They could certainly be independent of search, it would be super easy to position them somewhere else and use them to filter the list of visible patches. I agree that would make more sense anyways.

PR #21, opened 2019-08-08
EVG-6550: add task details to build view

dominoweir:
One tiny thing left to do on this ticket- a few of the fields I use here are cast from Time to Date because they are defined as being the Time type in the model for APITasks. They use the Time type because that's how the class was defining them in Go. Should I update the type in evergreen.js or stick to using this cast?

### DIFF
--- a/src/components/task/TaskPanel.tsx
+++ b/src/components/task/TaskPanel.tsx
@@ -1,11 +1,6 @@
 import { Card, CardActions, CardContent, Collapse, ExpansionPanelDetails, ExpansionPanelSummary, Grid, Link, Typography } from '@material-ui/core';
 import MuiExpansionPanel from '@material-ui/core/ExpansionPanel';
-import CalendarIcon from '@material-ui/icons/DateRangeOutlined';
-import ComputerIcon from '@material-ui/icons/DesktopMacOutlined';
-import ExpandLessIcon from '@material-ui/icons/ExpandLess';
-import ExpandMoreIcon from '@material-ui/icons/ExpandMore';
-import HourglassIcon from '@material-ui/icons/HourglassEmptyOutlined';
-import ClockIcon from '@material-ui/icons/ScheduleOutlined';
+import * as Icon from '@material-ui/icons';
 import { withStyles } from '@material-ui/styles';
 import { APITask, APITest, ConvertToAPITests } from 'evergreen.js/lib/models';
 import * as moment from 'moment';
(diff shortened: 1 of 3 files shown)


src/components/task/TaskPanel.tsx
```diff
@@ -1,9 +1,14 @@
-import { Card, CardActions, CardContent, Collapse, ExpansionPanelDetails, ExpansionPanelSummary, Grid, Typography } from '@material-ui/core';
+import { Card, CardActions, CardContent, Collapse, ExpansionPanelDetails, ExpansionPanelSummary, Grid, Link, Typography } from '@material-ui/core';
 import MuiExpansionPanel from '@material-ui/core/ExpansionPanel';
+import CalendarIcon from '@material-ui/icons/DateRangeOutlined';
```
john-m-liu: I think the icon imports can be consolidated. Can you try doing `import * as icons from '@material-ui/icons';` and then when you need to access them, do `icons.ExpandLess`

src/components/task/TaskPanel.tsx
```diff
@@ -84,6 +89,33 @@ export class TaskPanel extends React.Component<Props, State> {
 
   public render() {
 
+    const TaskDetails = () => (
+      <div>
+        <Typography>
+          <ClockIcon className="task-detail-icon" />
+          {this.stringifyNanoseconds(this.props.task.time_taken_ms * 1e6, true, true)}
```
john-m-liu: Instead of doing this, can you make the function take milliseconds?

src/components/task/TaskPanel.tsx
```diff
@@ -235,6 +274,42 @@ export class TaskPanel extends React.Component<Props, State> {
         break;
     }
   }
+
+  // stringifyNanoseconds takes an integer count of nanoseconds and
+  // returns it formatted as a human readable string, like "1h32m40s"
+  // If skipDayMax is true, then durations longer than 1 day will be represented
+  // in hours. Otherwise, they will be displayed as '>=1 day'
+  private stringifyNanoseconds = (input: any, skipDayMax: boolean, skipSecMax: boolean) => {
```
john-m-liu: Input can be one of 2 types can't it? You can specify this with a union type

src/components/task/TaskPanel.tsx
```diff
@@ -235,6 +274,42 @@ export class TaskPanel extends React.Component<Props, State> {
...
+    const NS_PER_MS = 1000 * 1000; // 10^6
+    const NS_PER_SEC = NS_PER_MS * 1000
+    const NS_PER_MINUTE = NS_PER_SEC * 60;
+    const NS_PER_HOUR = NS_PER_MINUTE * 60;
+
+
+    if (input === 0) {
```
john-m-liu: I think we can clean up this entire block by using a switch case

src/components/task/TaskPanel.tsx
dominoweir: Fixed

src/components/task/TaskPanel.tsx
dominoweir: Fixed

src/components/task/TaskPanel.tsx
```diff
@@ -235,6 +269,44 @@ export class TaskPanel extends React.Component<Props, State> {
         break;
     }
   }
+
+  // stringifyNanoseconds takes an integer count of nanoseconds and
```
john-m-liu: Forgot to change this comment. Also could you add a couple of unit tests for this function?

dominoweir: Done

dominoweir:
Okay, unit tests have been added. Take a look and let me know what you think.

PR #22, opened 2019-08-09
EVG-6552: expand first page of patches by default and create expand all option

dominoweir:
I'm still working on the styling for the expand all button (it's surprisingly hard to pin things to the bottom right of a div), but the functionality is all there. I also think it might be worth adding some code to one of the tests to confirm that the first page is in fact expanded on initial load, and another test to make sure that checking and unchecking the expand all button works. Let me know what you think.

### DIFF
--- a/src/components/patch/PatchContainer.tsx
+++ b/src/components/patch/PatchContainer.tsx
@@ -1,4 +1,4 @@
-import { Checkbox, FormControl, FormControlLabel, FormLabel, Grid, Input, InputBase, InputLabel, ListItemText, MenuItem, Paper, Select, Typography } from '@material-ui/core';
+import { Button, Checkbox, FormControl, FormLabel, Grid, Input, InputBase, InputLabel, ListItemText, MenuItem, Paper, Select, Typography } from '@material-ui/core';
 import SearchIcon from '@material-ui/icons/Search';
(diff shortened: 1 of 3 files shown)


src/components/patch/PatchContainer.tsx
```diff
@@ -83,45 +85,62 @@ export class PatchContainer extends React.Component<Props, State> {
...
+                  ))}
+                </Select>
+              </FormControl>
+            </div>
+          </Grid>
+          <Grid item={true} xs={2} className="expand-container">
+            <div className="expand">
+              <FormControlLabel
+              control={
+                <Checkbox
```
john-m-liu: A checkbox seems like the wrong UI element for this. It implies that there are only 2 valid states, but actually if I check it and then collapse something, it leaves the UI ambiguous

dominoweir:
Update: it turns out I had to modify the "expanded state" unit test anyways, so I just threw in a check to make sure the first set of patches is expanded initially.

src/components/patch/PatchContainer.tsx
dominoweir: Okay, I've replaced these with expand/collapse buttons. Let me know how they look.

PR #23, opened 2019-09-12
EVG-6583: add filter on patch type

ybrill:
Use the patch alias to determine the type of a patch.


src/components/patch/PatchContainer.tsx
```diff
@@ -166,20 +189,21 @@ export class PatchContainer extends React.Component<Props, State> {
           newPatches.map(patch => {
             const status = patch.status;
             const project = patch.project;
+            const patchAlias = patch.alias;
```
john-m-liu: This part doesn't compile

ybrill:
Opened https://github.com/evergreen-ci/evergreen.js/pull/11 and https://github.com/evergreen-ci/evergreen/pull/2752 to add the alias field.

src/components/patch/PatchContainer.tsx
ybrill: Should be fixed now.

PR #24, opened 2019-09-20
EVG-6568 load less data

ybrill:
Accommodate the API changes from Evergreen to load only the data necessary for the patch page.
Goes with https://github.com/evergreen-ci/evergreen/pull/2687 and https://github.com/evergreen-ci/evergreen.js/pull/10.


ybrill:
compile fails because we have no modules in PR patches ☹️

PR #25, opened 2019-10-30
EVG-6830: fix patches filter functionality

tgrander:
**JIRA ticket:** https://jira.mongodb.org/browse/EVG-6830

## Summary
Filtering on patches page was broken; the displayed patches were not being filtered based on the selected filters. This was due to multiple versions of the patches data being stored in state to represent all patches, filtered patches, and expanded patches. Having three different sources of truth for the state of the patches caused the state used to display the patches components to become out-of-sync. The changes in this PR remove the `visiblePatches` field from state. Filtering is now accomplished by filtering the patches on every re-render from `setState` based on the selected filters. Using this approach assures the the patches displayed in the UI are always in sync with the patches in state by having one source of truth for the patches state, which is the `allPatches` field in state. 

## Tests
**Things that could break as a result of this code:**
- the filtering functionality for patches
- the Spruce app

**I assured that these things did not break by:**
- manually testing the filter functionality, including selecting filters as well as search
- adding 2 integration tests that test filtering and search as the software is intended to be used by a real user 
- ran patch on committed changes

### DIFF
--- a/src/components/patch/PatchContainer.test.tsx
+++ b/src/components/patch/PatchContainer.test.tsx
@@ -1,22 +1,28 @@
-import { ExpansionPanel, InputBase, Select } from '@material-ui/core';
+import { ExpansionPanel, InputBase, Select } from "@material-ui/core";
 import * as enzyme from "enzyme";
-import * as moment from 'moment';
+import * as moment from "moment";
 import * as React from "react";
-import * as InfiniteScroll from 'react-infinite-scroller';
+import * as InfiniteScroll from "react-infinite-scroller";
 import * as rest from "../../rest/interface";
-import { Variant } from '../variant/Variant';
-import { Patch, PatchProps } from './Patch';
+import { Variant } from "../variant/Variant";
+import { Patch, PatchProps } from "./Patch";
 import { PatchContainer } from "./PatchContainer";
 
 describe("PatchContainer", () => {
-  const wrapper = enzyme.mount(<PatchContainer client={rest.EvergreenClient("", "", "", "", true)} username={"admin"} onFinishStateUpdate={null} />);
+  const wrapper = enzyme.mount(
+    <PatchContainer
+      client={rest.EvergreenClient("", "", "", "", true)}
+      username={"admin"}
+      onFinishStateUpdate={null}
+    />
+  );
 
   const infiniteScroll = wrapper.find(InfiniteScroll);
   infiniteScroll.prop("loadMore")(0);
 
   const checkExpandedState = jest.fn(() => {
     wrapper.update();
-    expect(wrapper.state("expandedPatches")).toEqual( {
+    expect(wrapper.state("expandedPatches")).toEqual({
       "5d66d4aa97b1d3794298a82f": 1,
       "5d66f0fd9dbe32552f1c9166": 1,
       "5d66f2e89dbe32552f1c91f5": 1,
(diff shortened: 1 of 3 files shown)


src/components/patch/PatchContainer.test.tsx
```diff
@@ -97,58 +96,42 @@ describe("PatchContainer", () => {
...
-    });
-    for (const versionId of expectedResults) {
-      expect(visibleIds).toContain(versionId);
-    }
-    for (const versionId of notInResults) {
-      expect(visibleIds).not.toContain(versionId);
-    }
-  })
+    beforeEach(() => {
+      otherWrapper = enzyme.mount(<PatchContainer client={rest.EvergreenClient("", "", "", "", true)} username={"admin"} onFinishStateUpdate={null} />);
```
john-m-liu: It's not clear to me why we need two of these components. If you're worried about persistent state between tests, can we remount it between the test cases above?

src/components/patch/PatchContainer.test.tsx
```diff
@@ -1,21 +1,20 @@
...
+import { Patch, PatchProps } from './Patch';
 import { PatchContainer } from "./PatchContainer";
 
 describe("PatchContainer", () => {
-
   const wrapper = enzyme.mount(<PatchContainer client={rest.EvergreenClient("", "", "", "", true)} username={"admin"} onFinishStateUpdate={null} />);
+
   const infiniteScroll = wrapper.find(InfiniteScroll);
-  infiniteScroll.prop("loadMore")(0);
+    infiniteScroll.prop("loadMore")(0);
```
john-m-liu: nit: was there a reason you made this whitespace change? I think it was fine before

src/components/patch/PatchContainer.test.tsx
```diff
@@ -97,58 +96,42 @@ describe("PatchContainer", () => {
...
-    }
-    for (const versionId of notInResults) {
-      expect(visibleIds).not.toContain(versionId);
-    }
-  })
+    beforeEach(() => {
+      otherWrapper = enzyme.mount(<PatchContainer client={rest.EvergreenClient("", "", "", "", true)} username={"admin"} onFinishStateUpdate={null} />);
+    })
+    afterEach(() => {
+      otherWrapper.unmount()
```
john-m-liu: nit: we should be consistent about whether we're using semicolons or not. I don't have a strong opinion either way, but the existing code does have semicolons at the end of lines

src/components/patch/PatchContainer.test.tsx
tgrander: undid these changes but forget to push the commit

src/components/patch/PatchContainer.test.tsx
tgrander: tldr: remounting the wrapper between test cases is one option. however these tests should be refactored and until they are, i think encapsulating the `beforeEach` and `afterEach` code to the filtering tests that require them is best. that way, the code required for these tests are clearly bound to the scope of the tests. 

ideally, at the top of the file, wrapper would be mounted and unmounted before and after each test using the jest `beforeEach` and `afterEach` to ensure that each test runs against a "clean" wrapper. the way the tests are currently written does not allow this to happen because the code [here](https://github.com/evergreen-ci/spruce/pull/25/files#diff-9adefdedf71f173747e15c1a2dae3b0cL15) is not encapsulated to a test scope and therefore throws an undefined error. tldr: these tests should be refactored.

src/components/patch/PatchContainer.test.tsx
tgrander: i'm going to use Prettier as an code formatter, which will add the semicolons to the file automatically. however it will also format the entire document, which will cause this PR to look like it has way more changes than it does. if it's alright with you, i'll make a subsequent PR right after this with all the Prettier reformatting that will add semicolons everywhere.

john-m-liu: Sounds good to me and I'll install it as well. You can push a separate commit that's just the prettyfication so that the diff is still easy to read

PR #27, opened 2019-11-15
Refactor App + Use Context for global state

tgrander:
**JIRA ticket:** https://jira.mongodb.org/browse/EVG-6921
This ticket applies to number 7 on the list below

## Summary

First of a series of refactors to the Spruce app in an effort to make it more modular, easy to read and follow best practices. This PR primarily refactors the `<App/>` component so that it serves more as an index for the rest of the application. I grouped context changes into this PR to achieve the refactors made to `<App/>`.

Changes made in this PR:
1. Separate `<Navbar/>` into its own component
2. Remove all business logic from `<App/>` so that it acts more like an index for the application as a whole
3. Make the dropdown for accessing config uploads into its own component
4. Make `<PluginsMenu />` into its own component 
5. Create a `<ContextProvider/>` component that wraps the entire application to provide context to any child in the application
6. Elevate the api client state into context so that it is accessible by any component
7. Elevate username state into context


src/context/apiClient.tsx
```diff
@@ -0,0 +1,66 @@
...
+    tryLoadConfig: () => void;
+    updateConfig: (configObj: ClientConfig) => void;
+  };
+}
+
+const ApiClientContext = createContext<ApiClientContextType | null>(null);
+
+const ApiClientProvider: React.FC = ({ children }) => {
+  const [apiClient, setApiClient] = useState(
+    rest.EvergreenClient("", "", "", "")
```
john-m-liu: You need to pass in the values from the config here

src/context/apiClient.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { ClientConfig, IsValidConfig } from "../models/client_config";
+
+const { createContext, useState } = React;
+
+const configPath = "/config.json";
+
+interface ApiClientContextType {
+  apiClient: rest.Evergreen;
+  actions: {
+    tryLoadConfig: () => void;
```
john-m-liu: This isn't called from anywhere as far as I can tell, but it should be one of the first things that happens

src/components/app/App.tsx
```diff
@@ -1,30 +1,29 @@
...
-  Menu,
-  MenuItem,
-  Toolbar,
-  Typography
-} from "@material-ui/core";
-import * as MenuIcon from "@material-ui/icons/Menu";
+
+// Bugsnag docs recommend instantiating the BugsnagClient as soon as possible and before the rest of the module's imports
+const bugsnagClient = bugsnag({
+  apiKey: "2d77ffd5445ab337cd3c996a2400438a",
```
john-m-liu: The key should be added as a field to the config and read from there

src/components/app/App.tsx
```diff
@@ -1,30 +1,29 @@
...
+  notifyReleaseStages: ["production"]
```
john-m-liu: Should this vary depend on whether we're in dev mode? Though I don't think we currently have different release stages defined, so this may be a moot point

src/context/apiClient.tsx
tgrander: the value passed to `useState` represents the default state, which is usually null or in this case can be `rest.EvergreenClient` called with empty strings

src/context/apiClient.tsx
tgrander: good point--will update

src/components/app/App.tsx
tgrander: the reason i did it this way is because bugsnag recommends instantiating the Client as soon as possible, i.e. before component mounts and therefore before a request would be made to fetch the config details. What is the significance of fetching the api key from the config file as opposed to hard coding it?

src/components/app/App.tsx
tgrander: what `notifyReleaseStages` does is only log things to bugsnag if `process.env.NODE_ENV` is equal to `production`. When CRA builds I believe it automatically sets NODE_ENV to `production`...but i could be wrong. The code above will for sure work as expected as soon as I use `env-cmd` to set environment. In the meantime i wanted to make sure that no reporting to bugsnag was being done in dev, test or staging environment.

src/context/apiClient.tsx
tgrander: updated that so that `tryLoadConfig` is called when the `ApiClientProvider` mounts. That way all consumers of the api client context will be updated when apiClient is ready.

src/components/app/App.tsx
john-m-liu: Since we're making an open source product, we can't be hardcoding things specific to us. I think it's fine to load this after fetching the config, because if there's a problem with that the entire app is unusable anyway

tgrander: Also, this hardcoded api key will be replaced by `process.env.BUGSNAP_API_KEY` once I use `env-cmd` to set env

tgrander: gotcha! updating now.

tgrander: api keys like this should be accessed through environment variables, which is not possible with the way things currently are, i.e. fetches and uses the config file. since the api key is public and i will be removing the hardcoded value in my next PR to use `env-cmd` to set environment variables, is it possible to leave as is for now?

john-m-liu: Let's leave it blank for now and not deploy until both are in. Even though it's not a secret I'd like to avoid committing that string since there might be web crawlers looking for long hex strings

tgrander: Sounds good to me. Unfortunately tho can't leave it blank because bugsnag complains.

tgrander: what I'll do is replace the bugsnag error boundary with a React implementation fo ErrorBoundary until the `env-cmd` ticket is done

PR #29, opened 2019-11-20
EVG-6595: render project patches if specified in pathname

SupaJoon:
https://jira.mongodb.org/browse/EVG-6595

If the project name does not exist, then fetch the current user's patches.


src/components/patch/PatchContainer.tsx
```diff
@@ -230,6 +230,11 @@ export class PatchContainer extends React.Component<Props, State> {
           username = urlParams.get("user");
         }
       }
+      const pathname = window.location.hash.split("?")[0].split('/')
+      let projectName = ''
+      if(pathname.length === 4 && pathname[2] === 'project') {
+        projectName = pathname[3]
+      }
```
tgrander: might be better off using a library like [query-string](https://www.npmjs.com/package/query-string). Otherwise this function should be a util and unit tested. my vote is for the query-string approach.

tgrander:
have you used [prettier](https://prettier.io/) before? it's a vs code extension we use to format our code

src/components/patch/PatchContainer.tsx
```diff
@@ -230,6 +230,11 @@ export class PatchContainer extends React.Component<Props, State> {
           username = urlParams.get("user");
         }
       }
+      const pathname = window.location.hash.split("?")[0].split('/')
+      let projectName = ''
+      if(pathname.length === 4 && pathname[2] === 'project') {
```
john-m-liu: We didn't discuss how we wanted to take in parameters, but I think if the current URL is something like `https://spruce.mongodb.com/#/patches?user=john.liu` for a user's patches, `https://spruce.mongodb.com/#/patches?project=spruce` should be for a project's patches. It seems like you're trying to accommodate both query params here, but I think that's actually invalid input

src/components/patch/PatchContainer.tsx
```diff
@@ -55,9 +57,15 @@ interface State {
   selectedPatchTypes: string[];
 }
 
+interface paramsShape {
```
tgrander: interfaces are typically ThisCase. Is this interface describing the params passed from react router? a more descriptive variable name could help me as the reader determine what these params are for.

src/components/patch/PatchContainer.tsx
```diff
@@ -80,6 +88,17 @@ export class PatchContainer extends React.Component<Props, State> {
     };
   }
 
+  public componentDidUpdate(prevProps: Props) {
```
tgrander: i think i understand why you used `componentDidUpdate` -- so that any changes to the url are captured and responded to? however i believe how the app currently works is that a url must be entered that either has the `USER_PAGE` or `PROJECT_PAGE` params. According to the [docs](https://reactjs.org/docs/react-component.html#componentdidupdate), `componentDidUpdate` is not called after initial render, which is when this logic would occur. therefore we want to perform this logic on `componentDidMount`. 

Am i missing some details or am I right by thinking that nothing in the app can toggle the params of the url that would require performing login on `didUpdate`. Entering a new URL will always cause a rerender of the entire app; therefore doing this logic in `componentDidMount` is sufficient.

SupaJoon: good point, we could change the componentDidUpdate to componentDidMount. You're right that we cannot toggle the params from the app and right now the componentDidUpdate is unnecessary.

src/components/patch/PatchContainer.tsx
SupaJoon: thanks!

src/components/patch/PatchContainer.tsx
SupaJoon: I adjusted the routes `patches/project/{projectName}` and `patches/user/{username}`. If the username is empty, the page will fetch the currently logged in user's patches.

src/components/app/App.tsx
```diff
@@ -42,14 +42,16 @@ const theme = createMuiTheme({
 
 // These wrapper components are temporary components to pass their children the value of contexts using the useContext hook
 // TODO: refactor PatchContainer and BuildView to be functional components that consume their respective contexts
-const Patches = () => {
+const Patches = (props: any) => {
+  const { params } = props.match;
```
john-m-liu: I'm confused as to what this is used for. Would it be possible to give props a concrete type?

SupaJoon:
evergreen retry

src/components/patch/PatchContainer.tsx
```diff
@@ -55,9 +57,15 @@ interface State {
   selectedPatchTypes: string[];
 }
 
+interface RouterParamsShape {
```
tgrander: nit: you do not need to include 'shape' in interface name

package.json
```diff
@@ -50,7 +50,7 @@
     "eslint-plugin-import": "2.14.0",
     "eslint-plugin-jsx-a11y": "6.1.1",
     "eslint-plugin-react": "7.11.1",
-    "evergreen.js": "^1.2.1",
+    "evergreen.js": "^1.2.2",
```
john-m-liu: This will change the checksum of the package, so be sure to npm install and commit the changes to package-lock.json

PR #34, opened 2019-12-31
EVG-7126: Mock GraphQL server 

SupaJoon:
These code changes allow us to connect to or mock a GQL server. If the environment is development or test, the GQL server will be mocked and a GQL schema needs to be obtained to initialize the GQL client. The schema is obtained through one of two ways depending which environment variables are provided. 

1. Provide an absolute path to a file containing a GQL schema in `-schemaPath` for the npm start script (this will set the `REACT_APP_SCHEMA_STRING` environment variable to the contents of the file)
2. Provide the URL to a live server with introspection enabled in the `REACT_APP_GQL_URL` environment variable. 

If a GQL mock server cannot be initialized because of an invalid schema or GQL server url, the App will still render.


src/index.tsx
```diff
@@ -1,8 +1,64 @@
...
+import { isDevelopment, isTest, getSchemaString, getGQLUrl } from "./utils";
+import { printSchema } from "graphql/utilities/schemaPrinter";
+import { SchemaLink } from "apollo-link-schema";
 import Evergreen from "./components/app/App";
 import registerServiceWorker from "./registerServiceWorker";
-import "./styles.css";
+import resolvers from "./gql/resolvers";
+
+async function getClientLink(): Promise<HttpLink | SchemaLink | null> {
+  const gqlURI: string = getGQLUrl();
```
john-m-liu: you can avoid defining types in situations like this where it's inferred from the function

src/index.tsx
```diff
@@ -1,8 +1,64 @@
...
+import { introspectSchema, makeExecutableSchema } from "graphql-tools";
+import { isDevelopment, isTest, getSchemaString, getGQLUrl } from "./utils";
+import { printSchema } from "graphql/utilities/schemaPrinter";
+import { SchemaLink } from "apollo-link-schema";
 import Evergreen from "./components/app/App";
 import registerServiceWorker from "./registerServiceWorker";
-import "./styles.css";
+import resolvers from "./gql/resolvers";
+
+async function getClientLink(): Promise<HttpLink | SchemaLink | null> {
```
john-m-liu: Should the code you added in this file be moved to either one of the components or a utility function in the gql folder? I thought we were avoiding putting code in index.tsx, though I could be misremembering

tsconfig.json
```diff
@@ -8,6 +8,7 @@
     "lib": ["es6", "dom"],
     "sourceMap": true,
     "allowJs": true,
+    "allowSyntheticDefaultImports": true,
```
tgrander: why is this necessary?

SupaJoon: ApolloConsumer default imports the React library

src/gql/GQLWrapper.tsx
```diff
@@ -0,0 +1,105 @@
...
+      });
+      setClient(gqlClient);
+    }
+    getAndSetClient();
+  }, []);
+
+  return client ? (
+    <ApolloProvider client={client}>{children}</ApolloProvider>
+  ) : (
+    <></>
```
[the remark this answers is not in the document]
SupaJoon: @tgrander The client may not be set for a few moments during dev mode before introspection finishes. This won't cause the app to crash because we are not rendering children here. Another possibility is to initiate the useState with a plain client. What do you think?

src/gql/GQLWrapper.tsx
```diff
@@ -0,0 +1,105 @@
...
+    let executableSchema: GraphQLSchema | null = null;
+    try {
+      executableSchema = makeExecutableSchema({
+        typeDefs: schemaString
+          ? schemaString
+          : printSchema(await introspectSchema(httpLink)),
+        resolvers
+      });
+    } catch (e) {
+      console.log(
```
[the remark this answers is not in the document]
SupaJoon: @tgrander This catch block can only be entered during dev mode and the message is intended for the developer only. The message means that one of two variables should be set if the developer wants to dev against a mock server and serves as guidance if the developer is not seeing what they expect. 
It may be possible that this is what the developer intended because they wanted to see the outcome of failed requests or view the user experience when the App cannot connect to GQL. 
Does this change your mind about logging to the console? Otherwise I can remove the message.

src/gql/GQLWrapper.tsx
```diff
@@ -0,0 +1,105 @@
...
+        "Unable to create mock server. Provide a valid value for REACT_APP_GQL_URL or REACT_APP_SCHEMA_STRING environment variables.",
+        e
+      );
+      return new HttpLink();
+    }
+    return new SchemaLink({ schema: executableSchema });
+  } else {
```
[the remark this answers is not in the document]
SupaJoon: I moved the SchemaLink into the try

src/gql/GQLWrapper.tsx
```diff
@@ -0,0 +1,105 @@
...
+  isDevelopment,
+  isTest,
+  schemaString
+}: ClientLinkParams): Promise<HttpLink | SchemaLink> {
+  const httpLink: HttpLink = new HttpLink({
+    uri: gqlURL
+  });
+
+  if (isDevelopment || isTest) {
+    let executableSchema: GraphQLSchema | null = null;
```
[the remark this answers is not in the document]
SupaJoon: I moved the executableSchema into the try

src/gql/GQLClientProvider.tsx
```diff
@@ -0,0 +1,17 @@
+import * as React from "react";
```
john-m-liu: It's not clear to me what the difference is between this wrapper and the GQLWrapper. Aren't they both an apollo client wrapper around the app? Could they be combined into 1 component?

src/gql/GQLWrapper.tsx
```diff
@@ -0,0 +1,106 @@
...
+      console.warn(
+        "Unable to initiate mock server. If this was unintended, provide a valid value for the REACT_APP_GQL_URL or REACT_APP_SCHEMA_STRING environment variables."
+      );
+      return new HttpLink();
+    }
+  }
+  return httpLink;
+};
+
+const cache = new InMemoryCache();
```
john-m-liu: It seems weird to have this outside the scope of a function. I would move it into getGQLClient

src/gql/GQLWrapper.tsx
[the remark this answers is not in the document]
tgrander: i am still suspicious of putting console.logs in production even if in just dev mode. but if that's what you'd like to do then i won't stand in your way. can you change it to `console.error` though? also would `throw new Error()`be helpful to the dev or no? for example in CRA if there is an error in dev then it shows red screen of death that makes error very clear to developer which helps to resolve. this error will fail silently without dev knowing exactly what happened without opening console.

src/gql/GQLClientProvider.tsx
SupaJoon: @john-m-liu I wanted to keep the environment variable accessor function separate from GQLWrapper for when/if we decide to move GQLWrapper into evergreen.js

src/gql/GQLWrapper.tsx
tgrander: it's not weird. having it outside the scope of the function is technically more performant because it is only ever invoked once rather than on re-render

src/gql/GQLWrapper.tsx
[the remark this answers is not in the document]
tgrander: okay. returning a fragment feels odd to me but i will leave for you to decide

src/gql/GQLWrapper.tsx
john-m-liu: Do we want the same cache for each GQL client?

src/gql/GQLClientProvider.tsx
john-m-liu: Having separation makes sense, but does this wrapper actually do anything? Or to put it differently, if https://github.com/evergreen-ci/spruce/pull/34/files/67875b0400f887763004d1275e945851f8be91f0#diff-6403a7f3b1429f6cdbaef0a7240047bcR67 is a GQLWrapper component, does that change anything?

src/gql/GQLWrapper.tsx
tgrander: there's only one gql client

src/gql/GQLClientProvider.tsx
SupaJoon: we can do away with GQLClientProvider

src/gql/GQLWrapper.tsx
SupaJoon: the useEffect callback will only be called once for the lifetime of GraphQLWrapper. If for whatever reason GraphQLWrapper unmounts and remounts, it will use the same cache from before it unmounted. For now we don't expect GraphQLWrapper to unmount and remount for the lifetime of App.

john-m-liu: If that's the desired behavior then what you have right now is correct (I realize it's mostly a moot point anyway)

PR #36, opened 2020-01-07
EVG-7141: Option to connect to live GQL server when running spruce in dev mode

SupaJoon:
`npm run start:live`: makes GQL requests `localhost:8080/query` and results are not mocked

`npm run start:mock-live-schema`: obtains the schema at `localhost:8080/query` through introspection and query results are mocked 

`npm run start:mock-custom-schema`: uses the schema in `.build-dev-cmdrc.json` and query results are mocked


src/components/app/App.tsx
```diff
@@ -70,7 +70,7 @@ const App: React.FC = () => {
         isDevelopment={isDevelopment()}
         isTest={isTest()}
         schemaString={getSchemaString()}
-        credentials="include"
+        credentials={isDevelopment() || isTest() ? "" : "include"}
```
SupaJoon: This is dependent on whether we require credentials for our local gql server

scripts/start.js
```diff
@@ -22,8 +22,18 @@ if (process.env.npm_config_schemaPath) {
   } catch (e) {
     console.error("Unable to load GQL schema from provided path", e);
   }
-  process.env.REACT_APP_SCHEMA_STRING = schemaString || "";
+  if (schemaString) {
+    process.env.REACT_APP_SCHEMA_STRING = schemaString;
+  }
+}
+if (process.env.npm_config_enableGQLMockServer) {
```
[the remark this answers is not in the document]
SupaJoon: this allows the user to set `REACT_APP_SCHEMA_STRING` by setting the `-schemaPath` parameter in `npm run start` to a file path containing your custom schema string. I just pushed some commits that will set this variable using the cmdrc library instead since we plan on rolling an unejected version of CRA and changes in this file will be reset

package.json
```diff
@@ -12,6 +12,9 @@
     "lint-fix": "tslint --fix -c tslint.json 'src/**/*.ts?(x)'",
     "serve": "serve -s build",
     "start": "lessc src/styles.less src/styles.css && node scripts/start.js",
+    "start:live": "env-cmd -e liveGQL -r .build-dev-cmdrc.json npm run start",
```
tgrander: `start:live` is not clear to me. maybe `start:dev-server`?

src/components/app/App.tsx
tgrander: we do require them

PR #53, opened 2020-01-22
MCI-3108: Cypress tests run locally

tgrander:
- Install Cypress and write a sample test
- npm commands to open and run cypress

Having trouble running Cypress in Ubuntu 18.04 in the CI environment; therefore Cypress will only run locally for now. This will allow us to write e2e tests for the UI while the Ubuntu + Cypress issue is figured out. Once resolved, we can run all Cypress tests in the CI environment.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 # testing
 /coverage
 cypress/videos/
+cypress/fixtures/
 
 # production
 /build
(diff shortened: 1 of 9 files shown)


.gitignore
```diff
@@ -2,6 +2,7 @@
 
 # testing
 /coverage
+cypress/videos/
```
john-m-liu: Running cypress locally also added some files to cypress/fixtures/. I think that also should be gitignored

PR #55, opened 2020-01-22
MCI-3108: Run Cypress in Evergreen

tgrander:
- Run a local evergreen server in CI
- Start local spruce development server
- Run cypress tests


.gitignore
```diff
@@ -3,7 +3,6 @@
 # testing
 /coverage
 cypress/videos/
-cypress/fixtures/
```
tgrander: @john-m-liu fixtures is where mock data is stored and should be pushed to source control

.evergreen.yml
```diff
@@ -56,43 +140,73 @@ functions:
...
       args: [run, coverage]
 
-  attach-results:
+  run-cypress-tests:
+    command: shell.exec
+    params:
+      working_dir: spruce
+      script: |
+        Xvfb :1 -screen 0 1280x800x24 &
+        export DISPLAY=:1
```
tgrander: this is a workaround for a known bug for Cypress running on Ubuntu 18.04. Version 3.8.3 of Cypress fixes the issue.

tgrander:
> Why did some fields in package-lock lose the bundled: true? I wonder if it's worth regenerating package-lock.json to make sure everything in our package.json is reflected in package-lock.json

@SupaJoon i'm not sure but if i reset the package lock file to `new-ui-feature-branch` version then all tests immediately fail...

tgrander:
@SupaJoon I must have accidentally pushed a merge conflict into package lock in `new-ui-feature-branch`. will make a PR to fix

.evergreen.yml
```diff
@@ -4,20 +4,104 @@ ignore:
...
     params:
       directory: spruce
 
-  npm-install:
+  get-evergreen-project:
+    command: git.get_project
+    type: setup
+    params:
+      directory: spruce
+      token: ${github_token}
```
john-m-liu: This token won't be populated without the setup code in evergreen, but you don't actually need it

.evergreen.yml
```diff
@@ -4,20 +4,104 @@ ignore:
...
+      revisions:
+        evergreen: ${evergreen_rev}
+
+  setup-mongodb:
```
john-m-liu: None of these commands will work because there is no makefile. You do need these to run for a working evergreen though

.evergreen.yml
```diff
@@ -56,43 +140,73 @@ functions:
...
 
-  attach-results:
+  run-cypress-tests:
+    command: shell.exec
+    params:
+      working_dir: spruce
+      script: |
+        Xvfb :1 -screen 0 1280x800x24 &
+        export DISPLAY=:1
+        npx cypress run --record --key ${cypress_record_key}
```
john-m-liu: Is there a way to run this in non-interactive mode? Looking at the task for this PR, it's hanging and will probably time out, and I think it's because it's waiting for input

SupaJoon: I think interactive mode is set with `cypress open`. The commit this code change belonged to had reasonable output and passed tests: https://evergreen.mongodb.com/task/spruce_ubuntu1604_e2e_test_patch_10233104f2a3b4870baf72fba9a3632790dbae36_5e28a536562343334d346dcb_20_01_22_19_40_41

.evergreen.yml
tgrander: they've worked for me dozens of times. the makefile is from the cloned evergreen repo

.evergreen.yml
tgrander: Yeah it's not in interactive mode. The hanging issue must be caused by something else.

.evergreen.yml
tgrander: it's an environment variable

.evergreen.yml
tgrander: I think it had to do with the ubuntu 18 workaround because everything works fine on ubuntu 16. Seems like ubuntu 18 will cause problems until the fix that will be merged in Cypress v3.8.3. Until then we should use ubuntu 16

.evergreen.yml
john-m-liu: I had missed it being set as a project expansion, but it's still not needed unless you specifically want to clone over HTTPS (normally necessary if this is running in a docker container)

PR #57, opened 2020-01-23
EVG-7205: authentication

tgrander:
https://jira.mongodb.org/browse/EVG-7205

(this PR is only 245 lines of code without the package-lock changes)

- user can login
- user can logout
- if an unauthenticated user visits a private route, they are redirected to login
- user is redirected back to the private route they were trying to view after logging in
- user if redirected to home page by default after logging in if no prev referrer

**E2E tests are used to test the user flows described above**

### Follow up items:
- Automatically authenticate user: https://jira.mongodb.org/browse/EVG-7263
- Log errors: https://jira.mongodb.org/browse/EVG-7266
- Styled login route: https://jira.mongodb.org/browse/EVG-7267


src/pages/Login.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  return (
+    <Wrapper>
+      <label>Username</label>
+      <input
+        type="text"
+        name="username"
+        value={username}
+        onChange={e => {
+          setUsername(e.target.value);
```
tgrander: I know that putting the lambda in the `onChange` is "not performant" but this page is barebones and will be replaced once we have designs so it does not matter

SupaJoon:
I am logged out after refreshing the browser. The user should be logged in until they reset their browser or log out.

src/components/PrivateRoute/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+  component: React.FC<RouteComponentProps>;
+};
+
+export const PrivateRoute: React.FC<PrivateRouteProps> = ({
+  component: Component,
+  ...rest
+}) => {
+  const { isAuthenticated } = useAuthStateContext();
+
+  const render = props =>
```
SupaJoon: Declaring the render function for the render prop inside this function will construct a new render function when any prop to `PrivateRoute` changes which could lead to unnecessary renders. We should declare render outside of `PrivateRoute`

src/context/auth.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+interface State {
+  isAuthenticated: boolean;
+}
+
+const defaultState: State = {
+  isAuthenticated: false
+};
+
+type Action = { type: "authenticate" } | { type: "deauthenticate" };
```
SupaJoon: "authenticate" and "deauthenticate" should be consts

src/context/auth.tsx
```diff
@@ -0,0 +1,117 @@
...
+    [dispatch]
+  );
+
+  const dispatchContext: DispatchContext = {
+    login: loginHandler,
+    logout: logoutHandler
+  };
+
+  return (
+    <AuthDispatchContext.Provider value={dispatchContext}>
```
SupaJoon: What was your reasoning for having two separate contexts for AuthDispatch and AuthState? I think the information they provide is very similar and could be grouped into one context to make all auth related information simpler to access for consumers.

src/pages/Login.tsx
```diff
@@ -0,0 +1,66 @@
+import React, { useState, ChangeEvent } from "react";
+import styled from "@emotion/styled/macro";
+import { useAuthDispatchContext, useAuthStateContext } from "../context/auth";
+import { Redirect, RouteComponentProps } from "react-router-dom";
+import { Location } from "history";
+
+const getReferer = (location: Location) => {
+  if (location && location.state && "referer" in location.state) {
```
SupaJoon: referrer is missing an "r"

src/pages/Login.tsx
```diff
@@ -0,0 +1,66 @@
+import React, { useState, ChangeEvent } from "react";
+import styled from "@emotion/styled/macro";
+import { useAuthDispatchContext, useAuthStateContext } from "../context/auth";
+import { Redirect, RouteComponentProps } from "react-router-dom";
+import { Location } from "history";
+
+const getReferer = (location: Location) => {
```
SupaJoon: `referrer` should should be an optional type in `LocationState`

src/pages/Login.tsx
SupaJoon: You can refactor the body of this function to: `return location.state.referer ?? "/";`

src/utils/gql/GQLWrapper.tsx
```diff
@@ -58,13 +60,26 @@ export const getClientLink = async ({
 
 const cache = new InMemoryCache();
 
+const authLink = (logout: Logout) =>
+  onError(({ networkError }) => {
```
SupaJoon: I'm unsure the reasoning behind logging the user out if there is a 401

src/context/auth.tsx
tgrander: these are not strings in JS, they are actually types. therefore they will autocomplete as though they are consts. try it out

src/context/auth.tsx
tgrander: read this article from kent c dodds: https://kentcdodds.com/blog/how-to-use-react-context-effectively

the value of these contexts is split into state and the dispatchers that update the state. Every time the value of context changes, all consumers are re-rendered. therefore, if a component only need to consume a dispatcher from auth context, then it should not re-render when the state is updated. splitting context into state and dispatch prevents these unnecessary re-renders.

src/pages/Login.tsx
tgrander: `LocationState` is not a type here, but if you are referring to `Location` then that is a type defined by react router

src/utils/gql/GQLWrapper.tsx
tgrander: because 401 stands for unauthorized

src/components/PrivateRoute/index.tsx
tgrander: it needs the `isAuthenticated` value which is only available within the context of the component. Also react is really fast and this will not cause a problem in any capacity we need to be worried about

src/pages/Login.tsx
tgrander: i'm afraid that does not work because the state type in react router is an empty interface. therefore this annoying way of deriving its value must be done.

src/components/PrivateRoute/index.tsx
SupaJoon: I was recommending that `isAuthenticated` should be passed to the render function as a prop. By doing this we can avoid rerendering all children of the render component in certain cases with a small refactor.

src/pages/Login.tsx
SupaJoon: LocationState is a type passed to Location

src/components/PrivateRoute/index.tsx
tgrander: these micro optimzations to avoid re-renders, which will happen very seldomly given the nature and location of this component in the react tree, will have an immaterial difference in the performance of this application.

src/pages/Login.tsx
tgrander: in that case it is still a type declared by react router. i also don't see why it matters...

src/pages/Login.tsx
SupaJoon: You could do something like this: 
```
const getReferer = (location: Location<{ refer?: String }>) => {
  return location.state.refer ?? "/";
};
```

src/context/auth.tsx
john-m-liu: Shouldn't this be an enum instead though?

src/context/auth.tsx
```diff
@@ -0,0 +1,117 @@
...
+};
+
+type Action = { type: "authenticate" } | { type: "deauthenticate" };
+
+type Dispatch = (action: Action) => void;
+
+export type Logout = () => void;
+
+interface DispatchContext {
+  login: (LoginParams) => void;
```
john-m-liu: LoginParams is a param name here, but I think you meant to use it as a type. Also does it make sense to export both the Login and Logout types?

tgrander:
> Shouldn't this be an enum instead though?

@john-m-liu i've gone the enum route and it made it more difficult to use the dispatch method elsewhere. for example, using dispatch with the actions as they are types now will autocomplete anywhere. if using enums then you have to import the enums from this file. 

i've found that typing the actions like this is easiest + provides all the benefits as using enums

src/context/auth.tsx
tgrander: LoginParams is a `{}` so it would be equivalent to writing `login: ({username, password}: LoginParams) => void;`

i only export things that need to be exported--Logout type is used elsewhere but LoginParams is not

PR #58, opened 2020-01-27
EVG-7209: Set up route for Task Page

SupaJoon:
This PR sets up the route for the task page. If `tab` is empty, then the task page will set `tab` to `logs`. 

This code doesn't handle filtering/sort parameters that are relevant to the component in the table section of the design. The filtering/sort parameters will be read and set by the relevant table section component in a different PR.


src/pages/Task.tsx
```diff
@@ -0,0 +1,19 @@
+import React, { useEffect } from "react";
+import { RouteComponentProps } from "react-router-dom";
+import { useParams, useHistory } from "react-router-dom";
+
+const DEFAULT_TAB = "logs";
```
tgrander: this would be a good place to use TS enums or just a TS type. Something like:

```
enum Tabs {
  TESTS = tests
  LOGS = logs
  FILES = files
  BUILD_BARON = build-baron
}

const DEFAULT_TAB = Tabs.LOGS
```

Or...

```
type Tab = "logs" | "tests" | "files"
```
This way the value for tab can only ever be set to these values.

src/App.tsx
```diff
@@ -11,6 +11,7 @@ import {
   isTest,
   shouldEnableGQLMockServer
 } from "./utils/getEnvironmentVariables";
+import TaskPage from "pages/Task";
```
tgrander: can we do named imports only please? For components and all other modules I prefer this approach because then the import supports autocomplete + is imported as the same thing everwhere

src/pages/Task.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+const Task: React.FC<RouteComponentProps> = () => {
+  const params = useParams<{ tab?: string; taskID: string }>();
+  const history = useHistory();
+  useEffect(() => {
+    if (!params.tab) {
+      history.replace(`/task/${params.taskID}/${DEFAULT_TAB}`);
+    }
+  }, [params.tab, params.taskID, history]);
```
tgrander: prefer destructuring the values of `params` instead of using multiple dot notations. 

`const {tab, taskId} = useParams<{ tab?: string; taskID: string }>();`

PR #60, opened 2020-01-28
EVG-7208: Test Table Routing

SupaJoon:
This PR includes Test Table routing where url query params will be adjusted to default values if they do not contain valid values for `category`, `sort`, `page` and `limit`
This PR also installs `query-string` and `antd` which will be used to build the table. 

Upcoming pull requests for this ticket include: 
- Updating GQL TaskTests query variables from url query params
- Updating url query params from table buttons
- Paginating the table / infinite scroll


cypress/integration/test_table_route.js
```diff
@@ -0,0 +1,77 @@
...
+
+    it("Limit will truncate if it's a float", function() {
+      cy.visit("/task/taskID/tests?category=TEST_NAME&page=0&limit=4.5&sort=1");
+      cy.location().should(loc => {
+        expect(loc.pathname).to.equal("/task/taskID/tests");
+        expect(loc.search).to.include("category=TEST_NAME");
+        expect(loc.search).to.include("page=0");
+        expect(loc.search).to.include("limit=4");
+        expect(loc.search).to.include("sort=1");
+      });
```
tgrander: nice job testing edge cases

src/pages/task/TestTable.tsx
```diff
@@ -0,0 +1,60 @@
+import React, { useEffect } from "react";
```
tgrander: nit: can we rename component `TestsTable` since there are multiple tests

PR #61, opened 2020-01-28
EVG-7294: fix type error and remove unused file

SupaJoon:
https://jira.mongodb.org/browse/EVG-7294


SupaJoon:
@tgrander it looks like `const parsed` has type `queryString.ParsedQuery<string>`. I created a follow up ticket to make tests fail in the case of typescript error: https://jira.mongodb.org/browse/EVG-7296

PR #62, opened 2020-01-31
EVG-7208: Tests table

SupaJoon:
https://jira.mongodb.org/browse/EVG-7208
Render a task's tests inside a table. 

To see the table, navigate to to `/task/<taskID>` 
- Clicking on the table header sort icons will update URL query params which will cause sorted data to be fetched from Page 0
- Scrolling down the table will result in additional pages being fetched and displayed. 
- Sorting icons are disabled during a network request.
- If a user navigates to `/task/<taskID>/tests` without valid query params, default query params are pushed to the URL.
- The required query params are `category` and `sort`. Values for category are case insensitive and can be `test_name`, `duration` and `status`. Values for `sort` are `-1` (descending) and `1` (ascending)
  [](url)
Follow-up Tickets:
Fix bug where entries are duplicated across pages: 
https://jira.mongodb.org/browse/EVG-7313
Logs column from design:
https://jira.mongodb.org/browse/EVG-7210
Improve speed of gql req in cypress: 
https://jira.mongodb.org/browse/EVG-7319
Get Cypress to match on body of Post
https://jira.mongodb.org/browse/EVG-7318

Original Design
Screen Shot 2020-02-05 at 11 11 17 AM

What it looks like in this PR
Screen Shot 2020-02-05 at 11 10 06 AM 1


src/pages/task/TestsTableCore.tsx
```diff
@@ -0,0 +1,211 @@
...
+  });
+  const parsed = queryString.parse(search);
+  const category = (parsed[RequiredQueryParams.Category] || "")
+    .toString()
+    .toUpperCase();
+  const sort = parsed[RequiredQueryParams.Sort];
+  const prevCategory = usePrevious(category);
+  const prevSort = usePrevious(sort);
+  useEffect(() => {
+    const elements = document.querySelectorAll(
```
john-m-liu: Instead of doing this, can you style the header cell in the column definitions? This feels pretty janky

src/pages/task/TestsTableCore.tsx
```diff
@@ -0,0 +1,211 @@
...
+        (el as HTMLElement).style["pointer-events"] = "none";
+      });
+    } else {
+      elements.forEach(el => {
+        (el as HTMLElement).style["pointer-events"] = "auto";
+      });
+    }
+  }, [networkStatus]);
+
+  if ((sort !== prevSort || category !== prevCategory) && networkStatus >= 7) {
```
john-m-liu: NetworkStatus.ready instead of 7

src/utils/string/index.ts
```diff
@@ -0,0 +1,23 @@
+import { DurationInputArg1, duration } from "moment";
+
+export const msToDuration = (ms: DurationInputArg1): string => {
```
john-m-liu: Weird that moment doesn't have a built-in way to format durations. Since this is the only place we're using it currently, does it make sense to use a different time library?

src/pages/task/TestsTable.tsx
```diff
@@ -1,68 +1,54 @@
...
   Limit = "0"
 }
 
-export const TestsTable: React.FC = () => {
+export const TestsTable: React.FC<Limit> = ({ limit }) => {
   const { pathname, search } = useLocation();
   const { replace } = useHistory();
+  const [validInitialQueryParams, setValidInitialQueryParams] = useState<
+    ValidInitialQueryParams
+  >();
```
tgrander: there is no initial value being set here

src/pages/task/TestsTable.tsx
```diff
@@ -1,68 +1,54 @@
...
       parsed[RequiredQueryParams.Category] = DefaultQueryParams.Category;
       parsed[RequiredQueryParams.Sort] = DefaultQueryParams.Sort;
-      parsed[RequiredQueryParams.Page] = DefaultQueryParams.Page;
-      parsed[RequiredQueryParams.Limit] = DefaultQueryParams.Limit;
       const nextQueryParams = queryString.stringify(parsed);
       replace(`${pathname}?${nextQueryParams}`);
+    } else if (!validInitialQueryParams) {
+      setValidInitialQueryParams({
+        initialCategory: parsed[RequiredQueryParams.Category],
+        initialSort: parsed[RequiredQueryParams.Sort]
```
tgrander: this should probably be what the default value of `validInitialQueryParams` is set to. then if they're not valid they can fall back to the defaults.

src/pages/task/TestsTableCore.tsx
```diff
@@ -0,0 +1,211 @@
...
+  duration: number;
+}
+
+interface UpdateQueryArg {
+  taskTests: [TaskTests];
+}
+
+const loadMoreContent = (): JSX.Element => (
+  <div
+    style={{
```
tgrander: no inline styles please. should use emotion for all styling to be consistent

tgrander: This looks like a React element but is named as a function. The name also threw me off--when i first read it i thought it was going to be a function to load more content. But it's a loading component

src/pages/task/TestsTableCore.tsx
```diff
@@ -0,0 +1,211 @@
...
+    dataIndex: "testFile",
+    key: Categories.TestName,
+    sorter: true
+  },
+  {
+    title: "Status",
+    dataIndex: "status",
+    key: Categories.Status,
+    sorter: true,
+    render: (tag: string): JSX.Element => {
```
tgrander: i think the return type will be automatically inferred from the return value. you might not need to assert the return type in the function declaration.

src/pages/task/TestsTableCore.tsx
```diff
@@ -0,0 +1,211 @@
...
+        </span>
+      );
+    }
+  },
+  {
+    title: "Time",
+    dataIndex: "duration",
+    key: Categories.Duration,
+    sorter: true,
+    render: (text: number): string => {
```
tgrander: you don't need to declare return type here

src/pages/task/TestsTableCore.tsx
```diff
@@ -0,0 +1,211 @@
...
+      return msToDuration(text.toFixed());
+    }
+  }
+];
+const rowKey = ({ id }) => id;
+type Props = ValidInitialQueryParams & Limit;
+export const TestsTableCore: React.FC<Props> = ({
```
tgrander: this could use some whitespace

src/pages/task/TestsTableCore.tsx
tgrander: agreed. this is an unideal way of disabling the sort buttons while a query is loading

src/pages/task/types.ts
```diff
@@ -0,0 +1,26 @@
+export interface ValidInitialQueryParams {
```
tgrander: i feel like this file belongs i the root of pages because it relates to task page as a whole. putting in subdirectory of task page doesn't feel quite right

src/utils/gql/queries.ts
```diff
@@ -0,0 +1,23 @@
+import gql from "graphql-tag";
```
tgrander: queries are not a util. putting the gql code in the utils folder is something i have felt is wrong for a while now. maybe this is the time to move it to a different directory.

src/utils/hooks/index.ts
```diff
@@ -0,0 +1,9 @@
+import { useEffect, useRef } from "react";
```
tgrander: hooks are also not a util and should go in their own `hooks` directory

src/index.css
```diff
@@ -1,13 +1 @@
...
-    sans-serif;
-  -webkit-font-smoothing: antialiased;
-  -moz-osx-font-smoothing: grayscale;
-}
-
-code {
-  font-family: source-code-pro, Menlo, Monaco, Consolas, 'Courier New',
-    monospace;
-}
+@import "~antd/dist/antd.css";
```
tgrander: this imports the CSS for the entire antd library, which is 60kb. Ideally we will only import the css required by the components we use. The ant docs talk about ways around this in the [advanced guide](https://ant.design/docs/react/use-with-create-react-app#Advanced-Guides). I'm cool with this being merged for now as long as a ticket is created and prioritized next sprint.

src/utils/string/index.ts
tgrander: i second this... moment is a HUGE dependency and it does not look like we're accessing its full potential here

src/pages/task/TestsTable.tsx
SupaJoon: I actually intended the initial query params to be empty. In the current implementation, if the query params read from the URL are invalid, we replace the values in the URL , otherwise, we set `validInitialQueryParams` to what was in the the URL. The table won't render until the query params read from the URL are valid. I could provide default props if I had a third boolean useState that represents when the URL has valid query params. 
The reason I put the initial query params in useState to begin with is because I `useQuery` needs them as props for the initial fetch and also because the antd table needs to know to know the sort category and direction for its initial render.

src/index.css
SupaJoon: https://jira.mongodb.org/browse/MCI-3156

src/pages/task/TestsTableCore.tsx
SupaJoon: The column prop provides a way for us to hide the sorter buttons all together while a network request is in progress but doesn't allow a way to disable them. Hiding the buttons all together is confusing for the end user when the network request is taking a long time.
I disabled the sort buttons in the first place was because I was unsure if the data was rendering correctly in the table when there were multiple requests being made at the same time. Part of that issue is duplicate keys being returned from tests endpoint between 2 different pages which I'm debugging so there is a chance we may not need to disable the sorters.

src/hooks/index.ts
```diff
@@ -0,0 +1,9 @@
+import { useEffect, useRef } from "react";
+
+export const usePrevious = <T>(state: T): T | undefined => {
+  const ref = useRef<T>();
+  useEffect(() => {
+    ref.current = state;
+  });
```
tgrander: i believe this should have a `[]` as it's second param, otherwise it will perform this action on every re-render of its invoker

package.json
```diff
@@ -24,7 +25,10 @@
     "axios": "^0.19.2",
     "env-cmd": "^10.0.1",
     "graphql": "^14.5.8",
+    "graphql-tag": "^2.10.1",
     "graphql-tools": "^4.0.6",
+    "lodash.get": "^4.4.2",
```
tgrander: +1 on importing only the one function we need ⭐️

src/gql/GQLWrapper.tsx
```diff
@@ -11,7 +11,7 @@ import {
 import { printSchema } from "graphql/utilities/schemaPrinter";
 import { SchemaLink } from "apollo-link-schema";
 import { onError } from "apollo-link-error";
-import { useAuthDispatchContext, Logout } from "../../context/auth";
+import { useAuthDispatchContext, Logout } from "context/auth";
```
tgrander: is this an absolute import?

SupaJoon: yes

src/pages/task/TestsTableCore.tsx
```diff
@@ -0,0 +1,248 @@
...
+  if (
+    (sort !== prevSort || category !== prevCategory) &&
+    networkStatus === NetworkStatus.ready
+  ) {
+    fetchMore({
+      variables: {
+        cat: category,
+        dir: sort === Sort.Asc ? "ASC" : "DESC",
+        pageNum: 0,
+        limitNum: limit
```
tgrander: it looks to me like `limit` does not need to be passed down through multiple levels of props because its just hardcoded when `<TestsTable limit={10} />` is rendered. Is there a reason that `limit` needs to be passed as a prop?

src/pages/task/TestsTableCore.tsx
john-m-liu: I didn't realize this was the motivation behind that, but if that's the case should we have a consistent way to block input while network requests are in progress throughout the app? For example we can show a spinner which also visually and functionally disables the components below it. That would solve the problem of trying to interact while there is a network request, and also prevent a certain class of race conditions that are possible by launching another network request while one is in progress

src/pages/task/types.ts
john-m-liu: But presumably there will be many more pages in the future - spruce won't just be the task page

src/pages/task/TestsTableCore.tsx
```diff
@@ -0,0 +1,248 @@
...
+        case TestStatus.Succeeded:
+          color = "green";
+          break;
+        case TestStatus.Failed:
+          color = "red";
+          break;
+        case TestStatus.SilentlyFailed:
+          color = "volcano";
+          break;
+        case TestStatus.Skipped:
```
john-m-liu: Let's use a cooler color for this - I'm thinking a shade of blue

src/pages/task/TestsTableCore.tsx
SupaJoon: I thought this is one of the values we are most likely to adjust so I made it available as prop TestsTable so it's easier to find and fine tune.

src/pages/task/TestsTableCore.tsx
tgrander: we should use whatever color is in designs

src/pages/task/types.ts
tgrander: yes and that's fine with me. within a `types` dir we can have files for each page and then have all exported through and index file if we'd like

src/pages/task/TestsTableCore.tsx
tgrander: i don't think we need a full-on universal solution for this occurrence just yet. solving it for this use case first is sufficient imo. As more examples of where a consistent pattern occur we can then go ahead and design an implementation. We will also better understand _how_ to design something like this well when we understand more use cases.

tgrander: i'm more so surprised that the antd infinite scroll component does not have a `disable` prop for the header actions. the components in the ant lib itself usually have this functionality. Speaking of blocking input while network requests are in progress, this is how that should be accomplished.

src/pages/task/TestsTableCore.tsx
tgrander: i can't think of any reason that we would need to pass it down two levels of props. if it's being hardcoded and not passed in from state that changes then we don't need to pass the value down as props. you can leave `limit` as a prop and just give it a default value so that it does not need to be passed down in two places. 

Every prop that require to be passed to our components makes the api of the component more complex. therefore if there is no use case for this prop to be dynamic right now then we could remove it as a prop altogether and hardcode it in `TestTableCore`. Then when and if we need it we can expose it as a prop. Otherwise we just complicate the interface of our components for situations that may or may never exist.

src/pages/task/TestsTableCore.tsx
tgrander: What if the event handler you pass to `onFetch` is bound to the context of `TestsTableCore`. Then in the event handler you can do something like this: 
```
const {data, loading} = useQuery(QUERY)

const onFetch = () => {
    if (!loading) {
        ...fetch stuff...
    }
});
``

Comment (SupaJoon) [reply src/pages/task/TestsTableCore.tsx]: @tgrander That solution works to prevent network requests but the sorting button will still toggle and misrepresent the data being shown.

File: src/pages/task/types.ts (hunk 10 above)
Comment (SupaJoon) [reply src/pages/task/types.ts]: Exporting all of our interfaces/types for each page through an index file may make it more difficult to come up with simple unique names for interfaces/types for all our pages.

File: src/pages/task/TestsTableCore.tsx (hunk 1 above)
Comment (tgrander) [reply src/pages/task/TestsTableCore.tsx]: I'm going to look into this further

File: cypress/integration/test_table.ts [hunk 19]
```diff
@@ -0,0 +1,63 @@
...
+    cy.contains("Time").click();
+    cy.location().should(loc => {
+      expect(loc.pathname).to.equal(
+        "/task/mci_windows_test_agent_8a4f834ba24ddf91f93d0a96b90452e9653f4138_17_10_23_21_58_33/tests"
+      );
+      expect(loc.search).to.include("category=DURATION");
+      expect(loc.search).to.include("sort=-1");
+    });
+  });
+});
```
tgrander: there should be a test in here to assert that filters are disabled while another sort request is loading, yeah?

src/App.test.tsx
```diff
@@ -1,9 +1,14 @@
...
 import App from "App";
+import wait from "waait";
 
-it("renders without crashing", () => {
+it("renders without crashing", async () => {
   const div = document.createElement("div");
   ReactDOM.render(<App />, div);
+  await act(async () => {
+    await wait(0);
+  });
```
tgrander: what's this for?

src/gql/queries/get-task-tests.ts
```diff
@@ -0,0 +1,57 @@
+import gql from "graphql-tag";
+export const GET_TASK_TESTS = gql`
```
tgrander: nit: whitespace after imports

src/index.css
tgrander: i don't think we need this anymore after my PR where i import the ant css directly into App

src/index.tsx
```diff
@@ -1,6 +1,7 @@
 import React from "react";
 import ReactDOM from "react-dom";
 import App from "App";
+import "index.css";
```
tgrander: i replace the index.css file with a direct import of the ant css

src/pages/task/types.ts
tgrander: they don't have to be exported from an index. i still don't think the types belong in the page's subdirectoy tho

src/utils/string/index.ts
```diff
@@ -0,0 +1,23 @@
+export const msToDuration = (ms: number): string => {
+  const days = Math.floor(ms / (24 * 60 * 60 * 1000));
+  const daysMilli = ms % (24 * 60 * 60 * 1000);
+  const hours = Math.floor(daysMilli / (60 * 60 * 1000));
+  const hoursMilli = ms % (60 * 60 * 1000);
+  const minutes = Math.floor(hoursMilli / (60 * 1000));
+  const minutesMilli = ms % (60 * 1000);
+  const seconds = Math.floor(minutesMilli / 1000);
+  const milliseconds = ms % 1000;
```
john-m-liu: The math here implies ms is in microseconds, not milliseconds

SupaJoon:
> do we have a visual example of what happens in the name column when the test name is really long? according to ben's designs the column should expand to fit the longest test name

I synced with Ben about this and he said the width should be based on the average use case/name length. He is also against dynamic column widths based on name lengths as new data pages loads. 
Here is a screen shot of what the table looks like with a long name with new column width settings.

Screen Shot 2020-02-05 at 3 06 07 PM

cypress/integration/test_table.ts
```diff
@@ -0,0 +1,63 @@
...
+      cy.get("input[name=password]").type("password");
+      cy.get("button[id=login-submit]").click();
+    }
+    cy.visit("/login");
+    login();
+  });
+  it("Navigate to tests table and click on table header to adjust query params", function() {
+    cy.visit(
+      "/task/mci_windows_test_agent_8a4f834ba24ddf91f93d0a96b90452e9653f4138_17_10_23_21_58_33/tests"
+    );
```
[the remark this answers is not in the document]
SupaJoon: the table will display "No Data"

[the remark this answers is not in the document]
tgrander: is there a test for this?

cypress/integration/test_table.ts
```diff
@@ -0,0 +1,63 @@
...
+    cy.wait(15000);
```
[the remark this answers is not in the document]
SupaJoon: I just pushed a commit to polyfill fetch in cypress so we can wait on the route alias instead

cypress/integration/test_table.ts
```diff
@@ -0,0 +1,63 @@
...
+    cy.contains("Name").click();
+    cy.location().should(loc => {
+      expect(loc.pathname).to.equal(
+        "/task/mci_windows_test_agent_8a4f834ba24ddf91f93d0a96b90452e9653f4138_17_10_23_21_58_33/tests"
+      );
+      expect(loc.search).to.include("category=TEST_NAME");
+      expect(loc.search).to.include("sort=-1");
+    });
+    cy.wait(15000);
```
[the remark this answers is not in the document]
SupaJoon: i just pushed a commit to polyfill fetch in cypress so we can wait on the query route alias

src/App.test.tsx
SupaJoon: When testing, code that causes React state updates should be wrapped into act(...). In this case, we are waiting on the result of a promise that that causes changes to `GQLWrapper`.
https://github.com/evergreen-ci/spruce/blob/new-ui-feature-branch/src/utils/gql/GQLWrapper.tsx#L113

cypress/integration/test_table.ts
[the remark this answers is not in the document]
SupaJoon: not yet, good call

src/pages/task/TestsTableCore.tsx
```diff
@@ -0,0 +1,230 @@
...
+          return prev;
+        }
+        return fetchMoreResult;
+      }
+    });
+  }
+  const dataSource: [TaskTestsData] = get(data, "taskTests", []);
+  const onFetch = () => {
+    const pageNum = dataSource.length / LIMIT;
+    if (pageNum % 1 !== 0) {
```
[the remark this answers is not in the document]
SupaJoon: limit is the number of entries that are sent back with each request, not the total number entries available. I should probably adjust the query to return the total number of items possible and do that check here instead. 
I do modulus division to make sure I don't send non-integers as the page num.

src/pages/task/TestsTableCore.tsx
```diff
@@ -0,0 +1,230 @@
...
+      return;
+    }
+    fetchMore({
+      variables: {
+        pageNum,
```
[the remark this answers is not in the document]
SupaJoon: I do modulus division to make sure I don't send non-integers as the page num.

src/utils/string/index.ts
SupaJoon: there are a thousand milliseconds in a second, so I see how many milliseconds are leftover after we account for seconds. I think this works.

cypress/integration/test_table.js
```diff
@@ -0,0 +1,81 @@
...
+    cy.contains("Name").click();
+    cy.once;
+    cy.once("fail", err => {
+      expect(err.message).to.include(
+        "'pointer-events: none' prevents user mouse interaction."
+      );
+    });
+  });
+
+  it("Navigate to tests table and click on table header to adjust query params", () => {
```
tgrander: "Adjusts query params when table headers are clicked"

cypress/support/commands.js
```diff
@@ -23,3 +23,18 @@
...
+      username: "admin",
+      password: "password"
+    }
+  ) => {
+    cy.visit("/login");
+    cy.get("input[name=username]").type(username);
+    cy.get("input[name=password]").type(password);
+    cy.get("button[id=login-submit]").click();
+  }
+);
```
tgrander: :)))))

src/pages/task/TestsTableCore.tsx
```diff
@@ -0,0 +1,234 @@
...
+import styled from "@emotion/styled/macro";
+import {
+  RequiredQueryParams,
+  SortQueryParam,
+  ValidInitialQueryParams
+} from "pages/types/task";
+import get from "lodash.get";
+import queryString from "query-string";
+import { NetworkStatus } from "apollo-client";
+const LIMIT = 10;
```
tgrander: nit: whitespace between imports please

src/utils/string/index.ts
john-m-liu: You're using the wrong variable then - ms is the input (which I'm assuming to be milliseconds). I'm looking at a test which took 25 secs, but shows as 25ms in this table

src/pages/task/TestsTableCore.tsx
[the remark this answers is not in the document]
john-m-liu: I must be missing something then - I assume datasource.length is the # of results from a single page of the query, which will usually be the limit, and you're checking "are there more results" with this line. Is that not the case?

src/pages/task/TestsTableCore.tsx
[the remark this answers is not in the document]
john-m-liu: pageNum is the result of normal division. js doesn't have an integer division operator. Or am I missing what you're trying to do here?

cypress/integration/test_table.js
```diff
@@ -0,0 +1,81 @@
+/// <reference types="Cypress" />
+
+describe("tests table", function() {
+  this.beforeEach(() => {
+    cy.server();
+    cy.login();
```
tgrander: this makes it so that cypress logs in before each of the `it` statements below. looks to me like logging in once before all tests using `beforeAll` would suffice.

tgrander: tried writing some tests myself and it looks like because of how auth currently works that logging in before every approach is required for now. my first ticket this sprint is to automatically authenticate a user. once that is done we should only need to authenticate once per test suite.

src/pages/task/TestsTableCore.tsx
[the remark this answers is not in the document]
SupaJoon: datasource consists of all of the entries we have requested so far through pagination. The limit is the maximum number of entries we can request at once. 
I calculate the next page index by dividing how many entries we have requested so far by the maximum number of entries in a page. If that value is not an integer, that means the number of items requested was less than an entire page and that there are no more entries to request.

cypress/integration/test_table.js
SupaJoon: @tgrander do you still request changes here?

src/pages/task/TestsTableCore.tsx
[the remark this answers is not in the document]
SupaJoon: the function will returns before fetchMore executes if pageNum is not a whole number

src/utils/string/index.ts
SupaJoon: it looks like I'm using seconds as ms. I made a code change in spruce to convert seconds to ms before giving the number to `msToDuration`

src/utils/string/index.ts
```diff
@@ -0,0 +1,23 @@
...
+  if (days > 1) {
+    return `${Math.trunc(days)}d ${hours}h ${minutes}m ${seconds}s`;
```
john-m-liu: I think this is excessive - nothing should really take more than a day, and if it does it'll be a small number of days. I think display a # of hours > 24 in that scenario would be easier to comprehend

src/utils/string/index.ts
```diff
@@ -0,0 +1,23 @@
...
+  }
+  if (hours > 0) {
+    return `${hours}h ${minutes}m ${seconds}s`;
```
john-m-liu: I think seconds precision is unneeded here

src/utils/string/index.ts
```diff
@@ -0,0 +1,23 @@
...
+  }
+  if (minutes > 0) {
+    return `${minutes}m ${seconds}s`;
+  }
+  if (seconds > 0) {
+    return `${seconds}s ${milliseconds}ms`;
```
john-m-liu: I would express this as a decimal # of seconds

PR #63, opened 2020-02-04
EVG-7200: Select component with search input

tgrander:
### User stories covered in this PR:
- [x] When I click on the dropdown menu button, a Select should appear and cursor should be auto focused on input field. 
- [x] I should be able to use my UP and DOWN arrow keys to move through menu items

To see the rest of the user stories for the projects dropdown see the New UI Google Doc [here](https://docs.google.com/document/d/1rxWptM-e3w8t4254KuW-jJvNCdcsU2MXX_iQViF9qlg/edit?ts=5e398dbc#heading=h.m8k76qa75acs)

### The designs
Screen Shot 2020-01-13 at 9 11 11 AM

### How it looks using ANT
Screen Shot 2020-02-04 at 12 57 26 PM

**The colors of the border will all be changed to fir Mongo. Doing so tho requires modifying webpack config. How we are going to accomplish this with CRA is TBD.**

**Ben signs off on this UI**

### DIFF
--- a/src/components/Navbar/index.tsx
+++ b/src/components/Navbar/index.tsx
@@ -4,6 +4,9 @@ import {
   useAuthDispatchContext,
   useAuthStateContext
 } from "../../context/auth";
+import Select from "antd/es/select";
+
+const { Option, OptGroup } = Select;
 
 export const Navbar: React.FC = () => {
   const { logout } = useAuthDispatchContext();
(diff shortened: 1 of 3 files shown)


src/components/Navbar/index.tsx
```diff
@@ -12,6 +15,25 @@ export const Navbar: React.FC = () => {
   return (
     <Wrapper>
       <InnerWrapper>
+        {false && (
```
tgrander: this will make it so that projects dropdown is not rendered until it's completely ready. Simply set to `true` if you wish to interact with it.

tgrander:
@SupaJoon the projects when it was in favorites had the same `value` attribute as when it is in its group. This caused the selection to get weird. Favorites now have unique ids, should be fixed now.

src/components/Navbar/index.tsx
tgrander: i got rid of this. 1) because TS error and 2) because we don't really need it

PR #64, opened 2020-02-04
EVG-7200: query for projects data

tgrander:
### User stories covered in this PR:
- [x] The dropdown should show grouped projects
- [x] I am able to search projects by name in dropdown menu

### Technical Implementation
- [x] Use OptionGroup from Select ANT component to create groups. Query projects from gql and render them in Select.

### Screenshot
Screen Shot 2020-02-04 at 7 19 17 PM

### DIFF
--- a/src/components/Navbar/index.tsx
+++ b/src/components/Navbar/index.tsx
@@ -14,16 +14,8 @@ import {
 
 const { Option, OptGroup } = Select;
 
-const renderProjectOption = (isFavorite: boolean = false) => ({
-  identifier,
-  displayName
-}: Project) => (
-  // two Options cannot have the same value attribute or it breaks ability to scroll with keyboard.
-  // therefore "-favorite" is appended to the end of a favorite's value prop
-  <Option
-    key={identifier}
-    value={isFavorite ? `${identifier}-favorite` : identifier}
-  >
+const renderProjectOption = ({ identifier, displayName }: Project) => (
+  <Option key={identifier} value={identifier}>
     {displayName}
   </Option>
 );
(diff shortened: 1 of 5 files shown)


src/components/Navbar/index.tsx
```diff
@@ -5,32 +5,56 @@ import {
...
 
+const renderProjectOption = (isFavorite: boolean = false) => ({
+  identifier,
+  displayName
+}: Project) => (
+  // two Options cannot have the same value attribute or it breaks ability to scroll with keyboard.
+  // therefore "-favorite" is appended to the end of a favorite's value prop
+  <Option
+    key={identifier}
+    value={isFavorite ? `${identifier}-favorite` : identifier}
```
john-m-liu: You could run into the same problem on the off-chance someone names a project xxx-favorite. You should have an escape sequence or prevent this scenario in the first place (below)

src/components/Navbar/index.tsx
```diff
@@ -5,32 +5,56 @@ import {
...
-            <Option value="spruce">Spruce</Option>
-            <Option value="cloud">Cloud</Option>
-            <Option value="build">Build</Option>
-          </OptGroup>
+          {data && data.projects.favorites.length > 0 && (
+            <OptGroup label="Favorites">
+              {data.projects.favorites.map(renderProjectOption(true))}
+            </OptGroup>
+          )}
+          {data &&
```
john-m-liu: I believe this will duplicate the favorite projects won't it? I think having them in there twice would be more confusing than just sorting favorites to the top

SupaJoon:
The request to get the projects is being made only one time and it happens before the user is able to log in. This results in a 405 from the gql server. In order to populate the dropdown, we should either retry the query if it fails or even better, avoid executing the query until after the user has logged in.

tgrander:
> The request to get the projects is being made only one time and it happens before the user is able to log in. This results in a 405 from the gql server. In order to populate the dropdown, we should either retry the query if it fails or even better, avoid executing the query until after the user has logged in.

The dropdown won't be rendered unless the user is authenticated. That would be part of a different user story and a different PR. These PRs are for the component alone, as if it were rendered in isolation.

src/components/Navbar/index.tsx
tgrander: Yes, projects that are present in favorites are still present in in the rest of the dropdown. This seems like the right UX to me because favoriting something does not mean removing it from the group to which it belongs. But I agree filtering favorites out of the projects would make this feature simpler and prevent problems caused by duplication.

PR #65, opened 2020-02-06
ProjectSelect is its own component

tgrander:
No code changes. Project select is its own component.


src/components/ProjectSelect/index.tsx
```diff
@@ -0,0 +1,48 @@
+import React from "react";
+import { Select, Icon } from "antd";
```
john-m-liu: Icon isn't used

tgrander: thanks

PR #67, opened 2020-02-07
EVG-7270: Breadcrumb Components

tgrander:
I created two Breadcrumb components -- `TaskBreadcrumb` and `PatchBreadcrumb`. Originally in my tech spec I planned to create one `Breadcrumb` component that used the url to determine what to render. This posed a number of challenges when trying to implement, such as using not being able to use React Router hooks because the Breadcrumb component was not within any specific route. 

Two breadcrumb components has proven to be much easier to understand because the code is straight forward and requires no logic. It receives props from its parent rather than having to make its own queries. It's also much easier to test this way and easier to manage loading and error states.

### On the patch route
Screen Shot 2020-02-07 at 12 07 16 AM
"Patch" will be updated to the displayName of the patch. See follow items below.

### On the task route
Screen Shot 2020-02-07 at 12 06 51 AM
The "Patch" crumb will be updated to show the displayName of the patch--see "Add patch displayName to task query: https://jira.mongodb.org/browse/EVG-7328". The blue "Patch" link is what it looks like when it's being hovered. 

### Follow Up Items
- Add patch displayName to task query: https://jira.mongodb.org/browse/EVG-7328
- Patch query: https://jira.mongodb.org/browse/EVG-7327


src/contants/routes.ts
```diff
@@ -0,0 +1,13 @@
...
+  task: "/task",
+  patch: "/patch"
+};
+
+export const routes = {
+  login: paths.login,
+  myPatches: paths.myPatches,
+  task: `${paths.task}/:taskID/:tab?`,
+  patch: `${paths.patch}/:patchID`
+};
```
tgrander: I made routes into constants because we will be referencing these routes throughout the app. Using constants prevents accidentally changing a route and causing a cascade of errors from having the route hardcoded in other places.

src/pages/task/TaskBreadcrumb.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { Link } from "react-router-dom";
+import { Breadcrumb } from "antd";
+import { paths } from "contants/routes";
+
+interface Props {
+  version: string;
+  displayName: string;
+}
+
+export const TaskBreadcrumb: React.FC<Props> = ({ version, displayName }) => {
```
john-m-liu: I'm not seeing the need for 2 breadcrumb components. This one already contains the patch one embedded in it more or less, and having the props contain the breadcrumb trail as a list of { label : link } should solve the problem you mentioned initially.

tgrander: I explain why 2 is best implementation in summary but will give it another go:

1. Using one breadcrumb component requires rendering outside of any specific routes. This makes the React Router hooks unusable, which is required to render the correct breadcrumbs. 
2. One breadcrumb component means that the breadcrumb component needs to query for the task data if rendered on the task page. This already creates somewhat messy logic to pull off. Furthermore, the data it is querying is the same data queried for int he task page. However the data cannot be easily shared between the breadcrumb component and the task page. This is one of my least favorite parts of gql on the client so far--there's no easy way to subscribe to the same data like in Redux. The two queries could be grouped together although it would require complex configuration at the link level. As you can see, the code to achieve this is already becoming complex and hard to follow, making it prone to bugs.

The code that is shared between the two components is so low that it does not pose a material reason to converge them. When written this way they are incredibly straight forward with no conditional logic, create no extra queries, require no complex configuration at the link level, and are easy to test because there is no conditional logic.

As someone who tried for an hour and a half to pull off one breadcrumb component, I strongly think this is the best way to go.

tgrander: when i was thinking "one breadcrumb" i thought "one breadcrumb component rendered at the App level. Now I see what you were requesting, which is one breadcrumb component rendered in two places: Patch and Task. That I can do!

john-m-liu:
Reran it and attached the screenshots/videos: https://evergreen.mongodb.com/task/spruce_ubuntu1604_e2e_test_patch_d088dcfa94d58fb262253a57da3298f0f26ea42c_5e3dc21a3e8e8666b3f4d8e2_20_02_07_20_01_34

tgrander:
@john-m-liu all tests work locally now but fail in CI for known reasons. the local evergreen server is never starting and therefore can't be queried, causing all tests that require the server to fail. i have a ticket that i'm prioritizing next sprint to fix this.

PR #68, opened 2020-02-07
MCI-3108 compile evg with correct go version

john-m-liu:
The cypress tests still fail, likely due to how we're doing npm start rather than start:dev-server, but this fixes it so that we're actually running evergreen

### DIFF
--- a/.evergreen.yml
+++ b/.evergreen.yml
@@ -124,13 +124,41 @@ functions:
     params:
       working_dir: spruce
       script: |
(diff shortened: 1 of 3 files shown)


tgrander:
thanks! could you change `npm-start` on [this line](https://github.com/evergreen-ci/spruce/blob/new-ui-feature-branch/.evergreen.yml#L96) to `npm start:dev-server` please. then all tests should be fixed.

PR #69, opened 2020-02-10
EVG-7263: automatically authenticate user

tgrander:
https://jira.mongodb.org/browse/EVG-7263

When a user first comes to spruce, they are either authenticated or not authenticated. While determining which one the user is, show them a loading screen.

- If user is authenticated, continue to the page they were trying to view
- If user is not authenticated, redirect them to logout
- On all subsequent route transitions the user will remain logged in

### Implementation
Use apollo link afterware to automatically authenticate and deauthenticate a user. When user first gets to spruce, show a loading screen while response to top-level query is in flight. If response to the top-level query has data then we know user is authenticated (because otherwise the response would be a 401). We set auth to true and redirect to where they trying to go. If the response is a 401 then we redirect the user the login screen.

### Follow Up Items
- Full page loading screen: https://jira.mongodb.org/browse/EVG-7363


src/components/Loading/FullPageLoad.tsx
```diff
@@ -0,0 +1,18 @@
+import React from "react";
+import styled from "@emotion/styled/macro";
+
+export const FullPageLoad = () => {
+  return (
+    <FullPage>
+      <div>LOADING...</div>
```
tgrander: made a ticket to implement a loading page based on designs

PR #73, opened 2020-02-11
EVG-7289: fix e2e tests in ci

tgrander:
https://jira.mongodb.org/browse/EVG-7289

OUR E2E TESTS IN CI WILL FINALLY WORK 🎆

All e2e tests were failing in CI because the server was not properly setup to talk to the client. Almost all changes occurred in the server, which included:
- Using smoke server instead of local_evergreen
- Updating smoke server yml file to include cors origin for `localhost:3000`

The changes here are so that test data reflects data in smoke db


.evergreen.yml
```diff
@@ -237,11 +237,10 @@ tasks:
     - func: copy-cmdrc
     - func: run-make
       vars:
-        target: load-local-data
-    - func: log-evergreen-to-file
```
john-m-liu: Also delete the function definition

PR #74, opened 2020-02-11
EVG-7168: GQL Query linting

SupaJoon:
This PR installs eslint-plugin-graphql to enable gql query linting. 
Install the VSC extension ESLint by Dirk Baeumer to see query issues in your editor.
I also changed the tslint.json to .js so its easier to work with in the future.

I'd also like your thoughts on whether we should check in the schema or not. 

Follow up ticket: 
Update gql schema with an npm script
https://jira.mongodb.org/browse/EVG-7140


.eslintrc.js
```diff
@@ -0,0 +1,34 @@
...
+        // Import your schema JSON here
+        schemaJson: require("./schema.json")
+
+        // OR provide absolute path to your schema JSON (but not if using `eslint --cache`!)
+        // schemaJsonFilepath: path.resolve(__dirname, './schema.json'),
+
+        // OR provide the schema in the Schema Language format
+        // schemaString: printSchema(schema),
+
+        // tagName is gql by default
```
tgrander: do we need to leave the comments in or can they go? i always read comments, so if they don't provide value then they're just extra stuff i have to read.

SupaJoon: I removed the schema and these comments in this follow up PR: EVG-7140

SupaJoon:
I'm closing this PR because the follow up PR https://github.com/evergreen-ci/spruce/pull/76 removes the schema in favor of symlinking, adjusts some file comments and updates the readme.

PR #75, opened 2020-02-11
EVG-7341: Toast messages for errors, succes, info, etc

tgrander:
https://jira.mongodb.org/browse/EVG-7341

toast_gif

Using the toast feature allows us to give feedback to users (mostly for errors). Errors like network errors.


src/context/toast.tsx
```diff
@@ -0,0 +1,51 @@
...
+);
+
+enum Toast {
+  success = "success",
+  warning = "warning",
+  error = "error",
+  info = "info"
+}
+
+const showToastMessage = (t: keyof typeof Toast) => (toastMessage: string) =>
```
john-m-liu: this can just be t: Toast

tgrander: Screen Shot 2020-02-11 at 7 05 20 PM

that is a natural thought to have but unfortunately TS does not work that way. It has to be `keyof typof enum`

PR #78, opened 2020-02-12
Page skeleton for Patch and Task pages

tgrander:
When the data for the patch and task page are loading, this is what will be shown. 

skeleton-gif

### DIFF
--- a/src/components/Loading/PageSkeleton.tsx
+++ b/src/components/Loading/PageSkeleton.tsx
@@ -1,0 +1,34 @@
+import React from "react";
+import { BigTitle } from "components/styles/Typography";
+import { PageWrapper } from "components/styles/PageWrapper";
+import { SiderCard } from "components/styles/SiderCard";
+import { PageHeader } from "components/styles/PageHeader";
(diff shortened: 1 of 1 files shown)


SupaJoon:
Is this for the initial load only?

tgrander:
> Is this for the initial load only?

it's for when the patch and task page data is loading. however i realize that we may want to allow each component (metadata, build variants, depends on, table) to load independently. However at this point all the queries, at least on patch page, are grouped together. and we may want to keep it that way, or we may not. 

this works for a full-page load of the pages (as i have now). and then for individual loads, filtering a table, then that would render just a component-level load.

tgrander:
i think a good next iteration of this feature would be to break up each component into its own loading component and then can assemble them together for a full page load...this actually might be worth doing now rather than later. will update and push a commit.

SupaJoon:
> i think a good next iteration of this feature would be to break up each component into its own loading component and then can assemble them together for a full page load...this actually might be worth doing now rather than later. will update and push a commit.

Sweet!!

SupaJoon:
evergreen retry

PR #79, opened 2020-02-14
EVG-7139: Retry requests in case of network error

SupaJoon:
https://jira.mongodb.org/browse/EVG-7139

Don't retry if the error is 401 (Unauthorized). Otherwise retry up to a maximum of 5 times where the initial delay is 300ms and subsequent delay times are randomized. The Retry link currently uses default params except for the `retryIf` field.

### DIFF
--- a/src/gql/GQLWrapper.tsx
+++ b/src/gql/GQLWrapper.tsx
@@ -94,7 +94,7 @@ const retryLink = new RetryLink({
   },
   attempts: {
     max: 5,
-    retryIf: (error): boolean => error && error.response.status !== 401
+    retryIf: (error): boolean => error && error.response.status >= 500
   }
 });
 
--- a/src/gql/GQLWrapper.tsx
+++ b/src/gql/GQLWrapper.tsx
@@ -94,9 +94,7 @@ const retryLink = new RetryLink({
   },
   attempts: {
     max: 5,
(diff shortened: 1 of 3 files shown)


src/gql/GQLWrapper.tsx
```diff
@@ -85,6 +86,18 @@ const authenticateIfSuccessfulLink = (dispatch: Dispatch) =>
...
 
+const retryLink = new RetryLink({
+  delay: {
+    initial: 300,
+    max: Infinity,
+    jitter: true
+  },
+  attempts: {
+    max: 5,
+    retryIf: (error): boolean => error && error.response.status !== 401
```
john-m-liu: I think we want to retry for anything >=500. 4xx implies that the client did something wrong so that retrying won't help

src/gql/GQLWrapper.tsx
```diff
@@ -85,6 +86,18 @@ const authenticateIfSuccessfulLink = (dispatch: Dispatch) =>
     });
   });
 
+const retryLink = new RetryLink({
+  delay: {
+    initial: 300,
+    max: Infinity,
```
john-m-liu: This shouldn't come up at all since there's a max attempts, but you should set this to something finite

PR #80, opened 2020-02-18
EVG-7318: Allow Cypress to wait for specific GQL operations

SupaJoon:
This PR allows us to wait on specific gql operations in cypress with the `waitForGQL` function. Previously we've only been able to wait for the GQL endpoint without any visibility into which query we are waiting on, now we can wait on specific query operations.

### DIFF
--- a/cypress/utils/networking.js
+++ b/cypress/utils/networking.js
@@ -1,8 +1,8 @@
-// this fn will wait until a query with queryName is executed on the aliased route
+// this fn will wait until a gql query with queryName is executed on the aliased route
 export const waitForGQL = (alias, queryName) => {
   const waitOnce = () => {
     cy.wait(alias).then(xhr => {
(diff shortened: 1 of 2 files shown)


cypress/utils/networking.js
```diff
@@ -0,0 +1,12 @@
+// this fn will wait until a gql query with queryName is executed on the aliased route
+export const waitForGQL = (alias, queryName) => {
+  const waitOnce = () => {
+    cy.wait(alias).then(xhr => {
```
john-m-liu: You probably want a way for this to timeout so that running cypress doesn't wait indefinitely on a network problem

SupaJoon: `cy.wait` defaults to a 5000ms timeout 
https://docs.cypress.io/api/commands/wait.html#Timeouts

PR #81, opened 2020-02-19
EVG-7210: Render Logs column in task page tests table

SupaJoon:
This PR renders the logs column buttons and relies on https://github.com/evergreen-ci/evergreen/pull/3176.
Screen Shot 2020-02-19 at 5 37 37 PM

My tests don't include verifying that a new tab has been opened after clicking on a Log column button because Cypress docs recommends against it. Instead they recommend to test the existence of attributes that would cause a new tab to open (https://docs.cypress.io/guides/references/trade-offs.html#Multiple-tabs). 

The `Buttons in log column should have target=_blank attribute` test may not pass until I insert some test results with log urls in our test server.

### DIFF
--- a/src/pages/task/TestsTableCore.tsx
+++ b/src/pages/task/TestsTableCore.tsx
@@ -33,9 +33,6 @@ const SpinWrapper = styled.div({
   paddingBottom: 40,
   border: "1px solid #e8e8e8"
 });
-const ButtonWrapper = styled.span({
-  marginRight: 8
-});
 const loadMoreContent = (
   <SpinWrapper>
     <Spin tip="Loading..." />
@@ -96,19 +93,20 @@ const columns: Array<ColumnProps<TaskTestsData>> = [
     dataIndex: "logs",
     key: "logs",
     sorter: false,
-    render: ({
-      htmlDisplayURL,
-      rawDisplayURL
-    }: {
-      htmlDisplayURL: string;
-      rawDisplayURL: string;
-    }): JSX.Element => {
+    render: (
+      {
+        htmlDisplayURL,
+        rawDisplayURL
+      }: { htmlDisplayURL: string; rawDisplayURL: string },
+      record,
+      index
+    ): JSX.Element => {
       return (
         <>
           {htmlDisplayURL && (
             <ButtonWrapper>
               <Button
-                className="htmlBtn"
+                id={`htmlBtn${index}`}
                 size="small"
                 target="_blank"
                 variant="default"
@@ -120,7 +118,7 @@ const columns: Array<ColumnProps<TaskTestsData>> = [
           )}
           {rawDisplayURL && (
             <Button
-              className="rawBtn"
+              id={`rawBtn${index}`}
               size="small"
               target="_blank"
               variant="default"
@@ -281,3 +279,7 @@ export const TestsTableCore: React.FC<ValidInitialQueryParams> = ({
     </div>
   );
 };
+
+const ButtonWrapper = styled.span({
+  marginRight: 8
+});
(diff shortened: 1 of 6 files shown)


src/pages/task/TestsTableCore.tsx
```diff
@@ -77,17 +81,60 @@ const columns: Array<ColumnProps<TaskTestsData>> = [
...
+    }: {
+      htmlDisplayURL: string;
+      rawDisplayURL: string;
+    }): JSX.Element => {
+      return (
+        <>
+          {htmlDisplayURL && (
+            <ButtonWrapper>
+              <Button
+                className="htmlBtn"
```
SupaJoon: This className is applied so cypress can find this button

src/pages/task/TestsTableCore.tsx
```diff
@@ -77,17 +81,60 @@ const columns: Array<ColumnProps<TaskTestsData>> = [
...
+                variant="default"
+                href={htmlDisplayURL}
+              >
+                HTML
+              </Button>
+            </ButtonWrapper>
+          )}
+          {rawDisplayURL && (
+            <Button
+              className="rawBtn"
```
SupaJoon: This className is applied so cypress can find this button

src/pages/task/TestsTableCore.tsx
tgrander: what about using ids for cypress. we would only use ids for testing which makes it clear it's not for styling

src/pages/task/TestsTableCore.tsx
tgrander: what about using ids for cypress. we would only use ids for testing which makes it clear it's not for styling

src/pages/task/TestsTableCore.tsx
```diff
@@ -32,6 +33,9 @@ const SpinWrapper = styled.div({
   paddingBottom: 40,
   border: "1px solid #e8e8e8"
 });
+const ButtonWrapper = styled.span({
+  marginRight: 8
+});
```
tgrander: styled components should always go at bottom of file for consistency

SupaJoon:
> i've read about opening things in new tabs as being an anti pattern in UX. would like to get some discussion going around this as it relates to a few parts of the app. opening in new app breaks user flow, for example they are not able to use the 'back' button to go back to the previous page anymore. and if a user does wish to open in a new tab then they can always use `cmd + click` on the link

Ben recommends opening the Logs in a new tab but I see the benefit of giving the user both options.

src/pages/task/TestsTableCore.tsx
SupaJoon: Assigning an id to button will result in duplicate id's and I would like to keep all id's unique so we are within the HTML 5.1 spec. It also seems overkill to assign a classname just for testing purposes and I think I need to remove the classname and improve the selector in cypress.

john-m-liu:
evergreen retry

src/pages/task/TestsTableCore.tsx
john-m-liu: Is there a test case where you'd want to find the logs button in a specific row though? If that's the case I think it makes sense to generate an ID that contains the row index

tgrander: > Assigning an id to button will result in duplicate id's and I would like to keep all id's unique so we are within the HTML 5.1 spec. It also seems overkill to assign a classname just for testing purposes and I think I need to remove the classname and improve the selector in cypress.

just append an id to the end of each id

tgrander:
The test failures on this PR look like they fail because of actual errors in the code, which is great news considering our test failures on FE were meaningless for so long. i'm sure you're working on those and i will approve once i see dat green.

PR #82, opened 2020-02-20
EVG-7343: render patch metadata

tgrander:
https://jira.mongodb.org/browse/EVG-7343

Rendering the patch metadata on the patch route. Also beginning to build out the shared styled components including Typography components.

### Screenshot
Screen Shot 2020-02-20 at 10 45 29 AM

### Link to designs
https://mongodb.invisionapp.com/share/X2VZOTNFD46#/screens/405089967


cypress.json
```diff
@@ -3,5 +3,6 @@
   "projectId": "m45k1x",
   "reporterOptions": {
     "mochaFile": "junit.xml"
-  }
+  },
+  "chromeWebSecurity": false
```
tgrander: this is so that cypress doesn't freak out when we from spruce.com to evergreen.com

tgrander:
tests are failing because duration code is not yet merged into evergreen

src/pages/Patch.tsx
```diff
@@ -1,22 +1,77 @@
...
+  StyledLink,
+  PageContent,
+  PageLayout,
+  PageSider
+} from "components/styles";
+import { Divider } from "components/styles/Divider";
+import { useQuery } from "@apollo/react-hooks";
+import { GET_PATCH, PatchQuery } from "gql/queries/patch";
 
 export const Patch = () => {
```
SupaJoon: I know this code change isn't part of your PR but it's missing the React.FC type

cypress/integration/patch-route.js
```diff
@@ -0,0 +1,35 @@
...
+
+  it("Loads patch data and renders it on the page", function() {
+    cy.visit(`/patch/${patch.id}`);
+    cy.get("h1[id=patch-name]").should("include.text", patch.desc);
+  });
+
+  it("Clicking 'Base commit' link in metadata goes to version page of legacy UI", function() {
+    cy.visit(`/patch/${patch.id}`);
+    cy.get("a[id=patch-base-commit]").click();
+    cy.url().should(
```
SupaJoon: This test can be more robust if we replace this line to match the pathname only
 cy.location('pathname').should('eq', `https://evergreen.mongodb.com/version/${patch.id}`)

src/contants/colors.ts
```diff
@@ -0,0 +1,2 @@
+export const blueBase = "#007CAD";
```
SupaJoon: The colors in this file are available from the @leafygreen-ui/palette npm module. What do you think about importing the colors from there instead of this file?

cypress.json
john-m-liu: I don't think this is necessary - cypress tests should be using localhost:9090 as the app server, and that's already configured to allow CORS from localhost:3000

src/pages/Patch.tsx
```diff
@@ -1,22 +1,77 @@
...
+            <P2>Makespan: {makespan && makespan}</P2>
+            <P2>Time taken: {timeTaken && timeTaken}</P2>
+            <P2>Submitted at: {submittedAt}</P2>
+            <P2>Started: {started && started}</P2>
+            <P2>Finished: {finished && finished}</P2>
+            <P2>{`Submitted by: ${author}`}</P2>
+            <P2>
+              <StyledLink
+                id="patch-base-commit"
+                href={`https://evergreen.mongodb.com/version/${version}`}
```
john-m-liu: Don't hardcode the app server - get it from the cmdrc

src/pages/Patch.tsx
tgrander: if the compiler does not complain then it is not needed. `React.FC` component is required when component accepts props.

cypress/integration/patch-route.js
tgrander: sure

src/contants/colors.ts
tgrander: good idea

SupaJoon:
evergreen retry

SupaJoon:
I think all of the failed tests will pass if you pass in a patch ID that exists in the CI db. Those can be found in the evergreen repo under `testdata/smoke`

tgrander:
evergreen retry

tgrander:
evergreen retry

PR #83, opened 2020-02-21
get login domain based on env

tgrander:
the login domain was originally being set using `getUiUrl` which was returning `localhost:3000`. However the ui url refers to the evergreen ui, which means `getUiUrl` should return `loclahost:9090` in dev/test and `evergreen.mongodb.com` in prod.

I broke the auth tests when i updated the return value of `getUiUrl` to be `localhost:9090` and realized that some additional logic was necessary to determine the domain used to make HTTP requests to `/login`.

This PR adds `getLoginDomain` to the env var utility functions. 

- add `REACT_APP_SPRUCE_URL` env var to dev in `.cmd-rc` file equal to `localhost:3000`
- if in de/test - return `REACT_APP_SPRUCE_URL`
- if in prod/staging - return `REACT_APP_UI_URL`

The dev server currently proxies all requests to `localhost:9090` from `localhost:3000`. **An alternative to doing this would be to add CORS to the login/logout endpoints.**

### DIFF
--- a/config/.cmdrc_sample.json
+++ b/config/.cmdrc_sample.json
@@ -1,7 +1,9 @@
 {
   "devServer": {
     "REACT_APP_GQL_URL": "http://localhost:9090/graphql/query",
-    "REACT_APP_API_URL": "http://localhost:3000/api"
+    "REACT_APP_API_URL": "http://localhost:3000/api",
+    "REACT_APP_UI_URL": "http://localhost:9090",
+    "REACT_APP_SPRUCE_URL": "http://localhost:3000"
   },
   "mockIntrospectSchema": {
     "REACT_APP_GQL_URL": "http://localhost:9090/graphql/query",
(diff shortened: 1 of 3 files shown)


tgrander:
the tests are passing locally for me. failing tests seem to be due to the task being queried no longer existing in db...as if it vanished. 🤔 looking into it

tgrander:
@john-m-liu in that case let's just do it now 💯

tgrander:
i tried to add CORS to login endpoints but there were unexpected things happening. i created this follow up ticket to address the problem at a later time: https://jira.mongodb.org/browse/EVG-7432

SupaJoon:
> I think it makes more sense if we set `REACT_APP_API_URL=http://localhost:9090` instead of `http://localhost:3000/api` and use it in place of `REACT_APP_UI_URL`.
> I'm under the impression that we are not utilizing `localhost:3000/api` for any of our endpoints

Discussed offline

PR #86, opened 2020-02-28
EVG-7390: Test Files Table

SupaJoon:
Relies on https://github.com/evergreen-ci/evergreen/pull/3228.
View table at http://localhost:3000/task/evergreen_ubuntu1604_89/files

PR Screenshots
Screen Shot 2020-02-28 at 2 05 23 PM
Screen Shot 2020-02-28 at 2 05 31 PM
Screen Shot 2020-02-28 at 2 05 49 PM

Design
Screen Shot 2020-02-28 at 2 11 26 PM

Ben mentioned that the file names should be links and button in the right column should be removed.


src/pages/Task.tsx
```diff
@@ -62,6 +63,16 @@ export const Task: React.FC = () => {
     task: { displayName, version }
   } = data;
 
+  let BodyComp = <></>
```
john-m-liu: Does it make sense to refactor the tab contents here into a higher-order component? I can see this becoming somewhat messy as we add tabs, and a separate HOC might help with that

src/pages/task/FilesTables/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+          href={record.link}
+          rel="noopener noreferrer"
+          target="_blank"
+        >
+          {text}
+        </a>
+      );
+    },
+    defaultSortOrder: "ascend" as SortOrder,
+    sorter: (a: File, b: File): number => a.name.localeCompare(b.name)
```
john-m-liu: 👍 for making this collation-friendly

src/pages/task/TaskPageBody.tsx
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+import { Tab, TaskURLParams } from "pages/types/task";
+import { TestsTable } from "pages/task/TestsTable";
+import { FilesTables } from "pages/task/FilesTables";
+import { useParams } from "react-router-dom";
+
+export const TaskPageBody: React.FC = () => {
+  const { tab } = useParams<TaskURLParams>();
+  let BodyComp = <></>;
+  switch (tab) {
```
tgrander: switch statements should always have a default

src/pages/task/TaskPageBody.tsx
```diff
@@ -0,0 +1,19 @@
...
+    case Tab.Tests:
+      BodyComp = <TestsTable />;
+      break;
+    case Tab.Files:
+      BodyComp = <FilesTables />;
+      break;
+  }
+  return BodyComp;
+};
```
tgrander: i saw @john-m-liu's comment but this being a separate component adds unneeded complexity. a render function that returns tab component based on `useParams` would have been much more straight forward. because technically this is not really a component. it's conditional logic.

cypress/integration/files_tables.js
```diff
@@ -0,0 +1,51 @@
+/// <reference types="Cypress" />
+import { waitForGQL } from "../utils/networking";
+
+const waitForFilesQuery = () => waitForGQL("@gqlQuery", "taskFiles");
+const FILES_ROUTE = "/task/evergreen_ubuntu1604_89/files";
+const FILES_ROUTE_WITHOUT_FILES =
+  "/task/evergreen_ubuntu1604_test_model_commitqueue_patch_5e823e1f28baeaa22ae00823d83e03082cd148ab_5e4ff3abe3c3317e352062e4_20_02_21_15_13_48/files ";
+
+describe("tests table", function() {
```
tgrander: should be named files table instead of tests table?

src/pages/task/TaskPageBody.tsx
tgrander: taking a second look at this component and i really don't think it should be its own component right now. it's really simple and straight forward; therefore would be easier to read if all in the same file. when components begin to become complex is when it's time to refactor. premature refactoring components leads to unnecessary componentization. Leave things as they are until they begin to become complex.

SupaJoon: @tgrander I think this is a reasonable abstraction for now since this list will grow and the task page is already doing multiple things (making a request for the task, rerouting to default tab). We may potentially add additional logic to the Task page to render the tab labels.

tgrander: eh, sure. i still think a simple function could have done the trick. but why use `let` to reassign variables instead of just returning the component directly?

SupaJoon:
evergreen retry

SupaJoon:
made this ticket to fix flaky auth test: https://jira.mongodb.org/browse/EVG-7488

src/pages/Task.tsx
```diff
@@ -13,15 +13,22 @@ import {
...
 
-enum Tab {
-  Logs = "logs",
-  Tests = "tests",
-  Files = "files",
-  BuildBaron = "build-baron"
-}
 const DEFAULT_TAB = Tab.Logs;
 
+const getBodyComp = tab => {
```
tgrander: render functions are typically named beginning with render. Does not matter much because this will be replaced in my PR with the Tabs logic. But `renderTab` would be a much clearer name imo.

src/pages/Task.tsx
```diff
@@ -62,6 +69,7 @@ export const Task: React.FC = () => {
     task: { displayName, version }
   } = data;
 
+  const bodyComp = getBodyComp(tab);
```
tgrander: why assign this a variable? can just be called as jsx below

src/pages/task/FilesTables.tsx
```diff
@@ -0,0 +1,123 @@
...
+            rowKey={rowKey}
+            columns={columns}
+            dataSource={files}
+            pagination={false}
+            scroll={{ y: 196 }}
+          />
+        </Fragment>
+      );
+    });
+  const body = tables.length ? tables : <div>No files found</div>;
```
tgrander: why assign this to a variable? can put the assigned logic directly into jsx below

src/types/task.ts
```diff
@@ -14,3 +14,15 @@ export enum SortQueryParam {
...
+  Logs = "logs",
+  Tests = "tests",
+  Files = "files",
+  BuildBaron = "build-baron"
+}
+
+export interface TaskURLParams {
+  tab?: Tab;
+  taskID: string;
+}
```
tgrander: why not leave these types in the component file?

src/pages/task/FilesTables.tsx
```diff
@@ -0,0 +1,123 @@
...
+  }
+  if (error) {
+    return <div>{error.message}</div>;
+  }
+  if (!filteredData && data) {
+    // this will happen once per mount
+    return <></>;
+  }
+
+  const onSearch = (e): void => {
```
tgrander: e has type `any`

src/pages/task/FilesTables.tsx
```diff
@@ -0,0 +1,123 @@
...
+import { Input } from "antd";
+import debounce from "lodash.debounce";
+import { SortOrder } from "antd/es/table/interface";
+
+const columns = [
+  {
+    title: "Name",
+    dataIndex: "name",
+    key: "name",
+    render: (text: string, record: File, index: number): JSX.Element => {
```
tgrander: index is not used

src/pages/Task.tsx
```diff
@@ -13,15 +13,22 @@ import {
...
 const DEFAULT_TAB = Tab.Logs;
 
+const getBodyComp = tab => {
+  switch (tab) {
+    case Tab.Tests:
+      return <TestsTable />;
+    case Tab.Files:
+      return <FilesTables />;
+    default:
+      return <></>;
```
tgrander: should the default be a "nothing" ?

src/pages/task/FilesTables.tsx
```diff
@@ -0,0 +1,124 @@
...
+    return <div>{error.message}</div>;
+  }
+  if (!filteredData && data) {
+    // this will happen once per mount
+    return <></>;
+  }
+
+  const onSearch = (e: React.ChangeEvent<HTMLInputElement>): void => {
+    setFilterStr(e.target.value);
+  };
```
tgrander: the return type for functions like these do not need to be explicitly typed. the compiler knows that the function will return void based on the function body.

src/pages/task/FilesTables.tsx
tgrander: variables like this are unnecessary / make code harder to read. can put the variable value directly into jsx below.

PR #90, opened 2020-03-06
EVG-7441 add dummy data to cmdrc

john-m-liu:


tgrander:
why?

john-m-liu:
It seemed weird to have real URLs in a sample config file. I noticed it while writing part of the new hire guide

PR #96, opened 2020-03-17
EVG-7413: tree select

SupaJoon:
Screen Shot 2020-03-17 at 1 02 41 PM

Screen Shot 2020-03-17 at 12 54 43 PM
The first picture is when the component is imported into Spruce, the second is from storybook. 

This PR introduces a the TreeSelect shared component as well as storybook. For now, this component will be used to render the checkbox dropdown to filter the table in the task and patch page. 
To view the component launch storybook with `npm run storybook`.
My next PR will import this component to render the status filter dropdown on the task page.


john-m-liu:
Should we be submitting this to leafygreen rather than here? I'm not sure if you discussed this with design systems beforehand, but this seems like a good addition to their library

tgrander:
> Should we be submitting this to leafygreen rather than here? I'm not sure if you discussed this with design systems beforehand, but this seems like a good addition to their library

sounds good but let's not block progress here. we can merge this and then take additional action based on if leafygreen is receptive to the component.

src/hooks/useOnClickOutside.ts
```diff
@@ -0,0 +1,21 @@
...
+  cb: () => void
+) => {
+  useEffect(() => {
+    // Bind the event listener
+    document.addEventListener("mousedown", handleClickOutside);
+    return () => {
+      // Unbind the event listener on clean up
+      document.removeEventListener("mousedown", handleClickOutside);
+    };
+  });
```
tgrander: this `useEffect` has no dependencies listed, which means this event binding will occur on every rerender of the component in which it is called...sounds like potential for a memory leak. passing an empty array as dependency will cause event listener to be attached only on mount.

john-m-liu:
But this is exactly the kind of thing that should prompt us to wait. If we end up adding it to leafygreen, it'll likely have different reviewers, and the only change here will be to add a dependency

tgrander:
and to update the component to reflect correct UX. i'm not worried about it. let's see how it goes with leafygreen.

src/hooks/useOnClickOutside.ts
SupaJoon: > in the screenshot you provide, shapes is a nested group. not all of shapes's children are selected but its box is selected. this would indicate that all children are selected. the parent checkbox should only be checked if all its children are selected.
> 
> i confirmed this UX by using the ant tree select component [here](https://ant.design/components/tree-select/).

The component controlling value is meant to take care of this.

SupaJoon:
> in the screenshot you provide, shapes is a nested group. not all of shapes's children are selected but its box is selected. this would indicate that all children are selected. the parent checkbox should only be checked if all its children are selected.
> 
> i confirmed this UX by using the ant tree select component [here](https://ant.design/components/tree-select/).

The component controlling value is meant to take care of this.

john-m-liu:
Correct UX will vary depending on who you ask though. Continuing on this is doing work that has a decent chance of being reverted, whereas spending 15 minutes to talk to someone will prevent that

tgrander:
Screen Shot 2020-03-17 at 4 05 21 PM

they do not need it - we can carry on :)

Going forward let's check in with leafygreen before creating UI components.

src/components/TreeSelect.tsx
```diff
@@ -0,0 +1,168 @@
...
+  state,
+  onChange
+}: {
+  tData: TreeDataEntry[];
+  state: string[];
+  onChange: (v: [string]) => void;
+}) => {
+  const rows: JSX.Element[] = [];
+  let level = 0;
+  tData.forEach(entry => {
```
john-m-liu: This propagates the change handler for the entire tree even if you click on something other than the root. It seems to work so I'm probably missing something, but shouldn't you start on the node that's clicked? Possibly related, while testing I'm seeing that checking a parent doesn't check its children, but it should right?

SupaJoon: The checkboxes aren't nested so that's why there isn't any propagation. I added a `padding-left` based on the level. In the example, checking a parent doesn't check or uncheck it's children because I didn't handle it in the component that controls the state of TreeSelect.

SupaJoon:
I'm going to add some logic to handle the parent/children select/deselect

src/hooks/useOnClickOutside.ts
```diff
@@ -0,0 +1,20 @@
...
+  useEffect(() => {
+    function handleClickOutside(event) {
+      if (ref.current && !ref.current.contains(event.target)) {
+        cb();
+      }
+    }
+    // Bind the event listener
+    document.addEventListener("mousedown", handleClickOutside);
+    return () => {
+      // Unbind the event listener on clean up
```
tgrander: TLDR: Comments do not add value for code where it is clear from reading react docs / having FE knowledge what it does.

i appreciate you adding more comments! Sarah Drasner made a great tweet about comments which is what inspired me to write more of them. What she said is that comments should explain the "why" about a code. For example, regex should always have a code because no one can read it. Complex or unclear code should have a comment explaining why it's there so that a future coder knows not to remove it. Or so they know how to refactor it. 

Comments like `Bind the event listener` and `Unbind the event listener on clean up` however basically restate the code they describe. They answer the "what" instead of the "why". Therefore the reader has to read double: the code and the comment which both convey the same thing.

src/components/TreeSelect.tsx
```diff
@@ -0,0 +1,337 @@
...
+  onChange: (v: [string]) => void;
+}) => {
+  const rows: JSX.Element[] = [];
+  tData.forEach(entry => {
+    renderCheckboxesHelper({ rows, data: entry, onChange, state, tData });
+  });
+  return rows;
+};
+
+const renderCheckboxesHelper = ({
```
tgrander: we talked offline about render functions and both agree that they have purpose and value but that components will lead to better abstractions. this code is fine as is and we can implement our decision around render funcs going forward.

PR #98, opened 2020-03-19
EVG-7413: Status Selector for Task Page

SupaJoon:

This PR builds off of https://github.com/evergreen-ci/spruce/pull/96.

Status selector displays statuses used to filter the tests page. Statuses are pushed to statuses query param as comma separated values and represent the state up the status selector. The statuses query param is filtered and passed to statuses field in the taskTests gql query.

https://github.com/evergreen-ci/evergreen/pull/3311 includes data shown in the gif. 

Test url: 
http://localhost:3000/task/evergreen_ubuntu1604_test_model_patch_5e823e1f28baeaa22ae00823d83e03082cd148ab_5e4ff3abe3c3317e352062e4_20_02_21_15_13_48/tests?category=TEST_NAME&sort=-1

testtable

Next I will implement the test name filter input that sits to the left of the status selector.


cypress/integration/test_table.js
```diff
@@ -95,4 +95,86 @@ describe("tests table", function() {
...
+    ];
+
+    statuses.forEach(({ display, key }) => {
+      it(`Clicking on ${display} status checkbox adds ${key} status to URL and clicking again removes it`, () => {
+        cy.get(".cy-checkbox")
+          .contains(display)
+          .click();
+        cy.location().should(loc => {
+          expect(loc.pathname).to.equal(TESTS_ROUTE);
+          expect(loc.search).to.include(`statuses=${key}`);
```
tgrander: i think this test would still pass if clicking display added every status to url every time. should this test that loc.seach _only_ contains the selected status?

cypress/integration/test_table.js
```diff
@@ -95,4 +95,86 @@ describe("tests table", function() {
...
+          .contains(display)
+          .click();
+      });
+      cy.get(".cy-checkbox")
+        .contains("Pass")
+        .click();
+      cy.get(".cy-checkbox > input")
+        .first()
+        .should("not.be.checked");
+    });
```
tgrander: these tests are more for the TreeSelect component in general whereas these are the test-table tests

cypress/integration/test_table.js
```diff
@@ -95,4 +95,86 @@ describe("tests table", function() {
...
+      statuses.forEach(({ display }) => {
+        cy.get(".cy-checkbox")
+          .contains(display)
+          .click();
+      });
+      cy.location().should(loc => {
+        expect(loc.search).to.include("statuses=pass,silentfail,fail,skip,all");
+      });
+    });
+  });
```
tgrander: i'm not sure what this test is doing

src/pages/task/TestsTable.tsx
```diff
@@ -39,15 +40,25 @@ export const TestsTable: React.FC = () => {
...
     } else if (!validInitialQueryParams) {
+      const statuses = parsed[RequiredQueryParams.Statuses];
       setValidInitialQueryParams({
         initialCategory: parsed[RequiredQueryParams.Category],
-        initialSort: parsed[RequiredQueryParams.Sort]
+        initialSort: parsed[RequiredQueryParams.Sort],
+        initialStatuses: (Array.isArray(statuses)
+          ? statuses
+          : [statuses]
+        ).filter(v => v && v !== TestStatus.All)
```
tgrander: we discussed not needing to set initial query params a while back. instead of adding more initial query param logic, you could remove it altogether

src/pages/task/testsTable/TestsTableCore.tsx
```diff
@@ -165,30 +81,61 @@ export const TestsTableCore: React.FC<ValidInitialQueryParams> = ({
...
+
+  // this fetch is when url params change (sort direction, sort category, status list)
+  // and the page num is set to 0
+  useEffect(() => {
+    if (
+      (sort !== prevSort ||
+        category !== prevCategory ||
+        statuses !== prevStatuses) &&
+      networkStatus === NetworkStatus.ready &&
+      !error
```
tgrander: what is the value-add of checking if url are not equal to previous value after a url param change is triggered?

SupaJoon: the dependency list for the useEffect has many variables aside from url params such as network status and error. This logic is necessary to make sure we make a request only when the relevant query params have changed.

cypress/integration/test_table.js
SupaJoon: This tests shows that clicking on multiple statuses adds them all to the URL as opposed to one, some or none

cypress/integration/test_table.js
SupaJoon: sure

src/pages/task/testsTable/TestsTableCore.tsx
tgrander: instead you could remove this logic, which is a difficult to read, and replace it with `history.listen` which will attach an event listener to the url. within the cb you can call fetchMore which will only be called when the url changes.

tgrander: https://reacttraining.com/react-router/web/api/Hooks/usehistory

SupaJoon: `listen` seems good to use but I still think it's meaningful to only dispatch the request when a relevant parameter has changed. The callback passed to `listen` has a location argument that we can deconstruct to examine the query parameters but there isn't a way to see which query parameters have changed from the previous render.

cypress/integration/test_table.js
SupaJoon: I see how these tests are geared towards the TreeSelect because we are testing how TreeSelect actions affect the TreeSelect state. It may make more sense if these tests checked what's happening in the URL instead of the TreeSelect checkbox state after checking/unchecking.

src/pages/task/testsTable/TestsTableCore.tsx
tgrander: > I still think it's meaningful to only dispatch the request when a relevant parameter has changed

when will a parameter change without it being relevant?

>  I would still want to compare the previous query param to the current

when will a parameter change where it is the same as before?

cypress/integration/test_table.js
tgrander: what i'm saying is these tests should be for the TreeSelect component and not in the test-table component. otherwise this test would have to be in the patch page table too, which doesn't make much sense.

src/pages/task/TestsTable.tsx
SupaJoon: I made a ticket for this: https://jira.mongodb.org/browse/EVG-7611

src/pages/task/testsTable/TestsTableCore.tsx
SupaJoon: In the future we may have other query param in the URL alongside the query params that control the task table.

cypress/integration/test_table.js
tgrander: what do you think of updating the name of the test to say exactly that? because that i understand :)

src/pages/task/testsTable/StatusSelector.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { useLocation, useHistory } from "react-router-dom";
+import queryString from "query-string";
+import { RequiredQueryParams } from "types/task";
+import { TestStatus } from "types/task";
+import { TreeSelect } from "components/TreeSelect";
+
+export const StatusSelector = () => {
+  const { pathname, search } = useLocation();
+  const { replace } = useHistory();
+  const value = useStatuses(search);
```
tgrander: +1 to abstracting this into a custom hook. only question is about the name -- use statuses how?

src/pages/task/testsTable/StatusSelector.tsx
```diff
@@ -0,0 +1,65 @@
...
+      tData={treeData}
+      inputLabel="Test Status:  "
+      id="cy-test-status-select"
+    />
+  );
+};
+
+const arrayFormat = "comma";
+
+const treeData = [
```
tgrander: +1 to putting these things below the component so that the component is the first code in the file.

src/pages/task/testsTable/TestsTableCore.tsx
tgrander: that sounds hypothetical though and therefore an equal argument is that we could never have irrelevant query params that control the task table. 

right now all params are relevant and updating on change will have no negative performance consequences.

src/pages/task/testsTable/StatusSelector.tsx
SupaJoon: useStatuses returns the query param statuses

PR #99, opened 2020-03-19
EVG-7605: patch header and metadata loading and error views

tgrander:
### When patch query is loading
Show loading skeletons on patch header and in metadata card

patch loading

### When patch query errored
Display error message in metadata card and do NOT render breadcrumb and header

Screen Shot 2020-03-19 at 9 08 55 AM

### When patch query is successful
Normal view

Screen Shot 2020-03-19 at 9 12 17 AM

### DIFF
--- a/src/pages/patch/patchTabs/tasks/TasksTable.tsx
+++ b/src/pages/patch/patchTabs/tasks/TasksTable.tsx
@@ -45,19 +45,17 @@ export const TasksTable: React.FC<Props> = ({ networkStatus, data }) => {
   };
 
   return (
-    <>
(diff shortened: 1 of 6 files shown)


src/pages/patch/patchTabs/tasks/TasksTable.tsx
```diff
@@ -45,19 +45,17 @@ export const TasksTable: React.FC<Props> = ({ networkStatus, data }) => {
   };
 
   return (
-    <>
```
tgrander: removing this unnecessary fragment

PR #100, opened 2020-03-19
EVG-7413: test name filter

SupaJoon:
This is the final PR for EVG-7413. 

These code changes render the test name filter above the tests table in the task page. Typing into the input will adjust the testname query param in the URL which results in a GQL request to update the contents of the table. 

testnamefilter


src/pages/task/TestsTable.tsx
```diff
@@ -42,23 +47,51 @@ export const TestsTable: React.FC = () => {
...
+    replace(`${pathname}?${nextQueryParams}`);
+  };
+
   return (
     <>
-      <StatusSelector />
+      <FiltersWrapper>
+        <StyledInput
+          placeholder="Search Test Names"
+          onChange={onSearch}
```
john-m-liu: Should we debounce this input? Testresult queries can be expensive, so it'd be nice to not perform the search if it's not done

SupaJoon:
> looks great. i know you tried in the past to add a loading indicator to the table and it was difficult. i'm wondering if there are any further thoughts on that at this time? the UX feels incomplete without a loading indicator when the new tasks are being fetched after a user types into the search bar. it's super fast in our local environment but will take significantly longer in prod. regardless of latency, i think we should figure out the loading indicator for the tables sooner rather than later.

Since we don't alter the contents of the table until a search result has returned, I think it makes most sense to add some kind of loading icon in the input. I'll add this.

tgrander:
> > looks great. i know you tried in the past to add a loading indicator to the table and it was difficult. i'm wondering if there are any further thoughts on that at this time? the UX feels incomplete without a loading indicator when the new tasks are being fetched after a user types into the search bar. it's super fast in our local environment but will take significantly longer in prod. regardless of latency, i think we should figure out the loading indicator for the tables sooner rather than later.
> 
> Since we don't alter the contents of the table until a search result has returned, I think it makes most sense to add some kind of loading icon in the input. I'll add this.

that's a great idea

cypress/integration/test_table.js
```diff
@@ -155,4 +155,27 @@ describe("Tests Table", function() {
...
+    });
+    it("Typing in test name filter updates testname query param", () => {
+      cy.location().should(loc => {
+        expect(loc.search).to.include("testname=group");
+      });
+    });
+
+    it("Input value is incuded in the taskTests GQL request body under variables.testName ", () => {
+      waitForTestsQuery(); // this query is for the initial data load before any typing begins
+      waitForTestsQuery();
```
tgrander: it has to be done twice in a row?

src/pages/task/TestsTable.tsx
```diff
@@ -10,11 +14,27 @@ import {
...
+    search: string,
+    replace: (path: string) => void,
+    pathname: string
+  ) => {
+    const parsed = queryString.parse(search, { arrayFormat });
+    parsed[RequiredQueryParams.TestName] = testName;
+    const nextQueryParams = queryString.stringify(parsed);
+    replace(`${pathname}?${nextQueryParams}`);
+  },
+  500
```
tgrander: do you think 500 is too long? we do want it to feel instantaneous without actually being instantaneous. maybe 200-300?

SupaJoon: 500 is a little noticeable, i changed it to 250.

cypress/integration/test_table.js
SupaJoon: the first query is the initial load, the second one is after typing

SupaJoon: I can alter the util function so we can pass in gql variables we're looking for such as testName in this case. also for some reason this test passes locally but not in CI

PR #102, opened 2020-03-20
EVG-7489: fetch sorted tasks when patch headers are clicked

tgrander:
https://jira.mongodb.org/browse/EVG-7489

test table

#### User stories 
- When I click a task table header, sorted tasks are fetched and displayed corresponding to the clicked header

#### How this is achieved technically
- `history.listen` is used to listen to updates to the url location
- when url location changes, the query params in url `search` are parsed and then fed to the gql `patchTasks` query as sorters
- new tasks are fetched with `fetchMore` from `useQuery` and the tasks table is updated with the new tasks

### Up next

Fetch more tasks when user scrolls table: https://jira.mongodb.org/browse/EVG-7490

### DIFF
--- a/cypress/integration/patch-route.js
+++ b/cypress/integration/patch-route.js
@@ -113,31 +113,31 @@ describe("Patch route", function() {
         cy.visit(path);
       });
 
-      it("updates the url when column headers are clicked", () => {
+      it("Updates the url when column headers are clicked", () => {
         cy.visit(path);
 
-        cy.get("th.cy-task-table-col-name").click();
+        cy.get("th.cy-task-table-col-NAME").click();
         locationHasUpdatedParams("NAME", "ASC");
 
-        cy.get("th.cy-task-table-col-name").click();
+        cy.get("th.cy-task-table-col-NAME").click();
         locationHasUpdatedParams("NAME", "DESC");
 
-        cy.get("th.cy-task-table-col-name").click();
+        cy.get("th.cy-task-table-col-NAME").click();
         locationHasUpdatedParams("NAME");
 
-        cy.get("th.cy-task-table-col-variant").click();
+        cy.get("th.cy-task-table-col-VARIANT").click();
         locationHasUpdatedParams("VARIANT", "ASC");
 
-        cy.get("th.cy-task-table-col-variant").click();
+        cy.get("th.cy-task-table-col-VARIANT").click();
         locationHasUpdatedParams("VARIANT", "DESC");
 
-        cy.get("th.cy-task-table-col-variant").click();
+        cy.get("th.cy-task-table-col-VARIANT").click();
         locationHasUpdatedParams("VARIANT");
       });
 
       it("clicking task name goes to task page for that task", () => {
         cy.visit(path);
-        cy.get("td.cy-task-table-col-name:first").within(() => {
+        cy.get("td.cy-task-table-col-NAME:first").within(() => {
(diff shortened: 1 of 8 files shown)


cypress/integration/patch-route.js
```diff
@@ -108,36 +109,81 @@ describe("Patch route", function() {
...
+        cy.contains(TABLE_SORT_SELECTOR, "Name").click();
+        cy.once("fail", err => {
+          expect(err.message).to.include(
+            "'pointer-events: none' prevents user mouse interaction."
+          );
+        });
+      });
+
+      it("Fetches sorted tasks when table sort headers are clicked", () => {
+        ["NAME", "STATUS"].forEach(sortBy =>
```
tgrander: only testing these two because it felt redundant (and expensive) to test all four statuses. these are the two statuses i subjectively deem to be "the most likely"

src/pages/patch/patchTabs/Tasks.tsx
```diff
@@ -1,42 +1,97 @@
...
 };
+
+const getQueryVariablesFromUrlSearch = (
+  patchId: string,
+  search: string
+): PatchTasksVariables => {
+  // TODO: add 'statuses' var here when the UI is implemented
+  const { sortBy, sortDir, page } = queryString.parse(
+    search
+  ) as PatchUrlSearchKeys;
```
tgrander: i tried other options but `as` seemed to be the only option here 🙈

cypress/integration/patch-route.js
SupaJoon: I think testing the other sort headers is a useful test and not including them could mislead us. The code to handle moving query params from the event listener to the URL and query variables is specific and could break in such a way where NAME and STATUS pass but the other sort headers fail.

src/pages/task/testsTable/TestsTableCore.tsx
```diff
@@ -52,22 +53,7 @@ export const TestsTableCore: React.FC<ValidInitialQueryParams> = ({
...
-      elements.forEach(el => {
-        (el as HTMLElement).style["pointer-events"] = "none";
-      });
-    } else {
-      elements.forEach(el => {
-        (el as HTMLElement).style["pointer-events"] = "auto";
-      });
-    }
-  }, [networkStatus]);
+  useDisableTableSortersIfLoading(networkStatus);
```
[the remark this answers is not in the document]
tgrander: 😃

cypress/integration/patch-route.js
tgrander: i agree!

src/hooks/useDisableTableSortersIfLoading.ts
```diff
@@ -0,0 +1,21 @@
+import { useEffect } from "react";
+import { NetworkStatus } from "apollo-client";
+
+export const useDisableTableSortersIfLoading = (
+  networkStatus: NetworkStatus
+) => {
+  useEffect(() => {
+    const elements = document.querySelectorAll(
```
john-m-liu: I can see us wanting to make this more generalized in the near future, but this works for now

tgrander: i get what you mean but this is a hack we have to do to disable the ant table because it does not have the right API for us to do it via props. This is not the typical way to accomplish this kind of thing in React and will not be generalized.

PR #103, opened 2020-03-23
EVG-7490: fetch additional tasks when scrolling tasks table

tgrander:
https://jira.mongodb.org/browse/EVG-7490

Users can infinitely scroll the tasks table to view more tasks. The table stops requesting more tasks when all tasks have been fetched. 

scrolling


src/pages/patch/patchTabs/Tasks.tsx
```diff
@@ -53,17 +55,59 @@ export const Tasks: React.FC<Props> = ({ taskCount }) => {
     });
   }, [history, fetchMore, id, error, networkStatus]);
```
john-m-liu: data should be added as a dependency here

tgrander: removed data as a dependency altogether

cypress/integration/patch-route.js
```diff
@@ -161,10 +161,78 @@ describe("Patch route", function() {
           clickSorterAndAssertTasksAreFetched(sortBy)
         );
       });
+
+      it("Fetches additional tasks as the user scrolls", () => {
```
SupaJoon: This is an awesome test

cypress/integration/patch-route.js
```diff
@@ -161,10 +161,78 @@ describe("Patch route", function() {
...
+
+            // confirm additional tasks were appended to table items
+            cy.get(".ant-table-row").should(
+              "have.length.greaterThan",
+              $initialTasks.length
+            );
+          });
+      });
+
+      it("Task count increments by the number of additional tasks fetched", () => {
```
SupaJoon: the test description implies that the task count increases by an exact number but we are only verifying that the task count increases after we scroll. 
A way to test what the description claims could be to inspect the limit query variable or response length and assert that the new task count is equal to the original task count + response length or limit variable

cypress/integration/patch-route.js
```diff
@@ -161,10 +161,78 @@ describe("Patch route", function() {
...
+                expect(parseInt($newTaskCount)).is.gt(
+                  parseInt($initialTaskCount)
+                );
+              });
+          });
+      });
+
+      it("Stops fetching tasks when all tasks have been fetched", () => {
+        scrollTasksTableUntilAllTasksFetched({ hasMore: true });
+        cy.get(".ant-table-body").scrollTo("bottom", { duration: 300 });
```
SupaJoon: there is a chance that the table has fully scrolled to the bottom by line 206. It may be a better to scroll all the way up and all the way down again to confirm no additional data will be fetched from scrolling.

tgrander:
evergreen retry

src/pages/patch/patchTabs/Tasks.tsx
```diff
@@ -72,12 +116,17 @@ export const Tasks: React.FC<Props> = ({ taskCount }) => {
 const getString = (param: string | string[]): string =>
   Array.isArray(param) ? param[0] : param;
 
-const getQueryVariablesFromUrlSearch = (patchId: string, search: string) => {
+const getQueryVariablesFromUrlSearch = (
+  patchId: string,
+  search: string,
+  page: number
```
SupaJoon: nit: page isn't actually from the url search params and this function isn't accessing url search params directly. we could probably change the name of the function to exclude `UrlSearch`

PR #105, opened 2020-03-24
EVG-7620: Prevent app crash when switching back and forth between page tabs

SupaJoon:
https://jira.mongodb.org/browse/EVG-7620

`Unhandled Rejection (Invariant Violation): ObservableQuery with this id doesn't exist` was caused by `fetchMore` processing a response after the component containing the useQuery has unmounted. 

This happens when switching between tabs because `history.listen` in the currently rendered component executes its callback fn (and fetchMore) with the new page route before unregistering the listener. 

To prevent the app from crashing I surrounded all calls to fetchMore  in history.listen callbacks with a try/catch. 
Another solution could have been to execute fetchMore only if the location variable in the history.listen callback matches the component route - this solution prevents a superfluous request from being made but would require parsing the location variable and matching against a route constant. 

The reason we were getting a 422 error was because fetchMore was using invalid query params from a different page.

### DIFF
--- a/src/pages/patch/patchTabs/Tasks.tsx
+++ b/src/pages/patch/patchTabs/Tasks.tsx
@@ -35,24 +35,20 @@ export const Tasks: React.FC<Props> = ({ taskCount }) => {
 
   // fetch tasks when url params change
   useEffect(() => {
-    return history.listen(async location => {
+    history.listen(location => {
       if (networkStatus === NetworkStatus.ready && !error) {
-        // catch errors resulting from async requests
-        // that return after component unmounts
-        try {
-          await fetchMore({
-            variables: getQueryVariablesFromUrlSearch(id, location.search),
-            updateQuery: (
-              prev: PatchTasksQuery,
-              { fetchMoreResult }: { fetchMoreResult: PatchTasksQuery }
-            ) => {
-              if (!fetchMoreResult) {
-                return prev;
-              }
-              return fetchMoreResult;
+        fetchMore({
+          variables: getQueryVariablesFromUrlSearch(id, location.search),
+          updateQuery: (
+            prev: PatchTasksQuery,
+            { fetchMoreResult }: { fetchMoreResult: PatchTasksQuery }
+          ) => {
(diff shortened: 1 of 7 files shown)


src/pages/patch/patchTabs/Tasks.tsx
```diff
@@ -35,20 +35,26 @@ export const Tasks: React.FC<Props> = ({ taskCount }) => {
...
-        fetchMore({
-          variables: getQueryVariablesFromUrlSearch(id, location.search),
-          updateQuery: (
-            prev: PatchTasksQuery,
-            { fetchMoreResult }: { fetchMoreResult: PatchTasksQuery }
-          ) => {
-            if (!fetchMoreResult) {
-              return prev;
+        // catch errors resulting from async requests
+        // that return after component unmounts
```
tgrander: it is standard for all async/await code to be wrapped in a try/catch; therefore these comments are not necessary

src/pages/task/testsTable/TestsTableCore.tsx
```diff
@@ -72,25 +72,31 @@ export const TestsTableCore: React.FC<ValidInitialQueryParams> = ({
...
-          testName
-        },
-        updateQuery: (
-          prev: UpdateQueryArg,
-          { fetchMoreResult }: { fetchMoreResult: UpdateQueryArg }
-        ) => {
-          if (!fetchMoreResult) {
-            return prev;
+      // catch errors resulting from async requests
+      // that return after component unmounts
```
tgrander: same thing with this comment

SupaJoon:
> should the same thing be done for the patch page `history.listen`?

Yes, anywhere where we use history.listen with fetchMore should have a try/catch. I included the try/catch on the patch page here: https://github.com/evergreen-ci/spruce/pull/105/files#diff-345b351a35b76219866aa63f353fa764R38

SupaJoon:
evergreen retry

SupaJoon:
ticket for flakey test: https://jira.mongodb.org/browse/EVG-7649

PR #106, opened 2020-03-24
EVG-7631: map task status to color

tgrander:
https://jira.mongodb.org/browse/EVG-7631

All task statuses in the task table correspond to their respective color. 

task statuses

### Testing
This is exactly the kind of case where visual testing shines. We do not have it set up yet but will soon with Storybook + Cypress.


src/pages/patch/patchTabs/tasks/TasksTable.tsx
```diff
@@ -62,6 +64,65 @@ const getSortDirFromOrder = (order: "ascend" | "descend") =>
...
+      </Badge>
+    );
+  } else if (status in mapUnsupportedBadgeColors) {
+    return (
+      <StyledBadge key={status} {...mapUnsupportedBadgeColors[status]}>
+        {status}
+      </StyledBadge>
+    );
+  }
+  throw new Error(`Status '${status}' is not a valid task status`);
```
SupaJoon: I think we should have an error boundary around the task table so the user can continue using the App if this error is thrown.

tgrander: an error boundary would just contain the error to the task table, it would not make it silent. i decided to throw an error here because really this should never happen and if it does it should not be silent - on the contrary it should be loud. i made a ticket to report errors and their severity to bugsnag, which would be what is ideally done here: https://jira.mongodb.org/browse/EVG-7645

SupaJoon: Since we send errors to bugsnag, I don't think we need to sacrifice UX for error reporting. In the case that this error is thrown, the user won't be able to use the application because they will see an error screen. We have the Bugsnag error boundary component already installed in our App for this purpose.

src/components/ErrorBoundary/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+  }
+
+  static getDerivedStateFromError() {
+    return { hasError: true };
+  }
+
+  render() {
+    if (this.state.hasError) {
+      // TODO: render error banner
+      return <h1>Something went wrong.</h1>;
```
john-m-liu: Even if this is temporary, we should be log the error somewhere shouldn't we?

src/pages/patch/patchTabs/Tasks.tsx
```diff
@@ -91,7 +92,7 @@ export const Tasks: React.FC<Props> = ({ taskCount }) => {
     return <div>{error.message}</div>;
   }
   return (
-    <>
+    <ErrorBoundary>
```
john-m-liu: This is probably a generalized question, but when do you want to put the error boundary around a specific component like this, vs the entire app?

src/pages/patch/patchTabs/tasks/TasksTable.tsx
```diff
@@ -62,6 +64,65 @@ const getSortDirFromOrder = (order: "ascend" | "descend") =>
 
 const rowKey = ({ id }: { id: string }): string => id;
 
+const mapTaskStatusToBadgeVariant = {
+  [TaskStatus.Inactive]: Variant.LightGray,
+  [TaskStatus.Unstarted]: Variant.LightGray,
+  [TaskStatus.Undispatched]: Variant.LightGray,
+  [TaskStatus.Started]: Variant.Yellow,
+  [TaskStatus.Dispatched]: Variant.LightGray,
```
john-m-liu: Dispatched should be the same yellow that started is

src/pages/patch/patchTabs/tasks/TasksTable.tsx
```diff
@@ -62,6 +64,65 @@ const getSortDirFromOrder = (order: "ascend" | "descend") =>
...
+  [TaskStatus.Succeeded]: Variant.Green,
+  [TaskStatus.Failed]: Variant.Red,
+  [TaskStatus.StatusBlocked]: Variant.DarkGray,
+  [TaskStatus.StatusPending]: Variant.Yellow
```
john-m-liu: Pending should be the same grey that unstarted is

src/pages/patch/patchTabs/Tasks.tsx
tgrander: ErrorBoundaries can be used anywhere an error should be contained - so they can and will be used around the entire app as well as around specific components.

src/pages/patch/patchTabs/tasks/TasksTable.tsx
tgrander: thanks!!

src/pages/patch/patchTabs/tasks/TasksTable.tsx
tgrander: ty!

src/components/ErrorBoundary/index.tsx
tgrander: for sure. we have not logged any bugs yet. i made a ticket to go round and log bugs based on severity. https://jira.mongodb.org/browse/EVG-7645

PR #109, opened 2020-03-25
EVG-7509: show patch status badge in header

tgrander:
### Designs
Screen Shot 2020-03-25 at 12 06 10 PM

### Actual UI
Screen Shot 2020-03-25 at 12 13 36 PM


src/pages/Patch.tsx
```diff
@@ -51,3 +65,14 @@ export const Patch = () => {
     </PageWrapper>
   );
 };
+
+const BadgeWrapper = styled.span`
+  display: inline-flex;
+`;
+
+const mapPatchStatusToBadgeVariant = {
```
SupaJoon: super nit: i wouldn't include map in the variable name

tgrander: i would

PR #110, opened 2020-03-25
EVG-7520: variant input field updates variant url param onChange

tgrander:
https://jira.mongodb.org/browse/EVG-7520
#### Created a reusable hook based off the testName filter logic from the TestTable component that updates the url param with user input
- updates the VARIANT url param when user types into input
- removes VARIANT url param if user deletes all text from input

tasks variant filter input

### DIFF
--- a/cypress/integration/patch/patch-route/tasks-filters.js
+++ b/cypress/integration/patch/patch-route/tasks-filters.js
@@ -29,18 +29,10 @@ describe("Tasks filters", function() {
 
   describe("Variant input field", () => {
     it("Updates the VARIANT url search param when input changes", () => {
-      cy.get("[data-cy=variant-input]")
-        .type(variantInputValue)
-        .then(() => {
-          locationHasUpdatedVariantParam(variantInputValue);
-        })
-        .then(() => {
-          cy.get("[data-cy=variant-input]")
-            .clear()
-            .then(() => {
-              locationHasUpdatedVariantParam(null);
-            });
-        });
+      cy.get("[data-cy=variant-input]").type(variantInputValue);
+      locationHasUpdatedVariantParam(variantInputValue);
+      cy.get("[data-cy=variant-input]").clear();
+      locationHasUpdatedVariantParam(null);
     });
   });
 });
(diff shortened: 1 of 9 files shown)


src/hooks/useFilterInputChangeHandler.ts
```diff
@@ -0,0 +1,52 @@
...
+ * @param  {(path:string)=>void} replace method from react router to update ur
+ * @return {[string, (e: InputEvent) => void]}
+ * pass {string} return value to input component as `value`
+ * pass {(e: InputEvent) => void} return value to input component as onChange prop
+ */
+export const useFilterInputChangeHandler = (
+  urlSearchParam: string,
+  pathname: string,
+  search: string,
+  replace: (path: string) => void
```
SupaJoon: since this is a hook, you can import useHistory and remove replace

SupaJoon: we can import useHistory and remove this argument

cypress/integration/patch/patch-route/tasks-filters.js
```diff
@@ -0,0 +1,46 @@
...
+  describe("Variant input field", () => {
+    it("Updates the VARIANT url search param when input changes", () => {
+      cy.get("[data-cy=variant-input]")
+        .type(variantInputValue)
+        .then(() => {
+          locationHasUpdatedVariantParam(variantInputValue);
+        })
+        .then(() => {
+          cy.get("[data-cy=variant-input]")
+            .clear()
```
SupaJoon: you can chain off the parent `then` by returning `cy.get("[data-cy=variant-input]").clear()`

src/hooks/useFilterInputChangeHandler.ts
```diff
@@ -0,0 +1,52 @@
...
+);
+
+type InputEvent = React.ChangeEvent<HTMLInputElement>;
+
+/**
+ * useFilterInputChangeHandler updates the url query param (urlSearchParam) by the
+ * the value user enters into the input field.
+ * @param  {string} urlSearchParam the url search param that should update
+ * @param  {string} pathname url pathname
+ * @param  {string} search url search
```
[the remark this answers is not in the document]
tgrander: yeah true. the hook is going to be twice for the tasks so i figured it would be more performant to pass location as a parameter. but now that you mention it i don't think it will be a problem to call it in the hook.

tgrander:
@john-m-liu nope that's just a different ticket :)

PR #111, opened 2020-03-26
EVG-7518: fetch tasks filtered by variant

tgrander:
filter tasks variant


cypress/integration/patch/patch-route/tasks-filters.js
```diff
@@ -28,11 +16,38 @@ describe("Tasks filters", function() {
...
+      filteredTasksAreFetched("variant", variantInputValue);
       cy.get("[data-cy=variant-input]").clear();
       locationHasUpdatedVariantParam(null);
     });
   });
 });
+
+const filteredTasksAreFetched = (variable, value) => {
+  cy.wait(200);
+  waitForGQL("@gqlQuery", "PatchTasks");
```
SupaJoon: The third argument to waitForGQL is a dictionary to match an xhr object path to a key so you may not need `cy.wait(200)` in case you are running into a race condition waiting for request you want. 
https://github.com/evergreen-ci/spruce/blob/new-ui-feature-branch/cypress/integration/test_table.js#L175

tgrander: i thought that's what it might do but i was having a really hard time reading and understanding that code so I erred on the side of caution and did not pass options. that's the kind of code an explicit comment would do wonders for in answering the 'why'

SupaJoon: I included comments for the definition
https://github.com/evergreen-ci/spruce/blob/new-ui-feature-branch/cypress/utils/networking.js#L6

SupaJoon: I can change the name of the parameter to `match` instead of `options`. That might make it clearer

tgrander: i could not find any examples of `waitForGQL` being used with `options` param so i am going to merge as is and will edit in future PR when there is a comment and example

PR #112, opened 2020-03-26
EVG-7521: filter tasks by task name

tgrander:
### Input field that updates url with input value and fetches tasks filtered by task name

filter task nameeee


cypress/integration/patch/patch-route/tasks-filters.js
```diff
@@ -19,20 +19,30 @@ describe("Tasks filters", function() {
...
       cy.get("[data-cy=variant-input]").clear();
       locationHasUpdatedVariantParam(null);
     });
   });
+
+  describe("Task name input field", () => {
+    const taskNameInputValue = "test-cloud";
+    it("Updates url with input value and fetches tasks filtered by task name", () => {
+      cy.get("[data-cy=task-name-input]").type(taskNameInputValue);
+      locationHasUpdatedVariantParam(taskNameInputValue, "taskName");
```
SupaJoon: we should remove "Variant" from this function name since we are also using it for task name. i also think this could be useful fo other tests and could belong in the utils folder.

src/pages/patch/patchTabs/Tasks.tsx
```diff
@@ -16,6 +16,7 @@ import { P2 } from "components/Typography";
 import { ErrorBoundary } from "components/ErrorBoundary";
 import { TaskFilters } from "pages/patch/patchTabs/tasks/TaskFilters";
 import { TaskSortBy } from "gql/queries/get-patch-tasks";
```
SupaJoon: this is no longer used

PR #113, opened 2020-03-26
EVG-7395: Task page title and status badge

SupaJoon:
https://jira.mongodb.org/browse/EVG-7395

This PR is mostly refactoring the TaskStatusBadge so we can use it on the task page as well as the task table. 

Screen Shot 2020-03-26 at 2 50 52 PM

### DIFF
--- a/src/components/TaskStatusBadge.tsx
+++ b/src/components/TaskStatusBadge.tsx
@@ -1,0 +1,63 @@
+import React from "react";
+import Badge, { Variant } from "@leafygreen-ui/badge";
+import { TaskStatus } from "types/task";
+import styled from "@emotion/styled/macro";
+
+const mapTaskStatusToBadgeVariant = {
+  [TaskStatus.Inactive]: Variant.LightGray,
+  [TaskStatus.Unstarted]: Variant.LightGray,
+  [TaskStatus.Undispatched]: Variant.LightGray,
+  [TaskStatus.Started]: Variant.Yellow,
+  [TaskStatus.Dispatched]: Variant.Yellow,
+  [TaskStatus.Succeeded]: Variant.Green,
+  [TaskStatus.Failed]: Variant.Red,
+  [TaskStatus.StatusBlocked]: Variant.DarkGray,
+  [TaskStatus.StatusPending]: Variant.LightGray
+};
+
+const failureColors = {
+  text: "#800080",
+  border: "#CC99CC",
+  fill: "#E6CCE6"
+};
+
+// the status colors that are not supported by the leafygreen Badge variants
+const mapUnsupportedBadgeColors = {
+  [TaskStatus.SystemFailed]: failureColors,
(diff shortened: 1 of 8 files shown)


src/pages/Task.tsx
```diff
@@ -82,9 +88,21 @@ export const Task: React.FC = () => {
...
+      {loading ? (
+        <PageHeader>
+          <Skeleton active={true} paragraph={{ rows: 0 }} />
+        </PageHeader>
+      ) : displayName || status ? (
+        <PageHeader>
+          <H2 id="task-name">{displayName}</H2>
+          {"  "}
+          <BadgeWrapper>
+            <ErrorBoundary>
```
john-m-liu: Just curious - is there a specific reason you're adding an error boundary around this component?

SupaJoon: TaskStatusBadge throws an error
https://github.com/evergreen-ci/spruce/pull/113/files#diff-a6da40343c81cd82a8f283202322bfc9R62**

src/components/TaskStatusBadge.tsx
```diff
@@ -0,0 +1,63 @@
...
+  fill: string;
+  text: string;
+}
+
+// only use for statuses whose color is not supported by leafygreen badge variants, i.e. SystemFailed, TestTimedOut, SetupFailed
+const StyledBadge = styled(Badge)`
+  border-color: ${(props: BadgeColorProps) => props.border} !important;
+  background-color: ${(props: BadgeColorProps) => props.fill} !important;
+  color: ${(props: BadgeColorProps) => props.text} !important;
+`;
```
tgrander: if you're moving these somewhere to be shared two comonents then i would think it would be in a shared folder/file as opposed to another component.

src/pages/Task.tsx
```diff
@@ -82,9 +88,21 @@ export const Task: React.FC = () => {
...
+              <TaskStatusBadge status={status} />
+            </ErrorBoundary>
+          </BadgeWrapper>
+        </PageHeader>
+      ) : null}
```
tgrander: this is the code i'm referring to that should be made into a shared component

src/components/TaskStatusBadge.tsx
SupaJoon: src/components is the "shared" folder according to fractal pattern

SupaJoon: ah, I think I know what you mean. I didnt export styled badge bc right now it's only used by TaskStatusBadge. TaskStatusBadge is the shared component.

src/pages/Task.tsx
```diff
@@ -109,3 +123,8 @@ export const Task: React.FC = () => {
     </PageWrapper>
   );
 };
+
+const BadgeWrapper = styled.span`
+  top: -2px;
+  position: relative;
+`;
```
tgrander: seems like this could/should be used for the patch page title too?

PR #118, opened 2020-03-31
EVG-7526: patch page - base status dropdown to filter tasks

tgrander:
https://jira.mongodb.org/browse/EVG-7526

base statuses


cypress/integration/patch/patch-route/tasks-filters.js
```diff
@@ -32,111 +94,80 @@ describe("Tasks filters", function() {
...
-    });
+      it("Clicking on 'All' checkbox adds all statuses to URL. Clicking again removes all statuses.", () => {
+        cy.get(".cy-checkbox")
+          .contains("All")
+          .then($all => {
+            $all.click();
+            locationHasUpdatedFilterParam(allStatuses, urlParam);
+            filteredTasksAreFetchedAndRendered();
+            $all.click();
+            locationHasUpdatedFilterParam(null);
```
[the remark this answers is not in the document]
tgrander: look at the function - it does

src/pages/patch/patchTabs/tasks/TaskFilters.tsx
```diff
@@ -17,17 +18,22 @@ export const TaskFilters: React.FC = () => {
...
   );
+  const [baseStatusesVal, baseStatusesValOnChange] = useStatusesFilter(
+    PatchTasksQueryParams.BaseStatuses
+  );
 
   return (
     <FiltersWrapper>
-      <StyledInput
+      <Input
+        style={{ width: "25%" }}
```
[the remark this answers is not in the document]
tgrander: this is an antd prop and doesn't matter because is temporary until you make the fixed width fix

cypress/integration/patch/patch-route/tasks-filters.js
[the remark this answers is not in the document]
SupaJoon: if you call this function with a single param `null`
then we are asserting this logic:
expect(loc.search).to.not.include(undefined);
im not sure what that's supposed to test

john-m-liu:
> There is a bug where filtering by task base status or task status makes the table rows disappear.
> To reproduce:
> 
> 1. go to `http://localhost:3000/patch/5e4ff3abe3c3317e352062e4/tasks`
> 2. check Failures in Task Base Status
> 3. check Test Timed out in Task Base Status

I noticed the same thing, but it also happens in new-ui-feature-branch so I ignored it. Definitely worth a ticket though

cypress/utils/index.js
```diff
@@ -0,0 +1,79 @@
...
+    requestVariables: [],
+  }).then(() =>
+    urlSearchParamsAreUpdated({ pathname, paramName, search: null })
+  );
+};
+
+export const assertQueryVariables = (queryName, variables = {}) => {
+  const options = {};
+  Object.entries(variables).forEach(([variable, value]) => {
+    if (variable.includes("status")) {
```
SupaJoon, line 34: checking the variable name seems fragile since the variable name could change. we can change this condition to `Array.isArray(value)` and handle all arrays the same way.

cypress/utils/index.js
```diff
@@ -0,0 +1,79 @@
+import { waitForGQL } from "./networking";
+
+// used to test status and base status dropdown filters
+export const clickingCheckboxFetchesFilteredTasksAndUpdatesUrl = ({
```
SupaJoon, line 4: can we include something in this function name to indicate that it tests the rendered items? also we should remove "tasks" since we're using this in the test table as well.

src/pages/patch/patchTabs/Tasks.tsx
```diff
@@ -138,13 +138,31 @@ const statusesToIncludeInQuery = {
   [TaskStatus.Unstarted]: true
 };
 
+const getStatuses = (rawStatuses: string[] | string) => {
+  const statuses = getArray(rawStatuses).filter(
+    status => status in statusesToIncludeInQuery
+  );
+  if (
```
SupaJoon, line 145: we can remove this and apply the fix we came up with earlier where we save the variables for useQuery in a useState to ensure they don't change between renders.

src/components/TreeSelect.tsx
```diff
@@ -320,7 +322,7 @@ const OptionsWrapper = styled.div`
   box-shadow: 0 3px 8px 0 rgba(231, 238, 236, 0.5);
   position: absolute;
   z-index: 5;
-  width: 352px;
+  width: width: 100%;
```
SupaJoon, line 325: can we revert the width changes since I'll handle the width fix in a separate ticket?

cypress/utils/index.js
```diff
@@ -0,0 +1,79 @@
...
+        );
+      };
+    } else {
+      options[`requestBody.variables[${variable}]`] = value;
+    }
+  });
+  waitForGQL("@gqlQuery", queryName, options);
+};
+
+export const resultsAreFechedAndRendered = ({
```
SupaJoon, line 50: missing "t" in fetched

src/components/TreeSelect.tsx
tgrander: the tests fail due to elements with static widths overing each other. so no sorry

src/pages/patch/patchTabs/Tasks.tsx
tgrander: you already applied the fix in a previous PR of yours

src/components/TreeSelect.tsx
SupaJoon: there is a typo in this line

PR #122, opened 2020-04-02
EVG-7611: Test table uses default GQL query variables and does not push query params to URL

SupaJoon:
https://jira.mongodb.org/browse/EVG-7611
This PR changes the Test table so it uses default query variables in the taskTest GQL query if the url query params are not recognizable. 
This PR also adjusts the default table state to query Status Ascending because this will display failed tests at the top of the list.


cypress/integration/test_table.js
```diff
@@ -173,16 +193,57 @@ describe("Tests Table", function() {
...
+        .invoke("toArray")
+        .then($initialTasks => {
+          // need to overscroll to trigger fetch
+          cy.get(".ant-table-body").scrollTo(0, "101%", { duration: 500 });
+          assertQueryVariables("STATUS", "ASC", [], "", 1);
+          cy.get(".ant-table-row").should(
+            "have.length.greaterThan",
+            $initialTasks.length
+          );
+        });
```
tgrander, line 217: in future, i think it would be beneficial for pages/features of the app that are more or less identical to be tested in exactly the same way. for example, it's interesting that tests for this feature were already written via the task table scrolling tests. but you chose to rewrite the tests in a slightly different way. if you think this way is better then should this approach be generalized and applied to the task table? it seems inefficient to be rewriting tests for features that exactly the same give or take a url parameter.

SupaJoon: good point

PR #124, opened 2020-04-07
EVG-7549: add commit queue checkbox

dominoweir:


src/pages/MyPatches.tsx
```diff
@@ -1,5 +1,27 @@
...
-  return <div>My Patches Page</div>;
+  const { search, pathname } = useLocation();
+  const history = useHistory();
+  const parsed = queryString.parse(search);
+  let showCommitQueue = (parsed.commitQueue === "true");
+  return (
+    <div>
+      <div>My Patches Page</div>
+      <Checkbox
+        onChange={() => {
```
dominoweir, line 15: I'm not sure if this should be in its own separate function, left it inline for now

dominoweir:
My main lingering question on this is how to ensure the box is checked initially, just because I'm not clear on how someone arrives at this page/if the URL will already contain commitQueue=true when they do.

src/pages/MyPatches.tsx
dominoweir: moved to a function after discovering the linter takes issue with lambdas

SupaJoon:
> My main lingering question on this is how to ensure the box is checked initially, just because I'm not clear on how someone arrives at this page/if the URL will already contain commitQueue=true when they do.

The checkbox would be checked on initial render if the commitQueue query param was included in the URL on initial render. This scenario would happen if you were to copy and paste the URL to someone after clicking the checkbox.

src/pages/MyPatches.tsx
```diff
@@ -1,5 +1,30 @@
...
+  const { search, pathname } = useLocation();
+  const history = useHistory();
+  const parsed = queryString.parse(search);
+  let showCommitQueue = parsed.commitQueue === "true";
+
+  function onCheckboxChange() {
+    showCommitQueue = !showCommitQueue;
+    history.push({
+      pathname: pathname,
+      search: "?commitQueue=" + showCommitQueue
```
SupaJoon, line 16: this line will have to be updated as we add more query params to the page which could make things confusing to maintain later on. Check out this function that passes on existing query params https://github.com/evergreen-ci/spruce/blob/new-ui-feature-branch/src/pages/patch/patchTabs/tasks/TasksTable.tsx#L31

src/pages/MyPatches.tsx
```diff
@@ -1,5 +1,30 @@
...
+import { useLocation, useHistory } from "react-router-dom";
+import queryString from "query-string";
+import Checkbox from "@leafygreen-ui/checkbox";
 
 export const MyPatches = () => {
-  return <div>My Patches Page</div>;
+  const { search, pathname } = useLocation();
+  const history = useHistory();
+  const parsed = queryString.parse(search);
+  let showCommitQueue = parsed.commitQueue === "true";
```
SupaJoon: `commitQueue` should be saved inside an interface that defines query params for this page like what we did for Tasks table on the Patch page: https://github.com/evergreen-ci/spruce/blob/new-ui-feature-branch/src/types/task.ts#L10

src/pages/MyPatches.tsx
```diff
@@ -1,5 +1,35 @@
...
+import queryString from "query-string";
+import Checkbox from "@leafygreen-ui/checkbox";
+import { MyPatchesQueryParams } from "types/task";
 
 export const MyPatches = () => {
-  return <div>My Patches Page</div>;
+  const { search, pathname } = useLocation();
+  const { replace } = useHistory();
+  const parsed = queryString.parse(search);
+  let showCommitQueue = parsed.commitQueue
```
SupaJoon, line 11: this can be showCommitQueue = parsed.commitQueue === "true"

src/pages/MyPatches.tsx
```diff
@@ -1,5 +1,35 @@
...
+    ? parsed.commitQueue === "true"
+    : true;
+
+  function onCheckboxChange() {
+    showCommitQueue = !showCommitQueue;
+    replace(
+      `${pathname}?${queryString.stringify({
+        ...queryString.parse(search),
+        [MyPatchesQueryParams.CommitQueue]: showCommitQueue
```
SupaJoon, line 20: and here the value can be !showCommitQueue

src/pages/MyPatches.tsx
```diff
@@ -1,5 +1,35 @@
...
-  return <div>My Patches Page</div>;
+  const { search, pathname } = useLocation();
+  const { replace } = useHistory();
+  const parsed = queryString.parse(search);
+  let showCommitQueue = parsed.commitQueue
+    ? parsed.commitQueue === "true"
+    : true;
+
+  function onCheckboxChange() {
+    showCommitQueue = !showCommitQueue;
```
SupaJoon, line 16: we should avoid reassigning variables when we can

src/types/task.ts
```diff
@@ -17,6 +17,10 @@ export enum PatchTasksQueryParams {
   TaskName = "taskName"
 }
 
+export enum MyPatchesQueryParams {
```
SupaJoon, line 20: we should move enum to `src/types/patch.ts`

src/pages/MyPatches.tsx
```diff
@@ -1,5 +1,32 @@
...
+import { MyPatchesQueryParams } from "types/patch";
 
 export const MyPatches = () => {
-  return <div>My Patches Page</div>;
+  const { search, pathname } = useLocation();
+  const { replace } = useHistory();
+  const parsed = queryString.parse(search);
+  const showCommitQueue = parsed.commitQueue === "true" || parsed.commitQueue === undefined;
+
+  function onCheckboxChange() {
```
SupaJoon, line 13: can we convert this to an arrow function?

dominoweir:
Tests have been added 👍

cypress/integration/my_patches.js
```diff
@@ -0,0 +1,31 @@
...
+  });
+
+  it("Should render with Show Commit Queue box checked when commitQueue not indicated in URL query param", () => {
+    cy.visit(MY_PATCHES_ROUTE);
+    cy.get(".cy-checkbox")
+      .find("input")
+      .should("be.checked");
+  });
+
+  it("Should render with Show Commit Queue box checked when commitQueue is false in URL query param", () => {
```
tgrander, line 18: i think this should be "Should render with Show Commit Queue box UNchecked when commitQueue is false in URL query param"

cypress/integration/my_patches.js
```diff
@@ -0,0 +1,31 @@
...
+  "/my-patches";
+describe("my patches page", function() {
+  beforeEach(() => {
+    cy.server();
+    cy.login();
+  });
+
+  it("Should render with Show Commit Queue box checked when commitQueue not indicated in URL query param", () => {
+    cy.visit(MY_PATCHES_ROUTE);
+    cy.get(".cy-checkbox")
```
tgrander, line 13: some of the old selectors will be classes but we have moved to using data-cy attributes on the elements for testing.
```
// JSX
<div data-cy="commit-queue-checkbox">I'm the checkbox element</div>

// test file
cy.get("[data-cy=commit-queue-checkbox]")
```

PR #126, opened 2020-04-08
EVG-7351: Status of patch should auto update

dominoweir:


src/pages/Patch.tsx
```diff
@@ -19,12 +19,16 @@ import { PatchStatus } from "gql/queries/get-patch-tasks";
 
 export const Patch = () => {
   const { id } = useParams<{ id: string }>();
-  const { data, loading, error } = useQuery<PatchQuery>(GET_PATCH, {
-    variables: { id: id }
+  const { data, loading, error, stopPolling } = useQuery<PatchQuery>(GET_PATCH, {
+    variables: { id: id },
+    pollInterval: 500
```
tgrander, line 24: maybe change to 2 seconds? it's an arbitrary number that will reduce number of network requests being fired x4 while still appearing immediate

dominoweir: Sounds good to me, 500ms was an arbitrary number anyways

PR #129, opened 2020-04-09
EVG-7759: npm commands to facilitate deployment

tgrander:
### Before:
- Run build command for respective deployment env (prod or staging)
- Run deploy command for respective deployment env (prod or staging)

#### Problems with this approach
- Easy to build for the wrong environment, e.g. build for prod and deploy to staging
- Easy to forget to build before deploying
- Had to include the bucket name to deploy. What is the bucket name? Why do I need to include it every single time? 
  - Easy to forget to include bucket name

### After:
- Run deploy command for respective deployment env (prod or staging)

#### Problems solved by this approach
- You only have to run one command to deploy which abstracts all of the details required to deploy, i.e. bucket name and running the right build step first


tgrander:
@john-m-liu these are the exact same commands and exact same deployment process that currently exists but re-organized to be more straight forward

PR #140, opened 2020-04-20
Evg 7791 add commit queue page

khelif96:
Add the commit queue page

add mutation for removing patch from commit queue


src/gql/queries/patch.ts
```diff
@@ -1,27 +1,5 @@
 import gql from "graphql-tag";
-
-export interface Patch {
```
khelif96, line 3: moved these types to `types/patch.ts`

khelif96:
So im getting a linter error 
`[2020/04/20 20:02:43.662] ERROR: src/pages/commitqueue/codeChangesModule/CodeChangesModule.tsx:50:33 - Lambdas are forbidden in JSX attributes due to their rendering performance impact`

For doing something like this `<Component onClick={() => doSomething(someVar)} />`

Could we disable this rule its a pretty common pattern  and practice to create inline lambda functions for passing a function into a component. The rule in question is `jsx-no-lambda`

src/gql/queries/get-commit-queue.ts
```diff
@@ -0,0 +1,42 @@
+import gql from "graphql-tag";
+import { Patch } from "../../types/patch";
+export const GET_COMMIT_QUEUE = gql`
+  query CommitQueue($id: String!) {
+    commitQueue(id: $id) {
+      ProjectID
+      Queue {
+        Issue
+        EnqueueTime
+        Patch {
```
tgrander, line 10: must have not caught these in your backend PR - field names should never be uppercased

src/pages/CommitQueue.tsx
```diff
@@ -0,0 +1,81 @@
...
+import { PageWrapper } from "components/styles";
+import { useParams } from "react-router-dom";
+import { Subtitle } from "@leafygreen-ui/typography";
+import Badge from "@leafygreen-ui/badge";
+import { uiColors } from "@leafygreen-ui/palette";
+import { useQuery } from "@apollo/react-hooks";
+import get from "lodash/get";
+import { Skeleton } from "antd";
+
+import { CommitQueueCard } from "./commitqueue/CommitQueueCard";
```
tgrander, line 12: please always prefer absolute imports over relative imports

src/pages/CommitQueue.tsx
```diff
@@ -0,0 +1,81 @@
...
+import {
+  GET_COMMIT_QUEUE,
+  CommitQueueQuery
+} from "../gql/queries/get-commit-queue";
```
tgrander, line 16: same comment about absolute imports

src/pages/CommitQueue.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+const HR = styled("hr")`
+  background-color: ${gray.light2};
+  border: 0;
+  height: 3px;
+`;
+
+const ErrorWrapper = styled.div`
+  word-wrap: break-word;
+`;
```
tgrander, line 40: styles always at bottom of module please. order of module is typically props/component/styles/everything else

src/pages/CommitQueue.tsx
```diff
@@ -0,0 +1,81 @@
...
+    return (
+      <ErrorWrapper data-cy="metadata-card-error">{error.message}</ErrorWrapper>
+    );
+  }
+  const commitQueue = get(data, "commitQueue");
+  const queue = get(commitQueue, "Queue");
+  return (
+    <PageWrapper>
+      <Header>
+        <PageTitle> Commit Queue </PageTitle>
```
tgrander, line 63: page title from the other pages can be reused here

src/pages/CommitQueue.tsx
```diff
@@ -0,0 +1,81 @@
...
+  align-items: center;
+  margin-top: 24px;
+  margin-bottom: 16px;
+`;
+const PageTitle = styled(Subtitle)`
+  margin-right: 16px;
+  font-size: 18px;
+`;
+
+const HR = styled("hr")`
```
tgrander, line 32: what is HR?

src/pages/CommitQueue.tsx
```diff
@@ -0,0 +1,81 @@
...
+  if (error) {
+    return (
+      <ErrorWrapper data-cy="metadata-card-error">{error.message}</ErrorWrapper>
+    );
+  }
+  const commitQueue = get(data, "commitQueue");
+  const queue = get(commitQueue, "Queue");
+  return (
+    <PageWrapper>
+      <Header>
```
tgrander, line 62: these types of components can be reused from other pages

src/pages/commitqueue/CommitQueueCard.tsx
```diff
@@ -0,0 +1,75 @@
+import React from "react";
+import styled from "@emotion/styled";
+import { Subtitle, Body } from "@leafygreen-ui/typography";
+import { uiColors } from "@leafygreen-ui/palette";
+import Button from "@leafygreen-ui/button";
+import { CodeChangeModules } from "./codeChangesModule/CodeChangesModule";
+import { ModuleCodeChanges } from "types/patch";
+import { prettyDate } from "utils/string";
+interface Props {
```
tgrander, line 9: nit: space between imports and everything else

src/pages/commitqueue/CommitQueueCard.tsx
```diff
@@ -0,0 +1,75 @@
...
+  display: flex;
+  flex-direction: column;
+  grid-area: 1 / 1 / 2 / 2;
+  margin-left: 16px;
+  padding-bottom: 24px;
+  width: 100%;
+`;
+const CardMetaData = styled(Body)`
+  color: ${gray.dark2};
+`;
```
tgrander, line 36: same comment about styles being at bottom of module

src/pages/commitqueue/codeChangesModule/CodeChangeItem.tsx
```diff
@@ -0,0 +1,69 @@
...
+`;
+
+const FileDiffTextContainer = styled("span")`
+  ${(props: { type: string; hasValue: boolean }) =>
+    props.hasValue &&
+    (props.type === "+" ? `color: ${green.base};` : `color: ${red.base};`)}
+  &:nth-child(2) {
+    margin-left: 16px;
+  }
+`;
```
tgrander, line 31: same comment about styles being at bottom of module

src/utils/string/index.ts
```diff
@@ -21,3 +21,24 @@ export const msToDuration = (ms: number): string => {
...
+  hours = hours % 12;
+  hours = hours ? hours : 12; // the hour '0' should be '12'
+  const minutes = date.getMinutes();
+  const seconds = date.getSeconds();
+  const ampm = hours >= 12 ? "pm" : "am";
+
+  return `${formatField(month)}/${formatField(day)}/${year} at ${formatField(
+    hours
+  )}:${formatField(minutes)}:${formatField(seconds)} ${ampm}`;
+};
```
tgrander, line 44: we use date-fns to format dates on the client. you can delete this code and use that

src/pages/commitqueue/codeChangesModule/CodeChangesModule.tsx
```diff
@@ -0,0 +1,86 @@
...
+    <div>
+      {moduleCodeChanges.map(moduleCodeChange => (
+        <CodeChangeModule
+          key={moduleCodeChange.rawLink}
+          moduleCodeChange={moduleCodeChange}
+        />
+      ))}
+    </div>
+  );
+};
```
tgrander, line 86: is this component necessary? imo no because it's just a map function.

src/pages/commitqueue/CommitQueueCard.tsx
```diff
@@ -0,0 +1,75 @@
+import React from "react";
+import styled from "@emotion/styled";
+import { Subtitle, Body } from "@leafygreen-ui/typography";
+import { uiColors } from "@leafygreen-ui/palette";
+import Button from "@leafygreen-ui/button";
+import { CodeChangeModules } from "./codeChangesModule/CodeChangesModule";
```
tgrander, line 6: please replace all relative imports with absolute imports

src/pages/CommitQueue.tsx
khelif96: https://www.w3schools.com/tags/tag_hr.asp

khelif96: using it for the horizontal line that seperates the title bar from the the rest of the content

src/pages/CommitQueue.tsx
khelif96: thanks, It was the opposite on my last team. Styles came before components.

src/contants/routes.ts
```diff
@@ -3,11 +3,13 @@ export const paths = {
   myPatches: "/my-patches",
   task: "/task",
   patch: "/patch",
+  commitQueue: "/commitQueue"
```
john-m-liu, line 6: routes are not case-sensitive, so I think this would be better as commit-queue

src/pages/CommitQueue.tsx
```diff
@@ -0,0 +1,81 @@
...
+        <PageTitle> Commit Queue </PageTitle>
+        <Badge variant="darkgray">
+          {queue.length} Item {queue.length > 1 && "s"}
+        </Badge>
+      </Header>
+      <HR />
+      {queue.map((commit, i) => (
```
john-m-liu, line 69: the entries in this queue are not yet commits, so I would call them queueItems or entries

src/pages/CommitQueue.tsx
```diff
@@ -0,0 +1,81 @@
...
+        <CommitQueueCard
+          key={commit.Issue}
+          index={i + 1}
+          title={commit.Patch.description}
```
john-m-liu, line 73: the Patch field is nullable, so I would add some light null handling here

src/pages/commitqueue/CommitQueueCard.tsx
```diff
@@ -0,0 +1,75 @@
...
+  author,
+  commitTime,
+  moduleCodeChanges
+}) => {
+  return (
+    <Card>
+      <Subtitle>{index}.</Subtitle>
+      <CommitQueueCardGrid>
+        <CommitInfo>
+          <CardTitle>{title}</CardTitle>
```
john-m-liu, line 63: The titles should be links to the patch page

src/pages/commitqueue/codeChangesModule/CodeChangesModule.tsx
```diff
@@ -0,0 +1,86 @@
...
+const CodeChangeModule: React.FC<{ moduleCodeChange: ModuleCodeChanges }> = ({
+  moduleCodeChange
+}) => {
+  const { fileDiffs } = moduleCodeChange;
+  const { additions, deletions } = totalFileDiffs(fileDiffs);
+  const [toggleAccordian, setToggleAccordian] = useState(false);
+  return (
+    <CodeChangeModuleContainer>
+      <AccordianToggle onClick={() => setToggleAccordian(!toggleAccordian)}>
+        <Icon glyph={toggleAccordian ? "CaretDown" : "CaretUp"} />
```
john-m-liu, line 51: The collapsed state should have the right icon, not up

src/pages/commitqueue/CommitQueueCard.tsx
```diff
@@ -0,0 +1,75 @@
...
+          <CardMetaData>
+            <b>By {author}</b> on {prettyDate(commitTime)}
```
john-m-liu, line 65: "By" shouldn't be bolded

src/pages/commitqueue/codeChangesModule/CodeChangeItem.tsx
```diff
@@ -0,0 +1,69 @@
...
+interface FileDiffTextProps {
+  type: string;
+  value: number;
+}
+export const FileDiffText: React.FC<FileDiffTextProps> = ({ value, type }) => {
```
john-m-liu, line 36: Instead of creating your own component, did you look at generalizing the CodeChanges component? That one needs a mode where it doesn't display column headers, but otherwise these 2 components have the same data source and are functionally identical

src/gql/queries/get-commit-queue.ts
khelif96: just opened up a pr for this 
https://github.com/evergreen-ci/evergreen/pull/3485

src/gql/queries/patch.ts
SupaJoon: We should leave the Patch interface in this file because the structure is specific to the fields in the gql string literal. As our codebase scales, it is possible that we will have another query that relies on Patch with different required fields so it could become cumbersome if we include all those interfaces together in `types/patch.ts`

tgrander:
> So im getting this linter error
> `[2020/04/20 20:02:43.662] ERROR: src/pages/commitqueue/codeChangesModule/CodeChangesModule.tsx:50:33 - Lambdas are forbidden in JSX attributes due to their rendering performance impact`
> 
> For doing something like this `<Component onClick={() => doSomething(someVar)} />`
> 
> Could we disable this rule?
> 
> Its a pretty common pattern and practice to create inline lambda functions for passing a function into a component. The rule in question is `jsx-no-lambda`

Not using inline lambda funcs in react components is as equally a common pattern as using them. There is a lot of chatter in react community around this topic. inline lambdas are "bad" for performance because they always result in a re-render. inversely, they are the easiest way to pass callback functions to react components. 

I have not had any problems with this rule, agree that they should and can be avoided for most part, am in favor of leaving it in place for the very reason the linter provides, which is that "Lambdas are forbidden in JSX attributes due to their rendering performance impact". We can vote on it though! @SupaJoon what do you think?

SupaJoon:
> > So im getting this linter error
> > `[2020/04/20 20:02:43.662] ERROR: src/pages/commitqueue/codeChangesModule/CodeChangesModule.tsx:50:33 - Lambdas are forbidden in JSX attributes due to their rendering performance impact`
> > For doing something like this `<Component onClick={() => doSomething(someVar)} />`
> > Could we disable this rule?
> > Its a pretty common pattern and practice to create inline lambda functions for passing a function into a component. The rule in question is `jsx-no-lambda`
> 
> Not using inline lambda funcs in react components is as equally a common pattern as using them. There is a lot of chatter in react community around this topic. inline lambdas are "bad" for performance because they always result in a re-render. inversely, they are the easiest way to pass callback functions to react components.
> 
> I have not had any problems with this rule, agree that they should and can be avoided for most part, am in favor of leaving it in place for the very reason the linter provides, which is that "Lambdas are forbidden in JSX attributes due to their rendering performance impact". We can vote on it though! @SupaJoon what do you think?

I think we should keep the rule as is to preserve performance.

src/pages/CommitQueue.tsx
```diff
@@ -0,0 +1,81 @@
...
+  font-size: 18px;
+`;
+
+const HR = styled("hr")`
+  background-color: ${gray.light2};
+  border: 0;
+  height: 3px;
+`;
+
+const ErrorWrapper = styled.div`
```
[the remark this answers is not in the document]
tgrander: on that note, i believe all errors should utilize the banner - but that seems out of the scope of this ticket maybe. a ticket exists for exactly this already https://jira.mongodb.org/browse/EVG-7773

src/pages/commitqueue/codeChangesModule/CodeChangeItem.tsx
tgrander: all typography should also come from leafygreen https://mongodb.github.io/leafygreen-ui/?path=/story/typography--default

src/pages/commitqueue/codeChangesModule/CodeChangesModule.tsx
khelif96: Good catch

src/pages/commitqueue/CommitQueueCard.tsx
khelif96: Its bolded on the designs.

src/gql/queries/patch.ts
tgrander: yes was thinking this too

khelif96: I don't necessarily agree, i think it makes more sense to keep our commonly reused types located in one area so we have a single source of truth for what the type for a certain query or field is. Its more error prone if we start recreating types on a per use basis since its easy to mistype a field.  
It shouldn't be an issue if we have different queries that rely on different subsets of fields with in a specific type since the type checking will still be valid on the fields that we do request in our queries.

tgrander: what arjun is saying is that this type is specifically the type returned by the graphql query and therefore belongs in the graphql namespace.

tgrander: all of the types for gql are currently in the gql folder. therefore it would make less sense organizationally to put one gql type in a different place than the rest of them. consistency is key when it comes to what does and does not make sense in a project's file structure. therefore putting this one gql type in a different place than the rest of them goes against your own point of a single source of truth. the point you provide goes against your own argument because doing it the way you suggest creates two sources of truth for gql types.

khelif96: The type exported from the graphql query is still defined in this file. 
https://github.com/evergreen-ci/spruce/blob/3956831786a197bf1bebbbd5812f9742e6d95fd5/src/gql/queries/patch.ts#L4

I just moved the type used by it

khelif96: The reason i moved the patch type was because the commitQueue returns a patch  type and  i thought it made more sense if they both implemented the same patch type.

tgrander: it's still a graphql type bc it represents the shape of a query. i see your point tho and do think there could be better file structure around types. however not certain that moving this one gql type to a different location is way to go. 

i would like to code generate the types from the gql schema soon anyways. and in which case all gql types will be in a single file in the graphql directory.

tgrander: we talked offline and decided that best way forward is to codegen the TS types for gql. jira ticket: https://jira.mongodb.org/browse/EVG-7866

src/pages/commitqueue/CommitQueueCard.tsx
john-m-liu: That seems like a mistake to me - could you check with Lara?

src/components/ErrorWrapper/index.tsx
```diff
@@ -0,0 +1,6 @@
+import React from "react";
```
john-m-liu, line 1: This isn't used

src/pages/patch/patchTabs/CodeChanges.tsx
```diff
@@ -3,13 +3,12 @@ import { useQuery } from "@apollo/react-hooks";
 import { useParams } from "react-router-dom";
 import { Skeleton, Table } from "antd";
```
john-m-liu, line 4: Table is no longer used

src/components/ErrorWrapper/index.tsx
khelif96: Thanks it's a force of habit to always include that.

cypress/integration/commit_queue.ts
```diff
@@ -0,0 +1,24 @@
...
+  beforeEach(() => {
+    cy.server();
+    cy.login();
+  });
+
+  it("Should render the commit queue page with one card", () => {
+    cy.visit(COMMIT_QUEUE_ROUTE);
+    cy.get("[data-cy=commit-queue-card]").should("have.length", 1);
+  });
+  it("Should be able to click on Total Code changes and toggle a drop down table", () => {
```
tgrander, line 18: beginning all tests with "should" is unnecessary. prefer starting tests with active verbs that describe the interaction. for example, "Clicking 'Total Code Changes' shows code changes table for patch"

src/components/PageTitle.tsx
```diff
@@ -22,13 +22,13 @@ export const PageTitle: React.FC<Props> = ({
     </PageHeader>
   ) : hasData ? (
     <PageHeader>
-      <H2 data-cy="page-title">
-        <span>
+      <Subtitle>
+        <span data-cy="page-title">
```
tgrander, line 26: i am suspicious of generic data-cy values like this. i know that this is a page and therefore the chance of it being rendered with another component that has a page title is rare. but name spacing cy attributes likes this seems safer.

src/pages/CommitQueue.tsx
```diff
@@ -0,0 +1,73 @@
...
+import get from "lodash/get";
+import { Skeleton } from "antd";
+import { PageTitle } from "components/PageTitle";
+import { ErrorWrapper } from "components/ErrorWrapper";
+import { CommitQueueCard } from "./commitqueue/CommitQueueCard";
+
+import {
+  GET_COMMIT_QUEUE,
+  CommitQueueQuery,
+} from "gql/queries/get-commit-queue";
```
tgrander, line 17: why is this import separated from the rest?

src/pages/CommitQueue.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+const { gray } = uiColors;
+
+export const CommitQueue = () => {
+  const { id } = useParams();
+  const { data, loading, error } = useQuery<CommitQueueQuery>(
+    GET_COMMIT_QUEUE,
+    {
+      variables: { id: id },
```
tgrander, line 26: variables: { id }

src/pages/CommitQueue.tsx
```diff
@@ -0,0 +1,73 @@
...
+    }
+  );
+  if (loading) {
+    return <Skeleton active={true} title={true} paragraph={{ rows: 4 }} />;
+  }
```
tgrander, line 31: page title can still be displayed if data is loading. instead of blocking rendering of all the ui, pass the loading prop to `PageTitle` and conditionally render the parts of the app that require data.

src/pages/commitqueue/CommitQueueCard.tsx
```diff
@@ -0,0 +1,92 @@
...
+import styled from "@emotion/styled";
+import { Subtitle, Body } from "@leafygreen-ui/typography";
+import { uiColors } from "@leafygreen-ui/palette";
+import Button from "@leafygreen-ui/button";
+import { CodeChangeModule } from "pages/commitqueue/codeChangesModule/CodeChangesModule";
+import { StyledRouterLink } from "components/styles/StyledLink";
+import { ModuleCodeChanges } from "types/patch";
+import { paths } from "constants/routes";
+import { format } from "date-fns";
+const FORMAT_STR = "MM/dd/yy' at 'hh:mm:ss' 'aa";
```
tgrander, line 11: same comment: space between imports and everthing else ;)

src/pages/commitqueue/CommitQueueCard.tsx
```diff
@@ -0,0 +1,92 @@
...
+            By <b>{author}</b> on {format(new Date(commitTime), FORMAT_STR)}
+          </CardMetaData>
+          <div>
+            {moduleCodeChanges.map((moduleCodeChange) => (
+              <CodeChangeModule
+                key={moduleCodeChange.rawLink}
+                moduleCodeChange={moduleCodeChange}
+              />
+            ))}
+          </div>
```
tgrander, line 47: prefer a React.fragment where unstyled div is used to contain content. Shorthand for which is `<>{content}</>`

src/pages/commitqueue/codeChangesModule/CodeChangesModule.tsx
```diff
@@ -0,0 +1,70 @@
...
+  transition: max-height 0.3s ease-in-out;
+`;
+
+const DropDownTextStyle = styled("span")`
+  margin-right: 24px;
+`;
+
+const DropDownText = styled(Body)`
+  font-size: 16px;
+`;
```
tgrander, line 70: I do not believe we are meant to change the font sizes of Typography elements. sort of defeats the purpose of having a defined typography system.

src/pages/patch/patchTabs/CodeChanges.tsx
```diff
@@ -54,78 +53,18 @@ export const CodeChanges = () => {
...
-    key: "deletions",
-    render: (text: number) =>
-      text === 0 ? text : <Deletion>-{text}</Deletion>,
-  },
-];
-
-const rowKey = (record: FileDiff, index: number): string => `${index}`;
-
 const StyledButton = styled(Button)`
   margin-left: 16px;
```
tgrander, line 65: thank you

src/components/PageTitle.tsx
khelif96: That value  was always there i just moved it. but i think this component is generic enough where i think it isnt necessary to name space it, it reflects what component it belongs too. What would you suggest name spacing it too?

src/pages/CommitQueue.tsx
khelif96: Im not sure i could have sworn i removed that space in a previous commit

cypress/integration/commit_queue.ts
```diff
@@ -0,0 +1,24 @@
...
+describe("commit queue page", function() {
+  beforeEach(() => {
+    cy.server();
+    cy.login();
+  });
+
+  it("Should render the commit queue page with one card", () => {
+    cy.visit(COMMIT_QUEUE_ROUTE);
+    cy.get("[data-cy=commit-queue-card]").should("have.length", 1);
+  });
```
tgrander, line 17: these tests will fail if an additional item is added to queue test data. we experienced trouble in the past where a small change to test data created a cascade of test failures because tests and data were tightly coupled. to avoid this we write tests that are robust enough that any data set can be plugged in and tests will still pass. for example, testing that number of cards rendered matches number of cards from response data.

src/components/PageTitle.tsx
tgrander: sounds good to me

src/pages/CommitQueue.tsx
tgrander: coming back to haunt you

src/pages/commitqueue/codeChangesModule/CodeChangesModule.tsx
khelif96: The default font sizes don't match the designs 🙃

tgrander: in that case I would make it whatever My Patches title is and send a message in spruce chat so that Lara knows to always use typography system sizes in future designs

cypress/integration/commit_queue.ts
khelif96: That makes sense, thats an interesting approach on university we populated the db with the data we will need for the tests  so we always had the same data/ were able to control what data was used by tests

src/pages/commitqueue/codeChangesModule/CodeChangesModule.tsx
khelif96: Actually looks like theres a leafygreen provider component that we can use to update our typography sizes through out the app.

cypress/integration/commit_queue.ts
tgrander: we also use a predefined set of data but we don't think tests should fail if one field on the test data changes. our approach on tests is to make them resilient by matching rendered ui against network responses rather than matching directly on the predefined data.

PR #142, opened 2020-04-22
EVG-7870 Reinstall devdependencies properly

khelif96:
Noticed some dev dependencies weren't installed correctly


khelif96:
Strange i tested it locally before i made this pr. 

> e2e tests are hanging

khelif96:
Anyone know why they didnt rerun after i updated this pr?

SupaJoon:
evergreen retry

tgrander:
> Strange i tested it locally before i made this pr.
> 
> > e2e tests are hanging

ci and local are different environments that have different dependencies. changes like this causing a discrepancy between local and ci e2e tests is normal. same thing happened when i tried to upgrade cra.

khelif96:
> > Strange i tested it locally before i made this pr.
> > > e2e tests are hanging
> 
> ci and local are different environments that have different dependencies. changes like this causing a discrepancy between local and ci e2e tests is normal. same thing happened when i tried to upgrade cra.

Thats strange, Is `node_modules` cached on the ci server?
 dependency changes shouldn't make the ci server break. I tried downgrading the changes in this pr to the versions they were before with the only exception of some being a patch version change or 2 different.  and that shouldn't cause any issues. Unless the app is expecting one set of dependencies but receives another (Expecting specific versions of a dep but  received the older cached versions).

khelif96:
I took a look at the logs and i dont think the ci server is breaking i think the dev server is never starting and the ci hangs waiting for it to start.
image
This is the last log message we get back from our application.

I think it most likely is getting stuck trying to load dependencies.

tgrander:
what i mean is that testing the app in ci is different from testing the app locally. mostly because CI is a programmed set of instructions whereas on local everything needed already exists.

node_modules is not cached on the ci server. the version of node at the time of configuring the ci environment did not support `npm ci`. cacheing node modules other ways has caused a lot of headaches for other teams. installing node modules for every ci build at the moment seems to be working alright. 

if you would like to see the set of instructions for the ci env then open up evergreen.yml in spruce.

if ci works before your changes but not after then it is probably something that you changed :)

tgrander:
> I think it most likely is getting stuck trying to load dependencies.

not sure that's the problem because i can see [here]( independently test that the query has the correct variables) that `npm-start` evergreen command is not started until `npm install` command finishes. 

this is interesting tho because this is the exact same issue that occurred when i tried to upgrade create react app. it doesn't look to me like it's a problem with downloading dependencies. looks more like the development server itself hanging indefinitely. this can happen due to an error from `npm run dev`.

khelif96:
Do you know if its possible to ssh into an instance that the ci server runs on  so i can try running the commands manually?

tgrander:
yes and that's a great idea. you can spawn a host from your failing patch. https://evergreen.mongodb.com/spawn?distro_id=ubuntu1604-small&task_id=spruce_ubuntu1604_e2e_test_patch_b01f2383dfa9d11c3b1e44f04509ab63d850996b_5ea0faee562343041a4bec25_20_04_23_02_18_23

khelif96:
Thats awesome i didnt know that was even a feature.

tgrander:
evergreen is pretty incredible

PR #143, opened 2020-04-22
EVG-7547, EVG-7548: search input for patch name & dropdown menu for statuses

SupaJoon:
https://jira.mongodb.org/browse/EVG-7547
https://jira.mongodb.org/browse/EVG-7548

URL will update when inputs are interacted with and GQL query will be made. 

There are not many e2e tests for these code changes because the My Patches page is still being developed. The PR to include the virtualized list and Patch card 
(https://jira.mongodb.org/browse/EVG-7552 & https://jira.mongodb.org/browse/EVG-7551) will allow me to test the full user story in cypress - from input interaction -> url update -> rendered results.  

Screen Shot 2020-04-22 at 5 32 10 PM

### DIFF
--- a/cypress/integration/my_patches.js
+++ b/cypress/integration/my_patches.js
@@ -11,18 +11,6 @@ describe("My Patches Page", () => {
     cy.listenGQL();
   });
 
-  it("Initial load should make a userPatches gql query with default query params when none are specified in the URL", () => {
-    cy.visit(MY_PATCHES_ROUTE);
-    cy.waitForGQL("userPatches", {
-      "request.body.variables.$includeCommitQueue": (v) => v == true,
-      "request.body.variables.$limit": (v) => v === 10,
-      "request.body.variables.$page": (v) => v === 0,
-      "request.body.variables.$patchName": (v) => v === "",
-      "request.body.variables.$statuses": (v) =>
-        Array.isArray(v) && v.length === 0,
-    }).then((xhr) => console.log(xhr));
-  });
-
   describe("Show commit queue checkbox", () => {
     beforeEach(() => {
       cy.preserveCookies();
--- a/cypress/integration/my_patches.js
+++ b/cypress/integration/my_patches.js
@@ -1,23 +1,46 @@
 /// <reference types="Cypress" />
 
 const MY_PATCHES_ROUTE = "/my-patches";
-describe("my patches page", function () {
-  beforeEach(() => {
-    cy.server();
+describe("My Patches Page", () => {
+  before(() => {
     cy.login();
   });
 
-  it("Should render with Show Commit Queue box checked when commitQueue not indicated in URL query param", () => {
-    cy.visit(MY_PATCHES_ROUTE);
-    cy.get("[data-cy=commit-queue-checkbox]").should("be.checked");
+  beforeEach(() => {
+    cy.preserveCookies();
+    cy.listenGQL();
   });
 
-  it("Should render with Show Commit Queue box unchecked when commitQueue is false in URL query param", () => {
-    cy.visit(`${MY_PATCHES_ROUTE}?commitQueue=false`);
-    cy.get("[data-cy=commit-queue-checkbox]").should("not.be.checked");
+  it("Initial load should make a userPatches gql query with default query params when none are specified in the URL", () => {
+    cy.visit(MY_PATCHES_ROUTE);
+    cy.waitForGQL("userPatches", {
+      "request.body.variables.$includeCommitQueue": (v) => v == true,
+      "request.body.variables.$limit": (v) => v === 10,
+      "request.body.variables.$page": (v) => v === 0,
+      "request.body.variables.$patchName": (v) => v === "",
+      "request.body.variables.$statuses": (v) =>
+        Array.isArray(v) && v.length === 0,
+    }).then((xhr) => console.log(xhr));
   });
-  it("Should render with Show Commit Queue box checked when commitQueue is true in URL query param", () => {
-    cy.visit(`${MY_PATCHES_ROUTE}?commitQueue=true`);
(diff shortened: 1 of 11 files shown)


src/pages/MyPatches.tsx
```diff
@@ -1,36 +1,131 @@
...
+              return prev;
+            }
+            return fetchMoreResult;
+          },
+        });
+      } catch (e) {
+        // empty block
+      }
+    });
+  }, [networkStatus, error, fetchMore, listen]);
```
tgrander, line 64: this is third appearance of this code that is almost identical to Patch and Task page fetchMore handler func. not a blocker but this would do well as a shared function or hook between the three pages.

cypress/integration/my_patches.js
```diff
@@ -1,23 +1,46 @@
...
+  it("Initial load should make a userPatches gql query with default query params when none are specified in the URL", () => {
+    cy.visit(MY_PATCHES_ROUTE);
+    cy.waitForGQL("userPatches", {
+      "request.body.variables.$includeCommitQueue": (v) => v == true,
+      "request.body.variables.$limit": (v) => v === 10,
+      "request.body.variables.$page": (v) => v === 0,
+      "request.body.variables.$patchName": (v) => v === "",
+      "request.body.variables.$statuses": (v) =>
+        Array.isArray(v) && v.length === 0,
+    }).then((xhr) => console.log(xhr));
```
tgrander, line 23: i don't know if testing the shape of the request so strictly is a good or practical way to test. Testing the end result of the query and how it appears and behaves to the user is the goal of the test imo. we should not need to inspect the request body because testing the usability of the page as a result of it does that job and does it more effectively. 

How Kent C Dodds describes what to test is "Think less about the code you are testing and more about the use cases that code supports."

These tests tests an underlying and private functionality of the feature. Therefore it does not test how the software is _used_.

cypress/integration/my_patches.js
```diff
@@ -1,23 +1,46 @@
...
   });
-  it("Should render with Show Commit Queue box checked when commitQueue is true in URL query param", () => {
-    cy.visit(`${MY_PATCHES_ROUTE}?commitQueue=true`);
-    cy.get("[data-cy=commit-queue-checkbox]").should("be.checked");
+
+  describe("Show commit queue checkbox", () => {
```
tgrander, line 26: I very much like the pattern of splitting up feature tests of a page between initial state and then different functionalities on the page.

cypress/integration/my_patches.js
```diff
@@ -1,23 +1,46 @@
...
+    beforeEach(() => {
+      cy.preserveCookies();
+    });
+
+    it("Should render with Show Commit Queue box checked when commitQueue not indicated in URL query param", () => {
```
tgrander, line 31: tests do not need to start with "should render with". For example, the description can be changed to "Commit queue checkbox". and then this test name would be "Renders checkbox checked by default". Longer test names are not always more clear than short, simple ones.

src/components/styles/Layout.tsx
```diff
@@ -27,3 +28,7 @@ export const PageSider = styled(Sider)`
   max-width: ${siderWidth} !important;
   min-width: ${siderWidth} !important;
 `;
+
+export const SimplePageTitle = styled(H2)`
+  margin-bottom: 16px;
+`;
```
tgrander, line 34: these seem like one-off styled per-page that should not exist as shared components

src/components/styles/filters.tsx
```diff
@@ -4,7 +4,13 @@ import { Input } from "antd";
 export const FiltersWrapper = styled.div`
   display: flex;
   margin-bottom: 12px;
+  align-items: center;
 `;
+
+export const FiltersWrapperSpaceBetween = styled(FiltersWrapper)`
+  justify-content: space-between;
+`;
```
tgrander, line 12: same comment here about having styled components in the shared folder that are slight modifications of other shared components

cypress/integration/my_patches.js
SupaJoon: True e2e tests are not possible until the feature is complete. I agree this test should be replaced by a test that asserts the rows rendered in the table after user interaction with the filter inputs.

cypress/integration/my_patches.js
SupaJoon: This test entered the codebase from a previous PR.

src/components/styles/filters.tsx
SupaJoon: I'll remove this one from the shared components folder. It may not be used again for a little while.

src/components/styles/Layout.tsx
SupaJoon: This title component can be used for any page title that follows the style of the My Patches page (a page title without a badge). I can rename this component so we can find it easier.

src/pages/my-patches/StatusSelector.tsx
```diff
@@ -0,0 +1,53 @@
...
+    value: ALL_PATCH_STATUS,
+    key: ALL_PATCH_STATUS,
+  },
+  {
+    title: "Created",
+    value: PatchStatus.Created,
+    key: PatchStatus.Created,
+  },
+  {
+    title: "Started",
```
john-m-liu, line 39: This should be "running"

src/pages/my-patches/StatusSelector.tsx
```diff
@@ -0,0 +1,53 @@
...
+    value: PatchStatus.Started,
+    key: PatchStatus.Started,
+  },
+  {
+    title: "Success",
```
john-m-liu, line 44: I realize it's wrong in the db models, but could we make the tense in the tiles consistent? So that we have either success/fail or succeeded/failed. The task status select has the same thing so feel free to defer this and I can make a ticket

PR #146, opened 2020-04-27
EVG-7866 Custom typescript/gql code generation

khelif96:
I'll call out some findings throughout the code.

But some new rules in order to get code generation to work.
- Queries should now be declared with a query name so the code generation knows what to name the corresponding type.
- Each query and mutation should have a unique name.
- Since query analysis for type checking occurs statically we cant place dynamic variables with in query strings we instead have to hard code the variable in the query or pass it in as query variable. (This only happens once in the code base so it shouldn't be an issue.)


codegen.js
```diff
@@ -0,0 +1,22 @@
...
+      },
+    },
+  ],
+  documents: ["./src/**/*.ts"],
+  overwrite: true,
+  generates: {
+    "./src/gql/generated/types.ts": {
+      plugins: ["typescript", "typescript-operations"],
+      config: {
+        skipTypename: true,
```
khelif96, line 17: idk if we want to preserve typenames but it produces much cleaner types by skipping it. But it can be useful to have it if want to perform operations based off the corresponding type  like we do in `src/pages/task/logs/LogTypes.tsx`

config/.cmdrc_sample.json
```diff
@@ -3,7 +3,8 @@
     "REACT_APP_GQL_URL": "http://localhost:9090/graphql/query",
     "REACT_APP_API_URL": "http://localhost:3000/api",
     "REACT_APP_UI_URL": "http://localhost:9090",
-    "REACT_APP_SPRUCE_URL": "http://localhost:3000"
+    "REACT_APP_SPRUCE_URL": "http://localhost:3000",
+    "REACT_APP_GQL_COOKIE": "cookie-from-REACT_APP_GQL_URL"
```
khelif96, line 7: This should be the cookie thats used in requests when communicating with the graphql playground. Im not sure if this is different for everyone in local development and how often it changes if it does.

codegen.js
```diff
@@ -0,0 +1,22 @@
...
+        headers: {
+          cookie: process.env.REACT_APP_GQL_COOKIE,
+        },
+      },
+    },
+  ],
+  documents: ["./src/**/*.ts"],
+  overwrite: true,
+  generates: {
+    "./src/gql/generated/types.ts": {
```
khelif96, line 14: types are saved to `gql/generated/types.ts` if any one has a better name/location for this lmk.

src/gql/queries/get-patch-tasks.ts
```diff
@@ -17,7 +16,7 @@ export const GET_PATCH_TASKS = gql`
   ) {
     patchTasks(
       patchId: $patchId
-      limit: ${PATCH_TASKS_LIMIT}
+      limit: 10
```
khelif96, line 19: Can no longer embed variables in queries dynamically we must either hard code them or pass them in as a query param

src/gql/queries/get-projects.ts
```diff
@@ -1,7 +1,7 @@
 import gql from "graphql-tag";
 
 export const GET_PROJECTS = gql`
-  {
+  query Projects {
```
khelif96, line 4: Every query should have a custom name. Also im not sure if theres a naming convention for queries. There seems to be a mix of both lower cased and uppercased query names.

src/pages/CommitQueue.tsx
```diff
@@ -10,22 +10,22 @@ import { Skeleton } from "antd";
...
+} from "gql/generated/types";
 
 const { gray } = uiColors;
 
 export const CommitQueue = () => {
   const { id } = useParams();
-  const { data, loading, error } = useQuery<CommitQueueQuery>(
-    GET_COMMIT_QUEUE,
-    {
-      variables: { id: id },
```
khelif96, line 26: I swear i updated this in my commitQueue pr but it keeps coming back 👻

src/pages/task/logs/LogTypes.tsx
```diff
@@ -1,26 +1,41 @@
...
+} from "gql/generated/types";
 import { useQuery } from "@apollo/react-hooks";
 import { TaskEventLogLine } from "./logTypes/TaskEventLogLine";
 import { LogMessageLine } from "./logTypes/LogMessageLine";
 import { ApolloError } from "apollo-client";
 import { useParams } from "react-router-dom";
 import get from "lodash/get";
 
+interface TaskEventLogEntryType extends TaskEventLogEntry {
+  kind?: "taskEventLogEntry";
```
khelif96, line 28: This doesn't exist on the official graphql type but it did in our types. enabling the typeName flag would probably be useful here? But im not sure how often we use it. What do you think?

SupaJoon:
@khelif96 What are the steps to run the code generation?

src/gql/generated/types.ts
```diff
@@ -0,0 +1,622 @@
...
+/** All built-in and custom scalars, mapped to their actual values */
+export type Scalars = {
+  ID: string;
+  String: string;
+  Boolean: boolean;
+  Int: number;
+  Float: number;
+  Time: any;
+  Duration: any;
+  StringMap: any;
```
tgrander, line 11: scalars can be mapped to their actual values by mapping them to their respective types in the config file https://graphql-code-generator.com/docs/plugins/typescript#configuration

src/gql/queries/get-patch-tasks.ts
tgrander, line 19: should probably pass limit at a variable to query instead of hardcoding in query

src/gql/queries/get-projects.ts
tgrander: let's use CamelCase

src/gql/queries/get-task-logs.ts
```diff
@@ -1,7 +1,7 @@
 import gql from "graphql-tag";
 
 export const GET_EVENT_LOGS = gql`
-  query taskLogs($id: String!) {
+  query eventLogs($id: String!) {
```
tgrander, line 4: CamelCase?

src/gql/queries/get-task-logs.ts
```diff
@@ -33,7 +33,7 @@ export const GET_TASK_LOGS = gql`
 `;
 
 export const GET_AGENT_LOGS = gql`
-  query taskLogs($id: String!) {
+  query agentLogs($id: String!) {
```
tgrander, line 36: CamelCase?

src/gql/queries/get-task-logs.ts
```diff
@@ -45,7 +45,7 @@ export const GET_AGENT_LOGS = gql`
 `;
 
 export const GET_SYSTEM_LOGS = gql`
-  query taskLogs($id: String!) {
+  query systemLogs($id: String!) {
```
tgrander, line 48: CamelCase?

src/pages/task/logs/LogTypes.tsx
tgrander: @SupaJoon

codegen.js
```diff
@@ -0,0 +1,22 @@
+module.exports = {
+  schema: [
+    {
+      "http://localhost:9090/graphql/query": {
+        headers: {
+          cookie: process.env.REACT_APP_GQL_COOKIE,
```
john-m-liu, line 6: Does graphql-codegen support passing a schema file? Having it hit the introspection endpoint and hacking through auth like this will make it annoying to work with, since you have to have the evergreen server running, and you'll have to update this whenever you do something with your user

src/gql/queries/get-projects.ts
john-m-liu: I don't have an opinion on this, but just double checking - do we want camelCase or TitleCase?

src/gql/queries/get-task-logs.ts
khelif96: Do you mean TitleCase? its already in camelCase.

codegen.js
tgrander: hitting the introspection endpoint is certainly the way to go for this but i understand what john's saying. we could pass it the path to the local schema file in evergreen. or we could not require authentication for gql endpoint if ENV = dev

src/gql/queries/get-task-logs.ts
tgrander: yes TitleCase, thank you :)

codegen.js
tgrander: So the latter idea i gave of disabling auth for dev won't work. what about an auth-less endpoint that only returns the schema?

john-m-liu: What's the advantage of creating an endpoint to return the file if you have the file already?

khelif96: Also im not sure how much of an issue having the evergreen server running would be chances are if you are regenerating the types you are probably already doing development and need the server running anyways.

john-m-liu: The schema file is expected to exist as "sdlschema" in the spruce folder for query linting: https://github.com/evergreen-ci/spruce/commit/bc288b6c33ce4f289d5bb1b9468abe3bb2e8cbb1#diff-04c6e90faac2675aa89e2176d2eec7d8R26

khelif96: The schema file and introspection file aren't the same. `graphql-codegen` supports importing a introspection json file. But i cant find a way to import a plain  `.graphql` schema file.

tgrander: i'm on board with what john's saying. you can put the path to the local schema file as the value to the schema key in the config file. https://graphql-code-generator.com/docs/getting-started/schema-field#graphql-file

khelif96: Sounds good to me. Looks like the file has to be named  with a `.graphql` filename for graphql-codegen to pick it up

PR #148, opened 2020-04-29
EVG-7551, EVG-7552: PatchCard and PatchCard list

SupaJoon:
https://jira.mongodb.org/browse/EVG-7551
https://jira.mongodb.org/browse/EVG-7552
These code changes depend on schema changes in https://github.com/evergreen-ci/evergreen/pull/3535
MyPatches


cypress/integration/my_patches.js
```diff
@@ -9,6 +15,119 @@ describe("My Patches Page", () => {
...
+      pathname: MY_PATCHES_ROUTE,
+      paramName: "patchName",
+      search: inputVal,
+    });
+    resultsAreFetchedAndRendered({
+      queryName: "UserPatches",
+      responseName: "userPatches.patches",
+      requestVariables: { patchName: inputVal },
+      tableRow: "[data-cy=patch-card]",
+    });
```
tgrander, line 34: seeing these reused is heaven

src/gql/queries/my-patches.ts
```diff
@@ -15,16 +15,30 @@ export const GET_USER_PATCHES = gql`
...
+      activated
       builds {
-        id
         buildVariant
         status
-        predictedMakespan
-        actualMakespan
       }
     }
   }
```
tgrander, line 44: i like this. separating the live updating query from the static query.

src/hooks/useStopPatchPolling.ts
```diff
@@ -0,0 +1,25 @@
...
+}: Props) => {
+  if (
+    hasData &&
+    (status === PatchStatus.Failed ||
+      status === PatchStatus.Success ||
+      activated === false)
+  ) {
+    stopPolling();
+  }
+};
```
tgrander, line 25: perfect! i was thinking that as a precaution against memory leaks that we could add this to the mix: 
```
useEffect(() => stopPolling, [])
```
This way, when the component unmounts, `stopPolling` will always be called

src/pages/MyPatches.tsx
```diff
@@ -96,32 +109,48 @@ export const MyPatches = () => {
...
+          checked={getQueryVariables(search).includeCommitQueue}
         />
       </FiltersWrapperSpaceBetween>
+      <Pagination
+        onChange={onChange}
+        current={page}
+        pageSize={LIMIT}
+        total={get(data, "userPatches.filteredPatchCount", 0)}
+      />
+      <div>
```
tgrander, line 121: can this div be replaced with a Fragment?

src/pages/my-patches/PatchCard.tsx
```diff
@@ -0,0 +1,131 @@
...
+import { paths } from "constants/routes";
+import { useQuery } from "@apollo/react-hooks";
+import get from "lodash/get";
+import { useStopPatchPolling } from "hooks";
+import {
+  PatchBuildVariantsAndStatusQueryVariables,
+  PatchBuildVariantsAndStatusQuery,
+  Maybe,
+} from "gql/generated/types";
+interface Build {
```
tgrander, line 19: space between these two

tgrander:
oops, don't think i can test in staging until the backend PR is merged

src/pages/Patch.tsx
```diff
@@ -15,10 +15,11 @@ import { PatchTabs } from "pages/patch/PatchTabs";
 import { BuildVariants } from "pages/patch/BuildVariants";
 import get from "lodash/get";
 import { Metadata } from "pages/patch/Metadata";
-import Badge, { Variant } from "@leafygreen-ui/badge";
 import { PatchStatus } from "types/patch";
```
john-m-liu, line 18: no longer used

src/hooks/useStopPatchPolling.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export const useStopPatchPolling = ({
+  hasData,
+  status,
+  activated,
+  stopPolling,
+}: Props) => {
+  if (
+    hasData &&
+    (status === PatchStatus.Failed ||
```
john-m-liu, line 19: I realize it was like this before but the data can still change if one of these conditions is met. Is the intent that we want to stop polling if we're absolutely sure the data won't change anymore?

tgrander: right now the live updating is not sophisticated but i think that has to do with the behavior/UX of the entire app not being fully fleshed out yet. also, if a patch changes from `not running` to `running again` then an event occurred that will most likely trigger a refetch, such as restarting the query.

john-m-liu: Right, I was referring to the scenario when someone else makes changes to the patch you're looking at. Would it be so bad to let it poll constantly? Or perhaps we can stop it when the tab loses focus and restart it when it gains focus

SupaJoon: That's right. We want to stop polling if we are sure the data won't change. Do you know how we can adjust the condition to achieve this?

john-m-liu: It can always change since you can restart patches

cypress/integration/my_patches.js
```diff
@@ -9,6 +15,119 @@ describe("My Patches Page", () => {
...
+    resultsAreFetchedAndRendered({
+      queryName: "UserPatches",
+      responseName: "userPatches.patches",
+      requestVariables: { includeCommitQueue: true },
+      tableRow: "[data-cy=patch-card]",
+    });
+  });
+
+  describe("Pagination", () => {
+    [2, 3].forEach((pageNum) =>
```
khelif96, line 71: instead of using a for loop use 
```
cy.wrap(array).each((index) => {
doSomething(index)
})
```
For loops dont behave deterministically in cypress because of how cypress runs tests asynchronously .

https://stackoverflow.com/questions/52212868/cypress-io-writing-a-for-loop

src/hooks/useStopPatchPolling.ts
tgrander: let's increase poll - what do y'all think? 5 seconds? 10 seconds? And then we can let it poll constantly and use `useEffect(() => stopPolling, [stopPolling])` to ensure polling stops when component unmounts.

SupaJoon: I removed the useStopPatchPolling hook and applied the useEffects in this commit: 6c65a23. I think 5 seconds is reasonable.

cypress/integration/my_patches.js
```diff
@@ -32,3 +152,64 @@ describe("My Patches Page", () => {
...
+
+const resultsAreFetchedAndRendered = ({
+  queryName,
+  responseName,
+  requestVariables,
+  tableRow,
+} = {}) => {
+  return assertQueryVariables(queryName, requestVariables).then((xhr) => {
+    const { response } = xhr;
+    const numberOfResults = get(response, `body.data.${responseName}`, [])
```
khelif96, line 203: Is there a way to more deterministically check that  this data is valid. I think we want to move away from relying on the data we get back from the request and instead comparing the results with expected outcomes.

src/components/PatchStatusBadge.tsx
```diff
@@ -0,0 +1,27 @@
+import React from "react";
+import { PatchStatus } from "types/patch";
```
khelif96, line 2: Is there a generated type we can use thats already available inside of `generated/types` ?

src/gql/generated/types.ts
```diff
@@ -626,7 +626,14 @@ export type UserPatchesQueryVariables = {
 };
 
 
-export type UserPatchesQuery = { userPatches: { patches: Array<{ projectID: string, description: string, status: string, createTime?: Maybe<Date>, builds: Array<{ id: string, buildVariant: string, status: string, predictedMakespan: number, actualMakespan: number }> }> } };
+export type UserPatchesQuery = { userPatches: { filteredPatchCount: number, patches: Array<{ id: string, projectID: string, description: string, status: string, createTime?: Maybe<Date>, builds: Array<{ buildVariant: string, status: string }> }> } };
+
+export type PatchBuildVariantsAndStatusQueryVariables = {
+  id: Scalars['String'];
```
khelif96, line 632: I noticed this before its weird how its generating single quotes for you but double quotes for me. Is prettier running on this file?

src/pages/MyPatches.tsx
```diff
@@ -96,32 +109,48 @@ export const MyPatches = () => {
...
+      <>
+        {error ? (
+          <ErrorWrapper>{error}</ErrorWrapper>
+        ) : loading ? (
+          <StyledSkeleton active={true} title={false} paragraph={{ rows: 4 }} />
+        ) : get(data, "userPatches.patches", []).length !== 0 ? (
+          data.userPatches.patches.map((p) => <PatchCard key={p.id} {...p} />)
+        ) : (
+          <NoResults data-cy="no-patches-found">No patches found</NoResults>
+        )}
```
khelif96, line 130: can we clean this up a little bit and maybe move some logic out. Its difficult to read when we have 3 layers deep of ternaries

src/pages/my-patches/PatchCard.tsx
```diff
@@ -0,0 +1,128 @@
...
+          on <b>{projectID}</b>
+        </TimeAndProject>
+      </Left>
+      <Center>
+        <BadgeContainer>
+          <PatchStatusBadge status={status} />
+        </BadgeContainer>
+        <IconsContainer>
+          {builds.map((b, i) => (
+            <div key={i}>
```
khelif96, line 67: Its not generally a good  practice to use indexes as keys. https://reactjs.org/docs/lists-and-keys.html#keys 
Try using the build id or some other unique field.

src/pages/my-patches/patch-card/BuildStatusIcon/CreatedIcon.ts
```diff
@@ -0,0 +1,9 @@
+import { uiColors } from "@leafygreen-ui/palette";
+import styled from "@emotion/styled";
+
+export const CreatedIcon = styled.div`
+  border-radius: 50%;
+  border: 2px solid ${uiColors.gray.light1};
```
khelif96, line 6: can we destructure this.

tgrander: if `uiColors` is used once in this file then what is the value of destructuring this?

src/components/PatchStatusBadge.tsx
tgrander: tldr: no. this is a global evergreen type representing the different patch statuses as an enum. it's not in generated types not do i think it should it be.

SupaJoon: It would make sense to use a generated enum it existed but it doesn't exist since our Patch gql type status field is a string.

cypress/integration/my_patches.js
SupaJoon: I agree that we should move away from tracking responses from gql and examining responses. We will need to stub our gql queries responses to compare results with expected outcomes. That work is defined in this ticket: https://jira.mongodb.org/browse/EVG-7910

src/gql/generated/types.ts
SupaJoon: `npm run codegen` is outputting single quotes for me

src/components/PatchStatusBadge.tsx
khelif96: Thanks didn't know it wasnt.

src/pages/my-patches/PatchCard.tsx
SupaJoon: good call

src/gql/generated/types.ts
SupaJoon: saving with prettier will generate double quotes though

src/pages/my-patches/patch-card/BuildStatusIcon/CreatedIcon.ts
khelif96: I think i left the comment as a force of habit from years of always having the [prefer-destructuring](https://eslint.org/docs/rules/prefer-destructuring) rule enforced. Its fine if we dont use it.

src/gql/generated/types.ts
khelif96: Thats strange ill try to figure out whats going on. So its more consistent from now on

khelif96: What if we set up some precommit hooks to run prettier and eslint on our code before we commit or push up changes? @tgrander  @SupaJoon  what do you think?

tgrander: that would be great

cypress/integration/my_patches.js
tgrander: we're not stubbing it, it will exist as it is now but we will match against the data in local evergreen instead of what is returned in the response

khelif96:
Speaking of typography  the designs have slightly different font sizes compared to the defaults in the Leafygreen typography package. eg. The designs `Body` tag use 16px font sizes while the default leafygreen `Body` tag is 14px. Leafygreen has a [Leafygreen provider component ](https://github.com/mongodb/leafygreen-ui/tree/master/packages/leafygreen-provider) That lets us update these default attributes globally. How do you guys feel about bring that in and wrapping our app with it? @SupaJoon  @tgrander

src/pages/MyPatches.tsx
```diff
@@ -60,34 +69,55 @@ export const MyPatches = () => {
...
+        <StyledSkeleton active={true} title={false} paragraph={{ rows: 4 }} />
+      );
+    }
+    if (get(data, "userPatches.patches", []).length !== 0) {
+      return data.userPatches.patches.map((p) => (
+        <PatchCard key={p.id} {...p} />
+      ));
+    }
+    return <NoResults data-cy="no-patches-found">No patches found</NoResults>;
+  })();
```
khelif96, line 108: Instead of an IIFE you can just capitalize the first letter of the function and turn it into a jsx component. 
```
const TableBody = ({ error, loading, data }) => {
  if (error) {
    return <ErrorWrapper>{error}</ErrorWrapper>;
  }
  if (loading) {
    return (
      <StyledSkeleton active={true} title={false} paragraph={{ rows: 4 }} />
    );
  }
  if (get(data, "userPatches.patches", []).length !== 0) {
    return data.userPatches.patches.map((p) => <PatchCard key={p.id} {...p} />);
  }
  return <NoResults data-cy="no-patches-found">No patches found</NoResults>;
};
```

src/pages/my-patches/patch-card/BuildStatusIcon.tsx
```diff
@@ -0,0 +1,38 @@
...
+export const BuildStatusIcon: React.FC<Props> = ({ status, buildVariant }) => {
+  const icon = statusToIcon[status];
+  if (!icon) {
+    return null;
+  }
+
+  return (
+    <Tooltip
+      triggerEvent="hover"
+      trigger={<div>{icon}</div>}
```
khelif96, line 23: Is this div necessary. Can we just place the icon in instead?

SupaJoon:
> Speaking of typography the designs have slightly different font sizes compared to the defaults in the Leafygreen typography package. eg. The designs `Body` tag use 16px font sizes while the default leafygreen `Body` tag is 14px. Leafygreen has a [Leafygreen provider component ](https://github.com/mongodb/leafygreen-ui/tree/master/packages/leafygreen-provider) That lets us update these default attributes globally. How do you guys feel about bring that in and wrapping our app with it? @SupaJoon @tgrander

Sounds good to me. It will hopefully lead to less custom styling.

SupaJoon:
> Per the designs i do not think this typography is the right color

The color from the design is #89989B and is currently applied in the implementation

SupaJoon:
> This typography has a greater weight in the designs. Can you use leafygreen typography?

The closest typography element in leafygreen is Body which is bare bones. I think it's best to extend the StyledLink component and add a font-weight and font-size to it since it shares a lot of the styles.

SupaJoon:
> Screen Shot 2020-05-01 at 11 47 26 AM
> margin between title and status is 58px in designs

I actually see this as 24px. Is this what you mean?
Screen Shot 2020-05-01 at 1 28 08 PM

src/pages/MyPatches.tsx
SupaJoon: I rather keep this as an IIFE because passing props to the component seems repetitive when the variables are available in the closure.

khelif96:
> > This typography has a greater weight in the designs. Can you use leafygreen typography?
> 
> The closest typography element in leafygreen is Body which is bare bones. I think it's best to extend the StyledLink component and add a font-weight and font-size to it since it shares a lot of the styles.

I think once we use the leafygreen provider and update some of the defaults we should be able to match the designs more closely unless there were other modifications made.

src/pages/my-patches/patch-card/BuildStatusIcon.tsx
SupaJoon: It is necessary for the tooltips to show up on the svg icons.

src/pages/MyPatches.tsx
tgrander: OH now we using an IIFE??!!!?!??!?!???!?!?!?!?!?!

PR #149, opened 2020-04-29
EVG-7811: Schedule a patch from configure page

tgrander:
https://jira.mongodb.org/browse/EVG-7811

schedule-patch

### Summary 
- Clicking "schedule" button on configure page redirects user to patch page for scheduled patch on success
- Shows error if scheduling patch fails

#### YOU CANNOT TEST THIS FEATURE LOCALLY - YOU MUST USE STAGING
Scheduling a patch requires having a github token in evergreen, which ain't gonna work locally. Use staging to test this feature by submitting a patch to staging and scheduling the patch on the configure page.


cypress/integration/patch/configure-patch.ts
```diff
@@ -178,6 +179,49 @@ describe("Configure Patch Page", () => {
...
+        .clear()
+        .type(val);
+      cy.route({
+        method: "POST",
+        url: "/graphql/query",
+        response: mockedErrorConfigureResponse,
+      });
+      cy.get("[data-cy=schedule-patch]").click();
+      cy.location().should((loc) =>
+        expect(loc.pathname).to.include(
```
khelif96, line 218: nit. We could prob just compare the path names directly.

cypress/integration/patch/configure-patch.ts
```diff
@@ -193,3 +237,118 @@ const getTaskCountFromText = (text) => {
   }
   return parseInt(text);
 };
+
+const mockedErrorConfigureResponse = {
+  errors: [
+    {
+      message: "WAH WAH CHICKEN WAH",
```
khelif96, line 244: lol

package.json
```diff
@@ -64,7 +64,7 @@
     "coverage": "npm run test -- --coverage --watchAll=false",
     "storybook": "start-storybook -p 9009 -s public",
     "build-storybook": "build-storybook -s public",
-    "codegen": "env-cmd -e devServer -r config/.cmdrc.json graphql-codegen --config codegen.js"
+    "codegen": "graphql-codegen --config codegen.js"
```
khelif96, line 67: Thanks allot. Can you please remove the variable in `.cmdrc_sample.json` too

src/pages/configurePatch/ConfigurePatchCore.tsx
```diff
@@ -8,19 +8,36 @@ import { useTabs, useDefaultPath } from "hooks";
 import { CodeChanges } from "pages/patch/patchTabs/CodeChanges";
```
khelif96, line 8: nit. i know this wasnt you but we should start setting some patterns around how we import external deps vs internal ones. How about External first, Then internal and everything in alphabetical order. Im sure theres probably an eslint extension that will automate this for us. what do you think?

tgrander: sorting imports is a divisive topics in the js community. some devs care a lot, some don't care at all. i struggle to see the value in sorting imports.

cypress/integration/patch/configure-patch.ts
tgrander: so true

package.json
tgrander: done

src/pages/configurePatch/ConfigurePatchCore.tsx
khelif96: Although the order doesn't make a difference I think it just produces cleaner looking code and it's easy to look at the deps and see what you have and what you don't. Same with sorting components and styles it's nice to be able to quickly glance and see what you have and don't have.

cypress/integration/patch-route.ts
```diff
@@ -173,8 +173,9 @@ describe("Patch route", function() {
       });
     });
 
+    // flakey test - must fix
     ["NAME", "STATUS", "BASE_STATUS", "VARIANT"].forEach((sortBy) => {
-      it(`Fetches tasks sorted by ${sortBy} when ${sortBy} header is clicked`, () => {
+      xit(`Fetches tasks sorted by ${sortBy} when ${sortBy} header is clicked`, () => {
```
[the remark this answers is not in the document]
tgrander: how do i force that? according to the cypress video they appear to be running at different times / are each listed as separate tests that run one after the other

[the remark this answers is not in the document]
khelif96: try placing it with in a normal for loop. forEach does some funny things with scoping. https://alligator.io/js/foreach-vs-for-loops/.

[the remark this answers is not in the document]
khelif96: So the tests might be scoped differently and not running consecutively every time

src/pages/configurePatch/ConfigurePatchCore.tsx
tgrander: if it can be automated by a tool like prettier then i'm all for it

package.json
john-m-liu: Could you remove it from the readme too? Thanks for cleaning this up though!

cypress/integration/patch-route.ts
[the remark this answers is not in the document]
tgrander: according to the article for/forEach are almost the same except forEach is kinda better

package.json
tgrander: sure thing

cypress/integration/patch-route.ts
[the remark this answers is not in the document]
khelif96: Actually we shouldnt be using for loops in cypress at all.  we should use the built in cypress statement.
```
cy.wrap(array).each((index) => {
doSomething(index)
})
```
https://stackoverflow.com/questions/52212868/cypress-io-writing-a-for-loop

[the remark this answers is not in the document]
khelif96: ~ This is the only reliable way, at least if the order of code execution is important. It's worth noting that - due to the asynchronous nature of cypress - a classical for loop will likely result in a non-deterministic execution order of the cy.* calls inside the loop.

src/pages/configurePatch/configurePatchCore/ConfigureTasks.tsx
```diff
@@ -82,7 +87,15 @@ export const ConfigureTasks: React.FC<Props> = ({
...
-        <Button variant={Variant.Primary}>Schedule</Button>
+        <Button
+          data-cy="schedule-patch"
+          variant={Variant.Primary}
+          onClick={onClickSchedule}
+          disabled={isEmpty(selectedVariantTasks) || loading}
+        >
+          {loading && <StyledIcon type="loading" />}
+          Schedule
+        </Button>
```
khelif96, line 98: I saw this interaction in the thread on slack can we generalize this and move it to its own component a button with a loading state would be super useful

src/pages/configurePatch/ConfigurePatchCore.tsx
```diff
@@ -55,6 +91,12 @@ export const ConfigurePatchCore: React.FC<Props> = ({ patch }) => {
   }
   return (
     <>
+      {errorSchedulingPatch && (
+        // TODO: replace with error banner
+        <div data-cy="error-banner">
+          There was a problem trying to schedule patch.
+        </div>
+      )}
```
khelif96, line 99: Do you see errors being displayed in the ui on the component level where they occurred or as more of a global error at the page level. As a banner that appears across the top of the screen. or as toast notification

tgrander: i like toast notifications but the designers decided on banners, at least for now

khelif96: I think banners are used throughout mongo so they are probably trying to keep a similar design language. University and stitch both use them.

src/pages/configurePatch/configurePatchCore/ConfigureTasks.tsx
tgrander: yes!

khelif96: Great ill wait for that so i can use it in my pr

tgrander:
failing test is unrelated failing my patches page test. might be a flake, have notified arjun.

PR #150, opened 2020-04-29
EVG-7909 Wire up remove patch from commit queue mutation

khelif96:
Tests wont pass until the corresponding pr in evergreen is merged so i can get some data. 
https://github.com/evergreen-ci/evergreen/pull/3540

Also realized there also isn't an empty state for the commit queue. Lara is working on some designs

commitqueue

### DIFF
--- a/cypress/integration/commit_queue.ts
+++ b/cypress/integration/commit_queue.ts
@@ -66,8 +66,6 @@ describe("commit queue page", function() {
     cy.get("[data-cy=commit-queue-patch-button]").should("exist");
     cy.get("[data-cy=commit-queue-patch-button]").click();
     cy.waitForGQL("CommitQueue");
-    cy.get("@gqlQuery").then(() => {
-      cy.get("[data-cy=commit-queue-card]").should("not.exist");
-    });
+    cy.get("[data-cy=commit-queue-card]").should("not.exist");
   });
 });
--- a/cypress/integration/commit_queue.ts
+++ b/cypress/integration/commit_queue.ts
@@ -59,4 +59,15 @@ describe("commit queue page", function() {
       cy.get("[data-cy=commitQueue-card-error]").contains(errorMessage);
     });
   });
+
+  it("Clicking on remove a patch from the commit queue should work", () => {
+    cy.visit(COMMIT_QUEUE_ROUTE_1);
+    cy.get("[data-cy=commit-queue-card]").should("exist");
+    cy.get("[data-cy=commit-queue-patch-button]").should("exist");
+    cy.get("[data-cy=commit-queue-patch-button]").click();
+    cy.waitForGQL("CommitQueue");
+    cy.get("@gqlQuery").then(() => {
+      cy.get("[data-cy=commit-queue-card]").should("not.exist");
+    });
+  });
 });
--- a/src/pages/CommitQueue.tsx
+++ b/src/pages/CommitQueue.tsx
@@ -4,6 +4,7 @@ import { PageWrapper } from "components/styles";
 import { useParams } from "react-router-dom";
 import Badge from "@leafygreen-ui/badge";
 import { uiColors } from "@leafygreen-ui/palette";
+import { Body } from "@leafygreen-ui/typography";
 import { useQuery } from "@apollo/react-hooks";
 import get from "lodash/get";
 import { Skeleton } from "antd";
@@ -67,6 +68,7 @@ export const CommitQueue = () => {
             commitQueueId={commitQueue.projectId}
           />
         ))}
+      {!queue && <Body>There are no items in this queue. </Body>}
     </PageWrapper>
   );
 };
--- a/src/pages/CommitQueue.tsx
+++ b/src/pages/CommitQueue.tsx
@@ -44,26 +44,28 @@ export const CommitQueue = () => {
         title="Commit Queue"
         badge={
           <Badge variant="darkgray">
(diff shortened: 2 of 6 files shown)


cypress/integration/commit_queue.ts
```diff
@@ -59,4 +59,15 @@ describe("commit queue page", function() {
...
     });
   });
+
+  it("Clicking on remove a patch from the commit queue should work", () => {
+    cy.visit(COMMIT_QUEUE_ROUTE_1);
+    cy.get("[data-cy=commit-queue-card]").should("exist");
+    cy.get("[data-cy=commit-queue-patch-button]").should("exist");
+    cy.get("[data-cy=commit-queue-patch-button]").click();
+    cy.waitForGQL("CommitQueue");
+    cy.get("@gqlQuery").then(() => {
```
tgrander, line 69: i don't think you need to `get` the gql query if you're not using it to inspect the request/response

src/pages/commitqueue/CommitQueueCard.tsx
```diff
@@ -28,7 +35,26 @@ export const CommitQueueCard: React.FC<Props> = ({
...
+  const handleEnroll = async (e) => {
+    e.preventDefault();
+    setButtonDisabled(true);
+    try {
+      await removePatchFromCommitQueue({
+        variables: { patchId, commitQueueId },
+        refetchQueries: ["CommitQueue"],
+      });
+    } catch (error) {
+      setButtonDisabled(false);
```
tgrander, line 54: instead of using state, you could set button to disabled when `loading` from the mutation is `true`

cypress/integration/commit_queue.ts
```diff
@@ -59,4 +59,13 @@ describe("commit queue page", function() {
       cy.get("[data-cy=commitQueue-card-error]").contains(errorMessage);
     });
   });
+
+  it("Clicking on remove a patch from the commit queue should work", () => {
```
tgrander, line 63: what is meant to happen if there is an error removing the item?

khelif96: There arent any error states in the designs. But ideally if theres an error we would so some sort of universal error banner to indicate there was an error on the page

khelif96: Cant really write a test for an error if theres no error state.

khelif96: Just spoke to Lara and she showed me the error banners. If these arent implemented already. I can probably make them in a separate ticket and use them here.

src/pages/commitqueue/CommitQueueCard.tsx
```diff
@@ -28,7 +35,23 @@ export const CommitQueueCard: React.FC<Props> = ({
...
+  const [removePatchFromCommitQueue, { loading }] = useMutation<
+    RemovePatchFromCommitQueueMutation,
+    RemovePatchFromCommitQueueMutationVariables
+  >(REMOVE_PATCH_FROM_COMMIT_QUEUE);
+  const handleEnroll = async (e) => {
+    e.preventDefault();
+    try {
+      await removePatchFromCommitQueue({
+        variables: { patchId, commitQueueId },
+        refetchQueries: ["CommitQueue"],
```
SupaJoon, line 49: I think it's a good practice to include the query variables for the refetch

src/pages/CommitQueue.tsx
```diff
@@ -44,26 +45,30 @@ export const CommitQueue = () => {
         title="Commit Queue"
         badge={
           <Badge variant="darkgray">
-            {queue.length} Item {queue.length > 1 && "s"}
+            {queue ? queue.length : 0} Item
```
john-m-liu, line 48: I think "0 items" is actually the correct way to say it

cypress/integration/commit_queue.ts
tgrander: there is already a ticket for it and i'm working on it right now. a lot of things are not included in the designs that we as front end engineers must read between the lines for. how it looks is irrelevant. my question is about **what** should happen when there is an error because right now i do not believe anything in the ui will happen. you don't need to show a banner, just render an error div with a TODO for showing the banner.

khelif96: Nothing in the UI does happen currently but i will add a conditionally rendered block to display the error in the card ui. I did leave a comment in the component that says "Replace this with better error handling" https://github.com/evergreen-ci/spruce/pull/150/files#diff-1b4c4e95f91af6974a5c0ce1977feeaeR52

tgrander: a catch block is where we would log the error. all interactions should still handle loading/error on a ui level.

src/pages/CommitQueue.tsx
tgrander: i also had trouble reading this

src/pages/commitqueue/CommitQueueCard.tsx
khelif96: The Apollo docs say they are optional but if we want to adopt this practice i dont mind.  It does have drawbacks tho in that we will have to pass in the query variables into every refetchQueries instance and could involve passing  props to components just for the sake of  refetching the query. Up to you what do you think?

SupaJoon: @khelif96 that's a good point. let's leave it as is.

PR #151, opened 2020-04-29
EVG-7515: realtime updates for patch build variants

tgrander:
Screen Shot 2020-04-29 at 6 51 45 PM

Patch build variants query polls the server until all tasks have ran or finished. 

I tested in staging.


cypress/integration/patch/patch-route/tasks-filters.ts
```diff
@@ -75,7 +75,8 @@ const singularStatuses = [
   },
 ];
 
-describe("Tasks filters", function() {
+// TODO: refactor to not wait for / inspect gql query. Use defined data set and test against it.
```
khelif96, line 78: Could you make a ticket for these so we dont forget to come back to them in the future.

src/pages/patch/BuildVariants.tsx
```diff
@@ -12,15 +12,24 @@ import { Skeleton } from "antd";
 import { TaskSquare } from "./buildVariants/TaskSquare";
 import { H3, P1 } from "components/Typography";
 import styled from "@emotion/styled/macro";
+import every from "lodash/every";
+import { TaskStatus } from "types/task";
```
khelif96, line 16: Does this type exist in the generated types?

cypress/integration/patch/patch-route/tasks-filters.ts
tgrander: one already exists. https://jira.mongodb.org/browse/EVG-7923

src/pages/patch/BuildVariants.tsx
tgrander: this is not a gql type and should never change, therefore does not need to be generated. it's a global evergreen type representing the different task statuses.

tgrander:
failing tests = flakey my patches test

PR #153, opened 2020-04-30
EVG-7557: include file count in Files Tab

SupaJoon:
https://jira.mongodb.org/browse/EVG-7557
Screen Shot 2020-04-30 at 6 40 47 PM

### DIFF
--- a/cypress/integration/page_tabs.ts
+++ b/cypress/integration/page_tabs.ts
@@ -126,9 +126,8 @@ describe("Tabs", () => {
       const fileCountPath = "responseBody.data.taskFiles.fileCount";
       cy.waitForGQL("GetTask", {
         [fileCountPath]: valExists,
-      }).then((xhr) => {
-        cy.dataCy("files-tab-badge").contains(get(xhr, fileCountPath));
       });
+      cy.dataCy("files-tab-badge").contains("0");
     });
   });
 });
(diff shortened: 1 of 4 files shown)


cypress/integration/page_tabs.ts
```diff
@@ -121,5 +120,17 @@ describe("Tabs", () => {
...
+
+    it("Should display a badge with the number of files in the Files tab", () => {
+      cy.visit(taskRoute);
+      const fileCountPath = "responseBody.data.taskFiles.fileCount";
+      cy.waitForGQL("GetTask", {
+        [fileCountPath]: valExists,
+      }).then((xhr) => {
+        cy.dataCy("files-tab-badge").contains(get(xhr, fileCountPath));
+      });
+    });
```
tgrander, line 132: aren't we trying to move away from the matching req/res against ui method and toward a defined data set method? this will probably just have to be refactored later.

SupaJoon: I can change this `cy.dataCy("files-tab-badge").contains(get(xhr, fileCountPath));` to `cy.dataCy("files-tab-badge").contains("0");` but we will still need to wait for some time before the tab updates with the file count since we are relying on a live gql query. I think stubbing will fix this because we won't have to wait for a live network request with variable response times.

PR #157, opened 2020-05-04
EVG-7937 Pre commit and pre push hooks 

khelif96:
This adds `pre-commit` and `pre-push` hooks to our work flow. 
When ever you try to commit any changes a git hook will automatically run `eslint` and `tslint` on only the files that were staged for commit. 

When you try to push up changes it will automatically run `npm run test`.  Which will run any tests we have written/ will write. Once we start writing more tests we can add a flag to this so it will only run tests on relevant files. 

Also added a [rules of hooks eslint plugin ](https://reactjs.org/docs/hooks-rules.html) This will help as we write more hooks in the future and ensure we don't accidentally break any of the rules of hooks.

The majority of the changes in this pr revolve around fixing any tslint warnings from functions that dont have return types declared.


khelif96:
I spoke to @john-m-liu  and it should be possible to run eslint on the ci server ill try to get that setup

> We're currently only using tslint in CI so react-hooks will only be checked locally since it's in the eslint config file. The reason we have both eslint and tslint is because gql query linting is only available for eslint and we have been using that while developing locally. I'm not opposed to switching from tslint to eslint completely because it offers more features and gives us a chance to run gql query linting in ci.

khelif96:
> cool! we don't write tests other than cypress right now tho so i'm not sure `npm run test` will accomplish anything other than adding a few seconds to the time it takes to push changes.

Yeah i figured there werent much tests written only 3 test suites run and i think they are mainly meant to make sure the app still runs and test our gql wrapper.
I can update this so that the tests will only run if you actually wrote some tests/ updated something that touches the tests but wont run if your changes arent related to the tests.

.evergreen.yml
```diff
@@ -215,8 +229,10 @@ tasks:
   - name: lint
     commands:
     - func: get-project
+    - func: sym-link
     - func: npm-install
     - func: npm-lint
```
SupaJoon, line 234: can we add the rules we are using in tslint to eslint so we can remove tslint?

khelif96:
After speaking with @SupaJoon  i realized we weren't actually doing typechecking. `tslint` and `eslint` just ensure code is formatted properly and certain patterns and rules are enforced. But in order to do actual typechecking we must run the type script compiler to get what ever output it spits out. 
I added that ability as well as added it to the precommit hook. 

Now when ever you try to commit we will run both `eslint` and `prettier` on your files as well as run the typescript compiler on any typescript files you have staged for commit. Changes will only be linted and type checked if they are staged for commit. So if the file is not staged for a commit it wont be type checked. The checks also happen in parallel. 

Below is an example of the output you would receive if you tried to commit the file `MyPatches.tsx` and had an error in your typescript types. the commit will be aborted but you can by pass this if needed. using the `--no-verify` flag. 
image

This is what a successful commit will look like. Notice how the tasks didnt run because this commit does not have any file types that match what `lint-staged` is looking for.
image

package.json
```diff
@@ -47,7 +47,7 @@
   "scripts": {
     "start": "react-scripts start",
     "build": "react-scripts build",
-    "test": "react-scripts test",
+    "test": "react-scripts test --watchAll=false",
```
tgrander, line 50: is this the command that ensures that tests only run if  relevant file was changed? i don't think it's worth running `npm run test` since we do not write tests other than e2e right now. we can always turn on this feature when there are tests to run.

khelif96: Theres an additional flag that i didnt enable and that fair ill disable it for now.

khelif96: @tgrander  just disabled the pre push hooks so no tests will run on push

tgrander: awesome! will approve once tests are passing

PR #158, opened 2020-05-04
EVG-7772: render errors for patch page

tgrander:
Screen Shot 2020-05-04 at 2 58 02 PM

Render an error banner if error occurs while fetching patch

### Each page has its own banner context
I played with a lot of different approaches and settled on this one.

Why not use a global banners context for all component?
Because then banners from one page appear on another page because the banners state is global. This would require clearing the banners state whenever page changes, which I don't want to do.


src/pages/Patch.tsx
```diff
@@ -61,3 +83,11 @@ export const Patch = () => {
...
   );
 };
+
+export const Patch = () => {
+  return (
+    <BannerContextProvider>
+      <PatchCore />
+    </BannerContextProvider>
+  );
+};
```
tgrander, line 93: this is not ideal but i could not successfully get other approaches to work. i tried to automatically render each route wrapped in `BannerContextProvider` where all routes are rendered. Spent about an hour on that but it was not successful.

SupaJoon: I'm not sure if you tried this but another abstraction to clean this up could be to create a higher order component that returns the argument wrapped with BannerContextProvider. So it could look something like `export const Patch = addBanner(() => {...})`

khelif96: The higher order component idea is great. Wrapping each route with a `withBannerContext` HOC would work great. And allows us to declare which routes have the context and saves some repetition from declaring it in each page.

tgrander: such a good idea!

src/pages/Patch.tsx
```diff
@@ -34,12 +42,26 @@ export const Patch: React.FC = () => {
...
   const description = get(patch, "description");
   const activated = get(patch, "activated");
-
+  useEffect(() => {
+    if (error) {
+      dispatchBanner.error(
+        `There was an error loading the patch: ${error.message}`
+      );
+    }
+  }, [error]);
```
khelif96, line 51: You should add dispatchBanner as a dependency for the useEffect hook. 
image

khelif96: I think if you have a plugin that runs eslint --fix on save it should automatically handle resolving use effect dependencies for you.

src/pages/Patch.tsx
```diff
@@ -17,10 +17,18 @@ import get from "lodash/get";
 import { Metadata } from "pages/patch/Metadata";
 import { useHistory } from "react-router-dom";
 import { paths } from "constants/routes";
+import {
+  useBannerDispatchContext,
+  useBannerStateContext,
+} from "context/banners";
+import { Banners } from "components/Banners";
 import { PatchStatusBadge } from "components/PatchStatusBadge";
+import { withBannersContext } from "higherOrderComponents/withBannersContext";
```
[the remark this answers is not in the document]
SupaJoon: @khelif96 I'm on board with that

src/pages/Patch.tsx
SupaJoon: useQuery also accepts an onError callback in the options where we can dispatch the banner.

tgrander: i left it out on purpose because it calls `dispatchBanner` in an endless loop. i will use the `onError` callback.

PR #159, opened 2020-05-04
EVG-7947: shared button component with loading icon

tgrander:
### Shared button component

schedule-patch-loading

Adds some tasty features to the leafygreen button, like showing a loading indicator

### DIFF
--- a/src/components/Button/index.tsx
+++ b/src/components/Button/index.tsx
@@ -18,6 +18,7 @@ interface Props {
   onClick?: () => void;
   disabled?: boolean;
   dataCy?: string;
+  glyph?: React.ReactElement;
 }
 
 export const Button: React.FC<Props> = ({
@@ -27,6 +28,7 @@ export const Button: React.FC<Props> = ({
   onClick,
   disabled,
   dataCy,
+  glyph,
 }) => {
   return (
     <LeafyGreenButton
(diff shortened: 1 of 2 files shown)


src/components/Button/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+  dataCy,
+}) => {
+  return (
+    <LeafyGreenButton
+      data-cy={dataCy}
+      variant={mapVariantToLeafyGreenVariant[variant]}
+      onClick={onClick}
+      disabled={disabled}
+    >
+      {loading && <StyledIcon type="loading" />}
```
khelif96, line 38: I think you can pass the Icon directly into the leafygreen button it has a `glyph` prop. 

https://github.com/mongodb/leafygreen-ui/blob/6b4cd8fecaa09350b739b9d3cf246a6ba30dfb3d/packages/button/src/Button.tsx#L298

tgrander: leafygreen does not have this icon -- it's from antd

khelif96: It takes a `ReactElement` as an input. So you should be able to pass `<StyledIcon/>` Directly into it.
https://github.com/mongodb/leafygreen-ui/blob/6b4cd8fecaa09350b739b9d3cf246a6ba30dfb3d/packages/button/src/Button.story.tsx#L39

tgrander: oh neat

khelif96: haha i had to do this exact thing like 100 times during my stitch rotation.

khelif96:
> ### Shared button component
> schedule-patch-loading
> 
> Adds some tasty features to the leafygreen button, like showing a loading indicator

Do you think its possible to make the button not shift the surrounding buttons around it when the loading animation starts?

tgrander:
> Do you think its possible to make the button not shift the surrounding buttons around it when the loading animation starts?

Not sure this is possible since the icon takes up space. What are the downsides of the adjacent content shifting 8px?

PR #161, opened 2020-05-05
EVG-7511: render commit queue position in patch metadata

tgrander:
Show commit queue position and link to commit queue page

cq-position

### Failing tests
Due to cq data not being in local evergreen
[This PR](https://github.com/evergreen-ci/evergreen/pull/3558) must be merged for tests to pass

### DIFF
--- a/cypress/integration/patch-route.ts
+++ b/cypress/integration/patch-route.ts
@@ -55,6 +55,12 @@ describe("Patch route", function() {
     cy.get("#task-count").within(hasText);
   });
 
+  it("Shows commit queue position in metadata if patch is on commit queue", function() {
+    cy.visit(`/patch/${patch.id}`);
+    cy.dataCy("commit-queue-position").click();
+    cy.location("pathname").should("eq", "/commit-queue/mongodb-mongo-test");
+  });
+
   it("'Base commit' link in metadata links to version page of legacy UI", function() {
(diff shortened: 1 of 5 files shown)


cypress/integration/patch-route.ts
```diff
@@ -55,6 +55,12 @@ describe("Patch route", function() {
     cy.get("#task-count").within(hasText);
   });
 
+  it("Shows commit queue position in metadata if patch is on commit queue", function() {
```
khelif96, line 58: can we use arrow functions here so its consistent with the other tests.

src/gql/generated/types.ts
```diff
@@ -111,6 +116,7 @@ export type Mutation = {
   restartTask: Task;
   saveSubscription: Scalars['Boolean'];
   removePatchFromCommitQueue?: Maybe<Scalars['String']>;
+  restartPatch?: Maybe<Scalars['String']>;
```
john-m-liu, line 119: Not sure if this was intentional but looks like it generated some types from a different schema change (but I'm guessing that it'll need to be added anyway)

khelif96: Yeah i guess `graphql-codegen` generates the types for everything in the `schema.graphql` file we pass in. Regardless of if we use them or not

PR #167, opened 2020-05-11
EVG-7965 Add the patch modal to the patch page

khelif96:
May-13-2020 13-01-18

not really sure how this pr got to 26 files changed. 

But i Created the patch modal for restarting user patches.

While working on this i noticed the entire app gets rerendered every time a query is made. Turns out apollo link was updating the authentication context every query which causes the entire app to rerender. Solved that by adding a check to  not update the context if the users auth state hasnt changed. 

Also noticed eslint wasnt running on cypress tests fixed that in this pr and have a follow up pr to make all the necessary changes.
#168


src/hooks/usePatchStatusSelect.ts
```diff
@@ -0,0 +1,77 @@
+import { useState, useEffect } from "react";
+import { PatchBuildVariant } from "gql/generated/types";
+
+export const usePatchStatusSelect = (
```
tgrander, line 4: i really, really like that you extracted the state logic of a component into a hook. when a component has complex state management, it can get messy quickly, e.g. the configure patch page. using hooks to encapsulate that messy state logic is a great way to separate the different concerns of a component.

src/pages/Patch.tsx
```diff
@@ -65,6 +65,16 @@ const PatchCore: React.FC = () => {
...
         title={description || `Patch ${get(patch, "patchNumber")}`}
         badge={<PatchStatusBadge status={status} />}
+        buttons={
+          <ActionButtons
+            canAbort
+            canRestart
+            canSchedule
+            canUnschedule
+            canSetPriority
+            initialPriority={1}
```
tgrander, line 75: i'm confused. where are these coming from?

src/pages/patch/patchModal/PatchModal.tsx
```diff
@@ -0,0 +1,228 @@
...
+  onCancel: () => void;
+}
+export const PatchModal: React.FC<PatchModalProps> = ({
+  visible,
+  onOk,
+  onCancel,
+}) => {
+  const { successBanner, errorBanner, clearAll } = useBannerDispatchContext();
+  const { id } = useParams<{ id: string }>();
+  const [abortInProgressTasks, setAbortInProgressTasks] = useState(false);
```
tgrander, line 39: same comment about boolean names. suggestion: `shouldAbortInProgressTasks `

src/pages/patch/patchModal/PatchModal.tsx
```diff
@@ -0,0 +1,228 @@
...
+  const [restartPatch, { loading: mutationLoading }] = useMutation<
+    RestartPatchMutation,
+    RestartPatchMutationVariables
+  >(RESTART_PATCH, {
+    onCompleted: () => {
+      onOk();
+      successBanner(`Successfully restarted patch: ${id}`);
```
tgrander, line 46: we do not use the patch id anywhere in the app to identify the patch (other than the url). therefore i do not think we should use it here in the success message. We could use the patch number, the description, or none of those at all and just say `success!`

src/pages/patch/patchModal/PatchModal.tsx
```diff
@@ -0,0 +1,228 @@
...
+  const patchBuildVariants = get(data, "patchBuildVariants");
+  const [
+    selectedTasks,
+    validStatus,
+    { toggleSelectedTask, setValidStatus },
+  ] = usePatchStatusSelect(patchBuildVariants);
+
+  const handlePatchRestart = async (e): Promise<void> => {
+    e.preventDefault();
+    clearAll();
```
tgrander, line 64: can this be renamed too to be like the rest of them? `clearAllBanners`

src/pages/patch/patchModal/PatchStatusCheckboxContainer.tsx
```diff
@@ -0,0 +1,70 @@
...
+      data-cy="task-status-checkbox"
+      onChange={() => toggleSelectedTask(id)}
+      label={
+        <PatchStateItemWrapper>
+          {square} {name}
+        </PatchStateItemWrapper>
+      }
+      checked={selectedTasks.includes(id)}
+      bold={false}
+    />
```
tgrander, line 58: does this need to be its own component? there is not much logic or complexity involved with it--i do not see anything that makes it benefit from it being grouped into `PatchStatusCheckboxContainer`

cypress/integration/patch/patch-route/patch-modal.ts
```diff
@@ -0,0 +1,63 @@
...
+    cy.get(`[data-cy=patch-status-filter] > .cy-treeselect-bar`).click();
+
+    // ideally this would target the text field itself but leafygreen Body tags dont
+    // support cy-data elements currently
+    cy.dataCy("patch-modal").should(
+      "contain.text",
+      allTasksSelectedConfirmationMessage
+    );
+  });
+  it("Restarting a task should close the modal and display a success message if it occurs successfully", () => {
```
tgrander, line 54: what happens when it errors?

src/components/Accordian/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+import styled from "@emotion/styled";
+import Icon from "@leafygreen-ui/icon";
+
+interface AccordianProps {
+  title: React.ReactNode;
+  contents: React.ReactNode;
+}
+export const Accordian: React.FC<AccordianProps> = ({ title, contents }) => {
+  const [toggleAccordian, setToggleAccordian] = useState(false);
+  const toggleAccordianHandler = (): void =>
```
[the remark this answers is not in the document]
SupaJoon: this may be necessary for eslint

package.json
```diff
@@ -142,6 +144,7 @@
     "jest-junit": "10.0.0",
     "lint-staged": "10.2.2",
     "prettier": "1.19.1",
+    "react-test-renderer": "16.12.0",
```
SupaJoon, line 147: where are we using this module? I don't see it imported into any our tests

src/hooks/tests/usePatchStatusSelect.test.ts
```diff
@@ -0,0 +1,144 @@
+import { renderHook, act } from "@testing-library/react-hooks";
```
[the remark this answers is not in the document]
khelif96: I definitely agree e2e tests are extremely valuable as they allow you to test how the system operates as a whole and determine if the program works as intended. 

But i think its definitely still worth it and complimentary to write unit tests for complex functionality like this. Since you are able to ensure the function works as intended and its behavior is captured and defined by the test. When an e2e test breaks it might be difficult to determine why or how it broke but with more fine grained tests your able to determine if the features that build up to the e2e test are still behaving as expected and it makes it simple to debug why an e2e test is failing if you can determine what feature used in the test has broken or isnt behaving as intended.

Its also nice to be able to do TDD for something like this with unit tests. Since i can define the expected behavior and define edge cases for it And then build out and refine the functionality so i can capture the different scenarios this function may go through.

src/components/PageTitle.tsx
```diff
@@ -49,3 +51,7 @@ const PageHeader = styled.div`
   display: flex;
   justify-content: space-between;
 `;
+
+const TitleWrapper = styled.div`
+  width: 60%;
```
SupaJoon, line 56: I think this may be over restricting the title width: 
example from code change:
Screen Shot 2020-05-12 at 1 19 27 PM

design:
Screen Shot 2020-05-12 at 1 19 06 PM

src/pages/Patch.tsx
khelif96: These are the required props for the `ActionButtons` component on the patch page. Since their functionality isn't set up i just passed in these props as true

src/pages/patch/patchModal/PatchModal.tsx
khelif96: Sure!

khelif96: tbh tho after using destructuring to pull out these functions for the banners i feel like i prefer the non destructured way of calling these functions. 

i think `banner.warningBanner()` or `banner.clearAllBanners()` or  even `banner.createWarning()` is so much more clearer of the intent and is easier to keep track of whats going on. Just my opinion tho. What do you think? @SupaJoon @tgrander

cypress/integration/patch/patch-route/patch-modal.ts
khelif96: It should display an error banner but its kind of difficult to represent error states in cypress e2e tests since you would have to force the application to error in some way.

khelif96:
> Per the designs, the dropdown of task statuses is only meant to include three of the statuses and the checkbox is a "mixed state"
> https://mongodb.invisionapp.com/d/main/#/console/19647524/416805006/preview

I spoke to @john-m-liu  there should be more states represented here the designs don't feature them all. He also left a comment on the designs indicating that.
image

src/hooks/tests/usePatchStatusSelect.test.ts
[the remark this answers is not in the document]
SupaJoon: I think it makes sense to unit test shared hooks because it's easier to rule out issues when e2e tests fail.  With the correct unit tests in place, it may not be necessary to select every combo of inputs in e2e tests to ensure a feature works as intended. What do ya'll think? @tgrander @khelif96

src/pages/Patch.tsx
SupaJoon: I think these are just placeholder props from the ActionButtons in the task page. The contents of ActionButtons will have to reflect the actions we can perform on a patch. This is the closest ticket that describes the task: https://jira.mongodb.org/browse/EVG-7353

package.json
khelif96: Its used by `react-hooks-testing-library` its a peer dependency that doesnt come bundled with i

src/hooks/tests/usePatchStatusSelect.test.ts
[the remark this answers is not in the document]
tgrander: e2e tests that are written well in cypress, meaning they are modular and test single units of the feature at a time, should not be difficult to debug. cypress makes debugging failed tests easy and personally i have not had trouble diagnosing a failing test unless the problem is with cypress itself, i.e. async difficulties. 

i am not going to write unit tests for most of my custom hooks because i plant to test them via the e2e tests. this hook specifically exists for one component. therefore it is essentially the component logic abstracted to a hook for better readability. since it is apart of the component i would test it thoroughly through the e2e tests.

i'm not going to object to anyone writing more extensive tests for a hook. you are correct that unit tests can catch more edge cases because they are not limited to a predefined set of data like the e2e tests are. however i will probably not write any unit tests...maybe ever. we'll see.

src/pages/patch/patchModal/PatchModal.tsx
tgrander: this is how i intended for the context to be used so yes, i am on board with it. but as you will find with a lot of things in this regard, i do not have a strong opinion because my opinions are generally spent elsewhere.

cypress/integration/patch/patch-route/patch-modal.ts
tgrander: can you mock the error response?

src/pages/patch/PatchModal.tsx
```diff
@@ -0,0 +1,180 @@
...
+        "data-cy": "patch-modal",
+      }}
+    >
+      {!loading && patchBuildVariants && (
+        <>
+          <TreeSelect
+            onChange={setValidStatus}
+            state={validStatus}
+            tData={statusesTreeData}
+            inputLabel="Tasks Selected: "
```
john-m-liu, line 115: The tree select has a default label of "no filters selected" which doesn't make much sense in this context. Could you expose that default as a prop?

src/pages/patch/PatchModal.tsx
```diff
@@ -0,0 +1,180 @@
+import React, { useState } from "react";
```
john-m-liu, line 1: This file name is a little vague since there will likely be more patch modals in the future. Can you call it something like PatchRestartModal?

PR #169, opened 2020-05-12
EVG-7994 Add the notification tab

khelif96:
Created the notifications tab of the user preferences page. Cypress tests will follow once the mutations are created and everything is wired up and all 3 pages are made. 

image

### DIFF
--- a/src/pages/preferences/NotificationsTab.tsx
+++ b/src/pages/preferences/NotificationsTab.tsx
@@ -4,9 +4,8 @@ import Card from "@leafygreen-ui/card";
 import TextInput from "@leafygreen-ui/text-input";
 import Button, { Variant } from "@leafygreen-ui/button";
 import { Body } from "@leafygreen-ui/typography";
-import { Radio } from "antd";
 import { Notifications } from "gql/generated/types";
-import { notificationFields } from "./fieldMaps";
+import { NotificationField } from "./NotificationField";
 
 interface ProfileTabProps {
   slackUsername?: string;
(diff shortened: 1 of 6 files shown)


src/pages/preferences/NotificationsTab.tsx
```diff
@@ -0,0 +1,157 @@
...
+}) => {
+  const [slackUsernameField, setslackUsernameField] = useState(slackUsername);
+  const [notificationStatus, setNotificationStatus] = useState(notifications);
+
+  const hasFieldUpdates =
+    slackUsername !== slackUsernameField ||
+    notificationStatus !== notifications;
+  const omitTypename = (key, value) =>
+    key === "__typename" ? undefined : value;
+  const newPayload = JSON.parse(JSON.stringify(notifications), omitTypename);
```
khelif96, line 27: I iterate through the notification object to generate the table and __typename is included as a field so this just removes it before the fields are rendered.

src/pages/preferences/NotificationsTab.tsx
```diff
@@ -0,0 +1,157 @@
...
+    <GridField gridArea={`${2 + index} / 3 / ${2 + index} / 4`}>
+      <Radio value="slack" />
+    </GridField>
+    <GridField gridArea={`${2 + index} / 4 / ${2 + index} / 5`}>
+      <Radio value="email" />
+    </GridField>
+    <GridField gridArea={`${2 + index} / 5 / ${2 + index} / 6`}>
+      <Radio value="" />
+    </GridField>
+  </GridCapableRadioGroup>
```
tgrander, line 110: since this is its whole own component can it live in its own file?

PR #170, opened 2020-05-12
EVG-7950: Paginate Tasks & Tests Table

SupaJoon:
https://jira.mongodb.org/browse/EVG-7950
This PR introduces two new shared components to adjust the page size and page for the table. These components are used on the tests and tasks table.
page


cypress/integration/patch/task-table.ts
```diff
@@ -0,0 +1,205 @@
...
+    cy.contains(TABLE_SORT_SELECTOR, "Name").click();
+    cy.once("fail", (err) => {
+      expect(err.message).to.include(
+        "'pointer-events: none' prevents user mouse interaction."
+      );
+    });
+  });
+
+  // flakey test - must fix
+  ["NAME", "STATUS", "BASE_STATUS", "VARIANT"].forEach((sortBy) => {
```
khelif96, line 69: You can try using a combination of the cypress `.each` command with a promise to iterate through this array and run each test with out them overlapping each other
https://docs.cypress.io/api/commands/each.html#Promises

cypress/integration/test_table.ts
```diff
@@ -254,20 +255,57 @@ describe("Tests Table", () => {
...
+        dataCyPrevPage,
+        firstPageDisplayNames,
+        0,
+        dataCyTableRows
+      );
+    });
+  });
+
+  describe("Changing page size updates URL and renders less than or equal to that many rows ", () => {
+    [20, 10, 50, 100].forEach((pageSize) => {
```
khelif96, line 301: Is this test less flaky then the one above? They seem to be doing similar things.

cypress/utils/index.js
```diff
@@ -105,3 +105,41 @@ export const elementExistenceCheck = (
...
+) => {
+  cy.get(dataCyPageBtn).click();
+  cy.get(dataCyTableRows).each(($el, index) => {
+    cy.wrap($el).contains(tableDisplayNames[index]);
+  });
+  cy.location("search").should("include", `page=${pageQueryParamValue}`);
+};
+
+/**
+ * Assert limit query param and table length after clicking on a page size button
```
khelif96, line 129: nice documentation

src/components/Pagination.tsx
```diff
@@ -0,0 +1,44 @@
...
+  const { search, pathname } = useLocation();
+  const setPage = (page: number) =>
+    replace(
+      `${pathname}?${queryString.stringify(
+        {
+          ...queryString.parse(search, { arrayFormat }),
+          page,
+        },
+        { arrayFormat }
+      )}`
```
khelif96, line 29: im wondering if this is something we might want to pull out and turn into more of a global utility function. I've seen similar logic used in other parts of the app.

src/components/styles/Layout.tsx
```diff
@@ -33,6 +34,20 @@ export const PageTitle = styled(H2)`
   margin-bottom: 16px;
 `;
 
+export const TableContainer = styled("div")`
+  display ${(props: { hide: boolean }): string => (props.hide ? "none" : "")}
```
khelif96, line 38: I think you might be missing a `:` also it might be better to do something like this.

`${(props: {hide: boolean}): string => (props.hide && 'display:none;')}`

This way if hide is false the css wont resolve to `display: `

src/pages/patch/patchTabs/Tasks.tsx
```diff
@@ -38,83 +50,60 @@ export const Tasks: React.FC<Props> = ({ taskCount }) => {
...
-        return patchTasks;
-      },
-    });
-  };
-
   if (error) {
     return <div>{error.message}</div>;
   }
+  const { limit, page } = getQueryVariables(search);
+  const isLoading = networkStatus < NetworkStatus.ready;
```
khelif96, line 78: Apollo also exports a `isNetworkRequestInFlight`  function.

src/pages/task/testsTable/TestsTableCore.tsx
```diff
@@ -79,86 +83,66 @@ export const TestsTableCore: React.FC = () => {
...
 
   // initial table sort button state to reflect initial URL query params
-  const { cat, dir } = getQueryVariables(search);
+  const { cat, dir, pageNum, limitNum } = getQueryVariables(search);
+
   columns.find(({ key }) => key === cat).defaultSortOrder =
     dir === SortDirection.Asc ? "ascend" : "descend";
-  const filteredTestCount = get(data, "taskTests.filteredTestCount", "-");
-  const totalTestCount = get(data, "taskTests.totalTestCount", "-");
+  const isLoading = networkStatus < NetworkStatus.ready;
```
khelif96, line 111: same comment as above

cypress/integration/patch/task-table.ts
```diff
@@ -0,0 +1,205 @@
...
+  "test-cloud",
+  "test-service",
+  "test-thirdparty",
+  "test-thirdparty",
+  "test-model",
+  "generate-lint",
+  "js-test",
+  "test-agent",
+  "test-auth",
+];
```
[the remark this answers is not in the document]
khelif96: It does seem brittle but i think as long as we don't make updates the data this relies on it should be fine.  Since we are using this to assert the data we expect is being returned. This is the best way to determine the system actually operates correctly. If we had this project  in the same directory as the data we could have probably fetched it directly from the data we use to seed the db.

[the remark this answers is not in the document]
khelif96: I think it might make sense to move this test data outside of the test tho. Like how we do for the gql tests in evergreen. Let test data live in its own directory/file just to clean up these test files a little bit

cypress/integration/patch/task-table.ts
SupaJoon: this test flaked because `clickSorterAndAssertTasksAreFetched` is unreliable. How can you tell if the tests are overlapping? From the test runner, I can see that the tests are running synchronously.

cypress/integration/patch/task-table.ts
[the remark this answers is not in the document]
tgrander: in this case i think it would be smart to have separate data sets for developing locally and for e2e tests. that way we can edit the local data freely without worrying about destroying e2e test results. created a ticket for this: https://jira.mongodb.org/browse/EVG-8009

src/components/Pagination.tsx
SupaJoon: sure, this could be a hook

cypress/integration/patch/task-table.ts
```diff
@@ -0,0 +1,205 @@
...
+      expect(err.message).to.include(
+        "'pointer-events: none' prevents user mouse interaction."
+      );
+    });
+  });
+
+  // flakey test - must fix
+  ["NAME", "STATUS", "BASE_STATUS", "VARIANT"].forEach((sortBy) => {
+    xit(`Fetches tasks sorted by ${sortBy} when ${sortBy} header is clicked`, () => {
+      clickSorterAndAssertTasksAreFetched(sortBy);
```
khelif96, line 71: If i'm reading this test correctly we are only asserting that the request is requesting that the data is sorted based on the the correct fields. It isn't asserting on whether the data is displayed in the correct order, just that we requested the data in a certain way. 
 I know this test is temporarily disabled but if you do reenable it i think the test should actually test that the ui is behaving correctly because in the end thats what the user sees.

cypress/integration/patch/task-table.ts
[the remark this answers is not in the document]
khelif96: the data we use locally should closely resemble the data we use on the e2e tests because in the end the tests we write and run locally are ran on the e2e server it could get confusing if we use two seperate datasets locally and on the e2e server. Because one set of data might break and the other one wont.

PR #172, opened 2020-05-14
EVG-7311: only import required antd css

SupaJoon:
https://jira.mongodb.org/browse/EVG-7311
Screen Shot 2020-05-14 at 3 04 33 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
   "scripts": {
     "start": "react-scripts start",
     "build": "react-scripts build",
-    "test": "react-scripts test --watchAll=false",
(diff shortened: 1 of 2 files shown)


SupaJoon:
`use-babel-plugin-import` requires us to install `react-app-rewired` to override the babel config. This is not ideal because because we will lose the "guarantees" CRA provides. https://github.com/timarney/react-app-rewired#rewire-your-app-

tgrander:
i'm not sure this is the best way to do this either

SupaJoon:
@tgrander These code changes reduce the bundle size by ~4% which is significant. I don't see an issue with maintaining a list of styles for the limited number of antd components we use.

tgrander:
there are test failures that are unusual

PR #174, opened 2020-05-18
EVG-7908: redirect to evergreen auth from spruce if logged out

tgrander:
If a user is logged out and navigates to spruce in the production or staging environment, they will be redirected to the Evergreen login page. 

When they log in, they will see the Evergreen waterfall page. 

The user will not be redirected back to Spruce when they login. There is no straight forward way to do this and IMO seeing the Evergreen page is sufficient. User sessions expire once a year so this is not expected to happen often. Spruce and Evergreen sharing the same cookie is also a temporary solution.


src/pages/Login.tsx
```diff
@@ -28,10 +29,13 @@ export const Login: React.FC<RouteComponentProps> = ({ location }) => {
     cb(e.target.value);
   };
 
+  if (isProduction()) {
```
SupaJoon, line 32: I think this should go after the isAuthenticated condition

tgrander: good point. done.

src/pages/Login.tsx
```diff
@@ -31,7 +32,10 @@ export const Login: React.FC<RouteComponentProps> = ({ location }) => {
   if (isAuthenticated) {
     return <Redirect to={getReferrer(location)} />;
   }
-
+  if (isProduction()) {
+    window.location.href = `${getUiUrl()}/login`;
```
khelif96, line 36: Any Reason to not just return a `<Redirect/>` over here?

tgrander: because spruce login is still needed in test environment

khelif96: setting `window.location.href` redirects the page why not just use the built in `react-router` `<Redirect /> ` component? Since they do the same thing.

khelif96: when we are in the test environment the redirect wont happen because of the isProduction conditional and we will still use the spruce login.

tgrander: > Since they do the same thing.

☝️ I guess that's why. I didn't consider it because I guess I view view react-router as in-app routing. Is there any benefit to using the react router Redirect?

khelif96: setting window.location.href causes the browser to refresh on the redirect while using history.push() or `<Redirect/>` dont . 
https://stackoverflow.com/questions/51713513/window-location-href-vs-redirect-utility-of-react-router-dom

khelif96: well i guess if this redirects out to an external page it will refresh anyways.

PR #176, opened 2020-05-19
EVG-7353: wire up modify version mutations to patch action buttons

tgrander:
#### Scheduling a verision
schedule patch

#### Unscheduling a verision
unschedule patch

#### Setting a version's priority
patch priority


cypress/integration/patch/action-buttons.ts
```diff
@@ -0,0 +1,91 @@
...
+    cy.get(".ant-btn.ant-btn-primary.ant-btn-sm")
+      .contains("Yes")
+      .click({ force: true });
+    cy.dataCy("banner").should("exist");
+  });
+
+  it("Error scheduling a version shows error banner", () => {
+    cy.dataCy("schedule-patch").click();
+    mockErrorResponse({ errorMessage: "There was an error scheduling tasks" });
+    cy.get(".ant-btn.ant-btn-primary.ant-btn-sm")
```
khelif96, line 28: nit. what about selecting these with the data-cy tag instead.
data-cy or some variant using wrapProps should work. 
 https://github.com/ant-design/ant-design/issues/8561#issuecomment-397424511

hesitant on using class names in case we add another element  with a similar class in the future. I like the granularity data-cy allows us to have over selecting elements.

cypress/integration/patch/action-buttons.ts
```diff
@@ -0,0 +1,91 @@
...
+      .click({ force: true });
+    cy.dataCy("banner")
+      .contains(priority)
+      .should("exist");
+  });
+
+  it("Error scheduling a version shows error banner", () => {
+    cy.dataCy("ellipsis-btn").click();
+    cy.dataCy("prioritize-patch").click();
+    cy.get(".ant-input-number-input")
```
khelif96, line 78: same data-cy comment.

src/gql/mutations/index.ts
```diff
@@ -0,0 +1,11 @@
+export { ABORT_TASK } from "./abort-task";
```
khelif96, line 1: Thanks 🙏

cypress/utils/mockErrorResponse.ts
```diff
@@ -0,0 +1,34 @@
...
+    | "INTERNAL_SERVER_ERROR"
+    | "RESOURCE_NOT_FOUND"
+    | "FORBIDDEN"
+    | "INPUT_VALIDATION_ERROR";
+}
+
+// mockErrorResponse returns a graphql query with an error message.
+// Use this to test error responses from queries and mutations.
+// Call mockErrorResponse before performing action that dispatches XHR request.
+export const mockErrorResponse = ({
```
khelif96, line 14: Im pretty apposed to mocking data including error responses for e2e tests. Since its not an accurate representation of how a real life system would behave (The error message could change or some other variable could affect the result and the test would not catch it).
 It might be better to mock it in a unit test or integration test if the goal of this test is to just see if an error banner appears.

src/pages/patch/ActionButtons.tsx
```diff
@@ -1,70 +1,129 @@
...
+    refetchQueries,
+  });
+
+  const [setPatchPriority, { loading: loadingSetPatchPriority }] = useMutation<
+    SetPatchPriorityMutation,
+    SetPatchPriorityMutationVariables
+  >(SET_PATCH_PRIORITY, {
+    onCompleted: () => {
+      successBanner(`Priority was set to ${priority}`);
+      setPriority(initialPriority);
```
SupaJoon, line 82: I think we should keep the priority set to the new value. This way the user knows the current value is before changing and this is how the modal currently works in the EVG ui.

cypress/utils/mockErrorResponse.ts
tgrander: what is the difference between mocking in a unit test or mocking in a cypress test? mocking in a unit or integration tests just means more tests for the same component spread out across different files. and it's not just about showing a banner and the error message is irrelevant. it's about testing that when the gql response errors out that the app handles it gracefully and displays what it's supposed to.

khelif96: It might be just the connotation of this being an e2e test and not an integration test. Its fine if we do this in a cypress test since that is just the tool. But i think its important to have that distinction between the two. Having more tests for the same component isn't necessarily bad since we are testing different functionalities for the same component. Always mocking the error response from a query or mutation wont allow us to catch a scenario where the behavior  of that query or mutation changes. If something changes on the backend and that query no longer returns an error we wont catch it in a test.

tgrander: 1. not all errors are mocked, i.e. requesting a patch ID that does not exist.
2. all queries will always have the potential of returning an error. what that errors says or its code does not usually matter. the important piece of the test is testing that application handles that error. 
3. if the same capabilities as an integration test can be achieved via Cypress e2e tests then we can and should use those capabilities. Using an integration test just to use an integration test is not a valid pro.

src/pages/patch/ActionButtons.tsx
tgrander: totes agree. done.

cypress/integration/patch/action-buttons.ts
tgrander: the thing about `wrapProps` is that how does it know to pass which data-cy attributes to which children? for example the yes/no buttons on the popconfirm. 

i also tried using the `okButtonProps` prop on the popconfirm but it only accepts specified button props. 

i am fine with the current approach for now.

khelif96: Thats fair. I wish antd supported this more/provided more options for it.

PR #178, opened 2020-05-26
EVG-7553: user patches patch card dropdown menu

tgrander:
May-26-2020 12-28-00

Sorry about all those weird merge commits on my PR. Trying to eradicate them once and for all. 

**From the dropdown menu, a user can:** 
- Reconfigure patch
- Schedule patch
- Unschedule patch
- Restart patch

I purposefully left out `set priority` from the lis of menu items. 

### What's all the code for?

I created components for each of the buttons. Each component handles its own gql mutation that dispatches the action of the button. These shared components can now be used on the patches page as well as in the patch card dropdown. Creating this components also cleaned up the implementation of the patch action buttons significantly.


src/components/PatchActionButtons/UnschedulePatchTasks.tsx
```diff
@@ -0,0 +1,92 @@
...
+  });
+
+  return (
+    <Popconfirm
+      icon={null}
+      placement="left"
+      title={
+        <>
+          <StyledBody>Unschedule all tasks?</StyledBody>
+          <Checkbox
```
SupaJoon, line 62: the label and checkbox look unaligned
Screen Shot 2020-05-26 at 2 46 09 PM
you may be able to pass jsx into the label prop to fix this

src/pages/patch/PatchRestartModal.tsx
```diff
@@ -28,14 +28,19 @@ interface PatchModalProps {
...
 export const PatchRestartModal: React.FC<PatchModalProps> = ({
   visible,
   onOk,
   onCancel,
+  patchId: patchIdFromProps,
+  refetchQueries,
 }) => {
   const dispatchBanner = useBannerDispatchContext();
   const { id } = useParams<{ id: string }>();
+  const patchId = patchIdFromProps || id;
```
SupaJoon, line 43: this is a great place to use the nullish coalescing operator

tgrander:
> thanks for cleaning and reusing a lot of this code. An issue I noticed that the patch dropdown doesn't hide when you click the background you can open multiple dropdowns at once.

I left it like this on purpose. closing the menu on click requires [forwarding a ref](https://reactjs.org/docs/forwarding-refs.html) for the popconfirm to each of the components. Since the menus close once action has been completed i felt like that was enough. i weighed the code required to it against the chances of it being an issue and decided on non-issue. 

do you think it's necessary still? prob a question for design anyways but i'm not sure they will understand with the underlying reason.

src/components/PatchActionButtons/UnschedulePatchTasks.tsx
tgrander: Screen Shot 2020-05-26 at 5 44 45 PM

Weird, there's no disalignment on mine.

src/pages/patch/PatchRestartModal.tsx
tgrander: done!

src/components/ButtonDropdown.tsx
```diff
@@ -48,13 +48,14 @@ interface CardItemProps {
...
 
 export const DropdownItem = styled.div`
-  > p:hover {
+  > small:hover {
     text-decoration: underline;
     cursor: pointer;
   }
-  pointer-events:${(props: CardItemProps) => props.disabled && "none"}; 
-  > p {
+  pointer-events: ${(props: CardItemProps) => props.disabled && "none"};
```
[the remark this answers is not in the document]
tgrander: i do find this approach less readable but done!

SupaJoon:
> > thanks for cleaning and reusing a lot of this code. An issue I noticed that the patch dropdown doesn't hide when you click the background you can open multiple dropdowns at once.
> 
> I left it like this on purpose. closing the menu on click requires [forwarding a ref](https://reactjs.org/docs/forwarding-refs.html) for the popconfirm to each of the components. Since the menus close once action has been completed i felt like that was enough. i weighed the code required to do it against the chances of it being an issue and decided on non-issue.
> 
> do you think it's necessary still? prob a question for design anyways but i'm not sure they will understand with the underlying reason.

I think it's good idea to implement this since we already have a solution (https://github.com/evergreen-ci/spruce/blob/new-ui-feature-branch/src/pages/task/ActionButtons.tsx#L143-L149) and is something our users will probably notice/report quickly. We can defer the code changes to another ticket.

khelif96:
> > > thanks for cleaning and reusing a lot of this code. An issue I noticed that the patch dropdown doesn't hide when you click the background you can open multiple dropdowns at once.
> > 
> > 
> > I left it like this on purpose. closing the menu on click requires [forwarding a ref](https://reactjs.org/docs/forwarding-refs.html) for the popconfirm to each of the components. Since the menus close once action has been completed i felt like that was enough. i weighed the code required to do it against the chances of it being an issue and decided on non-issue.
> > do you think it's necessary still? prob a question for design anyways but i'm not sure they will understand with the underlying reason.
> 
> I think it's good idea to implement this since we already have a similar solution (https://github.com/evergreen-ci/spruce/blob/new-ui-feature-branch/src/pages/task/ActionButtons.tsx#L143-L149) and is something our users will probably notice/report quickly. We can defer the code changes to another ticket.

I agree with @tgrander  at least for the short term while it would be nice if they closed when you focused away i dont think it would be the best solution to use refs and pass them to other components to achieve that functionality.

tgrander:
> > > > thanks for cleaning and reusing a lot of this code. An issue I noticed that the patch dropdown doesn't hide when you click the background you can open multiple dropdowns at once.
> > > 
> > > 
> > > I left it like this on purpose. closing the menu on click requires [forwarding a ref](https://reactjs.org/docs/forwarding-refs.html) for the popconfirm to each of the components. Since the menus close once action has been completed i felt like that was enough. i weighed the code required to do it against the chances of it being an issue and decided on non-issue.
> > > do you think it's necessary still? prob a question for design anyways but i'm not sure they will understand with the underlying reason.
> > 
> > 
> > I think it's good idea to implement this since we already have a similar solution (https://github.com/evergreen-ci/spruce/blob/new-ui-feature-branch/src/pages/task/ActionButtons.tsx#L143-L149) and is something our users will probably notice/report quickly. We can defer the code changes to another ticket.
> 
> I agree with @tgrander at least for the short term while it would be nice if they closed when you focused away i dont think it would be the best solution to use refs and pass them to other components to achieve that functionality.

it took an hour, which was annoying, but the functionality and UX is significantly better when dropdowns always close on outside clicks (unless that click is on the popconfirm content). @khelif96 i'm not sure an alternative solution to using refs exists. React docs advise against using them unless they are needed and in this case they are.

khelif96:
> > > > > thanks for cleaning and reusing a lot of this code. An issue I noticed that the patch dropdown doesn't hide when you click the background you can open multiple dropdowns at once.
> > > > 
> > > > 
> > > > I left it like this on purpose. closing the menu on click requires [forwarding a ref](https://reactjs.org/docs/forwarding-refs.html) for the popconfirm to each of the components. Since the menus close once action has been completed i felt like that was enough. i weighed the code required to do it against the chances of it being an issue and decided on non-issue.
> > > > do you think it's necessary still? prob a question for design anyways but i'm not sure they will understand with the underlying reason.
> > > 
> > > 
> > > I think it's good idea to implement this since we already have a similar solution (https://github.com/evergreen-ci/spruce/blob/new-ui-feature-branch/src/pages/task/ActionButtons.tsx#L143-L149) and is something our users will probably notice/report quickly. We can defer the code changes to another ticket.
> > 
> > 
> > I agree with @tgrander at least for the short term while it would be nice if they closed when you focused away i dont think it would be the best solution to use refs and pass them to other components to achieve that functionality.
> 
> it took an hour, which was annoying, but the functionality and UX is significantly better when dropdowns always close on outside clicks (unless that click is on the popconfirm content). @khelif96 i'm not sure an alternative solution to using refs exists. React docs advise against using them unless they are needed and in this case they are.

Makes sense i was thinking it could be possible to create an invisible overlay around the page that captures any outside click events turning the dropdown into a modal of some sort.

tgrander:
> Makes sense i was thinking it could be possible to create an invisible overlay around the page that captures any outside click events turning the dropdown into a modal of some sort.

neat idea. i think the refs would nicely though.

tgrander:
evergreen retry

PR #181, opened 2020-05-27
EVG-7975: Disable Select Build Variants and Tasks when viewing Changes 

SupaJoon:
https://jira.mongodb.org/browse/EVG-7975
Screen Shot 2020-05-27 at 11 24 28 AM

Screen Shot 2020-05-27 at 4 46 40 PM


john-m-liu:
Would it be better to disable the section rather than hide it? It feels a bit jarring for the section to disappear

SupaJoon:
> Would it be better to disable the section rather than hide it? It feels a bit jarring for the section to disappear

I like it! done!

src/pages/patch/patchTabs/Tasks.tsx
```diff
@@ -182,3 +205,47 @@ const getQueryVariables = (
         : DEFAULT_PAGE_SIZE,
   };
 };
+
+const renderStatusBadge = (status): null | JSX.Element => {
```
john-m-liu, line 209: You moved this here in https://github.com/evergreen-ci/spruce/pull/182/files as well which I think is the one where you intended to make this change. You might want to revert the change in one of the PRs so you don't hit a merge conflict if the diffs don't end up matching

PR #182, opened 2020-05-27
EVG-7838: set initial state for table sorters 

SupaJoon:
https://jira.mongodb.org/browse/EVG-7838
useSetColumnDefaultSortOrder handles setting the defaultSortOrder for the antd table column prop. I refactored the TasksTable so getQueryVariables is available in the same file as the columns in order to get the sort direction and sort category for the defaultSortOrder.

### DIFF
--- a/src/hooks/tests/useSetColumnDefaultSortOrder.test.ts
+++ b/src/hooks/tests/useSetColumnDefaultSortOrder.test.ts
@@ -1,0 +1,81 @@
+import { renderHook } from "@testing-library/react-hooks";
+import { useSetColumnDefaultSortOrder } from "hooks";
+import { ColumnProps } from "antd/lib/table";
+import { TestResult, SortDirection } from "gql/generated/types";
+
+test("Should add defaultSortOrder key with supplied direction to column item with matching category", () => {
+  const columns: ColumnProps<TestResult>[] = [
+    {
+      title: "Name",
+      dataIndex: "testFile",
+      key: "testName",
+      sorter: true,
+    },
+    {
+      title: "Status",
+      dataIndex: "status",
+      key: "status",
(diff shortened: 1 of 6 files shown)


src/hooks/tests/useSetColumnDefaultSortOrder.test.ts
```diff
@@ -0,0 +1,81 @@
+import { renderHook } from "@testing-library/react-hooks";
```
khelif96, line 1: Thanks for writing Tests!

src/hooks/useSetColumnDefaultSortOrder.ts
```diff
@@ -0,0 +1,15 @@
...
+  columns: ColumnProps<T>[],
+  category: string,
+  direction: string
+) => {
+  const targetColumnHeader = columns.find(({ key }) => key === category);
+  if (targetColumnHeader) {
+    // eslint-disable-next-line no-param-reassign
+    targetColumnHeader.defaultSortOrder =
+      direction === SortDirection.Asc ? "ascend" : "descend";
+  }
```
khelif96, line 14: Im reluctant on modifying the param passed in  directly. The only reason this works here is because objects are passed in by reference. When columns is [initially declared](https://github.com/evergreen-ci/spruce/pull/182/files#diff-345b351a35b76219866aa63f353fa764R213) its declared as a `const` which implies it should be immutable. By doing this you are breaking that assumption.  I would instead have this hook return a duplicate object and reassign `columns` with the returned value. The `no-param-reassign` rule is there to help avoid that since mutating objects in this manner could lead to some unintended side effects.

SupaJoon: very good point

PR #183, opened 2020-05-27
EVG-8066 Add auth config to authentication card

khelif96:
image
Added the authentication info to the auth card in the cli tab

this is dependent on.
https://github.com/evergreen-ci/evergreen/pull/3625

THE API KEY IN THE PICTURE IS NOT MY REAL API KEY ITS FOR MY LOCAL EVERGREEN SERVER

### DIFF
--- a/src/gql/queries/index.ts
+++ b/src/gql/queries/index.ts
@@ -1,0 +1,18 @@
+export { GET_USER_CONFIG } from "./get-user-config";
+export { GET_USER_SETTINGS } from "./get-user-settings";
+export { GET_PATCH, GET_PATCH_CONFIGURE } from "./patch";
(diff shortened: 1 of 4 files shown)


src/gql/queries/index.ts
```diff
@@ -0,0 +1,18 @@
+export { GET_USER_CONFIG } from "./get-user-config";
```
john-m-liu, line 1: What's the advantage of having this over importing the queries from where they're defined?

khelif96: Its a popular pattern. It just cleans up imports instead of having to import multiple files from similar paths, you instead have one origin for them. We already do the same thing in the hooks folder. https://github.com/evergreen-ci/spruce/blob/new-ui-feature-branch/src/hooks/index.ts and in some other component folders.

PR #186, opened 2020-06-01
EVG-7853: Autoupdate table statuses

SupaJoon:
https://jira.mongodb.org/browse/EVG-7853
introduces usePollTableQuery to poll the current table page on the My Patches page, Tasks Table and Tests Table.
I switched the build variant tooltip from Leafygreen to Antd. Hovering over the Leafygreen tooltip when multiple patch card rows are rendered would cause the page to freeze and go unresponsive even after making optimizations to minimize re-renders.


src/hooks/usePollTableQuery.ts
```diff
@@ -0,0 +1,113 @@
...
+  refetch,
+  setIntervalId,
+  getQueryVariables,
+}) => {
+  const queryVariables = getQueryVariables(search, resourceId);
+  const intervalId = window.setInterval(() => {
+    try {
+      refetch(queryVariables);
+    } catch {
+      clearInterval(intervalId);
```
SupaJoon, line 109: this catch will be entered when the component that exposes the useQuery refetch function unmounts.

src/pages/configurePatch/configurePatchCore/ConfigureTasks.tsx
```diff
@@ -143,10 +143,10 @@ const Actions = styled.div`
   margin-bottom: 8px;
   display: flex;
   align-items: center;
-  & > :first-child {
+  & > :first-of-type {
```
SupaJoon, line 146: removes console warning

src/pages/configurePatch/configurePatchCore/ConfigureTasks.tsx
```diff
@@ -143,10 +143,10 @@ const Actions = styled.div`
...
     margin-right: 40px;
   }
-  & > :not(:first-child) {
+  & > :not(:first-of-type) {
```
SupaJoon, line 149: removes console warning

src/pages/task/Metadata.tsx
```diff
@@ -73,7 +74,7 @@ export const Metadata: React.FC<{
           <H3>Depends On</H3>
           <Divider />
           {reliesOn.map((props) => (
-            <DependsOn key={props.buildVariant} {...props} />
+            <DependsOn key={uuid()} {...props} />
```
SupaJoon, line 77: fix dupe key warning

SupaJoon:
> I think this can put someone in the weird situation where tasks get added and move everything down, but I don't see a way around that with this pattern

That is true. This ux could be beneficial in a case where a user wants to see the latest filter/sort results without having to refresh the page.

SupaJoon:
evergreen retry

src/hooks/usePollTableQuery.ts
```diff
@@ -0,0 +1,119 @@
...
+  getQueryVariables,
+  refetch,
+  search,
+}: {
+  networkStatus: NetworkStatus;
+  getQueryVariables: (search: string, id?: string) => ApolloQueryVariables;
+  refetch: (
+    variables?: ApolloQueryVariables
+  ) => Promise<ApolloQueryResult<ApolloQueryResultType>>;
+  search: string;
```
tgrander, line 22: when there is more than one or two and the types are long and take multiple generics, i think having a separate interface to represent the type creates cleaner, more readable code.

src/hooks/usePollTableQuery.ts
```diff
@@ -0,0 +1,119 @@
...
+    prevQueryVariables,
+    networkStatus,
+    refetch,
+    intervalId,
+    resourceId,
+    getQueryVariables,
+    search,
+    pathname,
+    initialPathname,
+  ]);
```
tgrander, line 98: a short comment above each of these `useEffect` would provide me with the quick ability to skim and understand their purpose.  what a `useEffect` is accomplishing is not always obvious to a reader. going forward, our code as a whole could benefit from more comments, especially around `useEffect` that are called directly in a component. **specifically a comment addressing _the why_**.

src/hooks/usePollTableQuery.ts
```diff
@@ -0,0 +1,119 @@
...
+  refetch,
+  setIntervalId,
+  getQueryVariables,
+}) => {
+  const queryVariables = getQueryVariables(search, resourceId);
+  const intervalId = window.setInterval(() => {
+    refetch(queryVariables);
+  }, pollInterval);
+  setIntervalId(intervalId);
+};
```
tgrander, line 119: why is `setInterval` being used? i am unable to understand its objective from reading the code in `usePollTableQuery`

src/pages/MyPatches.tsx
```diff
@@ -38,45 +38,26 @@ const MyPatchesComponent: React.FC = () => {
...
-            return fetchMoreResult;
-          },
-        });
-      } catch (e) {
-        // empty block
-      }
-    };
-    fetch();
-  }, [search, fetchMore]);
+  const { showSkeleton } = usePollTableQuery({
```
tgrander, line 55: should the name of `usePollTableQuery` be changed since it's used to poll things that are not a table, such as the list of user patches?

src/pages/MyPatches.tsx
```diff
@@ -102,7 +83,7 @@ const MyPatchesComponent: React.FC = () => {
         </PageWrapper>
       );
     }
-    if (loading) {
+    if (showSkeleton) {
```
tgrander, line 86: why call it `showSkeleton` as opposed to `loading`?

src/pages/my-patches/PatchCard.tsx
```diff
@@ -66,24 +44,17 @@ export const PatchCard: React.FC<Props> = ({
...
-            );
-          })}
+          {builds.map((b) => (
+            <div key={b.id}>
+              <BuildStatusIcon
+                status={b.status}
+                buildVariant={b.buildVariant}
+              />
+            </div>
+          ))}
```
tgrander, line 57: nit: prefer destructuring

src/pages/my-patches/patch-card/BuildStatusIcon.tsx
```diff
@@ -1,37 +1,37 @@
...
-}) => {
+export const BuildStatusIcon: React.FC<Props> = ({ status, buildVariant }) => {
+  const router = useHistory();
+  const onClick = useMemo(
+    () => () =>
+      router.push(
+        `${paths.patch}/${id}/${DEFAULT_PATCH_TAB}?${PatchTasksQueryParams.Variant}=${buildVariant}`
+      ),
+    [buildVariant, router]
+  );
```
tgrander, line 27: I have a hunch why `useMemo` is used here and it's because of the tooltip performance problems you reference in the PR summary. however i am not sure this onClick function calls for `useMemo`. When used improperly, `useCallback` and `useMemo` can actually be less performant than declaring function in component body. Essentially, there are only ever two scenarios in which `useMemo` and `useCallback` are a good idea: referential equality and expensive computations. This article from Kent C Dodds explains it all beautifully: https://kentcdodds.com/blog/usememo-and-usecallback

src/hooks/usePollTableQuery.ts
SupaJoon: setInterval calls passed in function after a set amount of time. in this case the callback is a refetch function. so this is how the query is being polled hence the function name `pollQuery`

src/pages/MyPatches.tsx
SupaJoon: showSkeleton doesn't represent one to one the query loading state. it represents that query variables have changes & the query is loading which means the user interacted with a control on the page to adjust what data they see. 
showSkeleton is false when the query is loading because of polling bc that happens automatically from the setInterval and we don't want to display a loading skeleton every 5 seconds.

src/pages/MyPatches.tsx
SupaJoon: > Did you file a ticket with Leafygreen to address the performance issue with their tooltip?

the Leafygreen lead engineer is aware of the issue

src/pages/my-patches/patch-card/BuildStatusIcon.tsx
SupaJoon: that's a great article. useMemo is not needed here

src/hooks/usePollTableQuery.ts
tgrander: are we unable to use the pollInterval via the query? i'm confused because we poll other things without using setInterval. Can you explain to me please why this implementation requires it?

SupaJoon: the pollInterval option doesn't allow you to specify new query variables from the original query

tgrander: i see. can we add a comment? this is a perfect example of a commenting answering _the why_. The dev reading this a year from now (might even be you) will thank past Arjun for the comment.

cypress/support/hooks.js
```diff
@@ -8,9 +8,7 @@ enableFetchWorkaround();
 // private helpers
 function enableFetchWorkaround() {
   let polyfill;
-  // TODO: This breaks the onError callback apollo provides and causes onError to
-  // Be flaky all tests should be refactored to remove their dependency on this and this
-  // should be turned off.
+
```
khelif96, line 11: Can we leave this comment here

src/hooks/usePollQuery.ts
```diff
@@ -0,0 +1,140 @@
...
+    intervalId,
+    resourceId,
+    getQueryVariables,
+    search,
+    pathname,
+    initialPathname,
+  ]);
+
+  return {
+    showSkeleton: queryVarDiffOccured,
```
khelif96, line 113: If this is the only thing this hook returns i think you can just return `queryVarDiffOccured` directly

cypress/support/hooks.js
SupaJoon: oops!! didn't mean to remove those

PR #188, opened 2020-06-01
EVG-8075 Use queried download links instead of static ones

khelif96:
The download links for the cli binaries are now fetched from the evergreen backend. Since not every one has the same evergreen binaries on their system some download links will be disabled locally. Prod should have all the evergreen binaries and all the download links should work. 

Dependent on this.
https://github.com/evergreen-ci/evergreen/pull/3637

### DIFF
--- a/src/pages/preferences/preferencesTabs/cliTab/DownloadCard.tsx
+++ b/src/pages/preferences/preferencesTabs/cliTab/DownloadCard.tsx
@@ -1,68 +1,121 @@
 import React from "react";
+import { useQuery } from "@apollo/react-hooks";
 import { Subtitle, Body } from "@leafygreen-ui/typography";
 import styled from "@emotion/styled";
 import { uiColors } from "@leafygreen-ui/palette";
 import Button from "@leafygreen-ui/button";
+import { Skeleton } from "antd";
+import get from "lodash/get";
 import { SiderCard, StyledLink } from "components/styles";
 import { Accordian } from "components/Accordian";
+import { cliDocumentationUrl } from "constants/externalResources";
+import { GET_CLIENT_CONFIG } from "gql/queries";
 import {
-  cliDocumentationUrl,
-  cliDownloadUrls,
-} from "constants/externalResources";
+  ClientConfigQuery,
+  ClientConfigQueryVariables,
+  ClientBinary,
+} from "gql/generated/types";
 
 const { gray } = uiColors;
 
-export const DownloadCard = () => (
-  <Container>
-    <Subtitle>Command-Line Client</Subtitle>
-    <CardDescription>
-      View the <StyledLink href={cliDocumentationUrl}>documentation</StyledLink>{" "}
-      or run{" "}
-      <InlinePre>evergreen --help or evergreen [command] --help</InlinePre> for
-      additional assistance.
-    </CardDescription>
-    <CardGroup>
-      <CliDownloadBox
-        title="Linux (64-bit)"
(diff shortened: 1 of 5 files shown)


khelif96:
evergreen retry

khelif96:
evergreen retry.

khelif96:
evergreen retry

src/pages/preferences/preferencesTabs/cliTab/DownloadCard.tsx
```diff
@@ -1,68 +1,121 @@
...
+}
+const ExpandableLinkContents: React.FC<ExpandableLinkContentsProps> = ({
+  clientBinaries,
+}) => (
   <LinkContainer>
-    <StyledLink href={cliDownloadUrls.linux_arm64}>Linux ARM 64-bit</StyledLink>
-    <StyledLink href={cliDownloadUrls.linux_s390x}>Linux zSeries</StyledLink>
-    <StyledLink href={cliDownloadUrls.linux_386}>Linux 32-bit</StyledLink>
-    <StyledLink href={cliDownloadUrls.linux_ppce64le}>
+    <StyledLink href={getBinaryUrl("linux", "arm64", clientBinaries)}>
```
john-m-liu, line 88: Why not map over the returned array for these components?

src/pages/preferences/preferencesTabs/cliTab/DownloadCard.tsx
```diff
@@ -1,68 +1,121 @@
...
   </LinkContainer>
 );
+
+const getBinaryUrl = (
+  os: string,
+  arch: string,
+  clientBinaries: ClientBinary[]
+) => {
+  let url = null;
+  clientBinaries.forEach((binary) => {
```
SupaJoon, line 112: we can use Array find() here to avoid using a mutable variable

src/pages/preferences/preferencesTabs/cliTab/DownloadCard.tsx
```diff
@@ -1,68 +1,121 @@
...
 interface CliDownloadBoxProps {
   title: string;
-  link: string;
+  link: string | null;
 }
 const CliDownloadBox: React.FC<CliDownloadBoxProps> = ({ title, link }) => (
   <CliDownloadCard>
     <CliDownloadTitle>{title}</CliDownloadTitle>
-    <CliDownloadButton href={link} as="a">
+    <CliDownloadButton href={link} disabled={link === null} as="a">
```
SupaJoon, line 75: nit: can use !link here?

src/pages/preferences/preferencesTabs/cliTab/DownloadCard.tsx
khelif96: Hey @john-m-liu  i did it this way intentionally. There are 3 elements that have their own prominent cards and the rest are displayed as links. So i would have to filter those elements out in order to iterate through the list. All the elements also have distinct display names that aren't present in the data. It seemed simpler to fetch this data manually instead of dealing with transforming the data so i can use it.

john-m-liu: You'll need to have logic to extract the display name in both approaches, but mapping over the returned array accounts for results that aren't listed here, either because we just added them or forgot to handle here. I think that approach will be better and more future-proof

khelif96: That makes sense just pushed up those changes.

PR #187, opened 2020-06-01
EVG-8060: merge new-ui-feature-branch into master

tgrander:
Here it is boys & girls, the merge of new-ui-feature-branch into master.

PR #190, opened 2020-06-01
EVG-8020: my patches route adjustment

SupaJoon:
https://jira.mongodb.org/browse/EVG-8020
`/user/patches` => `/user/{userId}/patches`
`/user/{userId}/patches` renders the "My Patches" page
next PR for this ticket will adjust the patch page route

I'll make this PR against master once https://github.com/evergreen-ci/spruce/pull/187 is merged


src/components/Content/index.tsx
```diff
@@ -11,32 +11,35 @@ import { routes } from "constants/routes";
...
+
   return (
     <PageLayout>
       <Navbar />
       <Switch>
         <PrivateRoute path={routes.task} component={Task} />
         <PrivateRoute path={routes.configurePatch} component={ConfigurePatch} />
         <PrivateRoute path={routes.patch} component={Patch} />
-        <PrivateRoute path={routes.myPatches} component={MyPatches} />
+        <PrivateRoute exact path={routes.myPatches} component={MyPatches} />
```
SupaJoon, line 41: I opted to create a component that redirects than do pass in userId from the GET_USER query because waiting for a valid response object before rendering these routes prevents PrivateRoute from redirecting to login when the user is not logged in

SupaJoon:
evergreen retry

src/components/Breadcrumb/index.tsx
```diff
@@ -21,7 +21,7 @@ export const BreadCrumb: React.FC<Props> = ({
     <StyledBreadcrumb>
       <Breadcrumb.Item>
         <StyledP1>
-          <StyledRouterLink id="bc-my-patches" to={paths.myPatches}>
+          <StyledRouterLink id="bc-my-patches" to={routes.myPatches}>
```
tgrander, line 24: can we change this route to `userPatches` ?

src/components/Content/index.tsx
```diff
@@ -11,32 +11,35 @@ import { routes } from "constants/routes";
...
+import { MyPatches } from "pages/MyPatches";
 
 export const Content: React.FC = () => {
   const { isAuthenticated, initialLoad } = useAuthStateContext();
 
   // this top-level query is required for authentication to work
   // afterware is used at apollo link level to authenticate or deauthenticate user based on response to query
   // therefore this could be any query as long as it is top-level
-  useQuery<ProjectsQuery>(GET_PROJECTS);
+  useQuery<GetUserQuery>(GET_USER);
```
tgrander, line 28: this query must complete for the app to render at all. if success, show page. if error, show error, and if not authenticated, show login. therefore waiting for it to complete does not change how long it takes for the page to load.

src/constants/routes.ts
```diff
@@ -1,22 +1,30 @@
+enum PageNames {
+  Patches = "patches",
+}
```
tgrander, line 3: enums are typically used as types whereas this is just a constant

src/pages/MyPatches.tsx
```diff
@@ -1,216 +1,14 @@
...
-    </PageWrapper>
-  );
+import { GetUserQuery } from "gql/generated/types";
+import { GET_USER } from "gql/queries";
+import { useHistory } from "react-router-dom";
+import { getUserPatchesRoute } from "constants/routes";
+
+export const MyPatches: React.FC = () => {
+  const router = useHistory();
+  const { data } = useQuery<GetUserQuery>(GET_USER);
```
tgrander, line 9: this is making an additional request for the same query as the top level query in the routes component. this should either be deleted altogether or subscribe to the cache value of userId by making a local query. I think doing both is actually not a bad idea.

src/components/Content/index.tsx
SupaJoon: the loading screen flashes and then goes away even when the user is not authenticated because of the "initialLoad" variable in the condition. 
The user is redirected to the login screen because we allow PrivateRoutes to mount and redirect the user even when they are not authenticated and when the result of the user data query is undefined. 

doing something like this doesn't work because the route does not update when userId has a value. 
```
<PrivateRoute exact path={routes.myPatches}>
   <Redirect to={`${paths.user}/${get(data, "user.userId")}/patches`} />
</PrivateRoute>
```

if we block mounting the routes when result of the query isn't defined, the user will never be be redirected to the login page.

src/components/Content/index.tsx
SupaJoon: it must complete but it doesn't have to be successful in order to mount the routes. the private routes redirect the user to the login page and they are mounted whether this query fails or not.

src/components/Content/index.tsx
tgrander: okay, makes sense

src/pages/MyPatches.tsx
SupaJoon: the fetchPolicy defaults to cache-first. so if the data has been fetched and exists in the cache, it will be utilized. otherwise a network request is made.

src/components/Breadcrumb/index.tsx
SupaJoon: there already is a userPatches route

SupaJoon:
> The document also mentions that /user/patches redirects to patches for the current user. Are you implementing that later? It also reveals a flaw in the design because we can't ever have someone named "patches"

this happens in the MyPatches component

src/pages/user-patches/PatchCard.tsx
```diff
@@ -1,13 +1,13 @@
 import React from "react";
```
khelif96, line 1: We use camelCase for all other folders and files.

src/pages/user-patches/patch-card/BuildStatusIcon.tsx
```diff
@@ -1,9 +1,9 @@
 import React from "react";
```
khelif96, line 1: Same comment about using camelCase

SupaJoon:
there is a ticket to refactor the failing the tests

PR #191, opened 2020-06-02
EVG-8114: massively decrease amount of rows shown for loading table

tgrander:
Descreased from 80 to 8. 80 is way too many and was causing the page to jump when using the filters because it would increase the size of the page.


tgrander:
evergreen retry

PR #192, opened 2020-06-02
EVG-8036: fix patch & task action buttons smooshed text

tgrander:
### Before
Screen Shot 2020-06-02 at 4 20 03 PM

### After
Screen Shot 2020-06-02 at 4 20 35 PM


tgrander:
> Thanks for making this change can you also make this change to the commit queue button. https://jira.mongodb.org/browse/EVG-7954

Same solution will not work for the cq button. That problem has to do with the fixed width created by the grid layouts.

PR #193, opened 2020-06-02
EVG-8020: adjust /patch route

SupaJoon:
https://jira.mongodb.org/browse/EVG-8020
/patch/:id/:tab? redirects to /version/:id/:tab?


john-m-liu:
Could you rebase so it's clear what changes are from the previous pr?

src/components/Content/index.tsx
```diff
@@ -37,7 +38,8 @@ export const Content: React.FC = () => {
       <Switch>
         <PrivateRoute path={routes.task} component={Task} />
         <PrivateRoute path={routes.configurePatch} component={ConfigurePatch} />
-        <PrivateRoute path={routes.patch} component={Patch} />
+        <PrivateRoute path={routes.patch} component={PatchRedirect} />
+        <PrivateRoute path={routes.version} component={Patch} />
```
tgrander, line 42: Now that `Patch` is no longer a page I think that changing the component name to `Version` is the right move

PR #196, opened 2020-06-03
EVG-8021: dynamically show patch task statuses in dropdown

tgrander:
Only show patch task statuses that are relevant to the patch. As the task statuses change, so do the dropdown options. 

I made `getCurrentStatuses` a util so it can be used for the task dropdown filters too

### DIFF
--- a/src/gql/queries/get-patch-task-statuses.ts
+++ b/src/gql/queries/get-patch-task-statuses.ts
@@ -1,0 +1,11 @@
+import gql from "graphql-tag";
+
+export const GET_PATCH_TASK_STATUSES = gql`
+  query GetPatchTaskStatuses($id: String!) {
+    patch(id: $id) {
+      id
+      taskStatuses
+      baseTaskStatuses
+    }
+  }
+`;
--- a/src/utils/statuses/getCurrentStatuses.ts
+++ b/src/utils/statuses/getCurrentStatuses.ts
@@ -1,0 +1,28 @@
+import { Status } from "pages/patch/patchTabs/tasks/TaskFilters";
+
+const allKey = "all";
+
+export const getCurrentStatuses = (statuses: string[], treeData: Status[]) => {
+  const currentStatuses = [];
+  treeData.forEach((status) => {
+    const { key, children } = status;
+    if (key === allKey) {
+      currentStatuses.push(status);
+    }
+    if (children) {
+      const currentChildStatuses = getCurrentStatuses(statuses, children);
+      if (currentChildStatuses.length > 1) {
+        const newStatus = { ...status };
+        newStatus.children = currentChildStatuses;
+        currentStatuses.push(newStatus);
+      }
+      if (currentChildStatuses.length === 1) {
+        currentStatuses.push(currentChildStatuses[0]);
+      }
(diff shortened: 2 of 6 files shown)


src/utils/statuses/getCurrentStatuses.ts
```diff
@@ -0,0 +1,28 @@
+import { Status } from "pages/patch/patchTabs/tasks/TaskFilters";
```
khelif96, line 1: I dont think this is general enough to  live inside the `utils` folder. I think its only used by the tasks filter component, So it should probably live near it.

src/gql/queries/get-patch-task-statuses.ts
```diff
@@ -0,0 +1,11 @@
+import gql from "graphql-tag";
+
+export const GET_PATCH_TASK_STATUSES = gql`
```
khelif96, line 3: Instead of creating a new query can we use the already existing Patch query since they both fetch the same info. So apollo knows to fetch the data from the cache. Having a separate query for this data doubles the amount of requests going to the network.

src/utils/statuses/getCurrentStatuses.ts
tgrander: This function can be used for the task test statuses too

src/gql/queries/get-patch-task-statuses.ts
tgrander: they do not fetch any of the same data

khelif96: Sorry i thought they did but just took a look at the patch query and your right.

src/utils/statuses/getCurrentStatuses.ts
khelif96: Didnt read your comment before posting this review sorry.

khelif96:
Could we add the fields  from the `GET_PATCH_TASK_STATUSES` query to the `GET_PATCH` query just so we can utilize the cache?

tgrander:
> Could we add the fields from the `GET_PATCH_TASK_STATUSES` query to the `GET_PATCH` query just so we can utilize the cache?

@khelif96 
No. This is an expensive query that is not needed to load the page; therefore lazily load it after parent components have rendered. The cache is utilized in either scenario just at different times.

khelif96:
> > Could we add the fields from the `GET_PATCH_TASK_STATUSES` query to the `GET_PATCH` query just so we can utilize the cache?
> 
> @khelif96
> No. This is an expensive query that is not needed to load the page; therefore lazily load it after parent components have rendered. The cache is utilized in either scenario just at different times.

https://github.com/evergreen-ci/spruce/blob/bea8a92ec1d390e365a0157fbbdca5cbd6d2b465/src/pages/Patch.tsx#L36

The Patch query is already polled at the top level patch page. We could add these extra fields to that query. And remove the polling that occurs in the TaskFilters component. When ever the top level patch polls it will update the cache and update the value the TaskFilters component uses. This way we have 1 query servicing both components. Instead of each component performing its own query. We are performing these queries regardless this way we only send out one network request.

PR #201, opened 2020-06-04
EVG-8160: Redirect user to user patches page from `user/:id`

tgrander:


src/components/Content/index.tsx
```diff
@@ -47,6 +47,9 @@ export const Content: React.FC = () => {
         <PrivateRoute exact path="/">
           <Redirect to={routes.myPatches} />
         </PrivateRoute>
+        <PrivateRoute path={`${paths.user}/:id`}>
+          <Redirect to={routes.myPatches} />
```
SupaJoon, line 51: this means the user/any_id will go to user/my_id/patches. 
 what do you think about this change instead
`<PrivateRoute path={[routes.userPatches, `${paths.user}/:id`]} component={UserPatches} />` so we route user/id to user/id/patches

tgrander: going to user/my_id/patches  by default is the point because user/:id is not a route

tgrander: going to /patches by default is the same logic used to default map to tab on patch or task page

PR #202, opened 2020-06-04
EVG-7974 Add support for selecting multiple tasks in the configure tab

khelif96:
image
You can select multiple  variants by command clicking on them

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -69,6 +69,10 @@ module.exports = {
     },
   },
   rules: {
+    "@typescript-eslint/no-unused-vars": [
+      errorIfStrict,
+      { vars: "all", args: "after-used", ignoreRestSiblings: true },
+    ],
     "arrow-body-style": [
       errorIfStrict,
       "as-needed",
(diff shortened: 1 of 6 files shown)


.eslintrc.js
```diff
@@ -69,6 +69,10 @@ module.exports = {
     },
   },
   rules: {
+    "@typescript-eslint/no-unused-vars": [
+      errorIfStrict,
+      { vars: "all", args: "after-used", ignoreRestSiblings: true },
+    ],
```
tgrander, line 75: what's dis?

cypress/integration/patch/configure-patch.ts
```diff
@@ -178,6 +178,64 @@ describe("Configure Patch Page", () => {
             .should("eq", $tasks.length);
         });
     });
+    describe("Selecting multiple build variants", () => {
+      it("Selecting multiple build variants should work", () => {
+        cy.get('[data-cy-name="linux-docker"]').click();
```
tgrander, line 183: why not use `.dataCy` here?

.eslintrc.js
khelif96: This was already here i just moved it. So the plugins are in alphabetical order

cypress/integration/patch/configure-patch.ts
khelif96: the tag was a `data-cy-name` not a `data-cy` so i cant use `.dataCy` for it.

PR #203, opened 2020-06-04
EVG-7337 Add link to legacy ui to nav

khelif96:
image
Added a link that sends you to the legacy evergreen ui.

### DIFF
--- a/src/components/Navbar/index.tsx
+++ b/src/components/Navbar/index.tsx
@@ -1,5 +1,4 @@
 import React from "react";
-import { useQuery } from "@apollo/react-hooks";
 import styled from "@emotion/styled/macro";
 import { useAuthStateContext } from "context/auth";
 import { Layout } from "antd";
@@ -7,23 +6,16 @@ import { EvergreenLogo } from "components/icons";
 import { StyledLink } from "components/styles";
 import { Subtitle } from "@leafygreen-ui/typography";
 import { uiColors } from "@leafygreen-ui/palette";
-import get from "lodash/get";
-import { Link, useLocation } from "react-router-dom";
+import { Link } from "react-router-dom";
 import { routes } from "constants/routes";
 import { useLegacyUIURL } from "hooks/index";
-import { GET_USER } from "gql/queries";
-import { GetUserQuery } from "gql/generated/types";
 
 const { Header } = Layout;
 const { white, gray } = uiColors;
 
 export const Navbar: React.FC = () => {
   const { isAuthenticated } = useAuthStateContext();
-  const location = useLocation();
-  const { pathname } = location;
-  const { data } = useQuery<GetUserQuery>(GET_USER);
-  const userId = get(data, "user.userId");
-  const legacyURL = useLegacyUIURL(pathname, userId);
+  const legacyURL = useLegacyUIURL();
 
   if (!isAuthenticated) {
     return null;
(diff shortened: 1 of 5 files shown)


src/components/Navbar/index.tsx
```diff
@@ -1,17 +1,30 @@
...
+import { GET_USER } from "gql/queries";
+import { GetUserQuery } from "gql/generated/types";
 
 const { Header } = Layout;
+const { white, gray } = uiColors;
 
 export const Navbar: React.FC = () => {
   const { isAuthenticated } = useAuthStateContext();
+  const location = useLocation();
+  const { pathname } = location;
```
tgrander, line 23: `pathname` can be destructured in the line above as `const { pathname } = useLocation();`

src/components/Navbar/index.tsx
```diff
@@ -1,17 +1,30 @@
...
+  const { data } = useQuery<GetUserQuery>(GET_USER);
```
tgrander, line 24: Is this making a query to the local cache for user data? if not this would be the third network request made for user data.

src/hooks/useLegacyUIURL.ts
```diff
@@ -0,0 +1,34 @@
+import { useState, useEffect } from "react";
+import { matchPath } from "react-router-dom";
+import get from "lodash/get";
+import { getUiUrl } from "utils/getEnvironmentVariables";
+
+import { routes } from "constants/routes";
+
+export const useLegacyUIURL = (route: string, userId: string) => {
```
tgrander, line 8: route and userId do not need to be injected dependencies to the function, unless that would make it easier to test. location and userId are only used for this hook; therefore `useLocation` and `useQuery` can be called within the hook.

src/components/Navbar/index.tsx
SupaJoon: this query will check the cache before making a network request. `options.fetchPolicy` defaults to `cache-first`.

khelif96: Yeah it defaults to checking the cache first.

SupaJoon: I made this ticket which will reduce GetUser dispatches to 1. https://jira.mongodb.org/browse/EVG-8165

src/components/Navbar/index.tsx
```diff
@@ -25,28 +38,32 @@ export const Navbar: React.FC = () => {
             <StyledSubtitle>Evergreen</StyledSubtitle>
           </Logo>
         </Link>
+        {legacyURL && (
+          <StyledLink href={legacyURL}>Switch to legacy UI</StyledLink>
```
john-m-liu, line 42: The contrast ratio on this is too low. Can you change the foreground color to something in the light spectrum? We can probably use the same white as the Evergreen title
Screen Shot 2020-06-05 at 9 43 32 AM

src/hooks/useLegacyUIURL.ts
```diff
@@ -0,0 +1,34 @@
...
+  const [id, setId] = useState("");
+  const [legacyUIUrl, setLegacyUIUrl] = useState(null);
+  const uiURL = getUiUrl();
+  const legacyUIMap = {
+    [routes.version]: `${uiURL}/version/${id}`,
+    [routes.configurePatch]: `${uiURL}/patch/${id}`,
+    [routes.userPatches]: `${uiURL}/patches/${id}`,
```
john-m-liu, line 15: this should either be ${uiURL}/patches/user/${id} or ${uiURL}/patches/mine

PR #207, opened 2020-06-05
EVG-7851 Add Nav Dropdown

khelif96:
image

### DIFF
--- a/src/components/Navbar/NavDropdown.tsx
+++ b/src/components/Navbar/NavDropdown.tsx
@@ -1,0 +1,62 @@
+import React from "react";
+import { useQuery } from "@apollo/react-hooks";
+import styled from "@emotion/styled";
+import { Menu, Dropdown } from "antd";
+import Icon from "@leafygreen-ui/icon";
+import { uiColors } from "@leafygreen-ui/palette";
+import get from "lodash/get";
+import { Link } from "react-router-dom";
+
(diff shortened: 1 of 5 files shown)


src/components/Navbar/NavDropdown.tsx
```diff
@@ -0,0 +1,62 @@
...
+      </NavDropdownTitle>
+    </Dropdown>
+  );
+};
+
+const MenuItems = () => {
+  const uiURL = getUiUrl();
+  return (
+    <Menu>
+      <Menu.Item>
```
john-m-liu, line 38: Are the contents here meant to allow navigation similar to the old UI without having to flip back and forth? There are a couple things missing that can be added later, but I think it would make sense to add a link to the user patches page here in this PR

khelif96: This is meant to provide links to the legacy ui for pages that arent yet available in spruce.

SupaJoon:
can we add a border to the entire dropdown?
Screen Shot 2020-06-08 at 1 44 14 PM

src/components/Navbar/NavDropdown.tsx
john-m-liu: From the slack thread, I think we should add a link to the my patches page as well. That's where the current patches page lives and is where users will be looking for it

PR #209, opened 2020-06-08
EVG-8193: script for safer prod deployment

tgrander:
Added safety to prevent accidental prod deploys while keeping it just as simple and easy as before. 

Running `npm run deploy:prod` from a branch that is not master will exit the shell script.
Screen Shot 2020-06-08 at 9 11 58 AM

Running `npm run deploy:prod` from master will prompt the user for yes/no before running the production deploy script.

### DIFF
--- a/cli-scripts/deploy-prod.js
+++ b/cli-scripts/deploy-prod.js
@@ -1,0 +1,54 @@
+const prompt = require("prompt");
+const simpleGit = require("simple-git");
+const colors = require("colors/safe");
+const promptRun = require("prompt-run");
+
+const git = simpleGit("/Users/treygranderson/Desktop/dev/spruce");
+
+prompt.start();
+
+git
+  .init()
+  .then(() => () => {})
+  .then(() => git.branchLocal())
+  .then(({ current }) => {
+    if (current !== "master") {
+      console.log("Current branch is", current);
+      console.log(
+        colors.red(
+          "Error: You must be on master branch to deploy to production"
+        )
+      );
+      prompt.stop();
+    }
+  });
+
+prompt.get(
+  {
+    properties: {
+      confirmDeploy: {
+        description: colors.magenta(
+          "Are you sure you want to deploy to production? [Y/N]"
+        ),
+      },
+    },
+  },
+  function(err, { confirmDeploy }) {
+    const confirmDeployLowerCased = confirmDeploy.toLowerCase();
+    const isConfirmed =
+      confirmDeployLowerCased === "yes" || confirmDeployLowerCased === "y";
+    if (isConfirmed) {
+      promptRun({
+        command: "npm run deploy-prod:do-not-use-directly",
+        options: {},
+        questions: {
+          env: [],
+          args: [],
+        },
+      });
+    } else {
+      console.log(colors.cyan("Okay, no deploy"));
+      prompt.stop();
+    }
+  }
+);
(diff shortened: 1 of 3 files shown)


cli-scripts/deploy-prod.js
```diff
@@ -0,0 +1,54 @@
+const prompt = require("prompt");
+const simpleGit = require("simple-git");
+const colors = require("colors/safe");
+const promptRun = require("prompt-run");
+
+const git = simpleGit("/Users/treygranderson/Desktop/dev/spruce");
```
SupaJoon, line 6: this won't work on my machine

tgrander: hahahahahah!!!!

cli-scripts/deploy-prod.js
```diff
@@ -0,0 +1,54 @@
...
+      },
+    },
+  },
+  function(err, { confirmDeploy }) {
+    const confirmDeployLowerCased = confirmDeploy.toLowerCase();
+    const isConfirmed =
+      confirmDeployLowerCased === "yes" || confirmDeployLowerCased === "y";
+    if (isConfirmed) {
+      promptRun({
+        command: "npm run deploy-prod:do-not-use-directly",
```
khelif96, line 42: You should add this script to here instead `npm run build:prod && BUCKET=evergreen.spruce npm run deploy`  So people cant just run it directly.

tgrander: we want it to still be able to be used directly if need be, for example, doing an emergency deploy. this is not a firewall between deploying to prod deirectly, it's just a safety measure. if someone still does runs a command with `do-not-use-directly` then, well, they should just not be allowed to deploy to prod again lol

cli-scripts/deploy-prod.js
tgrander: fixed

cli-scripts/deploy-prod.js
tgrander: eh, i actually agree wit you. if we need to deploy directly we can just copy the deploy command from the cli script

tgrander: but this is easier so i might leave it + there is very low rist

package.json
```diff
@@ -75,7 +75,8 @@
     "coverage": "npm run test -- --coverage --watchAll=false",
     "storybook": "start-storybook -p 9009 -s public",
     "build-storybook": "build-storybook -s public",
-    "codegen": "graphql-codegen --config codegen.js"
+    "codegen": "graphql-codegen --config codegen.js",
+    "deploy-prod:do-not-use-directly": "npm run build:prod && BUCKET=evergreen.spruce npm run deploy"
```
john-m-liu, line 79: I didn't catch this before, but S3 buckets that we own shouldn't be part of the build system in an open source project. Can you make the staging and prod buckets params in the .cmdrc?

tgrander: in the process of chaning, but why does exposing bucket name matter if the bucket requires authorization to access?

tgrander: zakhar on build team does not consider bucket names a secret. given this in addition to the level of security surrounding our buckets, exposing bucket names is not compromising.

john-m-liu: It's not a security concern, settings specific to our deployment belong in a configuration file rather than hardcoded in the repo

tgrander: it's not super straight forward to pull off from what i can tell. if it's important to you can you create a ticket please?

john-m-liu: Yep created EVG-8202

tgrander: thank you sir!

PR #210, opened 2020-06-08
EVG-8195 Fix breadcrumb link styling

khelif96:
image


khelif96:
> nice! we def need visual testing soon

I was thinking that something like this would have been caught by that.

PR #214, opened 2020-06-09
Tg analytics 2

tgrander:
**Spruce Analytics Spreadsheet**
https://docs.google.com/spreadsheets/d/1iiySLi4f6oWfqWCaVtW2p3OGykVdfNwGAZITtvDIlRw/edit#gid=0

The analytics will follow the Object/Action convention for naming, which you can read about [here](https://segment.com/academy/collecting-data/naming-conventions-for-clean-data/)

### Essentially, the app will be separated into OBJECTS Spruce's objects are
- Patch
- Task
- Breadcrumb
- Navbar
- UserPatches
- CommitQueue
- Configure

#### An object also has a defined set of properties. The properties for the Patch object are: 
- patchId
- patchStatus

### Each object has a list of ACTIONS that can be taken on it. The actions for the Patch object are: 
- Filter Tasks
- Restart
- Schedule
- Set Priority
- Unschedule
- Paginate
- Change Tab
- Add Notificaiton
- Click Task Square


src/utils/parseQueryString.ts
```diff
@@ -0,0 +1,4 @@
+import queryString from "query-string";
+
+export const parseQueryString = (search: string) =>
```
SupaJoon, line 3: thank you 👍

tgrander: right??! i plan on making a followup pr at some point to replace all other instances of querystring.parse with this util

tgrander:
evergreen retry

src/analytics/patch/tests.tsx
```diff
@@ -0,0 +1,134 @@
...
+import { GET_USER, GET_PATCH, GET_PATCH_TASK_STATUSES } from "gql/queries";
+import { PatchTabs } from "pages/patch/PatchTabs";
+import { ContextProviders } from "context/Providers";
+import { GET_PATCH_EVENT_DATA } from "analytics/patch/query";
+import { Router } from "react-router-dom";
+import { createMemoryHistory } from "history";
+
+/**
+ * UNABLE TO GET THESE TESTS TO WORK AFTER 5 HOURS OF EFFORT
+ * MADE A TICKET TO REVISIT INTEGRATION TESTS AT A LATER DATE
```
khelif96, line 14: Can you please link the TicketNumber in the comment for reference

src/analytics/addPageAction.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export const addPageAction = <A extends ActionType, P extends Properties>(
+  { name, ...actionProps }: A,
+  properties: P & RequiredProperties
+) => {
+  const { newrelic } = window;
+  if (typeof newrelic !== "object") {
+    return;
+  }
+  const { search } = window.location;
```
khelif96, line 43: can we use useLocation() in react router for this? You might need to pass it from the usePatchAnalytics() hook

tgrander: we cannot because this is a function, not a hook, and therefore cannot use hooks.

tgrander: also, the useLocation in the hook at the time this function called is the previous value. we need the value of search at the time the function is called to be the most up to date.

src/analytics/patch/tests.tsx
tgrander: done

src/analytics/addPageAction.ts
```diff
@@ -0,0 +1,49 @@
...
+  newrelic.addPageAction(name, {
+    ...properties,
+    ...parseQueryString(search),
```
SupaJoon, line 46: what's the benefit of parsing search before sending to new relic?

tgrander: all fields are sent to new relic as key/value pairs instead of a string, which does not provide as great of value when querying data.

src/analytics/addPageAction.ts
```diff
@@ -0,0 +1,49 @@
...
+  interface Window {
+    newrelic: typeof NewRelicAPI; // eslint-disable-line no-undef
+  }
+}
+
+export interface Analytics<Action> {
+  sendEvent: (action: Action) => void;
+}
+
+type AnalyticsObject =
```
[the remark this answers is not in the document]
tgrander: sure, when that's done it will be added here

PR #215, opened 2020-06-09
EVG-8106: hide logs and html buttons when no logs found or gql error occurs

SupaJoon:
https://jira.mongodb.org/browse/EVG-8106
instead of rendering a gql error message that results from the TaskLogs query, print "No logs found"

### DIFF
--- a/src/pages/task/Logs.tsx
+++ b/src/pages/task/Logs.tsx
@@ -1,33 +1,24 @@
 import React, { useEffect, useState } from "react";
-import { Radio, RadioGroup } from "@leafygreen-ui/radio-group";
-import { useLocation, useHistory } from "react-router-dom";
+import { useLocation } from "react-router-dom";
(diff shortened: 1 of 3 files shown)


src/pages/task/Logs.tsx
```diff
@@ -64,64 +54,15 @@ export const Logs: React.FC<Props> = ({ logLinks }) => {
...
-          Event Logs
-        </Radio>
-      </StyledRadioGroup>
-      {options[currentLog] || <></>}
-    </div>
+  const LogComp = options[currentLog];
+  return LogComp ? (
+    <LogComp htmlLink={htmlLink} rawLink={rawLink} currentLog={currentLog} />
+  ) : (
+    <></>
```
khelif96, line 62: You can just do `return logComp && <LogComp ... />`

PR #216, opened 2020-06-09
EVG-8186 Add the new ui toggle tab to preferences

khelif96:
Dependent on https://github.com/evergreen-ci/evergreen/pull/3676
image


src/pages/preferences/preferencesTabs/NewUITab.tsx
```diff
@@ -0,0 +1,74 @@
...
+}
+export const NewUITab: React.FC<NewUITabProps> = ({ useSpruceOptions }) => {
+  const { spruceV1 } = useSpruceOptions;
+  const dispatchBanner = useBannerDispatchContext();
+  const [updateUserSettings, { loading }] = useMutation<
+    UpdateUserSettingsMutation,
+    UpdateUserSettingsMutationVariables
+  >(UPDATE_USER_SETTINGS, {
+    onCompleted: () => {
+      dispatchBanner.successBanner(
```
john-m-liu, line 26: Controls that save data immediately typically do not give any feedback about it because it's a bit jarring/annoying, and I think that applies here as well. It's definitely weird that some of the tabs have a confirmation button and some don't, but I think the pattern is clear enough here to remove the success banner

khelif96: All the other preferences tabs do show a success banner when ever we interact with them. and perform some type of modification.

src/pages/preferences/preferencesTabs/NewUITab.tsx
```diff
@@ -0,0 +1,74 @@
...
+          userSettings: {
+            useSpruceOptions: {
+              spruceV1: !spruceV1,
+            },
+          },
+        },
+        refetchQueries: ["GetUserSettings"],
+      });
+    } catch (err) {}
+  };
```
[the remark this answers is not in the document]
SupaJoon: bugsnag will catch this in the errorLink

src/pages/preferences/preferencesTabs/NewUITab.tsx
john-m-liu: Right, but those have a confirmation button. This one doesn't

SupaJoon: I personally like seeing the confirmation banner but I also think this is a good question for design

john-m-liu: I'll open a ticket

PR #217, opened 2020-06-10
EVG-7852: task analytics

tgrander:
Analytics for all task actions. See analytics spreadsheet for more info: https://docs.google.com/spreadsheets/d/1iiySLi4f6oWfqWCaVtW2p3OGykVdfNwGAZITtvDIlRw/edit#gid=0

Using the `@client` directive in queries to query the cache is always returning `undefined`. This is odd because it worked at some point in the original analytics implementation when context was being used. I am going to try a couple different things, the first of which is to make the cache query from context instead of from a hook.


src/analytics/addPageAction.ts
```diff
@@ -37,13 +37,18 @@ export const addPageAction = <A extends ActionType, P extends Properties>(
...
+  const attributesToSend = {
     ...properties,
     ...parseQueryString(search),
     ...actionProps,
-  });
+  };
+
+  if (typeof newrelic !== "object") {
+    console.log(name);
+    console.log("attributesToSend", attributesToSend);
```
khelif96, line 49: I think you forgot these here while debugging.

PR #225, opened 2020-06-11
EVG-8239 Add waterfall link to NavBar

khelif96:
image

### DIFF
--- a/src/components/Navbar/index.tsx
+++ b/src/components/Navbar/index.tsx
@@ -7,9 +7,11 @@ import { StyledLink } from "components/styles";
 import { Subtitle } from "@leafygreen-ui/typography";
 import { uiColors } from "@leafygreen-ui/palette";
 import { Link } from "react-router-dom";
+import { useNavbarAnalytics } from "analytics";
 import { routes } from "constants/routes";
 import { useLegacyUIURL } from "hooks";
-import { useNavbarAnalytics } from "analytics";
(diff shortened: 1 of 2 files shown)


src/components/Navbar/index.tsx
```diff
@@ -18,32 +18,27 @@ const { white, blue, gray } = uiColors;
...
-        <Link
-          to={routes.myPatches}
-          onClick={() => navbarAnalytics.sendEvent({ name: "Click Logo Link" })}
-        >
-          <Logo>
-            <EvergreenLogo />
-            <StyledSubtitle>Evergreen</StyledSubtitle>
-          </Logo>
-        </Link>
+        <NavActionContainer>
```
john-m-liu, line 30: Was taking out the analytics intentional?

khelif96: no it wasn't must have been a bad merge

src/components/Navbar/index.tsx
```diff
@@ -18,32 +18,27 @@ const { white, blue, gray } = uiColors;
...
+          <Link to={routes.myPatches}>
+            <Logo>
+              <EvergreenLogo />
+              <StyledSubtitle>Evergreen</StyledSubtitle>
+            </Logo>
+          </Link>
+          <NavTitle href={`${uiURL}/waterfall`}>Waterfall</NavTitle>
```
SupaJoon, line 37: can we add analytics here

PR #243, opened 2020-06-16
EVG-8110 Remove waitforgql dep from commit queue tests

khelif96:


cypress/integration/commit_queue.ts
```diff
@@ -15,19 +15,12 @@ describe("commit queue page", () => {
   });
   beforeEach(() => {
     cy.server();
```
SupaJoon, line 17: can remove this line

PR #244, opened 2020-06-16
EVG-8277: Upload source maps to bugsnag in prod deploy script

SupaJoon:
https://jira.mongodb.org/browse/EVG-8277


src/components/ErrorBoundary/index.tsx
```diff
@@ -2,8 +2,10 @@ import React from "react";
 import Bugsnag from "@bugsnag/js";
 import { getBugsnagApiKey } from "utils/getEnvironmentVariables";
 import BugsnagPluginReact from "@bugsnag/plugin-react";
+import { version } from "../../../package.json";
 
 Bugsnag.start({
+  appVersion: version,
```
khelif96, line 8: This implies that we are gonna be versioning the app. Do you think we should?

SupaJoon: we've been referring to the current release of Spruce as v1 so we already have some idea of versioning. for now the version in package.json serves as additional metadata for bugsnag to organize errors and nothing more than that

SupaJoon:
evergreen retry

PR #254, opened 2020-06-17
EVG-8172 Add hook to update page title

khelif96:


src/pages/ConfigurePatch.tsx
```diff
@@ -18,6 +19,8 @@ export const ConfigurePatch: React.FC = () => {
   >(GET_PATCH_CONFIGURE, {
     variables: { id },
   });
+  usePageTitle(`Configure Patch - ${id}`);
```
john-m-liu, line 22: This doesn't seem useful since patch ID is a random string

src/pages/Patch.tsx
```diff
@@ -42,13 +43,15 @@ const PatchCore: React.FC = () => {
...
   useEffect(() => stopPolling, [stopPolling]);
+
   const patch = get(data, "patch");
   const status = get(patch, "status");
   const description = get(patch, "description");
   const activated = get(patch, "activated");
   if (activated === false) {
     router.replace(`${paths.patch}/${id}/configure`);
   }
+  usePageTitle(`Patch - ${description} - ${patch && patch.patchNumber}`);
```
john-m-liu, line 54: I don't think it makes sense to include the description since it can be quite long. Just patch # seems fine

khelif96:
> Also can you double check the ticket #?

Thanks just updated it

PR #255, opened 2020-06-17
EVG-8052: style subscription modal

SupaJoon:
https://jira.mongodb.org/browse/EVG-8052
all the components here are antd except the buttons. leafygreen doesn't support a Select component so instead of using half leafygreen and half antd for inputs and selects, or creating my own Leafygreen style select, I deferred to antd.
app:
Screen Shot 2020-06-17 at 6 03 07 PM
design:
Screen Shot 2020-06-17 at 6 02 26 PM


src/components/NotificationModal.tsx
```diff
@@ -141,26 +127,38 @@ export const NotificationModal: React.FC<ModalProps> = ({
...
-              }}
-              value={extraFieldInputVals[key]}
-            />
-          </ExtraFieldContainer>
-        ))}
+      </Section>
+      {extraFields && extraFields.length && (
+        <Section>
+          <div>
+            <Body weight="medium">Choose how to be Notified</Body>
```
john-m-liu, line 134: Too much copypasta from below :)

src/components/NotificationModal.tsx
```diff
@@ -141,26 +127,38 @@ export const NotificationModal: React.FC<ModalProps> = ({
...
+                value={extraFieldInputVals[key]}
+              />
+            </ExtraFieldContainer>
+          ))}
+        </Section>
+      )}
       <div>
-        then notify by:
+        <div>
+          <Body weight="medium">Choose how to be Notified</Body>
```
john-m-liu, line 157: Changing the label here makes the options in the select have the wrong tense. Can you remove -ing from the option labels?
Screen Shot 2020-06-18 at 9 14 18 AM

src/components/NotificationModal.tsx
```diff
@@ -182,7 +176,9 @@ export const NotificationModal: React.FC<ModalProps> = ({
       <div>
         {currentMethodControl && (
           <>
-            {label}
+            <SectionLabelContainer>
+              <InputLabel>{label}</InputLabel>
```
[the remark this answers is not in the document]
khelif96: https://accessibility.huit.harvard.edu/provide-accessible-labels-and-instructions

src/components/NotificationModal.tsx
```diff
@@ -93,37 +92,25 @@ export const NotificationModal: React.FC<ModalProps> = ({
...
       width="50%"
       wrapProps={{
         "data-cy": "task-notification-modal",
       }}
     >
-      <>
-        when
+      <ModalTitle>Add Subscription</ModalTitle>
+      <Section>
+        <div>
```
khelif96, line 108: Is this div necessary?

src/components/NotificationModal.tsx
```diff
@@ -93,37 +92,25 @@ export const NotificationModal: React.FC<ModalProps> = ({
...
-        </Button>,
-      ]}
       width="50%"
       wrapProps={{
         "data-cy": "task-notification-modal",
       }}
     >
-      <>
-        when
+      <ModalTitle>Add Subscription</ModalTitle>
```
khelif96, line 106: I think you can just remove the `ModalTitle` styling and just pass the H2 directly to the title prop in the modal since thats what we do for some other modals

src/components/NotificationModal.tsx
```diff
@@ -141,26 +128,33 @@ export const NotificationModal: React.FC<ModalProps> = ({
...
-            />
-          </ExtraFieldContainer>
-        ))}
+        {extraFields &&
+          extraFields.length &&
+          extraFields.map(({ text, key }) => (
+            <ExtraFieldContainer key={key}>
+              <SectionLabelContainer>
+                <InputLabel>{text}</InputLabel>
+              </SectionLabelContainer>
```
khelif96, line 137: Same comment about the labels being associated with form inputs

src/components/NotificationModal.tsx
```diff
@@ -141,26 +128,33 @@ export const NotificationModal: React.FC<ModalProps> = ({
...
+                    ...extraFieldInputVals,
+                    [key]: event.target.value,
+                  });
+                }}
+                value={extraFieldInputVals[key]}
+              />
+            </ExtraFieldContainer>
+          ))}
+      </Section>
       <div>
```
khelif96, line 151: Is this unstyled div necessary?

src/components/NotificationModal.tsx
```diff
@@ -93,37 +92,25 @@ export const NotificationModal: React.FC<ModalProps> = ({
   const targetPath = get(currentMethodControl, "targetPath");
 
   return (
-    <Modal
+    <StyledModal
+      centered
       data-test-id="subscription-modal"
-      title={<H2>Add Subscription</H2>}
+      footer={null}
```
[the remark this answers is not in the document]
SupaJoon: antd will include a default footer if we exclude the prop entirely
Screen Shot 2020-06-18 at 2 40 34 PM

src/components/NotificationModal.tsx
```diff
@@ -141,26 +128,33 @@ export const NotificationModal: React.FC<ModalProps> = ({
...
-                  [key]: event.target.value,
-                });
-              }}
-              value={extraFieldInputVals[key]}
-            />
-          </ExtraFieldContainer>
-        ))}
+        {extraFields &&
+          extraFields.length &&
+          extraFields.map(({ text, key }) => (
```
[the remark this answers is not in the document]
SupaJoon: good point

src/components/NotificationModal.tsx
```diff
@@ -230,3 +243,40 @@ const ExtraFieldContainer = styled.div`
...
+  padding-bottom: 24px;
+  margin-bottom: 22px;
+  border-bottom: ${borderBottom};
+`;
+
+const SectionLabelContainer = styled.div`
+  padding-top: 16px;
+`;
+
+const ModalTitle = styled(H2)`
```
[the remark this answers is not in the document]
SupaJoon: the leafygreen H2 heading is the closest size header. we use this size on the Modify Version modal

src/components/NotificationModal.tsx
SupaJoon: good catch

src/components/NotificationModal.tsx
SupaJoon: the ModalTitle component takes care of spacing before an after the line so I think it's necessary. here is what it looks like using H2 directly. 
Screen Shot 2020-06-18 at 3 21 55 PM

src/components/NotificationModal.tsx
```diff
@@ -141,27 +126,35 @@ export const NotificationModal: React.FC<ModalProps> = ({
...
+                  });
+                }}
+                value={extraFieldInputVals[key]}
+              />
+            </ExtraFieldContainer>
+          ))}
+      </Section>
       <div>
-        then notify by:
+        <Body weight="medium">Choose how to be Notified</Body>
```
john-m-liu, line 150: Notified should be lowercase

src/components/NotificationModal.tsx
khelif96: I was referring to doing something like this. 
image
image

src/components/NotificationModal.tsx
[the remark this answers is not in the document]
khelif96: See my reply for https://github.com/evergreen-ci/spruce/pull/255/files#r442450351

SupaJoon:
evergreen retry

src/components/NotificationModal.tsx
SupaJoon: when using the title prop, we don't have control over the padding and color for the line underneath the title so to match the modal designs I had to pass the title component into the body of the modal component
from the design:
Screen Shot 2020-06-23 at 11 11 43 AM
.

src/components/NotificationModal.tsx
```diff
@@ -203,7 +199,26 @@ export const NotificationModal: React.FC<ModalProps> = ({
...
+        <Button
+          key="cancel"
+          onClick={onCancel}
+          dataCy="cancel-subscription-button"
+        >
+          Cancel
+        </Button>
+        <Button
+          key="save"
+          dataCy="save-subscription-button"
```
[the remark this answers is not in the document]
SupaJoon: dataCy is actually the prop we use for Button https://github.com/evergreen-ci/spruce/blob/master/src/components/Button/index.tsx#L39

src/components/NotificationModal.tsx
```diff
@@ -203,7 +199,26 @@ export const NotificationModal: React.FC<ModalProps> = ({
           </span>
         ))}
       </div>
-    </Modal>
+      <Footer>
+        <Button
+          key="cancel"
+          onClick={onCancel}
+          dataCy="cancel-subscription-button"
```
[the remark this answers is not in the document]
SupaJoon: dataCy is actually the prop we use for Button https://github.com/evergreen-ci/spruce/blob/master/src/components/Button/index.tsx#L39

src/components/NotificationModal.tsx
```diff
@@ -230,3 +244,41 @@ const ExtraFieldContainer = styled.div`
 const ErrorMessage = styled(Body)`
   color: ${uiColors.red.base};
 `;
+
+const borderBottom = `1px solid ${uiColors.gray.light2};`;
+
+const Section = styled.div`
+  padding-bottom: 24px;
+  margin-bottom: 22px;
+  border-bottom: ${borderBottom};
```
khelif96, line 253: `borderBottom` has a semicolon at the end of the string so when the css is generated it will have 2 semicolons at the end of this style. You should remove one of the semi-colons.

src/components/NotificationModal.tsx
```diff
@@ -203,7 +200,26 @@ export const NotificationModal: React.FC<ModalProps> = ({
           </span>
         ))}
       </div>
-    </Modal>
+      <Footer>
```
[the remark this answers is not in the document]
SupaJoon: I create a shared Modal component that accepts a footer so we can use the same styling as the designs

src/components/Modal.tsx
```diff
@@ -0,0 +1,56 @@
...
+  footer,
+  title,
+  onCancel,
+  visible,
+  children,
+  "data-cy": dataCy,
+}) => (
+  <StyledModal
+    maskStyle={{
+      backgroundColor: "hsla(0, 100%, 0%, .9)",
```
khelif96, line 24: You can use the leafygreen black and an opacity for this like we use in. https://github.com/evergreen-ci/spruce/blob/master/src/components/WelcomeModal/WelcomeModal.tsx#L91

SupaJoon: i used hsla instead of color because setting the opacity through the opacity css property resulted in a flicker during the modal open animation
hsla

flicker

SupaJoon: just update the color because i see they dont match

PR #263, opened 2020-06-22
EVG-8242: Format task logs

SupaJoon:
https://jira.mongodb.org/browse/EVG-8242
Screen Shot 2020-06-22 at 5 09 42 PM

Not sure how to avoid setting the innerHtml since I am using a function to generate HTML from a log line formatted with ansi codes. This should only be a security risk if someone was able to upload their own logs with a malicious link. I can include a domain whitelist in the toHTML options validate function but will need some ideas on how to create the list.


src/pages/task/logs/logTypes/LogMessageLine.tsx
```diff
@@ -2,6 +2,9 @@ import React from "react";
 import { LogMessage } from "gql/generated/types";
 import { getLogLineWrapper } from "pages/task/logs/logTypes/logMessageLine/LogLines";
 import { format } from "date-fns";
+import linkifyHtml from "linkifyjs/html";
+var Convert = require("ansi-to-html");
```
john-m-liu, line 6: Is it not possible to import this? If not be sure to use let

SupaJoon: oops!!

SupaJoon:
> I would suggest a package like `htmr` or `html-react-parser` to handle string to html conversion. they allow you to safely swap html elements with react components. Without having to worry about potentially risking XSS. I've used them both in the past, `html-react-parser` is a little more powerful. https://www.npmjs.com/package/htmr
> https://www.npmjs.com/package/html-react-parser

These are cool libraries but I think we can avoid installing them since ansi-to-html will escape any HTML that came from the logs to help prevent XSS attacks.
https://github.com/evergreen-ci/spruce/pull/263/files#diff-c87089f9420c6f135dfeb63d10ba205cR8

PR #264, opened 2020-06-24
EVG-8218 Update deploy scripts

khelif96:
The deploy script will now print out all the changes since the last time we released. If there are no changes it will prompt the user to confirm that they want to release.
image

We keep track of changes using npm version and git tags. When ever we release the user will be prompted for a version for the release they have the option to enter major, minor or patch which follows [semantic versioning guidlines](https://semver.org/) The version number will be updated automatically depending on their choice. After making the choice the version is updated in the `package.json` and a new tag is added to git automatically. The deploy process kicks off after this. At the end the user is reminded to push up the git tag so that their changes are reflected for everyone else.

The deploy email has also changed and now shows any changes in this release. 
image
I had to change the order this email is sent so that it now sends before we update the tags so the email script can also know what changes have occured.


cli-scripts/notify-email.js
```diff
@@ -0,0 +1,75 @@
...
+const simpleGit = require("simple-git");
+const { gitDescribeSync } = require("git-describe");
+const path = require("path");
+
+const git = simpleGit(path.resolve(__dirname, ".."));
+
+// this script gets the current user email, the latest git commit hash of
+// the current branch and sends an email to a recipient that will forward
+// the message to a slack channel
+const postToSlack = () =>
```
john-m-liu, line 11: I would call this email since it doesn't directly do anything with slack

cli-scripts/deploy-prod.js
```diff
@@ -3,53 +3,160 @@ const simpleGit = require("simple-git");
...
+    console.log("Current branch is", current);
+    console.log(
+      colors.red("Error: You must be on master branch to deploy to production")
+    );
+    return false;
+  }
+  return true;
+};
+
+// Fetches all the commits since the last release tag if there are know commits
```
SupaJoon, line 23: know typo

cli-scripts/deploy-prod.js
```diff
@@ -3,53 +3,160 @@ const simpleGit = require("simple-git");
...
+      {
+        properties: {
+          confirmDeploy: {
+            description: colors.magenta(
+              "Are you sure you want to deploy to production? [Y/N]"
+            ),
+          },
+        },
+      },
+      (confirmDeployerr, { confirmDeploy }) => {
```
SupaJoon, line 82: this variable isn't used. what that intentional?

PR #265, opened 2020-06-26
EVG-8287 improve restart modal performance

john-m-liu:
- In the patch restart modal, change the array of tasks to a dictionary since we were constantly mapping through it looking for 1 value
- Split out the task square checkbox into its own component and memoize it so that it doesn't re-render if its state doesn't change


src/utils/string/index.ts
```diff
@@ -26,3 +26,7 @@ export const omitTypename = (object) =>
   JSON.parse(JSON.stringify(object), (key, value) =>
     key === "__typename" ? undefined : value
   );
+
+export interface selectedStrings {
```
khelif96, line 30: I'm not sure that this is the best place for this to live. It might be better suited living in the same file as usePatchStatusSelect hook.

john-m-liu: Done. I didn't think it belonged in the hook either but maybe there isn't a best place for it

khelif96:
Did you fix the tests? they were failing on the pr

PR #274, opened 2020-06-29
EVG-8287 improve restart modal performance

john-m-liu:
Changes from before:
- Remove the custom memo comparator
- Actually pass the checked tasks to the places that care about it. oops


SupaJoon:
the failing test is flakey

SupaJoon:
evergreen retry

PR #279, opened 2020-07-02
EVG-8145 Format timetaken fields in patch duration

khelif96:
Dependent on 
https://github.com/evergreen-ci/evergreen/pull/3744
image

becomes this

image

### DIFF
--- a/src/utils/string/index.ts
+++ b/src/utils/string/index.ts
@@ -26,3 +26,6 @@ export const omitTypename = (object) =>
   JSON.parse(JSON.stringify(object), (key, value) =>
     key === "__typename" ? undefined : value
   );
+
+export const formatDuration = (duration: string): string =>
+  duration.replace(/\d*[dhms]/g, (match) => `${match} `).trim();
--- a/src/pages/patch/Metadata.tsx
+++ b/src/pages/patch/Metadata.tsx
@@ -1,9 +1,10 @@
 import React from "react";
+import { ApolloError } from "apollo-client";
 import { P2 } from "components/Typography";
 import { StyledLink } from "components/styles";
 import { PatchQuery } from "gql/generated/types";
 import { getUiUrl } from "utils/getEnvironmentVariables";
-import { ApolloError } from "apollo-client";
+import { formatDuration } from "utils/string";
 import { MetadataCard } from "components/MetadataCard";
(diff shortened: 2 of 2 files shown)


src/utils/string/index.ts
```diff
@@ -26,3 +26,6 @@ export const omitTypename = (object) =>
   JSON.parse(JSON.stringify(object), (key, value) =>
     key === "__typename" ? undefined : value
   );
+
+export const formatDuration = (duration: string): string =>
```
john-m-liu, line 30: Can we do this using date-fns rather than writing our own formatter? It looks like they have some support for durations but the documentation is pretty poor https://date-fns.org/v2.14.0/docs/Duration

khelif96: From what i can tell I dont think the duration support is all there to apply this type of formatting for this use case. 

We get duration strings in the following format from the backend `xxhxxmxxs`. The Duration type from `date-fns` expects an object with the following structure.
```
{
hours: <number>,
minutes: <number>,
seconds: <number>
}
```
So in order to even convert the data from the format we get it in i would have to write a function to parse the Duration string received from the backend and convert it to the `Duration` type supplied by `date-fns`.  Since `date-fns` doesn't supply any util function to do this.

`date-fns` did introduce a `formatDuration` [function](https://date-fns.org/v2.14.0/docs/formatDuration) but it wont provide the type of formatting we are looking for. Since there is no support for changing the labels for the hours, minutes, seconds fields. So what is currently `4h 12m 10s` in the old ui would become `4 hours 12 minutes 10 seconds` 

Do you think it would be worth it to add a util function for parsing the duration strings we receive from the backend to a `date-fns` format and changing the format we display durations in the UI?

john-m-liu: Ah I had incorrectly assumed these were the Duration scalar type which is a # of milliseconds. I think either we send these up as the duration type and format on the client, or format the string entirely on the server. It doesn't make sense to format in both places

khelif96: I opened https://github.com/evergreen-ci/evergreen/pull/3757 which will handle formatting entirely on the server

khelif96:
Closing this PR since this formatting is now being handled on the backend per https://github.com/evergreen-ci/evergreen/pull/3757

PR #282, opened 2020-07-07
EVG-8288: do not allow configuration of commit queue patches

tgrander:
https://jira.mongodb.org/browse/EVG-8288

### Summary
Screen Shot 2020-07-07 at 2 37 38 PM

- 'Reconfigure' link in dropdown menu on patch card (user patches page) and on version page is disabled for patches that are on the commit queue. 
- Users are redirected away from configure page and to the version page if patch is on commit queue

### Testing
- E2E tests
- manually tested both features listed above in staging


src/pages/ConfigurePatch.tsx
```diff
@@ -10,17 +10,31 @@ import { ConfigurePatchCore } from "pages/configurePatch/ConfigurePatchCore";
...
     variables: { id },
   });
+
   usePageTitle(`Configure Patch`);
 
+  const router = useHistory();
+
+  // redirect to version page if patch is on commit queue
+  useEffect(() => {
+    const commitQueuePosition = data?.patch?.commitQueuePosition ?? null;
```
john-m-liu, line 31: I believe this check fails if the patch has already started. I would check if data.patch.alias = "__commit_queue" to determine if it's a commit queue patch. Same with the other spot

src/pages/ConfigurePatch.tsx
```diff
@@ -10,17 +10,31 @@ import { ConfigurePatchCore } from "pages/configurePatch/ConfigurePatchCore";
...
+
+    if (commitQueuePosition !== null) {
+      router.push(getVersionRoute(id));
```
SupaJoon, line 34: we should replace instead of push to keep the browser history clean

src/pages/Patch.tsx
```diff
@@ -42,16 +46,23 @@ const PatchCore: React.FC = () => {
...
 
   const patch = get(data, "patch");
   const status = get(patch, "status");
   const description = get(patch, "description");
   const activated = get(patch, "activated");
+
+  const commitQueuePosition = patch?.commitQueuePosition ?? null;
+  const isPatchOnCommitQueue = commitQueuePosition !== null;
+
   if (activated === false) {
```
SupaJoon, line 60: Does this condition need to be adjusted so we don't redirect if `isPatchOnCommitQueue` is true?

src/pages/ConfigurePatch.tsx
tgrander: great suggestion, thanks!

src/pages/ConfigurePatch.tsx
tgrander: will do

src/pages/Patch.tsx
tgrander: yeah!

PR #284, opened 2020-07-08
EVG-8092 add site wide banner 

khelif96:
image
Dependent on https://github.com/evergreen-ci/evergreen/pull/3749


src/components/Banners/SiteBanner.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const { text, theme } = get(data, "siteBanner", {});
+  const [showBanner, setShowBanner] = useState(false);
+  useEffect(() => {
+    if (text !== "" && Cookies.get(text) === undefined) {
+      setShowBanner(true);
+    }
+  }, [text]);
+  if (loading) return null;
+
+  const hideBanner = () => {
```
bsamek, line 24: The behavior in the old UI is to redisplay the banner as soon as a new banner with new text is set https://github.com/evergreen-ci/evergreen/blob/880b3fd3058567b85907ed0e0d45461f80c5eab9/public/static/js/menu.js#L36. Is that behavior supported here?

khelif96: Yup it is I took that into account. We set a cookie with the current banner text when they hide the banner if the banner text changes then the cookie wont match and we show the banner.

bsamek: Thanks. I'll remove myself as a blocker from the review.

src/components/Banners/SiteBanner.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+type BannerProps = {
+  theme: string;
+};
+
+const Banner = styled.div`
+  transition: max-height 0.3s ease-in-out;
+  align-items: center;
+  ${(props: BannerProps) =>
+    props.theme && `background-color: ${bannerTypeToColor[props.theme]}`};
```
john-m-liu, line 55: I think "" is also a valid option for the theme, which defaults to announcement

khelif96: I dont think thats an option to clear the bannerTheme on the legacy UI admin panel. 
image

But i added support for it anyways just in case.

src/components/Banners/SiteBanner.tsx
```diff
@@ -0,0 +1,78 @@
...
+import IconButton from "@leafygreen-ui/icon-button";
+import Icon from "@leafygreen-ui/icon";
+import Cookies from "js-cookie";
+import get from "lodash.get";
+import { GET_SITE_BANNER } from "gql/queries";
+import { SiteBannerQuery } from "gql/generated/types";
+
+const { yellow, blue, green, red } = uiColors;
+export const SiteBanner = () => {
+  const { data, loading } = useQuery<SiteBannerQuery>(GET_SITE_BANNER);
```
SupaJoon, line 14: I think it makes sense to poll this query. Eventually we will add an option to turn off polling in https://jira.mongodb.org/browse/EVG-12420 to accommodate users with bandwidth restrictions

src/components/Banners/SiteBanner.tsx
```diff
@@ -0,0 +1,78 @@
...
+      setShowBanner(true);
+    }
+  }, [text]);
+  if (loading) return null;
+
+  const hideBanner = () => {
+    // If a user sees a banner and closes it lets set a cookie with the banners text as the key.
+    // The cookie will be auto deleted after a week. This ensures if a new banner with different text is returned we dont accidently hide it
+    setShowBanner(false);
+    Cookies.set(text, "viewed", { expires: 7 });
```
SupaJoon, line 28: I can think of some edge cases that could result from deleting the cookie and using the text as an identifier for the banner if the text is not unique that you're probably aware of.
A user can miss a message if banner A, B, and then A are displayed in the same week but the user doesn't log in to see banner B, the user will miss seeing banner A the second time around. 
A user will see the same message if the same banner is being displayed after 7 days (or the user resets their browser which we can't control)
I think the only way to avoid the first edge case is to use a unique id for the banner. The second edge case can be avoided by adding a "lastBannerViewed" flag that has the id of the last banner the user viewed.

src/components/Banners/SiteBanner.tsx
tgrander: I do not think it needs to be polled. The banner wide notification is not something that we actively anticipate changes for. Also, new pages are opened often enough by clicking on a link or creating a patch that notifications will be seen

src/components/Banners/SiteBanner.tsx
khelif96: I don't agree with that logic. 
The way it's currently implemented, If a user views banner A and dismisses it. We will not display another banner with the SAME text for 1 Week. 
If another Banner B is displayed the User will still see it, but if they dismiss it it wont show up for another week (If the banner is still there). If we decide to reintroduce a banner with the same text as Banner A and it is within 1 week of the First time banner A was dismissed we wont show it . (Chances are its the same issue that was present earlier in the week which the user already acknowledged.) 
I dont think we should necessarily worry about a user deleting cookies since if a user does that there will be other issues besides them having to see a banner one more time.
If the user deletes cookies they will be un authenticated from the site since we also store auth tokens in cookies. 
image

The banner text serves as the Unique id in this case since each banner will have its own text.
IMO It seems overkill to store whether or not a user viewed a banner on the db, Especially since the banners are only ever there for a short period of time.

src/components/Banners/SiteBanner.tsx
khelif96: Yeah i agree with Trey, Banners arent something that change to often. They are only ever displayed when an event that affects the evergreen service occurs.

src/components/Banners/SiteBanner.tsx
```diff
@@ -0,0 +1,78 @@
...
+import Icon from "@leafygreen-ui/icon";
+import Cookies from "js-cookie";
+import get from "lodash.get";
+import { GET_SITE_BANNER } from "gql/queries";
+import { SiteBannerQuery } from "gql/generated/types";
+
+const { yellow, blue, green, red } = uiColors;
+export const SiteBanner = () => {
+  const { data, loading } = useQuery<SiteBannerQuery>(GET_SITE_BANNER);
+  const { text, theme } = get(data, "siteBanner", {});
```
tgrander, line 15: tldr: consider using null propagation and nullish coalescing instead of `get`, which would be
```
const siteBanner = data?.siteBanner
const text = siteBanner?.text ?? ""
const theme = siteBanner?.theme ?? ""
```

I think we should move away from using `get` because it prevents us from being able to utilize TS autocomplete when accessing properties on the data object. using null propagation and nullish coalescing will yield same result as `get` but with type safety. for example, if `data.siteBanner` were to change to `data.banner`, this line of code would not signify an error and would just return the empty object.

src/components/Banners/SiteBanner.tsx
```diff
@@ -0,0 +1,78 @@
...
+      props.theme
+        ? bannerTypeToColor[props.theme]
+        : bannerTypeToColor.announcement
+    }`};
+  display: flex;
+  justify-content: space-between;
+  padding-left: 15px;
+  padding-right: 15px;
+  padding-top: 5px;
+  padding-bottom: 5px;
```
tgrander, line 65: `padding: 5px 15px;`

src/components/Banners/SiteBanner.tsx
SupaJoon: thanks for the explanation - I was wrong about the first edge case I described. I was under the impression that we only ever use one cookie for this feature but I realize we create a cookie per unique banner text. I also think it makes sense to not accommodate for the second edge case if it's unlikely the same banner will be active after 7 days.

src/components/Banners/SiteBanner.tsx
khelif96: Great idea!

PR #288, opened 2020-07-10
EVG-8224 Add status banner for when the browser is offline

khelif96:
offline-mode


khelif96:
Handled in https://github.com/evergreen-ci/spruce/pull/306

PR #304, opened 2020-07-16
Estimated time to start fix added

maryamiqba:
Created fix for extimated scheduled time for the metadata card @khelif96

### DIFF
--- a/src/gql/queries/get-task.ts
+++ b/src/gql/queries/get-task.ts
@@ -12,7 +12,7 @@ export const GET_TASK = gql`
         baseTaskLink
       }
       createTime
-      scheduledTime
+      estimatedStart
       displayName
       finishTime
       hostId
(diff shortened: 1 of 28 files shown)


src/pages/task/Metadata.tsx
```diff
@@ -41,6 +42,12 @@ export const Metadata: React.FC<{
       <P2 data-cy="task-metadata-submitted-at">
         Submitted at: {getDateCopy(createTime)}
       </P2>
+      <P2>
+        Estimated time to start:{" "}
+        <span data-cy="task-metadata-estimated_start">
+          {getDateCopy(scheduledTime)}
```
khelif96, line 48: This should display time to start not scheduledTime to start. So it should be the time difference between `scheduledTime` and the current time

src/gql/queries/get-task.ts
```diff
@@ -12,7 +12,7 @@ export const GET_TASK = gql`
         baseTaskLink
       }
       createTime
-      scheduledTime
+      estimatedStart
```
khelif96, line 15: Looks like you added this field twice you only need it once

PR #305, opened 2020-07-16
EVG-12503: Task checkboxes remain unselected after clicking

SupaJoon:
https://jira.mongodb.org/browse/EVG-12503
These code changes fix a bug where the task checkboxes remain unselected after clicking a checkbox.
This bug was being caused by passing an improperly formatted 2nd parameter to `lodash.get` in the `getTaskCheckboxState` state function which caused the `getTaskCheckboxState` to always return "unchecked" for task names with periods.

### DIFF
--- a/src/pages/configurePatch/configurePatchCore/ConfigureTasks.tsx
+++ b/src/pages/configurePatch/configurePatchCore/ConfigureTasks.tsx
@@ -190,9 +190,15 @@ const getTaskCheckboxState = (
   selectedBuildVariantTasks: VariantTasksState,
   variants: ProjectBuildVariant[]
 ): CheckboxState => {
-  const checkedCb = (allChecked: boolean, buildVariantName: string): boolean =>
-    allChecked &&
-    !!get(selectedBuildVariantTasks, [buildVariantName, taskName], false);
+  const checkedCb = (
+    allChecked: boolean,
+    buildVariantName: string
+  ): boolean => {
+    const isTaskSelected =
+      selectedBuildVariantTasks[buildVariantName] &&
+      selectedBuildVariantTasks[buildVariantName][taskName];
+    return allChecked && isTaskSelected;
+  };
   const checked: boolean = selectedBuildVariants
     .filter((buildVariantName) =>
       taskExistsInVariant(taskName, buildVariantName, variants)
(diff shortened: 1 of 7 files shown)


src/pages/configurePatch/configurePatchCore/ConfigureTasks.tsx
```diff
@@ -197,3 +181,115 @@ export const cardSidePadding = css`
...
+
+const getTaskCheckboxState = (
+  taskName: string,
+  selectedBuildVariants: string[],
+  selectedBuildVariantTasks: VariantTasksState,
+  variants: ProjectBuildVariant[]
+): CheckboxState => {
+  const checkedCb = (allChecked: boolean, buildVariantName: string): boolean =>
+    allChecked &&
+    !!get(selectedBuildVariantTasks, [buildVariantName, taskName], false);
```
tgrander, line 195: i propose not using `get` going forward because it does not catch type errors at compile time. from the PR summary saying the bug was "caused by passing an improperly formatted 2nd parameter to lodash.get", it sounds like type safety with get could have been the issue?

`!!get(selectedBuildVariantTasks, [buildVariantName, taskName], false);` would become 
```
selectedBuildVariantTasks.includes(buildVariantName) && selectedBuildVariantTasks.includes(taskName)
```
or something like that

src/pages/configurePatch/configurePatchCore/ConfigureTasks.tsx
```diff
@@ -7,6 +7,8 @@ import Checkbox from "@leafygreen-ui/checkbox";
 import { css } from "@emotion/core";
 import isEmpty from "lodash/isEmpty";
 import { Button } from "components/Button";
+import get from "lodash/get";
```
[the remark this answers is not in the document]
SupaJoon: I believe lodash.get is utilizing a dependency listed our package.json dependency list whereas lodash/get is utilizing a peer dependency. I can tell because if uninstall lodash.every but replace all lodash.every imports to lodash/every, the app will still compile.

src/pages/configurePatch/configurePatchCore/ConfigureTasks.tsx
SupaJoon: this is so true. `get` also doesn't always infer the correct return type from the first parameter when accessing nested values

tgrander: for accessing props in gql data, we can do `const somethingFromData = data?.somethingFromData ?? []` which will do a similar thing to `get` while utilizing type safety

PR #306, opened 2020-07-17
EVG-8224 Fix issue where user is offline and page goes blank

khelif96:
This pr prevents spruce screens from going blank when there browser loses access to the internet. It stops polling when the browser fires an offline event. and reenables polling when the browser fires an online event.

image


tgrander:
oops, when i first opened this PR it showed me, john, arjun as reviewers...but now it does not. if you did not intend for this to be reviewed then please disregard my previous comment.

khelif96:
> oops, when i first opened this PR it showed me, john, arjun as reviewers...but now it does not. if you did not intend for this to be reviewed then please disregard my previous comment.

I opened it as a draft pr it wasnt ready to be reviewed yet.

tgrander:
I think WIP PRs are great, but because they dispatch emails to everyone, @bsamek instructed me and arjun to not use them.

bsamek:
> I think WIP PRs are great, but because they dispatch emails to everyone, @bsamek instructed me and arjun to not use them.

I think that was pre-subteams. One advantage of sub-teams is that they can make process decisions and do process experiments without affecting the other subteams. If y'all want to try draft PRs, go for it.

src/hooks/usePollMonitor.ts
```diff
@@ -0,0 +1,39 @@
+import { useEffect, useState } from "react";
+import { pollInterval } from "constants/index";
+
+interface UsePollMonitorType {
```
SupaJoon, line 4: I was wondering why use an interface here instead of a type?

src/pages/Task.tsx
```diff
@@ -63,19 +64,19 @@ const TaskCore: React.FC = () => {
...
+    pollInterval,
     onError: (err) =>
       dispatchBanner.errorBanner(
         `There was an error loading the task: ${err.message}`
       ),
   });
-  useEffect(() => stopPolling, [stopPolling]);
-
+  usePollMonitor(startPolling, stopPolling);
+  // useEffect(() => stopPolling, [stopPolling]);
```
SupaJoon, line 79: extra comment

src/hooks/usePollQuery.ts
```diff
@@ -95,13 +98,17 @@ export const usePollQuery = <ApolloQueryVariables, ApolloQueryResultType>({
     search,
   ]);
 
+  // Responsible for clearing the refresh interval if the browser is offline
+  useEffect(() => {
+    if (isOffline) {
```
SupaJoon, line 103: we can restart polling when connectivity is restored with this function call.
`      pollQuery({
        search,
        resourceId,
        refetch,
        setIntervalId,
        getQueryVariables,
      });`
we may have to store the previous value of isOffline since the dependency array will have other items

src/hooks/usePollMonitor.ts
```diff
@@ -0,0 +1,39 @@
...
+      setIsOffline(true);
+      if (stopPolling) stopPolling();
+    });
+    window.addEventListener("online", () => {
+      setIsOffline(false);
+      if (startPolling) startPolling(pollInterval);
+    });
+    return () => {
+      window.removeEventListener("offline", () => undefined);
+      window.removeEventListener("online", () => undefined);
```
SupaJoon, line 35: the second parameter in removeEventListener should have the same reference from addEventListener, otherwise the event listener may not be removed

src/hooks/usePollMonitor.ts
tgrander: was wondering this too. i see that it works with an interface, but it looks kinda weird

khelif96:
> i think an integration could serve this hook well

@tgrander  I tried to write tests for this but it is not simple to test. 
For integration tests: Cypress does not currently allow simulating offline mode https://github.com/cypress-io/cypress/issues/235. There is a proposed work around to create a proxy to block requests in that thread but that will not work in this case since this hook relies on an event listener that listens for a browser `offline` event. 



> can we add some jest tests for usePollMonitor hook?

@SupaJoon  As far as creating a jest test that is also difficult to do since we rely on an event listener to detect if the users browser is offline. Its difficult to trigger events in a hook test since events are typically triggered on elements on the dom.(like a click event on a button) I could not find a way to trigger an offline event. If you have any suggestions im open to them

tgrander:
> > i think an integration could serve this hook well
> 
> @tgrander I tried to write tests for this but it is not simple to test.
> For integration tests: Cypress does not currently allow simulating offline mode [cypress-io/cypress#235](https://github.com/cypress-io/cypress/issues/235). There is a proposed work around to create a proxy to block requests in that thread but that will not work in this case since this hook relies on an event listener that listens for a browser `offline` event.
> 
> > can we add some jest tests for usePollMonitor hook?
> 
> @SupaJoon As far as creating a jest test that is also difficult to do since we rely on an event listener to detect if the users browser is offline. Its difficult to trigger events in a hook test since events are typically triggered on elements on the dom.(like a click event on a button) I could not find a way to trigger an offline event. If you have any suggestions im open to them

for clarity, when i say integration i mean jest test. when i say e2e i mean cypress. 

can you give more description and detail as to how you tested that this works then please?

khelif96:
> > > i think an integration could serve this hook well
> > 
> > 
> > @tgrander I tried to write tests for this but it is not simple to test.
> > For integration tests: Cypress does not currently allow simulating offline mode [cypress-io/cypress#235](https://github.com/cypress-io/cypress/issues/235). There is a proposed work around to create a proxy to block requests in that thread but that will not work in this case since this hook relies on an event listener that listens for a browser `offline` event.
> > > can we add some jest tests for usePollMonitor hook?
> > 
> > 
> > @SupaJoon As far as creating a jest test that is also difficult to do since we rely on an event listener to detect if the users browser is offline. Its difficult to trigger events in a hook test since events are typically triggered on elements on the dom.(like a click event on a button) I could not find a way to trigger an offline event. If you have any suggestions im open to them
> 
> for clarity, when i say integration i mean jest test. when i say e2e i mean cypress.
> 
> can you give more description and detail as to how you tested that this works then please?

Yeah so the easiest way to test is to open a spruce page with polling in it like the my patches page or a patch/task page and going into chrome dev tools, network tab and setting the browser to offline.  I did this and monitored the requests to make sure there weren't any being made while we were offline as well as monitored the renders using react dev tools to make sure the page wasn't rendering and trying to update any content while we were offline

SupaJoon:
> > @SupaJoon As far as creating a jest test that is also difficult to do since we rely on an event listener to detect if the users browser is offline. Its difficult to trigger events in a hook test since events are typically triggered on elements on the dom.(like a click event on a button) I could not find a way to trigger an offline event. If you have any suggestions im open to them

to test the pollMonitor hook, you can pass `() => window.dispatchEvent(new this.window.Event("offline"))` to the `act` function and assert the value of `isOffline`

khelif96:
> > > @SupaJoon As far as creating a jest test that is also difficult to do since we rely on an event listener to detect if the users browser is offline. Its difficult to trigger events in a hook test since events are typically triggered on elements on the dom.(like a click event on a button) I could not find a way to trigger an offline event. If you have any suggestions im open to them
> 
> to test the pollMonitor hook, you can pass `() => window.dispatchEvent(new this.window.Event("offline"))` to the `act` function and assert the value of `isOffline`

Thanks for the advice i was able to figure it out. using `fireEvent` in an `act`

PR #311, opened 2020-07-23
EVG-12530: hosts page pagination

tgrander:
- Add pagination components to the hosts page
- include pagination params in hosts query
- refetch hosts query if url search changes
- pagination e2e tests


src/gql/generated/types.ts
```diff
@@ -1,1233 +0,0 @@
-export type Maybe<T> = T | null;
```
khelif96, line 1: I think you accidentally deleted this file

tgrander: indeed. should be back now

PR #313, opened 2020-07-24
EVG-12487: Fix task_actions flake

SupaJoon:
https://jira.mongodb.org/browse/EVG-12487

I removed the flaking tests because I didn't think the test descriptions added value to our test suite


SupaJoon:
> is deleting it considered a fix? 🤔 😛

I don't think it's worth keeping this test around. I wasn't able to make this test stable without add `cy.wait` between the Restart and Schedule button clicks.

PR #314, opened 2020-07-27
EVG-8144: Started Task Status should display as "Running"

SupaJoon:
https://jira.mongodb.org/browse/EVG-8144
Use the display string "Running" to represent the Started Task Status. 
This PR also adjusts the Dispatched Task Status to display as "Dispatched" in the task dropdown filter


john-m-liu:
The "Matches status keys to status tree data" test passes on this PR when I'd expect it to fail because it's asserting dispatched -> running. Did you miss updating a spot?

SupaJoon:
> The "Matches status keys to status tree data" test passes on this PR when I'd expect it to fail because it's asserting dispatched -> running. Did you miss updating a spot?

The "Matches status keys to status tree data" test in `getCurrentStatuses.test.ts` passes because the test file is providing it's own TreeData input to the hook instead of using the TreeData input I changed in this PR

PR #316, opened 2020-07-28
EVG-12644 Add instructions for seeding data

khelif96:
Let me know if anything isn't clear or doesnt make sense

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,20 @@ We use Code generation to generate our types for our GraphQL queries and mutatio
 - Each query and mutation should have a unique name.
 - Since query analysis for type generation occurs statically we cant place dynamic variables with in query strings we instead have to hard code the variable in the query or pass it in as query variable.
 
+## How to get data for your feature
+If you need more data to be able to test out your feature locally the easiest way to do it is to populate the local db using real data from the staging or production environments.
+- First step is to locate where you want to fetch your data from if it exists in either the staging or production environments.
+- After identifying where to get your data from you should try to ssh into one of the db servers used for staging or prod. The urls for these dbs can be located in the `fabfile.py` located in the evergreen directory.
+- You should make sure you are connected to a secondary db before proceeding.
+- Open the mongo shell and identify what query returns the data you need.
+- Exit from the mongo shell and prepare to run `mongoexport`
+- On the staging db this can be run using `/var/lib/mongodb-mms-automation/mongodb-linux-x86_64-4.0.5/bin/mongoexport --db=mci --collection=<someCollection> --out=<outputFile>.json --query='<someQuery>'`
+- With this command a json file would be saved to your home directory with the results of the `mongoexport`
+- You can then transfer this json file to your local system by running the following command after quitting the ssh session. `scp <db you sshed into>:~/<outputFile>.json`
+- You should run this file through the scrambled-eggs script to sanatize it and remove any sensitive information `make scramble file=<path to file>.json`
+- Once you have this file you can copy the contents of it to the relevant `testdata/local/<collection>.json` file with in the evergreen folder
+- You can then delete `/bin/.load-local-data` with in the evergreen folder and run `make local-evergreen` to repopulate the local database with your new data.
+
 ## Deployment
 
 ### Requirements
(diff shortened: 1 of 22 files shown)


README.md
```diff
@@ -97,6 +97,20 @@ We use Code generation to generate our types for our GraphQL queries and mutatio
 - Each query and mutation should have a unique name.
 - Since query analysis for type generation occurs statically we cant place dynamic variables with in query strings we instead have to hard code the variable in the query or pass it in as query variable.
 
+## How to get data for your feature
+If you need more data to be able to test out your feature locally the easiest way to do it is to populate the local db using real data from the staging or production environments.
```
tgrander, line 101: i do not think prod and staging environments should be presented here as equals because in my experience, staging data is usually low quality and should not be used for local data. are there scenarios or examples you have in mind when using staging data is better than prod data?

README.md
```diff
@@ -97,6 +97,20 @@ We use Code generation to generate our types for our GraphQL queries and mutatio
...
+- First step is to locate where you want to fetch your data from if it exists in either the staging or production environments.
```
tgrander, line 102: the purpose of this line is unclear to me

README.md
```diff
@@ -97,6 +97,20 @@ We use Code generation to generate our types for our GraphQL queries and mutatio
...
+- After identifying where to get your data from you should try to ssh into one of the db servers used for staging or prod. The urls for these dbs can be located in the `fabfile.py` located in the evergreen directory.
```
tgrander, line 103: i think this list would benefit from being a numbered list of terse instructions. 
1. SSH into the production server - ssh url can be in the fabfile located here ('here' should be a link)
2. Ensure you are connected to a secondary node before proceeding
3. Run `mongo` command to open the mongo shell
4. Query your data, e.g. - show example of a mongo query maybe ?

...

README.md
```diff
@@ -97,6 +97,20 @@ We use Code generation to generate our types for our GraphQL queries and mutatio
...
+- You should make sure you are connected to a secondary db before proceeding.
+- Open the mongo shell and identify what query returns the data you need.
```
SupaJoon, line 105: nit: I thought these 2 steps were kind of confusing since I may already know the collection and query for the data I need. If we decide to keep these lines, I think we should mention something about identifying the collection along with the query.

README.md
```diff
@@ -97,6 +97,20 @@ We use Code generation to generate our types for our GraphQL queries and mutatio
...
+- Exit from the mongo shell and prepare to run `mongoexport`
+- On the staging db this can be run using `/var/lib/mongodb-mms-automation/mongodb-linux-x86_64-4.0.5/bin/mongoexport --db=mci --collection=<someCollection> --out=<outputFile>.json --query='<someQuery>'`
+- With this command a json file would be saved to your home directory with the results of the `mongoexport`
+- You can then transfer this json file to your local system by running the following command after quitting the ssh session. `scp <db you sshed into>:~/<outputFile>.json`
```
SupaJoon, line 109: I was unable to run this command without identifying a target file. My command looked something like this: 
` scp <db you sshed into>:~/<outputFile>.json <localOutputFilePathOrName>.json`

bsamek: If you'd like to put it in your current directory, you can `scp <db you sshed into>:~/<outputFile>.json .` The `.` is the current directory.

john-m-liu:
Does this make sense as part of the readme for an open-source product? The instructions are quite specific to our deployment (ie. someone might not even have a staging environment, and also the binary paths are specific to where we have files). I think this makes more sense as an internal document

tgrander:
> Does this make sense as part of the readme for an open-source product? The instructions are quite specific to our deployment (ie. someone might not even have a staging environment, and also the binary paths are specific to where we have files). I think this makes more sense as an internal document

if this OS project were being viewed or used by the outside world then i think your question would be valid. but as of now, the readme is read and used by evergreen team members; therefore i do think it belongs here because it adds value and would be lost anywhere else.

README.md
```diff
@@ -97,6 +97,44 @@ We use Code generation to generate our types for our GraphQL queries and mutatio
...
+## How to get data for your feature
+If you need more data to be able to test out your feature locally the easiest way to do it is to populate the local db using real data from the staging or production environments.
+
+1. You should identify if the data you need is located in the staging or prod db and ssh into them (You should be connected to the office network or vpn before proceeding). The urls for these db servers can be located in the `fabfile.py` located in the evergreen directory or [here](https://github.com/10gen/kernel-tools/blob/master/evergreen/fabfile.py).
+2. You should ensure you are connected to a secondary node before proceeding.
+3. Run `mongo` to open the the mongo shell.
+4. Identify the query you need to fetch the data you are looking for. 
+
+    ```
+    mci:SECONDARY> rs.slaveOk() // Allows read operations on a secondary node
```
tgrander, line 109: just flagging the casual racism in mongo terminology 🏴

khelif96: I believe Theres an effort on the way to eliminate it or there at least should be.

PR #317, opened 2020-07-28
EVG-12607 Add ami field to task metadata

khelif96:
Dependent on https://github.com/evergreen-ci/evergreen/pull/3838


src/pages/task/Metadata.tsx
```diff
@@ -30,6 +30,7 @@ export const Metadata: React.FC<{
   const baseCommit = get(task, "revision", "").slice(0, 10);
   const reliesOn = get(task, "reliesOn");
   const baseTaskMetadata = get(task, "baseTaskMetadata");
+  const ami = get(task, "ami");
```
tgrander, line 33: we can start using `task?.ami ?? ""` instead of `get` here. you could even replace all of these `get`s with optional chaining if you're up for it.

khelif96: Sure done.

src/pages/task/Metadata.tsx
```diff
@@ -19,22 +18,23 @@ export const Metadata: React.FC<{
...
+  const hostLink = task?.hostLink;
+  const startTime = task?.startTime;
+  const estimatedStart = task?.estimatedStart;
+  const timeTaken = task?.timeTaken;
+  const baseCommit = task?.revision.slice(0, 10);
+  const reliesOn = task?.reliesOn;
+  const baseTaskMetadata = task?.baseTaskMetadata;
+  const ami = task?.ami;
+  const baseTaskDuration = baseTaskMetadata?.baseTaskDuration;
+  const baseTaskLink = baseTaskMetadata?.baseTaskLink;
```
tgrander, line 34: some of these had default values which should be preserved. for example, `task?.revision.slice(0, 10);` will throw an error when revision is `undefined`. the default value should be `const revision = task?.revision ?? ""` then `revision.slice(0,10)`

khelif96: Thanks nice catch, Accidentally  took it out

src/pages/task/Metadata.tsx
```diff
@@ -78,6 +79,7 @@ export const Metadata: React.FC<{
           </StyledLink>
         </P2>
       )}
+      {ami && <P2 data-cy="task-metadata-ami">AMI: {ami}</P2>}
```
john-m-liu, line 82: I think just displaying AMI isn't very useful without context. This should probably be the distro instead with a link to the distro page, and the AMI as subtext afterwards

src/gql/generated/types.ts
```diff
@@ -104,6 +111,49 @@ export type Host = {
...
+  instanceType?: Maybe<Scalars["String"]>;
+  homeVolumeID?: Maybe<Scalars["String"]>;
+  user?: Maybe<Scalars["String"]>;
+  distro?: Maybe<DistroInfo>;
+  availabilityZone?: Maybe<Scalars["String"]>;
+  instanceTags?: Maybe<Array<Maybe<InstanceTag>>>;
+  expiration?: Maybe<Scalars["Time"]>;
+};
+
+export type HostEventLogData = {
```
john-m-liu, line 126: This will probably go away once everyone merges to master and resolves conflicts, but just a heads up you have some of Chaya's changes in here

src/pages/task/Metadata.tsx
khelif96: I added the distro link but Would a link to the Distro be anymore useful?  I believe the linked to Distro would always display the latest ami associated with the distro. So if the AMI changes for a distro it wont be reflected on the Distro page.

john-m-liu: I think the distro is the main piece of information since it answers questions like "what sort of machine did my task run on." The AMI is mostly a supplement to that in case the AMI has changed and the user is trying to exactly reproduce their task

src/pages/Hosts.tsx
```diff
@@ -60,7 +61,7 @@ const Hosts: React.FC = () => {
   }, [searchChanged, search, refetch]);
 
   const hosts = hostsData?.hosts;
-  const hostItems = hosts?.hosts ?? [];
+  const hostItems = (hosts?.hosts as Host[]) ?? [];
```
khelif96, line 64: Needed to do a type assertion here. Im not sure why it was needed tho since typescript should have properly asserted the correct type. cc. @tgrander  @SupaJoon

src/gql/generated/types.ts
```diff
@@ -104,6 +111,49 @@ export type Host = {
...
+  taskPid: Scalars["String"];
+  taskStatus: Scalars["String"];
+  execution: Scalars["String"];
+  monitorOp: Scalars["String"];
+  user: Scalars["String"];
+  successful: Scalars["Boolean"];
+  duration: Scalars["Duration"];
+};
+
+export type HostEventLogEntry = {
```
malikchaya2, line 145: Have you tried merging master recently? I added a HostEventLogEntry like 5 days ago.

src/pages/Hosts.tsx
khelif96: Figured this out. The hostsTable component was expecting a prop of type `Host[]` but the Hosts query didnt return data that matched the Host Type. I updated the Hosts query to add 2 fields so that it more closely matches the Host type.

src/gql/generated/types.ts
```diff
@@ -104,6 +111,49 @@ export type Host = {
...
+  id: Scalars["String"];
+  resourceType: Scalars["String"];
+  processedAt: Scalars["Time"];
+  timestamp?: Maybe<Scalars["Time"]>;
+  eventType?: Maybe<Scalars["String"]>;
+  data: HostEventLogData;
+  resourceId: Scalars["String"];
+};
+
+export type HostEvents = {
```
malikchaya2, line 155: I added a HostEvents type as well

src/gql/generated/types.ts
khelif96: I have, This file is autogenerated from the schema.graphql file and the queries. It likely just regenerated parts of this file and the git diff caught this as a new change. I double checked the file these fields weren't redeclared twice

PR #318, opened 2020-07-29
EVG-12527: reusable table filters

tgrander:
https://jira.mongodb.org/browse/EVG-12527

### The reusable table filters that will be used for filtering tables on hosts page & all other pages with a table that can be filtered

table-filters

### This PR includes
- CheckboxGroup component used for filtering statuses
- The table filter dropdown UI for selecting filters, update button and reset button
- Hooks used to store filter state and update url query params
- Jest integration tests that test the UI and the table filter hooks together

### Up Next
Adding the UI and hooks introduced in this PR to the hosts table

### DIFF
--- a/src/components/Checkbox/CheckboxGroup.tsx
+++ b/src/components/Checkbox/CheckboxGroup.tsx
@@ -1,0 +1,34 @@
+import React from "react";
+import styled from "@emotion/styled";
+import Checkbox from "@leafygreen-ui/checkbox";
+import { TreeDataEntry } from "components/TreeSelect";
+
+interface CheckboxesProps {
+  data: TreeDataEntry[];
+  value: string[];
+  onChange: (e: React.ChangeEvent<HTMLInputElement>, key: string) => void;
+}
+
+export const CheckboxGroup: React.FC<CheckboxesProps> = ({
+  data,
+  value,
+  onChange = () => undefined,
+}) => (
+  <CheckboxesWrapper>
+    {data.map(({ key, title, value: checkboxValue }) => (
+      <Checkbox
+        key={key}
+        className="cy-checkbox"
+        onChange={(e: React.ChangeEvent<HTMLInputElement>) => onChange(e, key)}
+        label={title}
+        checked={value.includes(checkboxValue)}
+        bold={false}
+        data-testid={title}
+      />
+    ))}
+  </CheckboxesWrapper>
+);
+
+const CheckboxesWrapper = styled.div`
+  padding: 4px;
+`;
--- a/src/components/Checkbox/CheckboxGroup.tsx
+++ b/src/components/Checkbox/CheckboxGroup.tsx
@@ -16,7 +16,7 @@ export const CheckboxGroup: React.FC<CheckboxesProps> = ({
 }) => (
   <CheckboxesWrapper>
     {data.map(({ key, title, value: checkboxValue }) => (
-      <Checkbox
+      <StyledCheckbox
         key={key}
         className="cy-checkbox"
         onChange={(e: React.ChangeEvent<HTMLInputElement>) => onChange(e, key)}
@@ -31,4 +31,11 @@ export const CheckboxGroup: React.FC<CheckboxesProps> = ({
 
 const CheckboxesWrapper = styled.div`
   padding: 4px;
+`;
+const StyledCheckbox = styled(Checkbox)`
+  margin-bottom: 8px;
+
+  :last-of-type {
+    margin-bottom: 0;
+  }
 `;
--- a/src/components/Table/Filters.tsx
+++ b/src/components/Table/Filters.tsx
@@ -99,7 +99,7 @@ export const CheckboxFilter: React.FC<CheckboxFilterProps> = ({
   </FilterWrapper>
 );
 
-export const getColumnTreeSelectFilterProps = ({
+export const getColumnCheckboxFilterProps = ({
   statuses,
   value,
   onChange,
--- a/src/components/Table/Filters.tsx
+++ b/src/components/Table/Filters.tsx
@@ -24,7 +24,6 @@ export const InputFilter: React.FC<InputFilterProps> = ({
 }) => (
   <FilterWrapper>
     <Input
-      aria-label="chicken"
       data-cy={dataCy}
       placeholder={placeholder}
       value={value}
--- a/src/components/Table/Filters.tsx
+++ b/src/components/Table/Filters.tsx
@@ -4,6 +4,7 @@ import { Button } from "components/Button";
 import { TreeDataEntry } from "components/TreeSelect";
 import { CheckboxGroup } from "components/Checkbox";
 import styled from "@emotion/styled";
+import { uiColors } from "@leafygreen-ui/palette";
 
 export interface InputFilterProps {
   dataCy?: string;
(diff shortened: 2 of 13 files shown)


src/components/Checkbox/CheckboxGroup.tsx
```diff
@@ -0,0 +1,34 @@
...
+import Checkbox from "@leafygreen-ui/checkbox";
+import { TreeDataEntry } from "components/TreeSelect";
+
+interface CheckboxesProps {
+  data: TreeDataEntry[];
+  value: string[];
+  onChange: (e: React.ChangeEvent<HTMLInputElement>, key: string) => void;
+}
+
+export const CheckboxGroup: React.FC<CheckboxesProps> = ({
```
SupaJoon, line 12: CheckboxGroup doesn't handle nested checkboxes that are found in the `children` of `TreeDataEntry`. Can you check with design if this is what we want to do moving forward?

src/components/Table/Filters.tsx
```diff
@@ -0,0 +1,142 @@
...
+  placeholder,
+  value,
+  onChange,
+  dataCy,
+  updateUrlParam,
+  resetUrlParam,
+}) => (
+  <FilterWrapper>
+    <Input
+      aria-label="chicken"
```
SupaJoon, line 27: we should say something more descriptive here

src/utils/url/updateUrlQueryParam.ts
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  const nextQueryParams = stringifyQuery({
+    ...urlParams,
+    ...(resetPage && { page: 0 }),
+  });
+
+  replace(`${pathname}?${nextQueryParams}`);
+
+  sendAnalyticsEvent(urlSearchParam);
```
SupaJoon, line 27: I think this could be even more useful if we can somehow include the component id that the user interacted with to update the URL. I hope if we do this, then we don't have to call `sendAnalytics` in the input onChange handler.

src/components/Checkbox/CheckboxGroup.tsx
tgrander: that feature can be added. my designs don't need it

src/utils/url/updateUrlQueryParam.ts
tgrander: analytics will be done in a different PR. please only review what i listed in the summary

src/components/Table/Filters.tsx
tgrander: lol typo

src/utils/url/updateUrlQueryParam.ts
tgrander: also, it's not in the input onChange handler, it's in the update function, which is only called when the user clicks the "Search" button. This is exactly when and where the analytics event should be called.

tgrander: for `useFilterInputChangeHandler` , where this function is called in the input onChange handler,
1. the analytics event must be sent onChange because onChange is when the url is updated and filters are dispatched in a query
2. `useFilterInputChangeHandler` will soon be deprecated when all table filters are replaced with the hooks i introduce in this PR

src/components/Checkbox/CheckboxGroup.tsx
tgrander: additionally, i would argue that the component for nested checkboxes in the form a tree should be its entirely own component since it is much more complex than this one.

SupaJoon: > that feature can be added in a future PR. my designs don't need it

We should add in the nested checkboxes as part of this ticket: https://jira.mongodb.org/browse/EVG-12550

> additionally, i would argue that the component for nested checkboxes in the form a tree should be its entirely own component since it is much more complex than this one. grouping everything into a single component for the sake of reusability can make code significantly less readable.

That makes sense to me. We should remove the word "tree" from variable names used throughout this PR since the ChecboxGroup doesn't support nested checkboxes.  https://jira.mongodb.org/browse/EVG-12550 can introduce a TreeCheckboxGroup.

src/hooks/useTableFilters.ts
```diff
@@ -0,0 +1,129 @@
...
+      urlSearchParam,
+      value,
+      search,
+      replace,
+      pathname,
+      sendAnalyticsEvent,
+      true
+    );
+
+    sendAnalyticsEvent(urlSearchParam);
```
SupaJoon, line 111: This same analytics event is being sent in the updateParams function. This seems like a bug.

src/hooks/useTableFilters.ts
```diff
@@ -0,0 +1,129 @@
...
+  };
+
+  const resetQueryParam = () => {
+    setValue("");
```
SupaJoon, line 53: I think we should be setting the value to the original query param values that are reflected in the filter component before any selections or deselections were made. 
It seems like `setValue("")` will deselect all of the checkboxes.

tgrander: the line you're commenting on is for the input filters, not checkboxes. 

reset = remove filter = setting it to the default value, which in this case is an empty string

src/hooks/useTableFilters.ts
tgrander: you're right about this not belonging here - i removed it. but i do not think removing the analytics event handlers is necessary because that's all they are - event handlers. they do not do anything at the moment and will not do anything until i do the analytics pr.

src/hooks/tests/useTableFilters/useTableFilters.test.tsx
```diff
@@ -0,0 +1,77 @@
+import React from "react";
```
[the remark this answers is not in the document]
tgrander: yeah! testing like a user would rather than implementation

src/components/Table/Filters.tsx
```diff
@@ -0,0 +1,141 @@
...
+        placeholder,
+        value,
+        onChange,
+        updateUrlParam,
+        resetUrlParam,
+      }}
+    />
+  ),
+  filterIcon: () => (
+    <Icon type="search" style={{ color: value ? "#1890ff" : undefined }} />
```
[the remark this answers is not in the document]
tgrander: yes!! thanks for catching

src/hooks/tests/useTableFilters/TestComponent.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  const { search } = useLocation();
+  const queryParams = parseQueryString(search);
+
+  return (
+    <>
+      <div>host id from url: {queryParams[hostIdUrlParam] ?? "N/A"}</div>
+      <InputFilter
+        {...{
+          dataCy: "",
```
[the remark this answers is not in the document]
tgrander: removed

src/hooks/tests/useTableFilters/TestComponent.tsx
```diff
@@ -0,0 +1,84 @@
+import React from "react";
+import { useTableInputFilter, useTableTreeSelectFilter } from "hooks";
+import { useLocation } from "react-router";
```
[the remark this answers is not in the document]
tgrander: personally i do not find value in sorting these. but if you do then please setup eslint to automate sorting imports :)

PR #321, opened 2020-07-31
EVG-12517 Display title and status 

malikchaya2:
My changes: 
image
The Design: 
image


src/types/host.ts
```diff
@@ -0,0 +1,27 @@
...
+  HostUninitialized = "initializing",
+  HostBuilding = "building",
+
+  // doesn't show up on the host page
+  HostStatusSuccess = "success",
+  HostStopping = "stopping",
+  HostStopped = "stopped",
+  HostStatusFailed = "failed",
+  HostExternalUserName = "external",
+}
```
tgrander, line 27: the "Host" at the beginning of each of these statuses is redundant, let's remove them. for example, `HostRunning` becomes `Running`

src/pages/Host.tsx
```diff
@@ -1,8 +1,34 @@
...
   const { id } = useParams<{ id: string }>();
+  // Query host data
+  const { data, loading } = useQuery<HostQuery, HostQueryVariables>(GET_HOST, {
+    variables: { id },
+  });
+  const host = get(data, "host");
+  const hostUrl = get(host, "hostUrl");
+  const status = get(host, "status");
+  usePageTitle(`Host${hostUrl ? ` - ${hostUrl}` : ""}`);
+  console.log(data);
```
tgrander, line 22: left in a console.log

src/pages/Host.tsx
```diff
@@ -1,8 +1,34 @@
...
 export const Host: React.FC = () => {
   const { id } = useParams<{ id: string }>();
+  // Query host data
+  const { data, loading } = useQuery<HostQuery, HostQueryVariables>(GET_HOST, {
+    variables: { id },
+  });
+  const host = get(data, "host");
+  const hostUrl = get(host, "hostUrl");
+  const status = get(host, "status");
+  usePageTitle(`Host${hostUrl ? ` - ${hostUrl}` : ""}`);
```
tgrander, line 21: nit: space between unrelated code

src/components/PageTitleJumbo.tsx
```diff
@@ -0,0 +1,59 @@
+import React from "react";
```
tgrander, line 1: the only difference between this component and `PageTitle` is the use of `Subtitle` versus the use of an `H2`, therefore this does not need to recreate the component. 

```
type Size = "large" | "medium";

interface Props {
  loading: boolean;
  hasData: boolean;
  title: string;
  badge: JSX.Element;
  buttons?: JSX.Element;
  size?: Size;
}

interface TitleTypographyProps {
  size: Size;
}

const TitleTypography: React.FC<TitleTypographyProps> = ({
  children,
  size,
}) => {
  if (size === "large") {
    return <H2>{children}</H2>;
  } else if (size === "medium") {
    return <Subtitle>{children}</Subtitle>;
  }
};

export const PageTitle: React.FC<Props> = ({
  loading,
  hasData,
  title,
  badge,
  buttons,
  size = "medium",
}) => (
  <>
    {!hasData && loading && (
      <PageHeader>
        <Skeleton active paragraph={{ rows: 0 }} />
      </PageHeader>
    )}
    {hasData && !loading && (
      <PageHeader>
        <TitleWrapper>
          <TitleTypography size={size}>
            <span data-cy="page-title">
              {title}
              {"  "}
              <BadgeWrapper>{badge}</BadgeWrapper>
            </span>
          </TitleTypography>
        </TitleWrapper>
        {buttons ?? null}
      </PageHeader>
    )}
  </>
);
```

src/components/HostStatusBadge.tsx
```diff
@@ -0,0 +1,53 @@
+import React from "react";
+import { HostStatus } from "types/host";
+import Badge, { Variant } from "@leafygreen-ui/badge";
+import styled from "@emotion/styled";
+
+interface Props {
+  status: string;
```
tgrander, line 7: status should be of type `HostStatus`

src/components/HostStatusBadge.tsx
```diff
@@ -0,0 +1,53 @@
...
+}
+
+export const HostStatusBadge: React.FC<Props> = ({ status }) => (
+  <StyledBadge variant={statusToBadgeVariant[status as HostStatus]}>
```
tgrander, line 11: you will not need the `as HostStatus` when you make status in Props of type `HostStatus`

src/components/HostStatusBadge.tsx
```diff
@@ -0,0 +1,53 @@
...
+    {hostStatusToCopy[status as HostStatus]}
```
tgrander, line 12: same comment

src/components/PageTitleJumbo.tsx
malikchaya2: thank you! I was wondering how I would do something like that! 
it's not only the H2 vs Subtitle that's different. The css for BadgeWrapper and for PageHeader is different too. Still worth it?

tgrander: yup, just pass a prop to the styled component that conditionally renders some css

src/components/PageTitle.tsx
```diff
@@ -1,57 +1,111 @@
...
-const BadgeWrapper = styled.span`
-  display: inline-flex;
-  position: relative;
-  top: -2px;
+const BadgeWrapper = styled.span<TitleTypographyProps>`
+  ${({ size }) =>
+    size === "medium" &&
+    css`
+      display: inline-flex;
+      position: relative;
```
tgrander, line 69: remove these from the conditional since it is present regardless of the `size` variable.

src/components/PageTitle.tsx
```diff
@@ -1,57 +1,111 @@
...
+      top: -2px;
```
tgrander, line 70: instead of having two separate string literals, this can be done: 

`top: ${props => props.size === "medium" ? "-2px" : "-4px"}`;

src/components/PageTitle.tsx
```diff
@@ -1,57 +1,111 @@
...
+    `}
+  ${({ size }) =>
+    size === "large" &&
+    css`
+      display: inline-flex;
+      position: relative;
+      top: -4px;
+      margin-left: 20px;
```
tgrander, line 78: this will become the only valuable left in the conditional

src/components/PageTitle.tsx
```diff
@@ -1,57 +1,111 @@
...
+      display: flex;
+      justify-content: space-between;
+    `}
+  ${({ size }) =>
+    size === "large" &&
+    css`
+      margin-bottom: 15px;
+      margin-top: 15px;
+      display: flex;
+      justify-content: space-between;
```
tgrander, line 96: same comments for this css

src/components/PageTitle.tsx
```diff
@@ -1,57 +1,111 @@
...
+  ${({ size }) =>
+    size === "medium" &&
+    css`
+      width: 70%;
+    `}
+  ${({ size }) =>
+    size === "large" &&
+    css`
+      width: 100%;
+    `}
```
tgrander, line 110: same comments for this css

src/pages/Host.tsx
```diff
@@ -1,8 +1,36 @@
...
+import { PageTitle } from "components/PageTitle";
+import { HostStatus } from "types/host";
 
 export const Host: React.FC = () => {
   const { id } = useParams<{ id: string }>();
+  // Query host data
+  const { data, loading } = useQuery<HostQuery, HostQueryVariables>(GET_HOST, {
+    variables: { id },
+  });
+  const host = get(data, "host");
```
khelif96, line 19: We started moving away from lodash.get and have adopted optional chaining so that we can preserve typescript typing.

you can do 
```
const host = data?.host;
const hostUrl = host?.hostUrl;
```

src/components/HostStatusBadge.tsx
```diff
@@ -0,0 +1,53 @@
...
+  [(HostStatus.Decommissioned,
+  HostStatus.Quarantined,
+  HostStatus.ProvisionFailed,
+  HostStatus.Uninitialized,
+  HostStatus.Building,
+  HostStatus.StatusSuccess,
+  HostStatus.Stopping,
+  HostStatus.Stopped,
+  HostStatus.StatusFailed,
+  HostStatus.ExternalUserName)]: Variant.LightGray,
```
khelif96, line 29: Thats interesting first time ive seen mapping multiple keys within an array to single items as values.

khelif96: Are you sure this works? I tried this out on a smaller scale and it seems like only the last key in the array gets mapped. 
image

PR #326, opened 2020-08-03
EVG-12665: hosts table sorting

tgrander:
https://jira.mongodb.org/browse/EVG-12665

Aug-03-2020 10-52-57

### Summary

Clicking hosts table column headers updates url `sortBy` and `sortDir` params

- `useSetColumnDefaultSortOrder` hook that retuns onChange handler to be passed to antd table component
- jest integration tests for `useSetColumnDefaultSortOrder` hook
- use `useSetColumnDefaultSortOrder` for tasks table and delete `TasksTable` component


src/hooks/useUpdateUrlSortParamOnTableChange.ts
```diff
@@ -0,0 +1,37 @@
...
+    // when order is undefined, sort should be reset; therefore removed from the url
+    if (!order) {
+      replace(pathname);
+    }
+
+    const queryParams = parseQueryString(search);
+
+    const nextQueryParams = {
+      ...queryParams,
+      [PatchTasksQueryParams.SortDir]:
```
SupaJoon, line 23: I think it makes sense to create an enum specific to Tables instead of using what exists in PatchTaskQueryParams

tgrander: i agree that it makes sense although I do not think it's necessary. i left it as `PatchTasksQueryParams` because the sortDir and sortBy params should all be the same. however making an enum specifically for the url query params will possibly be more explicit.

tgrander:
> How do you feel about giving the id column a consistent width as i click through the sort orders the id column width keeps changing which is a little jarring since the buttons end up moving around
> image
> image
> image

yeah totally, that happens when the table is filtered too. i'm doing atomic PRs so this PR is only for the feature of using the column headers to update the sort. i have a list of tickets in the epic backlog for polish & finish items at the end, including this one. https://jira.mongodb.org/browse/EVG-12728

PR #328, opened 2020-08-03
EVG-12712 Remove user select none to allow copying of build variants

khelif96:


khelif96:
Gonna merge failing test is the flaky filtering  e2e test

PR #329, opened 2020-08-04
EVG-12513 Evergreen Host Metadata Card Component

malikchaya2:
I'm not a fan of how the sshCommand wraps. Please let me know if you have ideas. I think I should change it to just "SSH".
Design: 
image
Implementation: 
image

------------**Update**----------
image
Changed how the sshCommand is displayed. Waiting on approval from design


src/pages/host/Metadata.tsx
```diff
@@ -0,0 +1,67 @@
...
+        Current Task:{" "}
+        <StyledLink data-cy="task-distro-link" href={taskLink}>
+          {runningTaskName}
+        </StyledLink>
+      </P2>
+    </HostMetaDataCard>
+  );
+};
+
+const DATE_FORMAT = "MMM d, yyyy, h:mm:ss aaaa";
```
john-m-liu, line 56: Does it make sense to put this somewhere in util so that we don't have to copy it in places?

src/components/styles/Layout.tsx
```diff
@@ -21,13 +21,14 @@ export const PageLayout = styled(Layout)`
...
 
-const siderWidth = "275px";
-
 export const PageSider = styled(Sider)`
   ${whiteBackground}
-  width: ${siderWidth} !important;
-  max-width: ${siderWidth} !important;
-  min-width: ${siderWidth} !important;
+  width: ${(props: { width?: number }): string =>
+    props.width ? `${props.width}px` : "275px"}; !important;
```
khelif96, line 27: I think we can remove this width css styling according to the antd docs they provide a prop to update the width. https://ant.design/components/layout/#Layout.Sider 

I think it would be better to just  use the width prop instead of hacking the css. 

```
export const PageSider = styled(Sider)`
  ${whiteBackground}
`;
PageSider.defaultProps = { width: 275 }; 
```

This should accomplish the same effect.

src/pages/host/Metadata.tsx
```diff
@@ -0,0 +1,67 @@
...
+const getDateCopy = (d: Date): string =>
+  d ? format(new Date(d), DATE_FORMAT) : "";
+
+export const P2 = styled.p`
```
khelif96, line 60: This looks similar to the P2 in 
https://github.com/evergreen-ci/spruce/blob/329538d5431e16336b3416049d1047f848004efa/src/components/Typography/index.ts#L47

You can probably use that one instead of recreating it here. You should be able to just import it and use it.

malikchaya2: I put it here because I had to change the css.

src/components/styles/Layout.tsx
```diff
@@ -21,13 +21,14 @@ export const PageLayout = styled(Layout)`
   ${whiteBackground}
 `;
 
-const siderWidth = "275px";
```
SupaJoon, line 24: nit: it could be easier to organize the styles if we keep this variable and use it for lines 27-31

src/components/HostMetaDataCard.tsx
```diff
@@ -0,0 +1,36 @@
+import React from "react";
```
tgrander, line 1: this component should go in the pages directory and should follow fractal pattern. it takes a second to learn fractal pattern, but it's straight forward once you do. give this blog post a read: https://hackernoon.com/fractal-a-react-app-structure-for-infinite-scale-4dab943092af

src/pages/host/Metadata.tsx
malikchaya2: thanks! moved it.

src/pages/host/Metadata.tsx
```diff
@@ -0,0 +1,63 @@
+import React from "react";
```
tgrander, line 1: oh ok!! i see you are already following fractal pattern here--nice job.

src/components/HostMetaDataCard.tsx
```diff
@@ -0,0 +1,36 @@
...
+}
+
+export const HostMetaDataCard: React.FC<Props> = ({
+  children,
+  error,
+  loading,
+}) => (
+  <SiderCard>
+    {loading && !error && (
+      <Skeleton active title={false} paragraph={{ rows: 0 }} />
```
SupaJoon, line 21: 0 rows will make hide the skeleton. we should add some rows so we can show it while it's loading

src/pages/host/Metadata.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { getUiUrl } from "utils/getEnvironmentVariables";
+import styled from "@emotion/styled/macro";
+import { getDateCopy } from "utils/string";
+
+export const Metadata: React.FC<{
+  loading: boolean;
+  data: HostQuery;
+  error: ApolloError;
+}> = ({ loading, data, error }) => {
+  const host = data ? data.host : null;
```
tgrander, line 16: `const host = data?.host ?? null`

src/pages/host/Metadata.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const lastCommunicationTime = host?.lastCommunicationTime;
+
+  const runningTask = host?.runningTask;
+  const runningTaskId = runningTask?.id;
+  const runningTaskName = runningTask?.name;
+  const sshCommand = `ssh ${user}@${hostUrl}`;
+
+  const taskLink = `${getUiUrl()}/task/${runningTaskId}`;
+  const distroLink = `${getUiUrl()}/distros##${distroId}`;
+  return (
```
tgrander, line 32: nit: add a space here

src/pages/host/Metadata.tsx
tgrander: you can change the css of an existing styled component by doing `const StyledP2 = styled(P2)`

SupaJoon: the P2 component in Typography seems to be mainly used with the shared MetadataCard component which is styled differently than the host metadata card. I think it makes sense to export this P2 from HostMetadataCard and also export the P2 in typography from the shared MetadataCard component and maybe call it something else like "Row".

src/pages/host/Metadata.tsx
SupaJoon: you can also do `const host = data?.host`

src/components/Typography/index.ts
```diff
@@ -3,9 +3,9 @@ import { uiColors } from "@leafygreen-ui/palette";
 
 import { css } from "@emotion/core";
 
-const { gray } = uiColors;
+export const { gray } = uiColors;
```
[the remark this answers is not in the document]
khelif96: I think you forgot to remove this.

src/pages/host/Metadata.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const sshCommand = `ssh ${user}@${hostUrl}`;
+
+  const taskLink = `${getUiUrl()}/task/${runningTaskId}`;
+  const distroLink = `${getUiUrl()}/distros##${distroId}`;
+
+  return (
+    <HostMetaDataCard error={error} loading={loading}>
+      <P2>User: {user}</P2>
+      <P2>Host Name: {hostUrl}</P2>
+      <P2>SSH Command: {sshCommand}</P2>
```
khelif96, line 36: There are 2 ways i can think of to clearly display the ssh command and make it usable you can 
A: Use a leafygreen [code component ](https://github.com/mongodb/leafygreen-ui/tree/master/packages/code) It will supply a formatted code component with a copy button.

B: Use the `pre` tag and format it to show the data. Heres a [link](https://github.com/evergreen-ci/spruce/blob/04dc0f31244de7c554bbb7fffac085ce02007438/src/pages/preferences/preferencesTabs/cliTab/VerifyCard.tsx#L51) to it in our code base
I used a combination of both on the [preferences/cli page](https://spruce.mongodb.com/preferences/cli)

malikchaya2: image
I proposed this in the design channel and waiting for feedback.

src/pages/Host.tsx
```diff
@@ -4,20 +4,27 @@ import { GET_HOST } from "gql/queries/get-host";
...
+    GET_HOST,
+    {
+      variables: { id },
+    }
+  );
+
   const host = data?.host;
   const hostUrl = host?.hostUrl;
   const status = host?.status as HostStatus;
+  const sshCommand = "ssh mci@macos-1014-68.macstadium.build.10gen.cc";
```
khelif96, line 27: I think you accidently forgot this value here while testing you should use the real ssh command

src/pages/Host.tsx
```diff
@@ -30,6 +37,14 @@ export const Host: React.FC = () => {
         hasData
         size="large"
       />
+      <PageLayout>
+        <PageSider width={350}>
+          <Metadata loading={loading} data={data} error={error} />
+          <Code language="shell" multiline={false}>
```
khelif96, line 43: I think setting multiline to true would be better for this so we get that overflow which i think is fine since the user has a copy button now.
image

malikchaya2: good point! thank you!

PR #330, opened 2020-08-05
EVG-12730 Host Events Query

malikchaya2:
[EVG-12730](https://jira.mongodb.org/browse/EVG-12730)

### DIFF
--- a/src/pages/Host.tsx
+++ b/src/pages/Host.tsx
@@ -1,8 +1,14 @@
 import React from "react";
 import { useParams } from "react-router-dom";
 import { GET_HOST } from "gql/queries/get-host";
+import { GET_HOST_EVENTS } from "gql/queries/get-host-events";
 import { useQuery } from "@apollo/react-hooks";
-import { HostQuery, HostQueryVariables } from "gql/generated/types";
+import {
+  HostQuery,
+  HostQueryVariables,
+  HostEventsQuery,
+  HostEventsQueryVariables,
+} from "gql/generated/types";
 import { usePageTitle } from "hooks/usePageTitle";
 import { PageWrapper, PageSider, PageLayout } from "components/styles";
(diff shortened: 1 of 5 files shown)


src/pages/Host.tsx
```diff
@@ -12,13 +18,27 @@ import { HostStatus } from "types/host";
...
 
+  // Query host data
+  const { data: hostEventData, loading: hostEventLoading } = useQuery<
+    HostEventsQuery,
+    HostEventsQueryVariables
+  >(GET_HOST_EVENTS, {
+    variables: { id, tag },
+  });
+
+  console.log(hostEventData, hostEventLoading);
```
khelif96, line 40: I think you forgot a `console.log` statement here.

malikchaya2: I left it there because I need to do something with hostEventData, hostEventLoading so the linter doesn't complain. I am not displaying it yet, so I'm not using it yet. The tests are still failing so it's probably not a good solution anyway. Separating this out might not be a good idea.

SupaJoon: `console.log` will actually pass our lint tests with a warning. You can fix your other lint tests by running `npm run codegen` with the schema changes in https://github.com/evergreen-ci/evergreen/pull/3860

khelif96:
can you merge in master?

src/pages/Host.tsx
khelif96: image

The linter tests fail because tag did not exist on the host type when this test was run. Rerunning it should fix that. I think Its fine to leave the console.logs() here temporarily since you plan on updating it later and users dont see this page yet, and  since they just result in linter warnings.

malikchaya2: awesome, thanks!

PR #332, opened 2020-08-06
EVG-12559: render Manage Public Keys page

SupaJoon:
https://jira.mongodb.org/browse/EVG-12559
This PR also refactors some code to make things more type safe
App screen shot after code changes:
Screen Shot 2020-08-07 at 11 29 46 AM

### DIFF
--- a/src/pages/Preferences.tsx
+++ b/src/pages/Preferences.tsx
@@ -4,7 +4,7 @@ import { useQuery } from "@apollo/react-hooks";
 import styled from "@emotion/styled";
 import { SideNav, SideNavGroup, SideNavItem } from "@leafygreen-ui/side-nav";
 import get from "lodash/get";
-import { paths, preferencesTabRoutes } from "constants/routes";
+import { PreferencesTabRoutes, getPreferencesRoute } from "constants/routes";
 import { PageWrapper } from "components/styles";
 import { PreferencesTabs } from "pages/preferences/PreferencesTabs";
 import { GET_USER_SETTINGS } from "gql/queries/get-user-settings";
@@ -22,8 +22,9 @@ export const Preferences: React.FC = () => {
   >(GET_USER_SETTINGS);
   usePageTitle("Preferences");
   const userSettings = get(data, "userSettings");
-  if (tab === undefined) {
-    return <Redirect to={`${paths.preferences}/profile`} />;
+
+  if (!tabRouteValues.includes(tab as PreferencesTabRoutes)) {
+    return <Redirect to={getPreferencesRoute(PreferencesTabRoutes.Profile)} />;
   }
   return (
     <PageWrapper>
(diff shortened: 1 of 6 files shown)


tgrander:
deleted the flakey host e2e test causing your test failure here: https://github.com/evergreen-ci/spruce/pull/333

src/pages/Preferences.tsx
```diff
@@ -22,41 +22,51 @@ export const Preferences: React.FC = () => {
   >(GET_USER_SETTINGS);
   usePageTitle("Preferences");
   const userSettings = get(data, "userSettings");
-  if (tab === undefined) {
-    return <Redirect to={`${paths.preferences}/profile`} />;
+
+  if (!tabRouteValues.includes(tab as PreferencesTabRoutes)) {
```
tgrander, line 26: if you set change `const { tab } = useParams<{ tab: string }>();` above to `const { tab } = useParams<{ tab: PreferencesTabRoutes }>();` then does it work / remove the need to use `tab as PreferencesTabRoutes` here?

src/pages/Preferences.tsx
```diff
@@ -65,7 +75,7 @@ export const Preferences: React.FC = () => {
           </SideNavGroup>
         </SideNav>
         <PreferencesTabs
-          tabKey={tab}
+          tabKey={tab as PreferencesTabRoutes}
```
tgrander, line 78: same comment here as above

src/pages/preferences/PreferencesTabs.tsx
```diff
@@ -8,27 +8,22 @@ import {
...
 import { withBannersContext } from "hoc/withBannersContext";
 import { NotificationsTab } from "./preferencesTabs/NotificationsTab";
 import { ProfileTab } from "./preferencesTabs/ProfileTab";
 import { CliTab } from "./preferencesTabs/CliTab";
 import { NewUITab } from "./preferencesTabs/NewUITab";
+import { PublicKeysTab } from "./preferencesTabs/PublicKeysTab";
 
 interface PreferenceTabsProps {
-  tabKey: string;
+  tabKey: PreferencesTabRoutes;
```
tgrander, line 21: ✅

src/pages/preferences/PreferencesTabs.tsx
```diff
@@ -39,25 +34,56 @@ const Tabs: React.FC<PreferenceTabsProps> = ({
...
+  userSettings: UserSettings
+): { title: string; Component: React.FC } => {
+  const {
+    githubUser,
+    timezone,
+    region,
+    slackUsername,
+    notifications,
+    useSpruceOptions,
+  } = userSettings || {};
```
tgrander, line 62: can `||` be replaced with `??` ?

src/pages/preferences/PreferencesTabs.tsx
```diff
@@ -39,25 +34,56 @@ const Tabs: React.FC<PreferenceTabsProps> = ({
...
   );
 };
 
-const componentMap = {
-  profile: ProfileTab,
-  notifications: NotificationsTab,
-  cli: CliTab,
-  newUI: NewUITab,
+const getTitleAndComponent = (
+  tabKey: PreferencesTabRoutes,
```
tgrander, line 52: although `tabKey` being an enum provides greater safety, can we still provide it with a default value to be even safer? e.g. `tabKey: PreferencesTabRoutes = PreferencesTabRoutes.default`

src/pages/Preferences.tsx
SupaJoon: I tried this and `tab` is still type string

src/pages/preferences/PreferencesTabs.tsx
SupaJoon: the default value will only be applied if tabKey argument is undefined. This is pretty good but I see that the enum value mapping in this function could unintentionally could become out of sync with`PreferencesTabRoutes` where `PreferencesTabRoutes` contains more values than the mapping established in this function. To avoid returning `undefined` in this case, I return a default value in the function body

src/pages/preferences/PreferencesTabs.tsx
SupaJoon: yup

PR #338, opened 2020-08-10
EVG-12520: update host status modal

tgrander:
https://jira.mongodb.org/browse/EVG-12699

Aug-10-2020 16-14-30

### Summary

Modal that will be used on `/hosts` and `/host` pages for updating host status.

### DIFF
--- a/src/components/Hosts/UpdateStatusModal.tsx
+++ b/src/components/Hosts/UpdateStatusModal.tsx
@@ -10,7 +10,7 @@ import {
 } from "gql/generated/types";
 import { UPDATE_HOST_STATUS } from "gql/mutations";
 import { useBannerDispatchContext } from "context/banners";
-import { HostStatus } from "types/host";
+import { UpdateHostStatus } from "types/host";
 import { Body } from "@leafygreen-ui/typography";
 
 const { Option } = Select;
@@ -31,7 +31,7 @@ export const UpdateStatusModal: React.FC<Props> = ({
 }) => {
   const dispatchBanner = useBannerDispatchContext();
 
-  const [status, setHostStatus] = useState<HostStatus | null>(null);
+  const [status, setHostStatus] = useState<UpdateHostStatus | null>(null);
 
   const [notes, setNotesValue] = useState<string>("");
 
@@ -100,7 +100,7 @@ export const UpdateStatusModal: React.FC<Props> = ({
       <StyledSelect
         data-cy="host-status-select"
         value={status}
-        onChange={(s) => setHostStatus(s as HostStatus)}
+        onChange={(s) => setHostStatus(s as UpdateHostStatus)}
       >
         {hostStatuses.map(({ title, value, key }) => (
           <Option key={key} value={value} data-cy={`${value}-option`}>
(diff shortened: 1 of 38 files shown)


src/components/Hosts/UpdateStatusModal.tsx
```diff
@@ -0,0 +1,216 @@
...
+
+// HOSTS STATUSES DATA FOR SELECT COMPONENT
+interface Status {
+  title: keyof typeof HostStatus;
+  value: HostStatus;
+  key: HostStatus;
+}
+
+const hostStatuses: Status[] = [
+  {
```
khelif96, line 146: Some of these statuses dont make sense in this context. This modal should update the status of the host by performing some type of action such as start, stop, terminate. The statuses are the result of that action.
 It shouldn't be possible for a user to directly give a host a stopping state they instead must perform a stop action and if successful the state of the host should become stopping, then stopped. 

You can reference some of the current actions here. https://github.com/evergreen-ci/evergreen/blob/de3debcf884440844f6a3d71b44b1ec36496d9a4/service/spawn.go#L32-L34

tgrander: great! thanks

src/components/Hosts/UpdateStatusModal.tsx
```diff
@@ -0,0 +1,162 @@
...
+          <Button
+            dataCy="modal-update-button"
+            variant="primary"
+            disabled={!status}
+            loading={loadingUpdateHostStatus}
+            onClick={onClickUpdate}
+          >
+            Update
+          </Button>
+        </Buttons>
```
khelif96, line 97: No need to create a container for these modal buttons you can pass in an array of jsx elements and antd will handle positioning the modal buttons. Similar to how we do in https://github.com/evergreen-ci/spruce/blob/b53ab195c0d671c555d10f4ed3f076f7a4d20f5d/src/pages/patch/PatchRestartModal.tsx#L104

src/components/Hosts/UpdateStatusModal.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+export const UpdateStatusModal: React.FC<Props> = ({
+  visible,
+  dataCy,
+  hostIds,
+  closeModal,
+}) => {
+  const dispatchBanner = useBannerDispatchContext();
+
+  const [status, setHostStatus] = useState<UpdateHostStatus | null>(null);
```
khelif96, line 34: You can remove the `| null` from the type of the useState() I believe the typescript compiler will ignore it any ways simply using `useState<UpdateHostStatus>(null)` should accomplish the same effect.
image
image

src/components/Hosts/UpdateStatusModal.tsx
```diff
@@ -0,0 +1,162 @@
...
+  });
+
+  const onClickUpdate = () => {
+    updateHostStatus({ variables: { hostIds, status, notes } });
+  };
+
+  const onClickCancel = () => {
+    closeModal();
+    resetForm();
+  };
```
khelif96, line 73: nit: we can also inline these in the jsx components since they seem small enough and there isnt to much going on.

tgrander: `onClickCancel` is used in two different places and i prefer them this way

src/components/Hosts/UpdateStatusModal.tsx
tgrander: ok

src/components/Hosts/UpdateStatusModal.tsx
tgrander: done

PR #343, opened 2020-08-12
EVG-12509  Evergreen Host Events Table Component

malikchaya2:
Still left to do: 
- [done] Convert the time to the user's time zone. 
- [done] Shorten string for HostTask
- [[done](https://github.com/evergreen-ci/evergreen/pull/3893)] Add all host event types to the event_log local database and make sure it renders properly on the UI. 
hostTable


src/pages/host/HostCard.tsx
```diff
@@ -0,0 +1,59 @@
...
+  loading,
+  metaData,
+}) => (
+  <SiderCard metaData={metaData}>
+    {loading && !error && (
+      <Skeleton active title={false} paragraph={{ rows: 4 }} />
+    )}
+    {error && !loading && (
+      <ErrorWrapper data-cy="metadata-card-error">{error.message}</ErrorWrapper>
+    )}
```
tgrander, line 30: i don't think loading and error are ever truthy at the same time, so `loading && <Component/>` and `error && <Component/>` should be enough

src/pages/host/HostCard.tsx
```diff
@@ -0,0 +1,59 @@
...
+    {!loading && !error && children}
```
tgrander, line 31: if loading and error are checked then you can just render `children` here

src/pages/host/HostCard.tsx
```diff
@@ -0,0 +1,59 @@
...
+  </SiderCard>
+);
+
+export const metaDataCard = css`
+  > * {
```
tgrander, line 36: avoid this selector in css as it is not performant

src/pages/host/HostTable.tsx
```diff
@@ -0,0 +1,57 @@
...
+    ),
+  },
+  {
+    title: "Event",
+    dataIndex: "data",
+    render: (_, { eventType, data }: HostEventLogEntry): JSX.Element => (
+      <div>{getHostEventString(eventType, data)}</div>
+    ),
+  },
+];
```
tgrander, line 29: component should always be first thing in the file

src/pages/host/HostTable.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { getHostEventString } from "pages/host/getHostEventString";
+import styled from "@emotion/styled";
+import { Subtitle } from "@leafygreen-ui/typography";
+
+// TABLE COLUMNS
+const columnsTemplate: Array<ColumnProps<HostEventLogEntry>> = [
+  {
+    title: "Date",
+    dataIndex: "timestamp",
+    width: 230,
```
tgrander, line 17: using percentage instead of px will allow the table to be responsive

src/pages/host/HostTable.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const logEntries = hostEvents?.eventLogEntries;
+  return (
+    <HostCard error={error} loading={loading} metaData={false}>
+      <StyledSubtitle>Recent Events </StyledSubtitle>
+      <Table
+        data-test-id="tests-table"
+        dataSource={logEntries}
+        rowKey={rowKey}
+        columns={columnsTemplate}
+        pagination={false}
```
tgrander, line 48: will pagination be added? i did not see it in the PR summary "left to do"

src/pages/host/getHostEventString.tsx
```diff
@@ -0,0 +1,291 @@
...
+            <StyledCollapse bordered={false}>
+              <Panel header="Provisioning logs" key="1">
+                <Code showLineNumbers multiline={false} language="shell">
+                  {data.logs}
+                </Code>
+              </Panel>
+            </StyledCollapse>
+          ) : (
+            ""
+          )}
```
tgrander, line 96: all of this repeated code can be a component

src/utils/string/index.ts
```diff
@@ -24,11 +24,48 @@ export const msToDuration = (ms: number): string => {
   return `${milliseconds}ms`;
 };
 
+export const stringifyNanoseconds = (input, skipDayMax, skipSecMax) => {
```
tgrander, line 27: needs unit test

malikchaya2: This is copied from [here](https://github.com/evergreen-ci/evergreen/blob/master/public/static/app/waterfall/tasks_grid.jsx#L93). I just changed var to const and things like that.

src/pages/host/HostTable.tsx
malikchaya2: https://jira.mongodb.org/browse/EVG-12512

src/pages/host/HostCard.tsx
malikchaya2: Where are you worried about loading and error both being true at the same time? It seems to be [loading && !error], [error && !loading], [!loading && !error].

src/pages/host/HostCard.tsx
malikchaya2: changed it to > p

src/pages/host/HostCard.tsx
```diff
@@ -0,0 +1,59 @@
...
+  padding-right: ${(props) => (props.metaData ? "0px" : "28px")};
+  padding-left: ${(props) => (props.metaData ? "15px" : "28px")};
+  margin-left: ${(props) => (props.metaData ? "0px" : "20px")};
+  margin-right: ${(props) => (props.metaData ? "0px" : "20px")};
+  > * {
+    margin-top: ${(props) => props.metaData && "20px"};
+  }
+`;
+
+// ${(props) => (props.metaData ? { metaDataCard } : { tableWrapper })};
```
john-m-liu, line 59: this can be removed

src/pages/host/getHostEventString.tsx
```diff
@@ -0,0 +1,285 @@
...
+import { Collapse } from "antd";
+const { Panel } = Collapse;
+
+const getTerminationString = (monitorOp: string) => {
+  switch (monitorOp) {
+    case `decommissioned`:
+      return `host was decommissioned.`;
+    case `idle`:
+      return `host was idle.`;
+    case `excess`:
```
john-m-liu, line 17: copypasta for this case

src/pages/host/getHostEventString.tsx
```diff
@@ -0,0 +1,285 @@
...
+      return `Host start attempt ${data.successful ? "succeeded" : ""}`;
+    case `HOST_STOPPED`:
+      return `Host stop attempt ${data.successful ? "succeeded" : ""}`;
+    case `HOST_MODIFIED`:
+      return `Host modify attempt ${data.successful ? "succeeded" : ""}`;
+    case `HOST_AGENT_DEPLOYED`:
+      return (
+        <div>
+          {" "}
+          Agent deployed {data.agentRevision ? "with revision" : ""}{" "}
```
john-m-liu, line 51: You can make these strings a bit more succinct if you do "with revision ${data.agentRevision}" as a single template string

src/pages/host/getHostEventString.tsx
```diff
@@ -0,0 +1,285 @@
...
+          </b>{" "}
+        </div>
+      );
+    case `HOST_JASPER_RESTART_ERROR`:
+      return (
+        <div>
+          {" "}
+          Host encountered error when restarting Jasper service
+          {data.logs ? (
+            <HostEventLog title="Provisioning logs" logs={data.logs} isCode />
```
john-m-liu, line 91: This thought is still a bit half-baked in my mind so feel free to disregard if you've already tried this. To avoid having to conditionally return the same HostEventLog element in a bunch of these cases (and maybe even the div), what if the switch/case logic is its own function that returns an object with the text and additional logs, and a different function actually returns the jsx to display the event?

src/pages/host/getHostEventString.tsx
malikchaya2: thank you!!!

src/pages/host/getHostEventString.tsx
malikchaya2: discussed over slack

src/pages/host/getHostEventString.tsx
malikchaya2: eslint doesn't allow it :(

src/utils/string/index.ts
```diff
@@ -1,4 +1,53 @@
 import { format } from "date-fns";
+import moment from "moment";
+import "moment-timezone";
```
SupaJoon, line 3: we use date-fns instead of moment since it's more lightweight.

src/utils/string/index.ts
```diff
@@ -1,4 +1,53 @@
...
+import { GET_USER_SETTINGS } from "gql/queries";
+
+export const useUserTimeZone = (time) => {
+  const { data: userSettingsData } = useQuery<GetUserSettingsQuery>(
+    GET_USER_SETTINGS
+  );
+  const timezone = get(userSettingsData, "userSettings.timezone", true);
+
+  if (timezone) {
+    return moment(time)
```
SupaJoon, line 16: I think this may be unnecessary if `time` is UTC. the Date library actually knows the timezone offset and will take that into account when you construct a new Date object. 
https://www.w3schools.com/jsref/jsref_gettimezoneoffset.asp

src/utils/string/index.ts
```diff
@@ -1,4 +1,53 @@
...
+      .tz(timezone)
+      .format("MMM D, YYYY h:mm:ss a");
+  }
+
+  return getDateCopy(time);
+};
+
+export const shortenString = (
```
SupaJoon, line 24: can we add a short comment for this function?

src/utils/string/index.ts
malikchaya2: It seems like that would set it based on local time, as opposed to setting it based on the timezone that the user set. We want to display it based on the time the user set in the preferences. (or maybe not?)

malikchaya2: I'm thinking that if a user is in SF, but working with a NY based team, they may want to see it displayed in Eastern Time. 
However, if this is not done anywhere else in the codebase, then maybe I should do that for consistency. Thoughts?

malikchaya2: nvm, doing more research, I think I'm wrong about that

src/utils/string/index.ts
```diff
@@ -24,11 +67,61 @@ export const msToDuration = (ms: number): string => {
...
+  }
+  if (input < NS_PER_HOUR * 24 || skipDayMax) {
+    return `${Math.floor(input / NS_PER_HOUR)}h ${Math.floor(
+      (input % NS_PER_HOUR) / NS_PER_MINUTE
+    )}m ${Math.floor((input % NS_PER_MINUTE) / NS_PER_SEC)}s`;
+  }
+  if (input === "unknown") {
+    return "unknown";
+  }
+  return ">= 1 day";
```
khelif96, line 104: Can this fall back to using `formatDistanceToNow` if the time interval is >= 1 day. This way on really large values we can have a more accurate representation of the actual duration until the time is reached.

khelif96:

src/utils/string/index.ts
```diff
@@ -24,11 +67,61 @@ export const msToDuration = (ms: number): string => {
...
-  d ? format(new Date(d), DATE_FORMAT) : "";
+// const DATE_FORMAT = "MMM d, yyyy, h:mm:ss aaaaa'm";
+// export const getDateCopy = (d: Date): string =>
+//   d ? format(new Date(d), DATE_FORMAT) : "";
+
+export const getDateCopy = (time, timeZone?) => {
+  if (timeZone) {
+    return formatToTimeZone(time, "MMM D, YYYY h:mm:ss a", {
+      // eslint-disable-next-line object-shorthand
+      timeZone: timeZone,
```
khelif96, line 120: This eslint rule shouldnt be disabled. Since these values have the same name you can just do `{timeZone}`

src/utils/string/index.ts
```diff
@@ -24,11 +67,61 @@ export const msToDuration = (ms: number): string => {
...
   JSON.parse(JSON.stringify(object), (key, value) =>
     key === "__typename" ? undefined : value
   );
 
-const DATE_FORMAT = "MMM d, yyyy, h:mm:ss aaaa";
-export const getDateCopy = (d: Date): string =>
-  d ? format(new Date(d), DATE_FORMAT) : "";
+// const DATE_FORMAT = "MMM d, yyyy, h:mm:ss aaaaa'm";
+// export const getDateCopy = (d: Date): string =>
+//   d ? format(new Date(d), DATE_FORMAT) : "";
```
khelif96, line 114: you can delete this deprecated code.

src/utils/string/index.ts
malikchaya2: discussed over slack

malikchaya2: replied on [#349](https://github.com/evergreen-ci/spruce/pull/349#discussion_r472205124)

src/utils/string/index.ts
malikchaya2: using { timeZone } doesn't work as input to formatToTimeZone

src/pages/host/getHostEventString.tsx
```diff
@@ -0,0 +1,287 @@
...
+    case `provision_failed`:
+      return `provisioning failed.`;
+    case `expired`:
+      return `expiration time passed.`;
+    default:
+      return `${monitorOp}`;
+  }
+};
+
+export const getHostEventString = (
```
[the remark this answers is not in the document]
malikchaya2: Will do in a separate ticket

package.json
```diff
@@ -36,6 +36,7 @@
     "apollo-link-timeout": "1.4.0",
     "axios": "0.19.2",
     "date-fns": "2.14.0",
+    "date-fns-timezone": "^0.1.4",
```
khelif96, line 39: Can you remove the `^`

PR #344, opened 2020-08-12
EVG-12780 Create spawn house page layout

khelif96:
Splitting up the spawn host page work this pr just contains the basic structure of the spawn host page plus data fetching
image

### DIFF
--- a/src/pages/spawn/SpawnHost.tsx
+++ b/src/pages/spawn/SpawnHost.tsx
@@ -1,3 +1,78 @@
 import React from "react";
+import { useQuery } from "@apollo/react-hooks";
+import styled from "@emotion/styled";
+import { H2 } from "@leafygreen-ui/typography";
+import Badge, { Variant } from "@leafygreen-ui/badge";
+import { pollInterval } from "constants/index";
+import { useBannerDispatchContext } from "context/banners";
+import { MyHostsQuery, MyHostsQueryVariables } from "gql/generated/types";
+import { GET_MY_HOSTS } from "gql/queries";
+import { SpawnHostButton } from "pages/spawn/spawnHost/index";
+import { useNetworkStatus } from "hooks";
+import { HostStatus } from "types/host";
 
-export const SpawnHost = () => <div>spawn host page</div>;
+export const SpawnHost = () => {
+  const dispatchBanner = useBannerDispatchContext();
+
+  const { data, loading, startPolling, stopPolling } = useQuery<
+    MyHostsQuery,
+    MyHostsQueryVariables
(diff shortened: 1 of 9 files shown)


src/pages/spawn/SpawnHost.tsx
```diff
@@ -1,3 +1,78 @@
...
+      console.log(e);
+    },
+  });
+  useNetworkStatus(startPolling, stopPolling);
+
+  if (loading) {
+    return <b>loading</b>;
+  }
+  const hosts = data?.myHosts;
+  const runningHosts = hosts.filter(
```
john-m-liu, line 35: These don't handle if you have 0 hosts since data.myHosts is nullable

src/pages/spawn/SpawnHost.tsx
```diff
@@ -1,3 +1,78 @@
...
+  const { data, loading, startPolling, stopPolling } = useQuery<
+    MyHostsQuery,
+    MyHostsQueryVariables
+  >(GET_MY_HOSTS, {
+    pollInterval,
+    onError: (e) => {
+      dispatchBanner.errorBanner(
+        `There was an error loading the patch: ${e.message}`
+      );
+      console.log(e);
```
SupaJoon, line 26: was this intentional?

src/pages/spawn/SpawnHost.tsx
```diff
@@ -1,3 +1,78 @@
...
+    },
+  });
+  useNetworkStatus(startPolling, stopPolling);
+
+  if (loading) {
+    return <b>loading</b>;
```
SupaJoon, line 32: why not use a skeleton here?

src/pages/spawn/SpawnHost.tsx
SupaJoon: I'm looking at the resolver for myHosts and I see that it won't return nil unless an error occurred. We can change adjust the schema to return a non-nullable array

khelif96: Yes it does when there are 0 hosts myHosts will return an empty array

src/pages/spawn/SpawnHost.tsx
khelif96: yeah this is still a wip which will be addressed in a follow up pr

src/pages/spawn/SpawnHost.tsx
khelif96: Just updated the schema to reflect that
https://github.com/evergreen-ci/evergreen/pull/3889

PR #345, opened 2020-08-12
EVG-12560: Allow deleting a public key from the My Public Keys list

SupaJoon:
https://jira.mongodb.org/browse/EVG-12560
These code changes allow a user to view their public keys and delete them by clicking the trash can. 
Screen Shot 2020-08-12 at 5 18 26 PM


src/pages/preferences/preferencesTabs/PublicKeysTab.tsx
```diff
@@ -1,3 +1,128 @@
...
+    onError(error) {
+      dispatchBanner.errorBanner(
+        `There was an error removing the public key: ${error.message}`
+      );
+    },
+  });
+
+  useEffect(() => {
+    setTableData(myKeysData?.myPublicKeys ?? []);
+  }, [myKeysData]);
```
tgrander, line 46: why is this necessary?

src/pages/preferences/preferencesTabs/PublicKeysTab.tsx
```diff
@@ -1,3 +1,128 @@
...
+
+  const columns = [
```
tgrander, line 48: i don't see any filters in the designs for this table, therefore the columns data can be outside of the component

tgrander: ah, sorting does require the context of the component function body. 👍

src/pages/preferences/preferencesTabs/PublicKeysTab.tsx
SupaJoon: the payload from the myPublicKeys query and removePublicKey mutation represent a user's complete list public keys so I update the value in useState to render in the table.

tgrander: i think it could be better to use apollo client paradigms to update data after a mutation rather than using react state and hooks, which introduces a new paradigm of updating data after a mutation that we do not use anywhere else in the app. 

you could add a `refetchQueries` to the mutation, which will probably do just fine. or you could use the `update` function in the mutation to update the cache after the mutation succeeds, which will then update the value of `myKeysData` and rerender the table.

SupaJoon: wow, very cool. thanks for the idea!!

PR #346, opened 2020-08-13
EVG-8304 display previous executions

john-m-liu:
Depends on https://github.com/evergreen-ci/evergreen/pull/3890
- Add a dropdown to the left side of the task page which allows choosing an execution, if the task has more than 1
- Add a new query to retrieve data on all executions for a task, which is the source for the dropdown
- Track the chosen execution in the task page state. It's a bit messy because we store it 0-indexed internally, but display 1-indexed and also null = latest
- Pass the execution through to the queries for each tab on the task page

Untitled


src/gql/queries/get-task-all-executions.ts
```diff
@@ -0,0 +1,11 @@
+import gql from "graphql-tag";
+
+export const GET_TASK_ALL_EXECUTIONS = gql`
```
tgrander, line 3: export this from the queries directory index file and import it from there too

src/pages/Task.tsx
```diff
@@ -144,6 +180,45 @@ const TaskCore: React.FC = () => {
...
+                  <StyledP1>
+                    {" "}
+                    Execution {ExecutionAsDisplay(
+                      singleExecution.execution
+                    )} - {shortDate(singleExecution.createTime)}
+                  </StyledP1>
+                </Option>
+              ))}
+            </StyledSelect>
+          )}
```
tgrander, line 221: to avoid the messiness that adding this component to an existing component + all of its state and query hooks, it can be made into its own component that encapsulates the jsx and and all the state/query logic

john-m-liu: done

john-m-liu:
fixed the failing tests related to this change

khelif96:

PR #349, opened 2020-08-14
EVG-12576 Add spawn host Table 

khelif96:
image

This pr adds the spawn host table rendered with live data. Tests will follow once more functionality is implemented.


src/pages/spawn/spawnHost/SpawnHostTable.tsx
```diff
@@ -1,3 +1,149 @@
...
+  const isVirtualWorkStation = host?.distro?.isVirtualWorkStation;
+  return isVirtualWorkStation ? (
+    <FlexContainer>
+      <HostIdSpan>{host.id}</HostIdSpan>
+      <WorkstationBadge>WORKSTATION</WorkstationBadge>
+    </FlexContainer>
+  ) : (
+    <HostIdSpan>{host.id}</HostIdSpan>
+  );
+};
```
tgrander, line 119: seems overkill to have this be its own component

src/pages/spawn/spawnHost/SpawnHostTable.tsx
```diff
@@ -1,3 +1,149 @@
...
+      sorter: (a, b) => sortFunctionDate(a, b, "expiration"),
+      render: (expiration) =>
+        formatDistanceToNow(new Date(expiration)) || "Does not expire",
+    },
+    {
+      title: "Uptime",
+      dataIndex: "uptime",
+      key: "uptime",
+      sorter: (a, b) => sortFunctionDate(a, b, "uptime"),
+      render: (uptime) => formatDistanceToNow(new Date(uptime)),
```
tgrander, line 57: sorting requires tests

src/pages/spawn/spawnHost/SpawnHostTable.tsx
```diff
@@ -1,3 +1,149 @@
...
+  <ActionButtonContainer>
+    <SpawnHostActionButton host={host} />
+    <PaddedButton glyph={<Icon glyph="Trash" />} size={Size.XSmall} />
+    <CopySSHCommandButton host={host} />
+    <PaddedButton size={Size.XSmall}>Edit</PaddedButton>
+  </ActionButtonContainer>
+);
+
+const CopySSHCommandButton: React.FC<{ host: Host }> = ({ host }) => {
+  const sshCommand = `${host.user}@${host.hostUrl}`;
```
SupaJoon, line 87: this is missing "ssh"

src/utils/string/index.ts
```diff
@@ -32,3 +33,36 @@ export const omitTypename = (object) =>
...
 export const getDateCopy = (d: Date): string =>
   d ? format(new Date(d), DATE_FORMAT) : "";
+
+export const copyToClipboard = (str: string) => {
+  const el = document.createElement("textarea");
+  el.value = str;
+  document.body.appendChild(el);
+  el.select();
+  document.execCommand("copy");
+  document.body.removeChild(el);
```
SupaJoon, line 43: this is interesting. the append and remove happen way too fast to even notice

src/utils/string/index.ts
```diff
@@ -32,3 +33,36 @@ export const omitTypename = (object) =>
...
+};
+
+export const sortFunctionString = (a, b, key) => {
```
SupaJoon, line 46: i think it makes sense to have some unit tests for `sortFunctionString` and `sortFunctionDate`

tgrander: i think unit tests are less important than ensuring the sorting functions do as they should in the context in which they are used. i don't mean don't write them when i say 'less important', i just mean that integration and e2e tests have higher value in asserting things work as they should.

src/pages/spawn/spawnHost/SpawnHostTable.tsx
khelif96: I dont think so, I think its cleaner this way  then trying to inline the logic inside of the render method of the columns field of the spawn host table

src/utils/string/index.ts
```diff
@@ -32,3 +33,36 @@ export const omitTypename = (object) =>
 const DATE_FORMAT = "MMM d, yyyy, h:mm:ss aaaa";
 export const getDateCopy = (d: Date): string =>
   d ? format(new Date(d), DATE_FORMAT) : "";
+
+export const copyToClipboard = (str: string) => {
```
john-m-liu, line 37: Can we use something like react-copy-to-clipboard instead of implementing this ourselves?

src/pages/spawn/spawnHost/SpawnHostTable.tsx
```diff
@@ -1,3 +1,149 @@
...
+      sorter: (a, b) => sortFunctionString(a, b, "status"),
+      render: (status) => <HostStatusBadge status={status} />,
+    },
+    {
+      title: "Expires In",
+      dataIndex: "expiration",
+      key: "expiration",
+      sorter: (a, b) => sortFunctionDate(a, b, "expiration"),
+      render: (expiration) =>
+        formatDistanceToNow(new Date(expiration)) || "Does not expire",
```
john-m-liu, line 50: I almost left the opposite comment on Chaya's pr, but this function is less precise than what we had before, and I think is not quite precise enough for these data elements. Could you use the function that she created to format durations? Probably need to coordinate some git logistics depending on who merges first

src/pages/spawn/spawnHost/SpawnHostTable.tsx
tgrander: if that's the case, then if it deserves to be its own component then it should be its own file too

src/pages/spawn/spawnHost/SpawnHostTable.tsx
khelif96: Just left a comment on her pr https://github.com/evergreen-ci/spruce/pull/343#discussion_r471778881 depending on when it is merged i can add this in a follow up pr.

src/utils/string/index.ts
khelif96: i disagree i dont think adding an extra package thats 37kb large is worth the extra bloat when we can accomplish the same functionality in 7 lines of code.

khelif96:

src/pages/spawn/spawnHost/SpawnHostTable.tsx
malikchaya2: I don't think it's a good idea to modify stringifyNanoseconds to fit this use case. stringifyNanoseconds takes a number as input, for examples something like 1183, and returns something like < 1 ms. This use case takes a date object pointing at some time in the future, and tries to determine how much time is left between now and then.

khelif96: I agree stringifyNanoseconds isnt suited for this use case. I think using formatDistanceToNow is is fine if we end up needing more granularity in the future we can probably create our own function but i dont think its as urgent.

PR #352, opened 2020-08-17
EVG-12562 & EVG-12563: Create/Update Modal for Public Key

SupaJoon:
This PR adds code for the public key modal. The modal can be accessed on http://localhost:3000/preferences/publickeys by clicking on "Add New Key" button or the pencil icon within a public key row. 
The submit button will be disabled when trying to add a key where the name already exists or update an existing key name to a name that already exists. 
There is basic validation for public keys and more robust validation on the server. 
Screen Shot 2020-08-19 at 3 21 02 PM


cypress.json
```diff
@@ -3,5 +3,6 @@
   "projectId": "m45k1x",
   "reporterOptions": {
     "mochaFile": "junit.xml"
-  }
+  },
+  "defaultCommandTimeout": 15000
```
SupaJoon, line 7: typing the public key always resulted in timeout so I increased the defaultCommandTimeout from a default of 4000. There is a option parameter on the type command to specify a different timeout value but cypress does not seem to respect that

src/pages/preferences/preferencesTabs/publicKeysTab/EditModal.tsx
```diff
@@ -0,0 +1,191 @@
...
+    setKeyName(initialKeyName);
+  }, [initialKeyName]);
+
+  useEffect(() => {
+    setKeyValue(initialKeyValue);
+  }, [initialKeyValue]);
+
+  useEffect(() => {
+    const inputErrors = [];
+    if (!keyName) {
```
john-m-liu, line 95: If you cancel the modal (I had valid data in, but I don't think it matters) 2 of these 3 errors flash briefly in the error container. It's probably because there is a brief rerender in response to setting the props to their defaults before the modal is hidden. Can you add a check so that it doesn't happen?

src/pages/preferences/preferencesTabs/publicKeysTab/EditModal.tsx
```diff
@@ -0,0 +1,191 @@
...
+    },
+  });
+  const [createPublicKey] = useMutation<
+    CreatePublicKeyMutation,
+    CreatePublicKeyMutationVariables
+  >(CREATE_PUBLIC_KEY, {
+    onError(error) {
+      dispatchBanner.errorBanner(
+        `There was an error creating the public key: ${error.message}`
+      );
```
khelif96, line 72: We should clear the error banner while making subsequent mutations. If a successful mutation is made after an errored mutation the error banner is still there.

src/pages/preferences/preferencesTabs/publicKeysTab/EditModal.tsx
```diff
@@ -0,0 +1,191 @@
...
+    },
+    update(cache, { data }) {
+      cache.writeQuery<GetMyPublicKeysQuery, GetMyPublicKeysQueryVariables>({
+        query: GET_MY_PUBLIC_KEYS,
+        data: { myPublicKeys: [...data.createPublicKey] },
+      });
```
khelif96, line 78: I think from a readability perspective this is confusing. Lets just call refetchQueries when we perform any mutation like we do elsewhere in the code base. Manually updating the apollo cache is error prone and in this case this is very cheap query so i dont see any advantage to updating the cache manually.

src/pages/preferences/preferencesTabs/publicKeysTab/EditModal.tsx
```diff
@@ -0,0 +1,191 @@
...
+    onError(error) {
+      dispatchBanner.errorBanner(
+        `There was an error editing the public key: ${error.message}`
+      );
+    },
+    update(cache, { data }) {
+      cache.writeQuery<GetMyPublicKeysQuery, GetMyPublicKeysQueryVariables>({
+        query: GET_MY_PUBLIC_KEYS,
+        data: { myPublicKeys: [...data.updatePublicKey] },
+      });
```
khelif96, line 62: Same comment as below about manually updating the apollo cache.

src/pages/preferences/preferencesTabs/publicKeysTab/EditModal.tsx
```diff
@@ -0,0 +1,191 @@
...
+  onCancel: () => void;
+}
+
+export const EditModal: React.FC<EditModalProps> = ({
+  replaceKeyName,
+  initialKeyName,
+  initialKeyValue,
+  existingKeys,
+  visible,
+  onCancel,
```
khelif96, line 45: I'm not convinced this is the best approach.
 I dont think we need to pass in all this information from the parent into this component. Doing it this way seems like we are managing this modals state in two separate places. I think it would be better to just pass in a publicKey as a prop. 
IF a publicKey is passed in then this would be an edit modal otherwise it would be a create new key modal. You can handle keeping track of the initial passed in values from within the modal itself.

src/pages/preferences/preferencesTabs/publicKeysTab/EditModal.tsx
```diff
@@ -0,0 +1,191 @@
...
+      <InputLabel>Public Key</InputLabel>
+      <TextArea
+        data-cy="key-value-input"
+        value={keyValue}
+        autoSize={{ minRows: 4, maxRows: 6 }}
+        onChange={(e) => setKeyValue(e.target.value)}
+      />
+      <ErrorContainer>
+        {errors.map((text) => (
+          <div key={uuid()} data-cy="error-message">
```
khelif96, line 168: Why have an unstyled div wrapping the error message?

src/pages/preferences/preferencesTabs/PublicKeysTab.tsx
```diff
@@ -54,19 +64,27 @@ export const PublicKeysTab: React.FC = () => {
     },
```
khelif96, line 64: Couldnt comment above but lines 38-43 i have the same comment about not manually updating the cache and instead using refetchQueries to update this data.

src/pages/preferences/preferencesTabs/publicKeysTab/EditModal.tsx
SupaJoon: hm, this is a good point. we can probably, store replaceKeyName inside the EditModal, initialKeyName and initialKeyValue can be combined into 1 prop and existingKeys can be read from EditModal.
After those changes the props could be `visible`, `onCancel`, and `publicKey?`

src/pages/preferences/preferencesTabs/publicKeysTab/EditModal.tsx
tgrander: i agree that refetchQueries will do fine here but i don't find this confusing. what about writing to the cache is error prone? all of our queries being statically typed makes it less error prone.

src/pages/preferences/preferencesTabs/publicKeysTab/EditModal.tsx
tgrander: same comment questioning your comment about manually updating the apollo cache

src/pages/preferences/preferencesTabs/publicKeysTab/EditModal.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+  useEffect(() => {
+    setKeyName(initialKeyName);
+  }, [initialKeyName]);
+
+  useEffect(() => {
+    setKeyValue(initialKeyValue);
+  }, [initialKeyValue]);
+
+  useEffect(() => {
```
tgrander, line 93: comments are very helpful when scanning a component with lots of hooks in them to know at a glance what each hook accomplishes

src/pages/preferences/preferencesTabs/publicKeysTab/EditModal.tsx
SupaJoon: this requires a little more code than using using `refetchQueries` but is documented well in the Apollo docs so I'm too concerned about readability. I think the tedious/error prone part is that there are 3 different gql endpoints that I use to get a list of public keys and they all have to be maintained. If I used `refetchQueries`, we only rely on a single query to get all of a users public keys.

src/pages/preferences/preferencesTabs/publicKeysTab/EditModal.tsx
SupaJoon: Since ErrorMessage extends the Body tag, I need to wrap them with a div so they go on a new line. I could also use <br/>

src/pages/preferences/preferencesTabs/publicKeysTab/EditModal.tsx
tgrander: i do think that refetchQueries is better for this query, but i don't want to rule out making writes to the cache. the feature itself is not error prone and works very well. Also, in the case where multiple queries rely on the same field - that is exactly what `fragments` are for

src/pages/preferences/preferencesTabs/publicKeysTab/EditModal.tsx
```diff
@@ -0,0 +1,199 @@
...
+        data: { myPublicKeys: [...data.createPublicKey] },
+      });
+    },
+  });
+
+  const [keyName, setKeyName] = useState<string>();
+  const [keyValue, setKeyValue] = useState<string>();
+
+  useEffect(() => {
+    setKeyName(replaceKey?.name ?? "");
```
john-m-liu, line 87: Won't doing this cause infinite rerenders? I think you need to use the initial state param above

src/pages/preferences/preferencesTabs/publicKeysTab/EditModal.tsx
khelif96: div is ok

src/pages/preferences/preferencesTabs/publicKeysTab/EditModal.tsx
SupaJoon: `replaceKey` is the initial state param and I can see how the name can be confusing. I followed the implementation I described in this comment: https://github.com/evergreen-ci/spruce/pull/352#discussion_r472251033
I can rename the key to `initialKey` or something to indicate that the prop represents an initial state

SupaJoon:
evergreen retry

SupaJoon:
evergreen retry

src/pages/preferences/preferencesTabs/publicKeysTab/EditModal.tsx
SupaJoon: @khelif96 I'm going to keep the current implementation since it's working properly and keep in mind `refetchQueries` for future mutations where the payload is simple/small

PR #360, opened 2020-08-20
EVG-12828: upgrade antd to v4

tgrander:


package.json
```diff
@@ -1,8 +1,9 @@
 {
   "name": "spruce",
-  "version": "0.7.1",
+  "version": "0.8.0",
```
khelif96, line 3: Didnt realize this didnt push up to upstream when you deployed.

package.json
```diff
@@ -1,8 +1,9 @@
...
   "private": true,
   "dependencies": {
+    "@ant-design/icons": "^4.2.2",
```
khelif96, line 6: `^`

package.json
tgrander: was confused by this

khelif96: When we issue a new release this version number gets increased it coincides with the released versions. https://github.com/evergreen-ci/spruce/releases

Its used by npm to determine what the next version of our project should be for the next time we release.

cypress/integration/patch/subscription-modal.ts
```diff
@@ -45,7 +45,7 @@ describe("Version Subscription Modal", () => {
       cy.dataTestId("1-display-name-option").should(
         "have.css",
         "user-select",
-        "none"
+        "auto"
```
SupaJoon, line 48: we should probably mute or adjust this test since checking css is no longer a reliable way to test if the item is disabled

PR #363, opened 2020-08-21
EVG-12615: apollo client 3

tgrander:


src/App.tsx
```diff
@@ -3,13 +3,6 @@ import { Global, css } from "@emotion/core";
 import GQLWrapper from "gql/GQLWrapper";
 import { BrowserRouter as Router } from "react-router-dom";
 import { ErrorBoundary } from "components/ErrorBoundary";
-import {
-  getGQLUrl,
-  getSchemaString,
-  isDevelopment,
-  isTest,
-  shouldEnableGQLMockServer,
-} from "utils/getEnvironmentVariables";
```
SupaJoon, line 12: can we remove any of these unused functions from `utils/getEnvironmentVariables` and anything related to `mockIntrospectSchema` and `mockCustomSchema` in our ReadMe?

src/App.tsx
```diff
@@ -36,14 +29,7 @@ const App: React.FC = () => (
   <ErrorBoundary>
     <ContextProviders>
       <Router>
-        <GQLWrapper
-          gqlURL={getGQLUrl()}
-          isDevelopment={isDevelopment()}
-          isTest={isTest()}
-          schemaString={getSchemaString()}
-          credentials="include"
```
SupaJoon, line 44: I think we may need this prop set to pass along the cookie in our GQL requests

tgrander: yup, that's why it's still included

src/App.tsx
tgrander: done

src/gql/GQLWrapper.tsx
```diff
@@ -1,64 +1,48 @@
...
 import { reportError } from "utils/errorReporting";
+import { getGQLUrl } from "utils/getEnvironmentVariables";
+
+const GQLWrapper: React.FC = ({ children }) => {
+  const [client, setClient] = useState(null);
+  const { logout, dispatch } = useAuthDispatchContext();
+
+  useEffect(() => {
+    async function getAndSetClient(): Promise<void> {
+      const gqlClient = await getGQLClient({
```
SupaJoon, line 21: `getGQLClient` is not longer async so we can don't need to import `useState` in this file

tgrander: yeah i thought so too which is why my original refactor was `const client = new ApolloClient()`. doing const client = new ApolloClient()  breaks authentication, which breaks e2e tests. the error is caused by infinite logout requests to the server, which are caused by infinite initial requests from the initial load logic. 

asynchronously setting client in state prevents the infinite rerenders.

SupaJoon: ah, I see what you mean by the infinite renders when we dont save the result of getGQLClient in state. We can actually do something like this since `getGQLClient` function is not asynchronous and shouldn't follow the `await` operator. 

```  const [client] = useState(
    getGQLClient({
      credentials: "include",
      gqlURL: getGQLUrl(),
      logout,
      dispatch,
    })
  );```

tgrander: there were consistent flakes for several retries when i changed it to your suggestion. when i reverted it back to my original implementation, the flakes stopped. the change you propose is cleaner but has no technical difference and, for whatever reason, created more flakes. I am fine merging it as my original implementation.

SupaJoon: can we remove the await on line 21 since `getGQLClient` is not an async function? I created a ticket to cleanup the rest of GQLWrapper since we've accumulated some tech debt here. https://jira.mongodb.org/browse/EVG-12889

PR #365, opened 2020-08-25
EVG-12749  Add Cypress Tests

malikchaya2:


cypress/integration/host/host_events.ts
```diff
@@ -0,0 +1,125 @@
...
+    );
+    cy.dataCy("host-running-task-set").contains(
+      "Assigned to run task evergreen_ubuntu1604_test_command_patch_5e823e1f28..."
+    );
+    cy.dataCy("host-running-task-cleared").contains(
+      "Current running task cleared (was:evergreen_ubuntu1604_test_command_patch_5e823e1f28...)"
+    );
+    cy.dataCy("host-task-finished").contains(
+      "Task evergreen_ubuntu1604_test_command_patch_5e823e1f28... completed with status: test-timed-out"
+    );
```
tgrander, line 66: rather than asserting the correct data line by line, which is difficult to read and unlikely maintainable, you can generate tests in a loop. take a look at the hosts table tests for examples of this.

cypress/integration/host/host_events.ts
```diff
@@ -0,0 +1,125 @@
...
+  });
+
+  it("host events with logs display the correct text and the logs get displayed when available", () => {
```
tgrander, line 69: same comment for these tests

cypress/integration/host/host_sidebar.ts
```diff
@@ -0,0 +1,46 @@
...
+    cy.dataCy("current-running-task").contains("none");
+  });
+
+  it("sshCommand has the correct values", () => {
+    cy.visit(pathWithTask);
+    cy.dataCy("ssh-command").contains(
+      "ssh admin@ec2-54-146-18-248.compute-1.amazonaws.com"
+    );
+  });
+});
```
tgrander, line 46: adding a test for ensuring text is copied to clipboard will add greater value here

src/pages/host/getHostEventString.tsx
```diff
@@ -273,15 +308,17 @@ export const HostEventLog: React.FC<{
...
-        <Code multiline={false} language="shell">
-          {logs}
-        </Code>
-      ) : (
-        { logs }
-      )}
-    </Panel>
-  </StyledCollapse>
+  <span data-cy="host-event-logs">
+    <StyledCollapse bordered={false}>
```
SupaJoon, line 312: is it possible to add a data-cy attribute to StyledCollapse instead of nesting it inside of a span? this will make our HTML cleaner especially if the the outer span's primary purpose is to hold a data-cy attribute

src/pages/host/getHostEventString.tsx
```diff
@@ -35,20 +35,36 @@ export const getHostEventString = (
...
+      return (
+        <span data-cy="provision-error">
+          Host encountered error during provisioning
+        </span>
+      );
     case HostEvent.Started:
-      return `Host start attempt ${data.successful ? "succeeded" : ""}`;
+      return (
+        <span data-cy="started">
+          Host start attempt {data.successful ? "succeeded" : ""}
```
SupaJoon, line 50: I realize this was code existed before this diff but it is a good practice to save repeated copy in a variable to allow easier updates

cypress/integration/patch-route.ts
```diff
@@ -10,7 +10,7 @@ const badPatch = {
   id: "i-dont-exist",
 };
 
-const hasText = ($el) => {
+export const hasText = ($el) => {
```
SupaJoon, line 13: I don't see hasText used anywhere else in this diff. Does it need to be exported?

cypress/integration/host/host_core.ts
```diff
@@ -15,3 +15,22 @@ describe("Host load page with nonexistent host", () => {
...
+  });
+
+  beforeEach(() => {
+    cy.listenGQL();
+    cy.preserveCookies();
+  });
+
+  it("title shows the host name", () => {
+    cy.visit(`host/macos-1014-68.macstadium.build.10gen`);
+    cy.get("[data-cy=page-title").should(
```
SupaJoon, line 31: the first function call can replaced with `cy.dataCy("page-title)`

cypress/integration/host/host_sidebar.ts
malikchaya2: Based on a brief google search it seems to be a bit complicated to write in cypress. Considering that the copy functionality is not very important to the page, and is part of a leafy-green component which I assume is well tested on their end, it probably becomes a less worthwhile pursuit. Please lmk if you disagree.

malikchaya2:
> the e2e tests you wrote are failing in CI

https://github.com/evergreen-ci/evergreen/pull/3936
This should fix it

PR #369, opened 2020-08-27
EVG-12512 Handle pagination for host events table

malikchaya2:
pagination


malikchaya2:
note on tests: 
this relies on the back end change for the tests to pass. https://github.com/evergreen-ci/evergreen/pull/3937

cypress/integration/host/host_events.ts
```diff
@@ -12,6 +12,8 @@ describe("Host events", () => {
 
   it("host events display the correct text", () => {
     cy.visit(pathWithEvents);
+    cy.dataTestId("host-event-table-page-size-selector").click();
+    cy.dataTestId("host-event-table-page-size-selector-100").click();
```
tgrander, line 16: this test asserts that pagination component can be clicked but it does not assert that the correct number of hosts appear after pagination is selected

cypress/integration/host/host_events.ts
```diff
@@ -12,6 +12,12 @@ describe("Host events", () => {
...
+
+    const dataCyTableRows =
+      "[data-test-id=host-events-table] tr td:first-child";
+    cy.get(dataCyTableRows).should("have.length.of.at.most", 100);
```
tgrander, line 20: i think this will always pass because it checks if task queue table length is less than or equal to 100. since you're working with static data, you can assert the exact length that the table should be

malikchaya2: good point. changed it

PR #370, opened 2020-08-28
EVG-8136 Add an eta timer in a tasks metadata

khelif96:
This adds a countdown/countup timer in the task metadata card that adds information about how long a task has been running for/ expected to run for.
I also added some conditionals so that some unpopulated fields dont show up unless they have data.

eta2


src/pages/task/Metadata.tsx
```diff
@@ -39,29 +42,46 @@ export const Metadata: React.FC<{
...
+          <span data-cy="task-metadata-estimated_start">
+            {msToDuration(estimatedStart)}
+          </span>
+        </P2>
+      )}
+      {/* Can only show the time running and eta if the task is running and 
+      it has a baseTaskDuration to calculate the eta with */}
+      {status === TaskStatus.Started && baseTaskDuration && (
+        <P2>
+          Running Time / ETA:{" "}
```
SupaJoon, line 64: nit: I think ETA refers to a date but we are actually showing a duration. I think "ETA" should actually be "Time Left"

src/pages/task/Metadata.tsx
```diff
@@ -39,29 +42,46 @@ export const Metadata: React.FC<{
...
-        <span data-cy="task-metadata-started">{getDateCopy(startTime)}</span>
-      </P2>
-      <P2 data-cy="task-metadata-finished">
-        Finished:{" "}
-        <span data-cy="task-metadata-started">{getDateCopy(finishTime)}</span>
-      </P2>
-      <P2 data-cy="task-metadata-duration">
-        Duration: {msToDuration(timeTaken)}{" "}
-      </P2>
+      {estimatedStart > 0 && (
```
SupaJoon, line 52: can we test this condition in the `task_metadata_card.ts`?

src/pages/task/Metadata.tsx
```diff
@@ -39,29 +42,46 @@ export const Metadata: React.FC<{
...
+        <P2>
+          Estimated time to start:{" "}
+          <span data-cy="task-metadata-estimated_start">
+            {msToDuration(estimatedStart)}
+          </span>
+        </P2>
+      )}
+      {/* Can only show the time running and eta if the task is running and 
+      it has a baseTaskDuration to calculate the eta with */}
+      {status === TaskStatus.Started && baseTaskDuration && (
```
SupaJoon, line 62: can we test this condition in the `task_metadata_card.ts`?

khelif96:
Couldnt think of a way to reliably test this component in cypress since it relies on real time to pass and its difficult to mock that in cypress. So i wrote a unit test using `@testing-library/react;` I also wrote some scaffolding and custom render functions for some common use cases such as needing react router. They also support queries of the dom through our `data-cy` attributes that we already use on our cypress tests

src/pages/task/metadata/ETATimer.tsx
```diff
@@ -0,0 +1,43 @@
...
+  const [runningTime, setRunningTime] = useState(
+    differenceInMilliseconds(parsedStartTime, Date.now())
+  );
+
+  useEffect(() => {
+    const timer = setTimeout(() => {
+      const currentTime = Date.now();
+      setEta(differenceInMilliseconds(estimatedCompletionTime, currentTime));
+      setRunningTime(differenceInMilliseconds(currentTime, parsedStartTime));
+    }, 1000);
```
john-m-liu, line 32: this fixes https://jira.mongodb.org/browse/EVG-2279 👍

PR #371, opened 2020-08-31
EVG-12912 Update eslint rules about import order

khelif96:
This pr makes a change to the eslint rules so that imports are automatically sorted in the following order.
```
import React from "react";
import {someThing} from "@somePackageThatStartsWith@";
import {someOtherThing} from "anExternalDep";
import {someOtherOtherThing} from "anAliasedInternalDep";
import {someOtherOtherOtherThing} from "./AReletavePath";
```


khelif96:
Failing test is the flaky public_key_management one which is not affected by this pr so im gonna merge.

PR #372, opened 2020-08-31
EVG-12724 task createTime -> ingestTime

ybrill:
The createTime for a task in a patch is actually the time its base version was created. The ingestTime field is the time the task was created.


ybrill:
@khelif96 I merged in master

PR #379, opened 2020-09-01
EVG-12759: scroll to task in url

tgrander:
### When task ID is present in url, a ref is attached to that task in the table and automatically scrolled to. 

Aug-28-2020 13-15-36


src/pages/taskQueue/TaskQueueTable.tsx
```diff
@@ -36,6 +39,18 @@ export const TaskQueueTable = () => {
...
+  const taskRowRef = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
+    if (taskRowRef.current) {
+      taskRowRef.current.scrollIntoView({
+        behavior: "smooth",
+        block: "start",
+      });
+    }
+  });
```
tgrander, line 52: Why is there no dependency array here?
1. Gives console warning if `[taskRowRef.current]` is put as dependency
2. Does not trigger scroll if only `[taskRowRef]` is put as dependency
3. No dependency array = always updates on re-render = 🆗 & no performance hits as far as i can tell

src/pages/taskQueue/TaskQueueTable.tsx
```diff
@@ -79,9 +105,24 @@ export const TaskQueueTable = () => {
...
       rowKey={({ id }: { id: string }): string => id}
+      rowSelection={{
+        hideSelectAll: true,
+        selectedRowKeys: [taskId],
+        renderCell(...[, { id }]) {
+          if (id === taskId) {
+            return <div ref={taskRowRef} />;
+          }
+          return null;
+        },
```
khelif96, line 116: you can also do this in a 1 liner. 
` return id === taskId ? <div ref={taskRowRef} /> : null;`

PR #387, opened 2020-09-02
EVG-12874: task queue position on task metadata

tgrander:
Screen Shot 2020-09-02 at 5 03 14 PM


src/pages/task/Metadata.tsx
```diff
@@ -43,6 +52,13 @@ export const Metadata: React.FC<{
 
   const distroLink = `${getUiUrl()}/distros##${distroId}`;
 
+  const { data: taskQueuePositionData } = useQuery<
+    TaskQueuePositionQuery,
+    TaskQueuePositionQueryVariables
+  >(TASK_QUEUE_POSITION, { variables: { taskId } });
+
+  const taskQueuePosition = taskQueuePositionData?.task?.minQueuePosition ?? 0;
```
tgrander, line 60: I decided to make a separate query for the task queue position because it is non-critical data that I do not want to slow down the top-level task query, even if it is only a matter of milliseconds.

john-m-liu (approved):

PR #391, opened 2020-09-08
EVG-12899 Change LastCommunication Format on /host/<hostID>

malikchaya2:
image

### DIFF
--- a/cypress/integration/host/host_sidebar.ts
+++ b/cypress/integration/host/host_sidebar.ts
@@ -27,7 +27,7 @@ describe("Host page title and sidebar ", () => {
 
     cy.visit(pathWithTask);
     cy.dataCy("host-last-communication").contains(
-      "Last Communication: Jul 20, 2020 9:16:36 am"
+      "Last Communication: about 2 months ago"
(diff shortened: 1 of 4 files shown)


SupaJoon (approved):

cypress/integration/host/host_sidebar.ts
```diff
@@ -27,7 +27,7 @@ describe("Host page title and sidebar ", () => {
 
     cy.visit(pathWithTask);
     cy.dataCy("host-last-communication").contains(
-      "Last Communication: Jul 20, 2020 9:16:36 am"
+      "Last Communication: about 2 months ago"
```
khelif96, line 30: I think the tests will break when this changes. Next month this will be 3 months ago

khelif96 (changes requested):

cypress/integration/host/host_sidebar.ts
malikchaya2: Thanks for spotting this!

khelif96 (approved):

PR #393, opened 2020-09-08
EVG-7693: title case all badge copy

SupaJoon:
https://jira.mongodb.org/browse/EVG-7693
Screen Shot 2020-09-08 at 5 54 44 PM
Screen Shot 2020-09-08 at 5 54 26 PM


src/components/SmallCapsBadge/index.tsx
```diff
@@ -0,0 +1,6 @@
+import styled from "@emotion/styled";
+import Badge from "@leafygreen-ui/badge";
```
khelif96, line 2: I think if we want all our badges follow this pattern it might be ok to just name this Badge and just reexport the leafygreen badge from here. 
```
import Badge, {Variant} from "@leafygreen-ui/badge";

 const SmallCapsBadge = styled(Badge)`
  text-transform: lowercase;
`;
export default SmallCapsBadge;
export {Variant} 
```

khelif96: This way we can just import like this.

```
import Badge, {Variant} from "components/Badge"
```
instead of 
```
import {Variant} from "@leafygreen-ui/badge";
import {Badge} from "components/SmallCapsBadge";
```

khelif96 (changes requested):

src/components/Badge/index.tsx
```diff
@@ -0,0 +1,9 @@
+import styled from "@emotion/styled";
+import Badge, { Variant } from "@leafygreen-ui/badge";
+
+const StyledBadge = styled(Badge)`
+  text-transform: none;
+`;
```
khelif96, line 6: Is this still needed?

SupaJoon: yup, we need override this property: https://github.com/mongodb/leafygreen-ui/blob/master/packages/badge/src/Badge.tsx#L27

khelif96 (approved):

PR #394, opened 2020-09-09
EVG-12910 Display related Jira tickets in the build baron

malikchaya2:
image

update: 
image


src/utils/string/index.ts
```diff
@@ -110,14 +110,19 @@ export const omitTypename = (object) =>
     key === "__typename" ? undefined : value
   );
 
-export const getDateCopy = (time: Date, tz?: string) => {
+export const getDateCopy = (time: Date, tz?: string, dateOnly?: boolean) => {
   if (tz) {
-    return formatToTimeZone(time, "MMM D, YYYY h:mm:ss a", {
+    const dateFormat = dateOnly ? "MMM D, YYYY" : "MMM D, YYYY h:mm:ss a";
```
SupaJoon, line 115: we should use a lowercase "d" since `D` represents day of year, where as `d` represents the day of month  
https://date-fns.org/v2.16.1/docs/format

src/pages/Task.tsx
```diff
@@ -197,49 +230,61 @@ const TaskCore: React.FC = () => {
...
-                      "Files"
-                    )}
-                  </span>
-                }
-                id="task-files-tab"
-              >
-                <FilesTables />
-              </Tab>
-            </StyledTabs>
+            {showBuildBaronTab !== undefined && (
```
SupaJoon, line 233: this condition will always be true since `showBuildBaronTab` will never be undefined

src/components/Buildbaron/BuildBaronTable.tsx
```diff
@@ -0,0 +1,98 @@
...
+const { blue, gray } = uiColors;
+
+export const BuildBaronTable: React.FC<{
+  eventData: BuildBaron;
+  timeZone: string;
+}> = ({ eventData, timeZone }) => {
+  const searchReturnInfo = eventData?.searchReturnInfo;
+  const jiraIssues = searchReturnInfo?.issues;
+
+  const columns = [
```
SupaJoon, line 19: we usually create the columns variable outside of our function components

SupaJoon (changes requested):

src/pages/Task.tsx
```diff
@@ -129,6 +140,28 @@ const TaskCore: React.FC = () => {
...
+    BuildBaronQueryVariables
+  >(GET_BUILD_BARON, {
+    variables: { taskId: id, execution },
+    pollInterval,
+  });
+  const buildBaron = buildBaronData?.buildBaron;
+  const buildBaronConfigured = buildBaron?.buildBaronConfigured;
+
+  // logic for displaying the build baron tab
+  const timeZone = useUserTimeZone();
```
SupaJoon, line 155: we don't need this variable since the date-fns format function can handle timezone differences as long as the time in UTC (i think it is)

SupaJoon (changes requested):

src/pages/Task.tsx
malikchaya2: discussed over slack

SupaJoon (approved):

src/components/Buildbaron/BuildBaronTable.tsx
```diff
@@ -0,0 +1,96 @@
...
+      pagination={false}
+    />
+  );
+};
+
+const JiraTicketRow: React.FC<{
+  jiraKey: String;
+  fields;
+}> = ({ jiraKey, fields }) => {
+  const url = `https://jira.mongodb.org/browse/${jiraKey}`;
```
john-m-liu, line 44: This URL should be queried from the db since it is configurable

src/components/Buildbaron/BuildBaronTable.tsx
```diff
@@ -0,0 +1,96 @@
...
+import { Table } from "antd";
+import { StyledLink } from "components/styles";
+import { BuildBaron, JiraTicket } from "gql/generated/types";
+import { getDateCopy } from "utils/string";
+
+const { blue, gray } = uiColors;
+
+const columns = [
+  {
+    title: "Related tickets from JIRA",
```
john-m-liu, line 14: Jira is actually not an acronym (I think it's short for Gojira) so this shouldn't be all caps

src/components/Buildbaron/BuildBaronTable.tsx
```diff
@@ -0,0 +1,96 @@
...
+];
+
+export const BuildBaronTable: React.FC<{
+  eventData: BuildBaron;
+}> = ({ eventData }) => {
+  const searchReturnInfo = eventData?.searchReturnInfo;
+  const jiraIssues = searchReturnInfo?.issues;
+
+  return (
+    <Table
```
john-m-liu, line 30: Could you override the table header color in the antd table? It's a slight shade of grey which I don't think goes well with the background, and the design does not have a special background-color on it: https://mongodb.invisionapp.com/share/C2WKQWRWG5Y#/screens/415337714

john-m-liu (changes requested):

src/components/Buildbaron/BuildBaronTable.tsx
malikchaya2: thanks! I was also wondering about why it was all caps. I'll fix it.

src/components/Buildbaron/BuildBaronTable.tsx
malikchaya2: https://jira.mongodb.org/browse/EVG-12988

src/components/Buildbaron/BuildBaronTable.tsx
malikchaya2: image

john-m-liu (approved):

src/components/Buildbaron/index.ts
```diff
@@ -0,0 +1,3 @@
+import { loadable } from "components/SpruceLoader";
+
+export const BuildBaron = loadable(() => import("./BuildBaron"));
```
khelif96, line 3: nice.

khelif96 (approved):

src/components/Buildbaron/index.ts
malikchaya2: thanks! It's technically part of the Task page, and not used by any other pages, so should this be moved to /pages/Task instead?

PR #396, opened 2020-09-09
EVG-12960: fix task files input style

SupaJoon:
https://jira.mongodb.org/browse/EVG-12950
Screen Shot 2020-09-09 at 3 59 00 PM

### DIFF
--- a/src/pages/task/FilesTables.tsx
+++ b/src/pages/task/FilesTables.tsx
@@ -112,12 +112,13 @@ export const FilesTables: React.FC = () => {
 
   return (
     <>
-      <StyledInput
-        placeholder="Search File Names"
-        onChange={onSearch}
-        suffix={<Icon glyph="MagnifyingGlass" />}
-      />
-      <br />
+      <div>
+        <StyledInput
+          placeholder="Search File Names"
+          onChange={onSearch}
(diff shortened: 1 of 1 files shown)


src/pages/task/FilesTables.tsx
```diff
@@ -117,6 +117,7 @@ export const FilesTables: React.FC = () => {
         onChange={onSearch}
         suffix={<Icon glyph="MagnifyingGlass" />}
       />
+      <br />
```
khelif96, line 120: Why the br. Can we just do the padding with a margin on the styled input?

SupaJoon: the br is actually so the input is on a separate line from the contents below. an alternative would be to wrap the input with a div

khelif96: I like wrapping the parent with a flex column div approach a little better. The br seems a little out of place, In this context.
Typically `<br/>`  is [used to break up lines of  text instead of elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/br#:~:text=The%20HTML%20element%20produces,division%20of%20lines%20is%20significant.)

khelif96 (approved):

PR #399, opened 2020-09-10
EVG-12571 Edit Spawn Host modal

khelif96:
Changes in this pr
Made the spawn volume button reusable,
Started the edit spawn host modal.
Made some fields like expiration and volume reusable 

Things to come in follow up prs,
Adding a User tags field,
Adding some tests,
Adding input validation,

edithostmodal


SupaJoon (approved):

src/pages/spawn/spawnHost/EditSpawnHostModal.tsx
```diff
@@ -0,0 +1,184 @@
...
+  HostExpirationField,
+  VolumesField,
+} from "pages/spawn/spawnHost/fields";
+
+const { Option } = Select;
+
+interface editSpawnHostState {
+  expiration: Date;
+  noExpiration: boolean;
+  displayName?: string;
```
SupaJoon, line 28: I think it could make sense to this required since displayName since the defualt value in defaultEditSpawnHostState is an empty string

src/pages/spawn/spawnHost/EditSpawnHostModal.tsx
```diff
@@ -0,0 +1,184 @@
...
+  instanceType?: string;
```
SupaJoon, line 29: this can also be required since a default value exists

src/pages/spawn/spawnHost/fields/HostExpirationField.tsx
```diff
@@ -0,0 +1,106 @@
...
+  data: ExpirationDateType;
+  onChange: React.Dispatch<React.SetStateAction<any>>;
+}
+export const HostExpirationField: React.FC<HostExpirationFieldProps> = ({
+  onChange,
+  data,
+}) => {
+  const { expiration, noExpiration } = data;
+
+  const updateDate = (_, dateString) => {
```
john-m-liu, line 25: For these, I think the first parameter is the actual date object right? So wouldn't it be cleaner to use that and combine with time rather than parsing the string?

john-m-liu (changes requested):
The modal doesn't seem to block input, so clicking around on it is toggling the stuff below. Do you know what's causing that?
Untitled

khelif96:
> The modal doesn't seem to block input, so clicking around on it is toggling the stuff below. Do you know what's causing that?
> Untitled

Thats really odd looks like the click event is propagating down to the table below it.

src/pages/spawn/spawnHost/fields/HostExpirationField.tsx
khelif96: Thanks good suggestion. not sure why we missed it last time.

khelif96:
> > The modal doesn't seem to block input, so clicking around on it is toggling the stuff below. Do you know what's causing that?
> > Untitled
> 
> Thats really odd looks like the click event is propagating down to the table below it.

@john-m-liu  I looked into it and it has to do with how react handles event propagation. https://github.com/ant-design/ant-design/issues/16004 , https://github.com/facebook/react/issues/11387 I found a workaround tho i'm going to be wrapping the modals with a div that has a simple on Click handler to prevent event propagation.

src/pages/spawn/spawnHost/fields/HostExpirationField.tsx
```diff
@@ -0,0 +1,98 @@
...
+  onChange: React.Dispatch<React.SetStateAction<any>>;
+}
+export const HostExpirationField: React.FC<HostExpirationFieldProps> = ({
+  onChange,
+  data,
+}) => {
+  const { expiration: expirationString, noExpiration } = data;
+  const expiration = new Date(expirationString);
+  const updateDate = (d: Date) => {
+    const year = d.getFullYear();
```
john-m-liu, line 26: I think you need null handling in both places. I'm not exactly sure what setup I used to cause this, but I got a runtime error when d is undefined by clicking the x to clear the date

khelif96: thanks i was able to replicate this I removed the option to clear the date entirely the expiration should be specified by default or no expiration should be set.

john-m-liu (approved):

PR #400, opened 2020-09-14
EVG-12868: Give ability to turn on and off trend charts on task pages…

zamj:
… in Spruce

### DIFF
--- a/config/.cmdrc_sample.json
+++ b/config/.cmdrc_sample.json
@@ -1,5 +1,6 @@
 {
   "devServer": {
+    "REACT_APP_TREND_CHARTS_ENABLED": "true",
     "REACT_APP_GQL_URL": "http://localhost:9090/graphql/query",
     "REACT_APP_API_URL": "http://localhost:3000/api",
     "REACT_APP_UI_URL": "http://localhost:9090",
@@ -10,10 +11,12 @@
     "REACT_APP_ENABLE_GQL_MOCK_SERVER": "true"
   },
   "staging": {
+    "REACT_APP_TREND_CHARTS_ENABLED": "false",
     "REACT_APP_API_URL": "https://www.example.com",
     "REACT_APP_UI_URL": "https://www.example.com"
   },
   "prod": {
+    "REACT_APP_TREND_CHARTS_ENABLED": "false",
     "REACT_APP_DEPLOYS_EMAIL": "some-email@something.com", 
     "REACT_APP_SITE_URL": "https://www.example.com",
     "REACT_APP_BUGSNAG_API_KEY": "this-is-the-api-key",
--- a/README.md
+++ b/README.md
@@ -47,16 +47,19 @@ Follow these directions to enable query linting during local development so your
 ```js
 {
   "devServer": {
+    "REACT_APP_TREND_CHARTS_ENABLED": "true",
     "REACT_APP_GQL_URL": "http://localhost:9090/graphql/query",
     "REACT_APP_API_URL": "http://localhost:3000/api",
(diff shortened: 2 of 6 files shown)


ancostas:
Hi all reviewers! For context, is this the right way to config things in Spruce? We are basically just adding a feature flag here -- we looked for where Spruce gets config from evergreen but weren't really sure where that is happening so have done it here for now.

zamj:
> Hi all reviewers! For context, is this the right way to config things in Spruce? We are basically just adding a feature flag here -- we looked for where Spruce gets config from evergreen but weren't really sure where that is happening so have done it here for now.

In addition, we've had some interest from teams that aren't currently using our performance infrastructure that they might like to start using it. Turning trend charts on and off globally makes their onboarding flow much easier than the old style where teams had to opt in manually.

bsamek:
> In addition, we've had some interest from teams that aren't currently using our performance infrastructure that they might like to start using it. Turning trend charts on and off globally makes their onboarding flow much easier than the old style where teams had to opt in manually.

What about teams that we don't want to display trend charts for?

config/.cmdrc_sample.json
```diff
@@ -10,10 +11,12 @@
     "REACT_APP_ENABLE_GQL_MOCK_SERVER": "true"
   },
   "staging": {
+    "REACT_APP_TREND_CHARTS_ENABLED": "false",
     "REACT_APP_API_URL": "https://www.example.com",
     "REACT_APP_UI_URL": "https://www.example.com"
   },
   "prod": {
+    "REACT_APP_TREND_CHARTS_ENABLED": "false",
```
bsamek, line 19: Although a feature flag is fine, I don't think that's the intent of EVG-12868. Currently the admin page lets Evergreen administrators turn on and off the trend charts for specific projects.

bsamek (changes requested):

zamj:
> > In addition, we've had some interest from teams that aren't currently using our performance infrastructure that they might like to start using it. Turning trend charts on and off globally makes their onboarding flow much easier than the old style where teams had to opt in manually.
> 
> What about teams that we don't want to display trend charts for?

The Trend Charts tab will still be there but our thinking was we'd either have a message to the effect of "No data found for this project" or whatever data we had for this project was displayed there. If people aren't interested in Trend Charts but they are sending perf data for whatever reason, they won't have to see the trend charts since it is in that separate tab. 

We can chat offline real quick if you'd like and hammer out the expected workflow.

ancostas:
@bsamek Relating to the evergreen admin page, do you know where this can be read in this app if we set it through the old UI? Regardless where we set the thing, I think the real pain point here is we don't know how to read from the evergreen admin settings in Spruce, although it seems like it should be pretty simple?

bsamek:
So, there are two issues here, I think, and we should keep them separate:
1. Should we always show a trend charts tab even if there is no trend chart data? I think the answer there is no, since expanded metrics is an external service. Evergreen treats it (correctly, I think) as a plugin. I think therefore it should only exist if someone has opted in.
2. How can you access the admin data in Spruce. This is something that someone who has worked on the Spruce code base can help you with. The best place to reach out might be #evergreen-spruce.

john-m-liu (changes requested):
The way you've configured it here, the list of projects must be saved locally, and even then you can't change it unless you redeploy spruce. You should instead query the admin config like in https://github.com/evergreen-ci/evergreen/blob/de36105b383fa56f8cf49dfd129a307a20ee8b68/graphql/resolvers.go#L1343

zamj:
Sure, we'll go ahead and go the more fine grained route. Closing for now

PR #401, opened 2020-09-15
EVG-12958 Create User tags field  for edit spawn host modal

khelif96:
This pr creates the user tags portion of the edit spawn host modal.

This UI only shows the user editable portion of a hosts instance tags.
Users are able to create new tags, delete tags, and update existing tags
They can update a tags value and only the tag itself will change. But if they update the key, we treat it as if they created a new tag and delete the old key and replace it with a new one.
usertags


src/pages/spawn/spawnHost/fields/UserTagsField.tsx
```diff
@@ -0,0 +1,87 @@
...
+  data: UserTagsFieldStateType;
+  onChange: React.Dispatch<React.SetStateAction<any>>;
+  instanceTags: InstanceTag[];
+}
+export const UserTagsField: React.FC<UserTagsFieldProps> = ({
+  data,
+  onChange,
+  instanceTags,
+}) => {
+  const userTags = instanceTags.filter((tag) => tag.canBeModified);
```
SupaJoon, line 22: i get this `TypeError: Cannot read property 'filter' of null` when I open the edit modal with evergreen running on the master branch locally

SupaJoon (changes requested):

src/pages/spawn/spawnHost/EditSpawnHostModal.tsx
```diff
@@ -160,8 +183,8 @@ const FlexContainer = styled.div`
 `;
 
 const SectionContainer = styled(FlexContainer)`
-  align-items: center;
-  margin-top: 15px;
+  <<<<<<<head=======align-items: center;
```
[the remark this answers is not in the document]
khelif96: merge conflict :(

SupaJoon (changes requested):
We should create an "+ Add Instance Tag" that generates a new input input for the user to edit instead of having a dedicated input row

src/pages/spawn/spawnHost/fields/UserTagsField.tsx
```diff
@@ -0,0 +1,87 @@
...
+  onChange: React.Dispatch<React.SetStateAction<any>>;
+  instanceTags: InstanceTag[];
+}
+export const UserTagsField: React.FC<UserTagsFieldProps> = ({
+  data,
+  onChange,
+  instanceTags,
+}) => {
+  const userTags = instanceTags?.filter((tag) => tag.canBeModified);
+  const [visibleTags, setVisibleTags] = useState(userTags);
```
SupaJoon, line 23: visibleTags will not be updated on subsequent polled myHosts queries

khelif96: Dont think this is necessarily an issue since this is a user edited field and likely wont be affected by polling since the user has to manually update this field anyways.

src/pages/spawn/spawnHost/fields/UserTagsField.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+export interface UserTagsFieldStateType {
+  addedInstanceTags?: InstanceTag[];
+  deletedInstanceTags?: InstanceTag[];
+}
+
+interface UserTagsFieldProps {
+  data: UserTagsFieldStateType;
+  onChange: React.Dispatch<React.SetStateAction<any>>;
+  instanceTags: InstanceTag[];
```
john-m-liu, line 15: I think you need a handler to reset the state, so that the cancel button on the modal can use it right?

src/pages/spawn/spawnHost/fields/userTagsField/UserTagRow.tsx
```diff
@@ -0,0 +1,142 @@
...
+                data-cy="user-tag-value-field"
+              />
+            </Section>
+          </FlexColumnContainer>
+
+          <IconButtonContainer>
+            {hasEditedTag ? (
+              <IconButton variant="light" aria-label="Edit Tag">
+                <Icon
+                  glyph="Edit"
```
john-m-liu, line 87: I don't understand this icon choice. This can display for new tags, which if you click it will add the tag. For that scenario, I think we want an icon that implies "confirm" or "accept," like a checkmark. For editing an existing tag, you can already edit it by typing in the box, so I would think we also want an icon that implies "confirm" don't we?

src/pages/spawn/spawnHost/fields/userTagsField/UserTagRow.tsx
```diff
@@ -0,0 +1,142 @@
...
+  return (
+    <FlexColumnContainer
+      data-cy={shouldShowTagInput ? "user-tag-row" : "add-tag-button-row"}
+    >
+      {shouldShowTagInput && (
+        <FlexContainer>
+          <FlexColumnContainer>
+            <Section>
+              <InputLabel htmlFor={`tag_key_${tagId}`}>Key</InputLabel>
+              <Input
```
john-m-liu, line 63: How do you feel about having a keypress handler for both these inputs that will call the create/update handler if you hit enter? I think it might feel natural to expect that enter does that

john-m-liu (changes requested):

src/pages/spawn/spawnHost/fields/userTagsField/UserTagRow.tsx
khelif96: Good idea

src/pages/spawn/spawnHost/fields/userTagsField/UserTagRow.tsx
khelif96: nice suggestion. Just added

SupaJoon (approved):
Do you think we should disable the save button until the user has confirmed their tags by clicking the checkmark next to the input? I have a feeling some users may unknowingly throw away their new tag by clicking save before confirming their tag by clicking the checkmark.

SupaJoon (changes requested):
I noticed two bugs:
1) sometimes it takes more than one trash btn click to delete a tag
2) a new tag is created when supplying a key for the initial tag. To replicate: a) create a new tag b) enter a value but do not enter a key c) save d) enter a value

src/pages/spawn/spawnHost/fields/UserTagsField.tsx
john-m-liu: It doesn't seem to reset the state completely. If you have an unsaved row, canceling then going back in leaves the row there

khelif96:
> I noticed two bugs:
> 
> 1. sometimes it takes more than one trash btn click to delete a tag
> 2. a new tag is created when supplying a key for the initial tag. To replicate: a) create a new tag b) enter a value but do not enter a key c) save d) enter a value

1. Cant figure out why the click fails occasionally my best guess is it might be leafy green related. Since this is a leafygreen component. I tried doing some logging and when the click fails go through the onclick handler is never even called. So i dont think its related to anything im doing. I dont think this is to big of an issue worst case the user might have to click the button one more time. I will bring it up with the leafygreen team however to see if they know why its failing.

2. I was able to eliminate this by just doing some input validation. And not allowing empty inputs.

SupaJoon (approved):

src/pages/spawn/spawnHost/fields/5-userTagsField.stories.tsx
```diff
@@ -0,0 +1,42 @@
...
+    value: "idk",
+    canBeModified: false,
+  },
+];
+export const PublicKeyFormView = () => {
+  const [state, setState] = useState<UserTagsFieldStateType>({
+    addedInstanceTags: [],
+    deletedInstanceTags: [],
+  });
+  console.log({ state });
```
john-m-liu, line 30: debug line here

john-m-liu (approved):

src/pages/spawn/spawnHost/fields/5-userTagsField.stories.tsx
khelif96: yup this is a story so i think its fine to have this here since im using this to confirm it produces the correct state

PR #419, opened 2020-09-23
EVG-13082: task link in task queue item goes to legacy UI for mainline commits

tgrander:
https://jira.mongodb.org/browse/EVG-13082

Currently, the Spruce workflow is only officially supported for patch builds. However, the links to the patch details from the Task Queue page lead to Spruce both for patch builds and for mainline commits.

Task link should go to legacy UI if task type is `commit`

### DIFF
--- a/src/pages/taskQueue/TaskQueueTable.tsx
+++ b/src/pages/taskQueue/TaskQueueTable.tsx
@@ -72,18 +72,30 @@ export const TaskQueueTable = () => {
       key: "displayName",
       className: "cy-task-queue-col-task",
       width: "30%",
-      render: (_, { displayName, id, project, buildVariant }) => (
+      render: (_, { displayName, id, project, buildVariant, requester }) => (
         <TaskCell>
           <Body>
-            <StyledRouterLink
-              data-cy="current-task-link"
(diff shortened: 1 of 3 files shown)


khelif96 (approved):
That commit history tho.

tgrander:
> That commit history tho.

i know right!!! trying to squash now

src/pages/taskQueue/TaskQueueTable.tsx
```diff
@@ -72,18 +72,30 @@ export const TaskQueueTable = () => {
...
+                onClick={() =>
+                  taskQueueAnalytics.sendEvent({ name: "Click Task Link" })
+                }
+              >
+                {displayName}
+              </StyledRouterLink>
+            ) : (
+              <StyledLink
+                data-cy="current-task-link"
+                href={`${getUiUrl()}/task/${id}`}
```
SupaJoon, line 91: we should make this a shared url generator since we use the same URL structure here: https://github.com/evergreen-ci/spruce/blob/master/src/pages/host/Metadata.tsx#L29

SupaJoon (approved):

PR #421, opened 2020-09-24
EVG-12407 Use error banner for configure patch page

khelif96:
This pr adds the error banner back. But removes the test for it since the apollo on error callback flakes in cypress


SupaJoon (approved):

khelif96:
Merged failing test was a flake thats failing on all branches

PR #423, opened 2020-09-24
EVG-8050 Add OOM Tracker info to the new task page

malikchaya2:


src/pages/task/Metadata.tsx
```diff
@@ -183,6 +186,12 @@ export const Metadata: React.FC<Props> = ({ loading, data, error, taskId }) => {
           ))}
         </span>
       ) : null}
+      {oomTracker && oomTracker.detected && (
+        <P2>
+          Out of Memory Kill detected
+          {oomTracker.pids ? `(PIDs: ${oomTracker.pids.join(", ")}` : ""} )
```
SupaJoon, line 192: `oomTracker.pids` should actually be `oomTracker?.pids` since `oomTracker` could be undefined based on the assignment in line 54

SupaJoon, line 192: oomTracker.pids should actually be oomTracker?.pids since oomTracker could be undefined based on the assignment in line 54

SupaJoon (changes requested):

khelif96 (approved):
Apart from arjuns comment lgtm

src/pages/task/Metadata.tsx
```diff
@@ -183,6 +186,12 @@ export const Metadata: React.FC<Props> = ({ loading, data, error, taskId }) => {
           ))}
         </span>
       ) : null}
+      {oomTracker && oomTracker.detected && (
+        <P2>
```
khelif96, line 190: Check with design but maybe this should look different to emphasis that this event happened

malikchaya2: I reached out to design. Thanks!

src/pages/task/Metadata.tsx
malikchaya2: doesn't the check on 189 make sure that oomTracker is defined by the time it reaches line 192?

SupaJoon: you're right!!

SupaJoon (approved):

PR #426, opened 2020-09-25
EVG-12988  bb Jira URL should be queried from the db since it is configurable

malikchaya2:


khelif96 (changes requested):
There Seems like theres a ton of prop drilling to get the jira host to the component that its needed in. I dont necessarily feel that this url should be included in the build baron query since its so general, and we likely would reuse it again. I think It might make more sense to include it as apart of the spruceConfig query https://github.com/evergreen-ci/evergreen/blob/master/graphql/schema.graphql#L734

john-m-liu (approved):

SupaJoon:
We can use the jira host by querying the SpruceConfig: https://github.com/evergreen-ci/evergreen/blob/master/graphql/schema.graphql#L735
https://github.com/evergreen-ci/evergreen/blob/master/graphql/schema.graphql#L735

khelif96 (approved):

PR #427, opened 2020-09-25
EVG-13062 Wire up mutation for spawn host status update buttons

khelif96:
hostaction


john-m-liu (approved):

src/pages/spawn/spawnHost/SpawnHostActionButton.tsx
```diff
@@ -1,23 +1,110 @@
...
-  ) : null;
+  // When the UPDATE_SPAWN_HOST_STATUS mutation occurs the host state is not immediately updated, It gets updated a few seconds later.
+  // Since the GET_MY_HOSTS query on this components parent polls at a slower rate, this component triggers a poll at a faster interval for that
+  // query when it returns an updated host status the polling is halted. This allows the query to poll slowly and not utilize unnecessary bandwith
+  // except when an action is performed and we need to fetch updated data.
+  const [getMyHosts, { startPolling, stopPolling }] = useLazyQuery<
+    MyHostsQuery,
+    MyHostsQueryVariables
+  >(GET_MY_HOSTS, {
+    pollInterval: 1000,
```
SupaJoon, line 34: do we also need to add `fetchPolicy: "network-only"`. I wasn't able to find anything in the apollo docs about what `fetchPolicy` defaults when a pollInterval is set.

src/pages/spawn/spawnHost/SpawnHostActionButton.tsx
```diff
@@ -1,23 +1,110 @@
...
+    },
+  });
+  useNetworkStatus(startPolling, stopPolling);
+
+  // Stop polling when we get updated host data
+  useEffect(() => {
+    if (stopPolling) {
+      stopPolling();
+    }
+  }, [host]); // eslint-disable-line react-hooks/exhaustive-deps
```
SupaJoon, line 48: this seems to work because the MyHost query in SpawnHost.tsx only creates a new reference to data and data.myHosts.hosts when the data fields change. 
Do you think it makes sense diff `host.id` and `host.status` instead of the entire host object so we can be more specific about when to stop polling? 
This could save us from an edgecase where we stop polling prematurely after we update two host states and stop polling for both hosts as soon as the references inside the `data.myHost.hosts` change once.

SupaJoon (approved):

src/pages/spawn/spawnHost/SpawnHostActionButton.tsx
khelif96: It most likely goes to network only since there would be no point to poll from  from the cache.

src/pages/spawn/spawnHost/SpawnHostActionButton.tsx
khelif96: I dont think this can happen. When we render each row in the spawn host table we use the host id as the row key. By doing this react knows to update only the row with the matching host. So there shouldn't be an issue from host ids changing and effecting other rows and causing other components props to change.

 We pass the individual host into each spawn Host action button and each one performs its own query and has its own hook to manage polling. Therefore the stopPolling call of one hook would not affect another component.

PR #429, opened 2020-09-25
EVG-13109 Only query the build baron once the execution is retrieved

malikchaya2:


src/pages/Task.tsx
```diff
@@ -130,6 +126,9 @@ const TaskCore: React.FC = () => {
 
   usePageTitle(`Task${displayName ? ` - ${displayName}` : ""}`);
 
+  const [failedTask, setFailedTask] = useState(undefined);
+  const [showBuildBaronTab, setshowbuildbarontab] = useState(!failedTask);
```
SupaJoon, line 130: I think we originally included `showBuildBaronTab` to conditionally render all of the leafygreen tabs at the same time. This helped us avoid a bug in leafygreen where we were unable to remove or add tabs under the Tab provider after the Tab provider's initial render. 
I think leafygreen patched that bug and we no longer need `showBuildBaronTab`.

src/pages/Task.tsx
```diff
@@ -108,6 +99,11 @@ const TaskCore: React.FC = () => {
       dispatchBanner.errorBanner(
         `There was an error loading the task: ${err.message}`
       ),
+    onCompleted: () => {
+      setFailedTask(
+        task?.status?.includes("failed") || task?.status?.includes("timed-out")
```
SupaJoon, line 104: this can be clearer if we directly compare the task status against [task status constants defined in spruce](https://github.com/evergreen-ci/spruce/blob/master/src/types/task.ts#L49)

SupaJoon: I also don't think this will work as expected since onCompleted is creating a closure around the `task` variable from line 109. You would want to look at the data object passed in a the first argument to `onCompleted`. https://www.apollographql.com/docs/react/data/queries/#options

src/pages/Task.tsx
```diff
@@ -275,14 +252,12 @@ const TaskCore: React.FC = () => {
...
                 </Tab>
-                {showBuildBaronTab ? (
-                  <Tab name="Build Baron" id="task-build-baron-tab">
-                    <BuildBaron
-                      data={buildBaronData}
-                      error={buildBaronError}
-                      taskId={id}
-                    />
-                  </Tab>
+                {failedTask && execution !== undefined ? (
```
SupaJoon, line 255: I don't think you need to save failedTask in state since it is derived from the most recent data from `GetTaskQuery`

SupaJoon (changes requested):

src/components/Buildbaron/BuildBaronTab.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { useQuery } from "@apollo/client";
+import { Tab } from "@leafygreen-ui/tabs";
+import BuildBaron from "components/Buildbaron/BuildBaron";
+import { BuildBaronQuery, BuildBaronQueryVariables } from "gql/generated/types";
+import { GET_BUILD_BARON } from "gql/queries";
+
+interface Props {
+  taskId: string;
+  execution: number;
+  setshowbuildbarontab: React.Dispatch<React.SetStateAction<Boolean>>;
```
khelif96, line 11: camel case please.

khelif96: I also dont think this needs to be passed in. The component can handle rendering itself.

src/pages/Task.tsx
khelif96: Yup i think this can be removed

khelif96 (changes requested):

src/pages/Task.tsx
malikchaya2: I did that since it is undefined at first, and I want it to re-render once it becomes defined. Is that not necessary?

src/components/Buildbaron/BuildBaronTab.tsx
malikchaya2: I need it to rerender when it changes.

src/pages/Task.tsx
SupaJoon: the component is expected to rerender when the value for `data` is updated so you can access fields from the `data` object directly. If you log `task?.status?.includes("failed") || task?.status?.includes("timed-out")` or render it as an html in your jsx, you should see that the component is rendering the latest values of `data` so you can avoid saving them state.

src/components/Buildbaron/BuildBaronTab.tsx
```diff
@@ -0,0 +1,48 @@
...
+export const BuildBaronTab: React.FC<Props> = ({
+  taskId,
+  execution,
+  setshowbuildbarontab,
+}) => {
+  const {
+    data: buildBaronData,
+    error: buildBaronError,
+    loading: buildBaronLoading,
+  } = useQuery<BuildBaronQuery, BuildBaronQueryVariables>(GET_BUILD_BARON, {
```
SupaJoon, line 23: the execution number can be changed from the task page from the execution select dropdown
Screen Shot 2020-09-29 at 12 20 38 PM
We should refetch build baron data within a `useEffect` when the execution number changes

SupaJoon (changes requested):

src/components/Buildbaron/BuildBaronTab.tsx
SupaJoon: that setState will actually cause a render in the parent component by accessing a data from a request that was dispatched from the child. I think this variable and the useState hook that it originated from can be removed because leafygreen can now handle tabs being conditionally mounted/unmounted

src/components/Buildbaron/BuildBaronTab.tsx
```diff
@@ -0,0 +1,49 @@
...
+    loading: buildBaronLoading,
+    refetch,
+  } = useQuery<BuildBaronQuery, BuildBaronQueryVariables>(GET_BUILD_BARON, {
+    variables: { taskId, execution },
+  });
+
+  const prevExecution = usePrevious(execution);
+  if (execution !== prevExecution) {
+    refetch();
+  }
```
khelif96, line 27: I dont think this is necessary if you just place execution in a useeffect hook it should do the same thing.

```
useEffect(() => refetch(), [execution])
```

src/components/Buildbaron/BuildBaronTab.tsx
```diff
@@ -0,0 +1,49 @@
...
+      <Tab name="Build Baron" id="task-build-baron-tab">
+        <BuildBaron
+          data={buildBaronData}
+          error={buildBaronError}
+          taskId={taskId}
+        />
+      </Tab>
+    );
+  }
+  return null;
```
khelif96, line 48: `return 
showTab ? <Tab ... /> : null`

src/pages/Task.tsx
```diff
@@ -127,6 +119,10 @@ const TaskCore: React.FC = () => {
   const logLinks = get(task, "logs");
   const isPerfPluginEnabled = false;
   const patchAuthor = data?.task.patchMetadata.author;
+  const failedTask =
+    task?.status?.includes(TaskStatus.Failed) ||
+    task?.status?.includes(TaskStatus.TaskTimedOut) ||
+    task?.status?.includes(TaskStatus.TestTimedOut);
```
khelif96, line 126: I dont think using  `.includes()` is necessary. Its better to just do a direct comparison. Using the .includes() will search the string for the result which is unnecessarily expensive. 
`task.status === TaskStatus.Failed || ....`

khelif96 (changes requested):
Spoke offline, Gonna leave some suggestions lmk when your ready for another review.

src/components/Buildbaron/BuildBaronTab.tsx
```diff
@@ -0,0 +1,49 @@
...
+    error: buildBaronError,
+    loading: buildBaronLoading,
+    refetch,
+  } = useQuery<BuildBaronQuery, BuildBaronQueryVariables>(GET_BUILD_BARON, {
+    variables: { taskId, execution },
+  });
+
+  const prevExecution = usePrevious(execution);
+  if (execution !== prevExecution) {
+    refetch();
```
SupaJoon, line 26: we need to add in the variables: `refetch({ taskId, execution })`

src/components/Buildbaron/BuildBaronVariables.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const {
+    data: buildBaronData,
+    error: buildBaronError,
+    loading: buildBaronLoading,
+    refetch,
+  } = useQuery<BuildBaronQuery, BuildBaronQueryVariables>(GET_BUILD_BARON, {
+    variables: { taskId, execution },
+  });
+
+  useEffect(() => {
```
SupaJoon, line 30: after testing this locally, I see that the useQuery on line 26 will create request new data when the taskId and execution variables on line 27 change. If we convert this function component to a hook, we will actually have to replace `useQuery` with `useLazyQuery` and call refetch with the taskId & execution number from props when taskFailed is true and execution is defined.

src/components/Buildbaron/BuildBaronVariables.tsx
```diff
@@ -0,0 +1,47 @@
...
+    refetch();
+  }, [execution, refetch]);
+
+  const buildBaron = buildBaronData?.buildBaron;
+  const buildBaronConfigured = buildBaron?.buildBaronConfigured;
+
+  // logic for displaying the build baron tab
+  const buildBaronIsProductionReady = true;
```
SupaJoon, line 38: I think we should keep this flag off until there are some tests for the BuildBaron Spruce code.

src/pages/Task.tsx
```diff
@@ -185,6 +164,15 @@ const TaskCore: React.FC = () => {
         banners={bannersState}
         removeBanner={dispatchBanner.removeBanner}
       />
+      {failedTask && execution !== undefined ? (
+        <BuildBaronVariables
```
SupaJoon, line 168: This function component can be converted into a react hook that returns `{showBuildBaronTab, buildBaronData, buildBaronError}`. In order to do that, you would have to handle this case `failedTask && execution !== undefined` within the hook because hooks cannot be conditionally rendered. 
A technicality that will come up when trying to do this is that we do not want to make buildBaron queries for failed tasks or when the execution is undefined. We can circumvent that by importing `useLazyQuery` instead of `useQuery` and only fetching/refetching when the props failedTask and execution are valid (true, defined).

SupaJoon (changes requested):

SupaJoon:
These work well but I think it can be improved a lot if we convert the BuildBaronVariables to a hook called useBuildBaronVariables. This would allow us to avoid returning `null` in the BuildBaronFunction component and allow us to interact with hook return values directly instead of setting state in the parent component. The main challenge to converting this component to a hook would be to handle the cases when taskFailed is false or execution is undefined inside the hook. That would require use to import useLazyQuery instead of useQuery and only fetch data when those values are valid.

src/components/Buildbaron/BuildBaronVariables.tsx
malikchaya2: ah yes, this should be false.

SupaJoon (approved):

khelif96 (approved):
LGTM 🚀 

Just 1 suggestion

PR #433, opened 2020-09-28
EVG-12901: Upgrade to Cypress 5 and enable test retries

SupaJoon:
https://jira.mongodb.org/browse/EVG-12901
I am open to suggestions for the retry count. It is possible to specify a retry count for specific tests and test suites. https://docs.cypress.io/guides/guides/test-retries.html#Global-Configuration
This PR also unmutes tests we marked as flakey

### DIFF
--- a/cypress/integration/patch/configure-patch.ts
+++ b/cypress/integration/patch/configure-patch.ts
@@ -382,7 +382,7 @@ describe("Configure Patch Page", () => {
         "eq",
         `/patch/${unactivatedPatchId}/configure/tasks`
       );
-      cy.get("[data-cy=error-banner]").should("exist");
+      cy.get("[data-cy=banner]").contains("WAH WAH CHICKEN WAH");
     });
   });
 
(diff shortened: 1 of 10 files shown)


cypress/integration/patch/configure-patch.ts
```diff
@@ -365,8 +365,7 @@ describe("Configure Patch Page", () => {
       );
     });
 
-    // TODO: Reactivate this test when were remove xhr dep on tests
-    xit("Shows error banner if unsuccessful and keeps data", () => {
+    it("Shows error banner if unsuccessful and keeps data", () => {
```
khelif96, line 368: This test wont ever work it depends on removing the xhr graphql intercept code from our test suite

SupaJoon: this test seems to work well in cypress 5. I can remove it if you think it's misleading

khelif96 (approved):

PR #436, opened 2020-09-30
EVG-12941 Do not display hostName and ssh Command when there is no HostUr

malikchaya2:


src/pages/Host.tsx
```diff
@@ -79,7 +80,7 @@ export const HostCore: React.FC = () => {
     boolean
   >(false);
 
-  usePageTitle(`Host${hostUrl ? ` - ${hostUrl}` : ""}`);
+  usePageTitle(`Host${hostId ? ` - ${hostId}` : ""}`);
```
SupaJoon, line 83: should we show `host?.displayName` when available?

src/pages/Host.tsx
```diff
@@ -90,7 +91,7 @@ export const HostCore: React.FC = () => {
       {host && (
         <>
           <PageTitle
-            title={`Host: ${hostUrl}`}
+            title={`Host: ${hostId}`}
```
SupaJoon, line 94: same comment about showing display name

SupaJoon (approved):

khelif96 (approved):

src/pages/Host.tsx
malikchaya2: I don't think it's necessary because it doesn't seem to be popular.

PR #447, opened 2020-10-01
EVG-12710 Performance improvements and refactoring for some spruce pages.

khelif96:
Changes in this PR by commit hash:
419917b Specify the key apollos in memory cache should use for the user query. Apollo was silently throwing a warning about this since it didnt know what key to use for the user query since there is no id supplied. (The key/id is used to identify the query in the inMemory cache)

0b6a0c6 Let the `useDisableTableSortersIfLoading` hook take in just the loading status as an input. We dont need to depend on the network status. Since we only care if the query is in the middle of loading.

8f2f262 Refactor most spruce pages to remove the dependence on the usePollQuery hook during my analysis of the performance on spruce. I noticed this hook had a performance hit on the pages it was used on, since it was creating several timers to manually trigger apollo refetches. It also added unnecessary complexity to the pages it was on. I was able to accomplish the same behavior with out it, by just reorganizing where/ how some variables are declared and using apollos built in polling functionality.

5113ad2 Some tests relied on a gql query being made when an action is performed. But this is no longer always the case. As part of the performance improvements we now rely on the cache more. When actions happen in quick succession such as filtering the results of a table. Apollo now checks the cache for an appropriate response before going to the network. If this is the first time an action is performed apollo goes to the network, but if the action has been performed before we can pull the results from the cache and display the results instantly.

The biggest performance improvement for large patches will come in a future pr/ ticket after design updates how build variants should be displayed.


tgrander (approved):
all of the instances of `refetch` were removed from the gql queries and everything still works as far as i can tell. does refetch not need to be called for the query to update with the new params?

khelif96:
> all of the instances of `refetch` were removed from the gql queries and everything still works as far as i can tell. does refetch not need to be called for the query to update with the new params?

No it doesn't! I don't think its documented well in the apollo docs but the query operates like most other hooks. If the input is changed it re executes the hook with the latest input. Previously the query variables were stored and frozen in state. (`initialQueryVariables`). So the useQuery hook was unable to tell that anything changed, and required you to use refetch.

khelif96:
@tgrander  I think this line in the apollo source code is responsible for triggering an update when new options are passed in. https://github.com/apollographql/apollo-client/blob/8f70d39dbb238ecfa626aca0d22355d31e1cb088/src/react/hooks/utils/useBaseQuery.ts#L26

src/pages/patch/patchTabs/Tasks.tsx
```diff
@@ -46,45 +46,36 @@ export const Tasks: React.FC<Props> = ({ taskCount }) => {
...
-    getQueryVariables,
-    refetch,
-    search,
-    isOffline,
-  });
+  useDisableTableSortersIfLoading(loading);
 
+  let showSkeleton = true;
+  if (data) {
+    showSkeleton = false;
```
SupaJoon, line 72: this means that the user will only see a loading skeleton when the table initially loads. I think this could lead to a smoother experience and we should do this for the other tables as well

src/pages/task/testsTable/TestsTableCore.tsx
```diff
@@ -45,38 +42,25 @@ export const TestsTableCore: React.FC = () => {
...
-  // poll task tests
-  const isOffline = useNetworkStatus();
-  const { showSkeleton } = usePollQuery({
-    networkStatus,
-    getQueryVariables,
-    refetch,
-    search,
-    isOffline,
-  });
+  useDisableTableSortersIfLoading(loading);
```
SupaJoon, line 63: `loading` won't represent loading states from polled queries unless we set `notifyOnNetworkStatusChange: true` in the query options. We should add this setting to disable the sorters whenever a table request is in flight.

src/pages/patch/patchTabs/Tasks.tsx
```diff
@@ -28,7 +29,6 @@ import {
 import { GET_PATCH_TASKS } from "gql/queries/get-patch-tasks";
 import {
   useDisableTableSortersIfLoading,
-  usePollQuery,
```
SupaJoon, line 31: a feature of usePollQuery was to refetch data from the db whenever table params changed. since we rely on the apollo cache, we should only allow cached data to exist as long the pollInterval. This will ensure that any table page we navigate to will be fresh data from within the last minute (or whatever the pollInterval is set to)

SupaJoon (changes requested):

src/pages/patch/patchTabs/Tasks.tsx
khelif96: This was actually not intended and was a typo on my part. The behavior is the same with or without it.

src/pages/task/testsTable/TestsTableCore.tsx
khelif96: I dont think we should do this. Polling happens in the back ground and if there is a change in the data it would update the UI to reflect that. If there is no change to the data the ui remains the same. 

Examining this hook i dont think this condition is ever hit/used. I slowed down the network request to watch what happens on the page. When we load new data a loading bar is displayed since its the first time its been requested. Which i think is fine since its in response to a user action. 
page swap

But when we poll in the background i don't think it should be noticeable to the user, We shouldn't indicate that to the user in anyway apart from updating the data on the screen if there are changes. `notifyOnNetworkStatusChange` Would visibly notify the user that something was loaded even tho nothing may have changed on screen.

src/pages/patch/patchTabs/Tasks.tsx
```diff
@@ -46,45 +42,30 @@ export const Tasks: React.FC<Props> = ({ taskCount }) => {
...
   );
 
-  const { data, error, networkStatus, refetch } = useQuery<
+  const { data, error, loading, startPolling, stopPolling } = useQuery<
     PatchTasksQuery,
     PatchTasksQueryVariables
   >(GET_PATCH_TASKS, {
-    variables: initialQueryVariables as PatchTasksQueryVariables,
-    notifyOnNetworkStatusChange: true,
-    fetchPolicy: "network-only",
```
SupaJoon, line 67: we should change this to `cache-and-network` since the default fetchPolicy is `cache-first`

src/pages/UserPatches.tsx
```diff
@@ -61,22 +61,15 @@ const UserPatchesComponent: React.FC = () => {
...
   );
 
-  const { data, refetch, networkStatus, error } = useQuery<
+  const { data, loading, startPolling, stopPolling, error } = useQuery<
     UserPatchesQuery,
     UserPatchesQueryVariables
   >(GET_USER_PATCHES, {
-    variables: initialQueryVariables,
-    notifyOnNetworkStatusChange: true,
-    fetchPolicy: "network-only",
```
SupaJoon, line 70: same comment about fetchPolicy

src/pages/task/testsTable/TestsTableCore.tsx
```diff
@@ -45,38 +42,23 @@ export const TestsTableCore: React.FC = () => {
...
 
   // initial request for task tests
-  const { data, refetch, networkStatus } = useQuery<
+  const { data, loading, startPolling, stopPolling } = useQuery<
     TaskTestsQuery,
     TaskTestsQueryVariables
   >(GET_TASK_TESTS, {
-    variables: initialQueryVariables,
-    notifyOnNetworkStatusChange: true,
-    fetchPolicy: "network-only",
```
SupaJoon, line 66: same comment about fetchPolicy

SupaJoon (changes requested):

SupaJoon (approved):

PR #451, opened 2020-10-06
EVG-12870: Add trend charts to Spruce task page

zamj:
There's still some work to be done before Trend Charts are complete so we're still going to leave it off for now. When the final feature PR is in, we'll have to add the prod url to the settings

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,7 @@ Follow these directions to enable query linting during local development so your
 ```js
 {
   "devServer": {
+    "REACT_APP_SIGNAL_PROCESSING_URL": "https://performance-monitoring-and-analysis.server-tig.staging.corp.mongodb.com",
     "REACT_APP_GQL_URL": "http://localhost:9090/graphql/query",
     "REACT_APP_API_URL": "http://localhost:3000/api",
     "REACT_APP_UI_URL": "http://localhost:9090",
@@ -57,6 +58,7 @@ Follow these directions to enable query linting during local development so your
     "REACT_APP_UI_URL": "https://evergreen-staging.corp.mongodb.com"
   },
   "prod": {
+    "REACT_APP_SIGNAL_PROCESSING_URL": "https://performance-monitoring-and-analysis.server-tig.prod.corp.mongodb.com",
     "REACT_APP_DEPLOYS_EMAIL": "something@something.com", 
     "REACT_APP_SITE_URL": "https://spruce.mongodb.com",
     "REACT_APP_BUGSNAG_API_KEY": "this-is-the-api-key",
--- a/config/.cmdrc_sample.json
+++ b/config/.cmdrc_sample.json
@@ -1,5 +1,6 @@
 {
   "devServer": {
+    "REACT_APP_SIGNAL_PROCESSING_URL": "https://performance-monitoring-and-analysis.server-tig.staging.corp.mongodb.com",
     "REACT_APP_GQL_URL": "http://localhost:9090/graphql/query",
     "REACT_APP_API_URL": "http://localhost:3000/api",
     "REACT_APP_UI_URL": "http://localhost:9090",
@@ -14,6 +15,7 @@
     "REACT_APP_UI_URL": "https://www.example.com"
   },
   "prod": {
+    "REACT_APP_SIGNAL_PROCESSING_URL": "https://performance-monitoring-and-analysis.server-tig.prod.corp.mongodb.com",
     "REACT_APP_DEPLOYS_EMAIL": "some-email@something.com",
     "REACT_APP_SITE_URL": "https://www.example.com",
     "REACT_APP_BUGSNAG_API_KEY": "this-is-the-api-key",
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "@leafygreen-ui/text-input": "1.1.1",
     "@leafygreen-ui/toggle": "3.0.1",
     "@leafygreen-ui/typography": "1.0.1",
+    "@mongodb-dev-prod/trend-charts-ui": "~0.3.1",
     "ansi_up": "4.0.4",
     "antd": "4.5.4",
     "antd-table-infinity": "1.1.6",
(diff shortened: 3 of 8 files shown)


README.md
```diff
@@ -57,6 +58,7 @@ Follow these directions to enable query linting during local development so your
     "REACT_APP_UI_URL": "https://evergreen-staging.corp.mongodb.com"
   },
   "prod": {
+    "REACT_APP_SIGNAL_PROCESSING_URL": "https://www.example.com",
```
ancostas, line 61: should this be prod?

config/.cmdrc_sample.json
```diff
@@ -14,6 +15,7 @@
     "REACT_APP_UI_URL": "https://www.example.com"
   },
   "prod": {
+    "REACT_APP_SIGNAL_PROCESSING_URL": "https://www.example.com",
```
ancostas, line 18: prod?

zamj:
@khelif96 Material ui has been removed from the trend-charts in favor of antd

package.json
```diff
@@ -24,6 +24,7 @@
     "@leafygreen-ui/text-input": "1.1.1",
     "@leafygreen-ui/toggle": "3.0.1",
     "@leafygreen-ui/typography": "1.0.1",
+    "@mongodb-dev-prod/trend-charts-ui": "^0.2.3",
```
SupaJoon, line 27: we should remove the `^` so we can load a previous version of trend-charts-ui and redeploy if needed by only making changes to spruce instead of spruce and trend-charts-ui

khelif96:
The compile and e2e test tasks are OOMing.

package.json
ancostas: This will coerce lockstep releases with spruce, right? Any patch update to the trend charts ui will have us coming here to update this line. If you want to load and redeploy a previous version by only changing spruce, it seems like this would be the line to edit at that point. On our end it would be convenient to just iterate major/minor/patch versions with some care, and this UI to consume updates as appropriate, e.g. fixed major/minor version and latest patch.

zamj:
> The compile and e2e test tasks are OOMing.

Fixed now

package.json
zamj: @SupaJoon @khelif96 Any thoughts on this? We'd prefer to not do lockstep releases so that we're not having to do two separate pr's whenever we make a minor change on the trend charts. If this is a hard requirement, that's fine though

SupaJoon: It's not a hard requirement for us to exclude the caret and I am for excluding the caret if it can benefit your workflow. 
Something to consider if we exclude the caret is that there isn't an easy way to tell which version of trend-charts-ui is deployed in spruce by looking at the build/dist files (you can see the output in the `build` folder by running `npm run build` in spruce). The most convenient way to understand which version of `trend-charts-ui` is deployed in Spruce would be to see what was available in the trend-charts-ui npm module at the time of the most recent spruce deploy.

zamj: That should be fine. We can do `~` instead if that makes you more comfortable so that we're only automatically updating on patch updates. That way we don't have to bother yall with bug fixes or small changes but are marking the times when we're making larger changes.

src/pages/Task.tsx
```diff
@@ -127,6 +125,16 @@ const TaskCore: React.FC = () => {
     taskStatus: task?.status,
   });
 
+  if (isPerfPluginEnabled) {
```
SupaJoon, line 128: when both plugins are enabled, the tabToIndexMap contains non-integer numbers and values repeat across the keys: `{"logs":0,"tests":1,"files":2,"trend-charts":3.5,"build-baron":3.5}`. I

package.json
```diff
@@ -51,8 +52,8 @@
     "waait": "1.0.5"
   },
   "scripts": {
-    "start": "react-scripts start",
-    "build": "REACT_APP_GIT_SHA=`git rev-parse HEAD` react-scripts build",
+    "start": "react-scripts --max_old_space_size=8196 start",
+    "build": "REACT_APP_GIT_SHA=`git rev-parse HEAD` react-scripts --max_old_space_size=8196 build",
```
khelif96, line 56: Do you guys know why we are running into this in the first place. Im curious what is causing the build process to OOM in the first place, that we need to increase this limit

zamj: It's probably the chart library that we use. We faced a similar issue in our repo.
https://github.com/plotly/react-plotly.js/issues/135

src/pages/Task.tsx
zamj: Whoops, nice catch! Should be fixed now and handle any combination of them being enabled

package.json
zamj: @khelif96 I went ahead and used the smallest partial bundle provided by that plotting library we could instead of what we had before. We still need a higher memory limit during building but it reduced the final bundle size, in our repo at least, by a lot.

package.json
SupaJoon: @zamj sounds good to me!

package.json
khelif96: Is there anything you can do to make the plotly library smaller. This plugin including the plotly library is larger then spruce, and all our deps combined. It causes spruces bundle to be twice as large. 
image

vs 
image

You might have already done this but if not build a custom version of plotly using only the modules you need. https://github.com/plotly/plotly.js/#modules

khelif96: Btw this chart can be viewed by looking at the files tab on the compile task

tgrander: Building a custom version of plotly sounds time consuming. It is known that graph charts tend to be huge. If their module is code split from the rest of the app then this should not be a problem because the performance on their page should be the concern of their team. also - if you are requesting that they build a custom version then benchmarks for the loading times with and without the plotly library would be helpful if not necessary. For example, if the load time with plotly lib is under a certain acceptable threshold, e.g. 2000ms, then there is no need to waste time over optimizing.

khelif96:
Good to go after we sort out the tabs not rendering properly

zamj:
> Good to go after we sort out the tabs not rendering properly

Should be good to try now

khelif96 (approved):

khelif96:
Thanks for all of this!

khelif96:
The compile task is still OOMing I think you accidently removed the flag to allocate it more memory

zamj:
> The compile task is still OOMing I think you accidently removed the flag to allocate it more memory

Whoops, good catch!

PR #452, opened 2020-10-06
EVG-12549: Improve Restart Version Modal performance

SupaJoon:
https://jira.mongodb.org/browse/EVG-12549
this PR virtualizes the list of task checkboxes and adds an event listener to the checkbox list container instead of each checkbox.

to see the performance improvement, compare restarting all tasks for localhost:3000/version/5ed2bbae9ccd4e6de99b9976/tasks & https://spruce.mongodb.com/version/5ed2bbae9ccd4e6de99b9976/tasks


src/components/PatchActionButtons/RestartPatch.tsx
```diff
@@ -36,7 +36,7 @@ export const RestartPatch = forwardRef<HTMLDivElement, RestartPatchProps>(
           </DropdownItem>
         )}
         {openModal && (
-          <div className={openModal && "ant-popover-open"} ref={ref}>
+          <div className={openModal ? "ant-popover-open" : ""} ref={ref}>
```
khelif96, line 39: I know you didn't add this class, but do you know what it does? I couldn't find anything about it from a search in the antd docs. And it doesn't look like a style is being applied to the element. Could we remove this?

src/pages/patch/patchRestartModal/PatchStatusCheckboxContainer.tsx
```diff
@@ -13,20 +13,44 @@ export const PatchStatusCheckboxContainer: React.FC<PatchStatusCheckboxContainer
...
-      />
-    ))}
-  </ScrollableContainer>
-);
+}) => {
+  const possibleListHeight = tasks.length * itemSize;
+  const listHeight =
+    possibleListHeight < maxListHeight ? possibleListHeight : maxListHeight;
+  const toggleHandler = (e) => {
+    // @ts-ignore
```
khelif96, line 21: Is this necessary? I dont see any typescript errors.

src/pages/patch/patchRestartModal/PatchStatusCheckboxContainer.tsx
```diff
@@ -13,20 +13,44 @@ export const PatchStatusCheckboxContainer: React.FC<PatchStatusCheckboxContainer
...
+    // eslint-disable-next-line jsx-a11y/no-static-element-interactions,jsx-a11y/click-events-have-key-events
+    <div data-cy="patch-status-selector-container" onClick={toggleHandler}>
+      <FixedSizeList
+        height={listHeight}
+        itemSize={itemSize}
+        itemCount={tasks.length}
+        itemData={tasks}
+      >
+        {({ style, data, index }) => {
+          const { id: taskId, status, name: displayName } = data[index];
```
khelif96, line 37: Could we just pass the task into `<TaskStatusCheckbox />` and let it use the task to render the info it needs?

src/pages/patch/patchRestartModal/PatchStatusCheckboxContainer.tsx
```diff
@@ -13,20 +13,44 @@ export const PatchStatusCheckboxContainer: React.FC<PatchStatusCheckboxContainer
...
+  const toggleHandler = (e) => {
+    // @ts-ignore
+    const { name } = e.target;
+    if (selectedTasks[name] !== undefined) {
+      toggleSelectedTask(name);
+    }
+  };
+  return (
+    // eslint-disable-next-line jsx-a11y/no-static-element-interactions,jsx-a11y/click-events-have-key-events
+    <div data-cy="patch-status-selector-container" onClick={toggleHandler}>
```
khelif96, line 29: Out of curiosity how does this click handler work? Is `<FixedSizeList/>` bubbling up the click event for each item in the virtualized list?

khelif96 (changes requested):
This is great! theres a significantly noticeable performance boost. I did notice some weird behavior and had a few questions.
weirdClicks

src/pages/patch/patchRestartModal/PatchStatusCheckboxContainer.tsx
khelif96: Is there an advantage to doing it this way the docs seem to place the click handler on the actual rows https://react-window.now.sh/#/examples/list/memoized-list-items

src/components/PatchActionButtons/RestartPatch.tsx
SupaJoon: the className is added when `openModal` is true so the parent component can see when the modal is open by checking `ref.current.className` instead of listening to changes on `openModal` variable. https://github.com/evergreen-ci/spruce/blob/master/src/pages/userPatches/patchCard/DropdownMenu.tsx#L33
I tried removing the className and the modal closed on click so I think we need it for now.

src/pages/patch/patchRestartModal/PatchStatusCheckboxContainer.tsx
SupaJoon: TaskStatusCheckbox is wrapped in `React.memo` which will memoize the render result by shallow comparing the props. Destructuring non-reference values from `data[index]` helps ensure the component only renders once when taskId, status or displayName change as opposed to rendering when the reference to data[index] changes.

src/pages/patch/patchRestartModal/PatchStatusCheckboxContainer.tsx
SupaJoon: Children of `<FixedSizeList />`  bubble their events to the `div` wrapper. 
This change helps the performance because we are attaching only one event listener to the dom instead of many.

SupaJoon:
@khelif96 thanks for pointing out the issue with clicking parent checkboxes. I see that the bug existed the code changes in this PR and created: https://jira.mongodb.org/browse/EVG-13192

src/components/Modal.tsx
```diff
@@ -19,15 +20,16 @@ export const Modal: React.FC<ModalProps> = ({
   onCancel,
   visible,
   children,
+  onOk,
   "data-cy": dataCy,
 }) => (
-  // Placing a stop propogation event handler around this modal to prevent click events from bubbling up to their parent component
   // eslint-disable-next-line
```
khelif96, line 26: I know this isn't from this change but can you give a specific rule to disable?

src/pages/patch/patchRestartModal/TaskStatusCheckbox.tsx
```diff
@@ -2,30 +2,34 @@ import React from "react";
...
 
 interface TaskStatusCheckboxProps {
-  task: PatchBuildVariantTask;
-  selectedTasks: selectedStrings;
-  toggleSelectedTask: (id: string) => void;
+  displayName: string;
+  status: string;
+  taskId: string;
+  checked: boolean;
+  style: React.CSSProperties;
```
khelif96, line 11: Would you mind adding a comment to this explaining why the style prop is here and that its used by react window. I can see us coming back to this in the future and wondering what it does.

khelif96 (approved):
Great job. This is significantly faster! then the old modal

PR #454, opened 2020-10-07
EVG-13083 Adding an new key now resets the public key form state

khelif96:


src/pages/spawn/spawnHost/spawnHostModal/PublicKeyForm.tsx
```diff
@@ -48,6 +48,8 @@ export const PublicKeyForm: React.FC<PublicKeyFormProps> = ({
           style={{ width: 200 }}
           placeholder="Select existing key"
           onChange={selectPublicKey}
+          value={keyName}
```
SupaJoon, line 51: we should also add `disabled={shouldAddNewKey}` to prevent the user from selecting an existing key while adding a new. I attached a video of the edgecase adding this prop can prevent.
new name equals old name

SupaJoon (changes requested):

SupaJoon (approved):

PR #455, opened 2020-10-07
EVG-13179 Replace build variant task squares with grouped task squares

khelif96:
image

The build variant grid now groups tasks by status. 
Clicking on a status will filter the tasks on the patch page based on the status you selected.
Clicking on a build variant will filter based on the variant

Dont merge until Monday!


src/pages/patch/buildVariants/GroupedTaskSquare.tsx
```diff
@@ -0,0 +1,63 @@
...
+        patchAnalytics.sendEvent({
+          name: "Click Grouped Task Square",
+          taskSquareStatus: status,
+        })
+      }
+    >
+      <Tooltip
+        title={
+          <span data-cy="task-square-tooltip">
+            {count} {count > 1 ? "task's" : "task"} with status {status}
```
SupaJoon, line 36: typo: we should remove the apostrophe

src/pages/patch/BuildVariants.tsx
```diff
@@ -35,27 +38,61 @@ export const BuildVariants: React.FC = () => {
...
+            data-cy="patch-build-variant"
           >
-            <P1>{displayName}</P1>
-            <VariantTasks>
-              {tasks.map((task) => (
-                <TaskSquare key={task.id} {...task} />
-              ))}
-            </VariantTasks>
+            <P1>
+              <Link to={`${getVersionRoute(id)}?page=0&variant=${variant}`}>
```
SupaJoon, line 47: we should analytics to this link as well

SupaJoon (approved):

src/pages/patch/buildVariants/GroupedTaskSquare.tsx
```diff
@@ -0,0 +1,63 @@
...
+        <TaskSquare status={status}>{count}</TaskSquare>
+      </Tooltip>
+    </StyledRouterLink>
+  );
+};
+
+interface TaskSquareProps {
+  status: string;
+}
+const TaskSquare = styled.div<TaskSquareProps>`
```
john-m-liu, line 49: Could we call this a GroupedTaskSquare or something to avoid confusion with the old individual task squares?

john-m-liu (changes requested):

src/pages/patch/buildVariants/GroupedTaskSquare.tsx
khelif96: The parent component is called GroupedTaskSquare. This is just the styled component, I think eventually we will delete the Other old Individual Task Squares if we decide to go with this approach.

src/pages/patch/BuildVariants.tsx
```diff
@@ -35,27 +38,61 @@ export const BuildVariants: React.FC = () => {
...
+
 const BuildVariant = styled.div`
   margin-bottom: 8px;
 `;
 const VariantTasks = styled.div`
   display: flex;
   flex-wrap: wrap;
 `;
+
+const groupTasksByStatus = (tasks: PatchBuildVariantTask[]) => {
```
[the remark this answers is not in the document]
khelif96: I added tooltips that describe the status of each square. You are also able to click on a square to filter just by the status. It might be confusing if 1 square now relates to 3 different statuses

src/pages/patch/BuildVariants.tsx
```diff
@@ -35,27 +45,86 @@ export const BuildVariants: React.FC = () => {
...
   display: flex;
   flex-wrap: wrap;
 `;
+
+const groupTasksByColor = (tasks: PatchBuildVariantTask[]) => {
+  const taskColors: {
+    [key: string]: { count: number; statuses: string[]; textColor: string };
+  } = {};
+  tasks.forEach((task) => {
+    if (taskColors[mapVariantTaskStatusToColor[task.status]]) {
```
john-m-liu, line 109: I would save mapVariantTaskStatusToColor[task.status] in a separate variable so that this is easier to read

src/pages/patch/buildVariants/GroupedTaskSquare.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const { id } = useParams<{ id: string }>();
+
+  const nextQueryParams = queryString.stringify(
+    {
+      statuses,
+      page: 0,
+    },
+    { arrayFormat }
+  );
+  const status = "test";
```
john-m-liu, line 33: This looks mostly unused now, but I think we now want to send the color in the event right?

john-m-liu (approved):

src/pages/patch/buildVariants/GroupedTaskSquare.tsx
khelif96: Nice catch!

tgrander (approved):

tgrander (commented):
there's some complex logic around sorting the tasks by status but there are no tests that assert that logic results in the correct value

khelif96:
> there's some complex logic around sorting the tasks by status but there are no tests that assert that logic results in the correct value

Good point! Just wrote some!

PR #460, opened 2020-10-09
EVG-13102 Part 1 Migrate over to yarn

khelif96:
The first step in setting up a better plugin infrastructure is to migrate our package manager on spruce from npm to yarn.
Yarn has a few advantages over npm
Yarn is faster than npm when it comes to installing packages, because it parallelizes install threads and also caches dependencies so fetching deps will be faster. 

Yarn also enables easier development of packages since it has better support for monorepos and workspaces. https://classic.yarnpkg.com/blog/2017/08/02/introducing-workspaces/


john-m-liu (commented):
I checked this pr out from a fresh folder and things worked flawlessly. Have you looked at the test failure though? Looks like everything is having problems logging in, though I can't see how it would be affected by this

khelif96:
> I checked this pr out from a fresh folder and things worked flawlessly. Have you looked at the test failure though? Looks like everything is having problems logging in, though I can't see how it would be affected by this

@john-m-liu  Yeah the problem is related to serving up a built version of spruce instead of running the dev server. There is no reverse proxy for serving the built version so we can't route the requests to the correct backend.
I did this for 2 reasons, 
1) It makes sense to e2e test on a built version of the app since that is what we actually serve to the users.
2) The dev server seems to be breaking a file watcher limit on CI since it has to watch every single file inside of spruce. 

I think i can fix the reverse proxy issue by writing a server to serve up the built version of spruce in CI

john-m-liu:
What back end is it using and where is that configured? From the videos you get the login page from the old UI, which likely means that you're getting the login redirect from a running evergreen server somewhere

khelif96:
This pr will work once the file watcher limit is increased on the ubuntu 1604 distro, This should allow the dev server to run properly. 
 https://jira.mongodb.org/browse/BUILD-12215

tgrander (approved):

SupaJoon (approved):
we should include something about yarn in our ReadMe

khelif96:
> we should include something about yarn in our ReadMe

I've replaced all references of npm to yarn in our readme. For all intents and purposes yarn should behave the same way to the end user.

PR #461, opened 2020-10-12
EVG-1256: Spawn Volume Modal (part 1)

SupaJoon:
https://jira.mongodb.org/browse/EVG-12586

These code changes render the Spawn Volume Modal along with selectors for volume size and availability zone. There will be jest tests to verify the shape of the request and button states. I adjusted some form labels as well as the title to make the modal consistent with the Spawn Host Modal. 

Screen Shot 2020-10-12 at 2 13 08 PM


john-m-liu (approved):
This is part of the antd table, but it's kind of weird. For the expand row renderer, the component looks like it's actually rendered as a new row rather than part of the row you expanded, so you get a confusing border and zebra striping. Do you know if there's a way to have it be part of the row you expanded?

Screen Shot 2020-10-12 at 1 59 56 PM

tgrander (approved):

SupaJoon:
> This is part of the antd table, but it's kind of weird. For the expand row renderer, the component looks like it's actually rendered as a new row rather than part of the row you expanded, so you get a confusing border and zebra striping. Do you know if there's a way to have it be part of the row you expanded?
> 
> Screen Shot 2020-10-12 at 1 59 56 PM
> 
> Disregard this if there isn't an easy way to do it or if you think it's fine

I will make sure these code changes are included in part 2.

PR #462, opened 2020-10-12
EVG-13188 Refactor Edit Spawn Host modal to use reducer

khelif96:
I wanted to start adopting this useReducer pattern a little bit more when it comes to complex state updates. It lets us be more explicit and deliberate about state changes.

### DIFF
--- a/src/pages/spawn/spawnHost/editSpawnHostModal/useEditSpawnHostModalState.ts
+++ b/src/pages/spawn/spawnHost/editSpawnHostModal/useEditSpawnHostModalState.ts
@@ -1,0 +1,67 @@
+import { useReducer } from "react";
+import { InstanceTag } from "gql/generated/types";
+import { MyHost } from "types/spawn";
+
+interface editSpawnHostState {
+  expiration?: Date;
+  noExpiration: boolean;
+  displayName?: string;
+  instanceType?: string;
+  volumeId?: string;
+  addedInstanceTags?: InstanceTag[];
+  deletedInstanceTags?: InstanceTag[];
+}
+
+export const useEditSpawnHostModalState = (host: MyHost) => ({
+  reducer: useReducer(reducer, host, init),
+  defaultEditSpawnHostState: init(host),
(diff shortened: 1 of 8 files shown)


src/pages/spawn/spawnHost/fields/UserTagsField.tsx
```diff
@@ -10,13 +10,11 @@ export interface UserTagsFieldStateType {
 }
 
 interface UserTagsFieldProps {
-  data: UserTagsFieldStateType;
   onChange: React.Dispatch<React.SetStateAction<any>>;
```
SupaJoon, line 13: This prop changed types

src/pages/spawn/spawnHost/editSpawnHostModal/useEditSpawnHostModalState.ts
```diff
@@ -0,0 +1,67 @@
...
+  reducer: useReducer(reducer, host, init),
+  defaultEditSpawnHostState: init(host),
+});
+
+const init = (host: MyHost) => ({
+  displayName: host.displayName,
+  expiration: host.expiration,
+  noExpiration: host.noExpiration,
+  instanceType: host.instanceType,
+  volumeId: "",
```
SupaJoon, line 25: I think we need to set the volumeId to what exists in host. Same with the instance tags so the user can the initial values.

SupaJoon (approved):

src/pages/spawn/spawnHost/editSpawnHostModal/useEditSpawnHostModalState.ts
SupaJoon: It looks like we don't actually have to do the same thing with instance tags because the UserTagsField component takes the instance tags from the `MyHost` as the initial display

khelif96: Yup

PR #463, opened 2020-10-12
EVG-13188 Wire up edit spawn host mutation

khelif96:


SupaJoon (approved):
I noticed that attaching a volume through the modal doesn't work as expected because of the ec2Manager.ModifyHost does not handle the volume input. I create a ticket for it [here](https://jira.mongodb.org/browse/EVG-13220). Aside from that, I was able to update all of the fields the modal allows.

tgrander (approved):

PR #464, opened 2020-10-13
EVG-8181 preserve query part of url on task page

john-m-liu:
The previous commit on this ticket exposed an existing bug where we would replace the URL search on the task page. Also removed a test that is no longer relevant now that the default tab is dynamic

This should fix the e2e tests


SupaJoon (approved):
Thank you!

PR #466, opened 2020-10-13
EVG-13179 Update build variant design and allow for filtering by variant on status

khelif96:
Further changes on the build variant grid. The squares are now rectangular and smaller. 
And clicking on a status filters for that specific variant 
image

### DIFF
--- a/src/hooks/useFilterInputChangeHandler.ts
+++ b/src/hooks/useFilterInputChangeHandler.ts
@@ -1,4 +1,4 @@
-import { useState } from "react";
+import { useState, useEffect } from "react";
 import debounce from "lodash.debounce";
 import queryString from "query-string";
 import { useLocation, useHistory } from "react-router-dom";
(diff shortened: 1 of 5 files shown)


src/pages/patch/buildVariants/GroupedTaskSquare.tsx
```diff
@@ -12,20 +12,23 @@ interface Props {
...
   textColor,
+  variant,
 }) => {
   const patchAnalytics = usePatchAnalytics();
   const { id } = useParams<{ id: string }>();
 
   const nextQueryParams = queryString.stringify(
     {
       statuses,
+      variant,
```
SupaJoon, line 31: i think we should expand and overwrite query params in case the patch page relies on additional query params in the future

SupaJoon (approved):

src/hooks/useFilterInputChangeHandler.ts
```diff
@@ -32,8 +32,13 @@ export const useFilterInputChangeHandler = (
 
   const parsed = queryString.parse(search, { arrayFormat });
```
tgrander, line 33: this should use the util

tgrander (commented):
looks like there should be more left and right padding for variant squares with 4 digit values. the numbers almost overflow

malikchaya2 (approved):

tgrander (approved):

SupaJoon (approved):

PR #471, opened 2020-10-14
EVG-12549: Additional Patch Restart modal improvements

SupaJoon:
https://jira.mongodb.org/browse/EVG-12549
These code changes prevent the modal from being unmounted/remounted on modal close & open and also fixes an issue where the checkbox label wraps to the next line when the container width is decreased.


src/pages/patch/patchRestartModal/TaskStatusCheckbox.tsx
```diff
@@ -44,4 +44,7 @@ const StateItemWrapper = styled("div")`
   display: flex;
   flex-direction: row;
   align-items: center;
+  white-space: nowrap;
+  overflow: hidden;
```
khelif96, line 48: Does overflow: scroll work? Just in case some one really wanted to view the full task name.

src/components/PatchActionButtons/RestartPatch.tsx
```diff
@@ -9,47 +9,56 @@ interface RestartPatchProps {
...
+  patchId,
+  refetchQueries,
+  hideMenu,
+  visibilityControl,
+}) => {
+  const fallbackVisibilityControl = useState(false);
+  const [isVisible, setIsVisible] =
+    visibilityControl !== undefined
+      ? visibilityControl
+      : fallbackVisibilityControl;
```
khelif96, line 27: I'm not totally convinced we should have two methods of controlling the state of this component. What would be the advantage of doing it this way vs only one method?

SupaJoon: Ideally, the RestartPatch should be the only component to control the visibility of the modal because that way we do not need to declare a useState to manage to visibility whenever we import RestartPatch (we are able to avoid that here because the prop is optional: https://github.com/evergreen-ci/spruce/blob/master/src/pages/patch/ActionButtons.tsx#L89).

src/pages/patch/patchRestartModal/TaskStatusCheckbox.tsx
SupaJoon: After removing this line, I see that it is not necessary to set overflow on the checkbox row because the checkbox container will dynamically adjust scrolling horizontally when the labels are long enough. 
scrolling gif

khelif96 (approved):

PR #472, opened 2020-10-14
EVG-13230: Clean up useEditSpawnHostState reducer types and make HostExpirationFields a shared component

SupaJoon:
https://jira.mongodb.org/browse/EVG-13230
These code changes allow us to be more specific about information processed in our reducer functions. It also enables us to only ever have expand the host state object in our reducer function instead of within our React function components. A possible next step could be to incorporate useReducer for our Spawn Host Modal.


khelif96 (approved):

khelif96:
Merging so it can be used in an upcoming pr

PR #485, opened 2020-10-23
EVG-12587: Edit Volume Modal

SupaJoon:
https://jira.mongodb.org/browse/EVG-12587
The best way to test these code changes is to point your local development server to production, spawn a volume, and then attempt to change it's settings by clicking the edit button in the table row. 
Screen Shot 2020-10-23 at 2 05 52 PM


khelif96 (approved):

src/pages/spawn/spawnVolume/spawnVolumeTable/spawnVolumeTableActions/editButton/EditVolumeModal.tsx
```diff
@@ -0,0 +1,128 @@
...
+    spawnAnalytics.sendEvent({
+      name: "Edited a Spawn Volume",
+      params: variables,
+    });
+    updateVolumeMutation({
+      variables,
+    });
+  };
+
+  const noChange =
```
john-m-liu, line 78: Does it make sense to make a custom hook that checks if the state is dirty? Maybe one that takes in an initial state and a current state, and does a deep compare of them? I imagine the pattern we're doing here could be generalized and used in a couple places we have forms

john-m-liu (commented):
Looks good, just had 1 question

tgrander (approved):

src/pages/spawn/spawnVolume/spawnVolumeTable/spawnVolumeTableActions/editButton/EditVolumeModal.tsx
SupaJoon: we actually have access to an [isEqual function](https://lodash.com/docs/4.17.15#isEqual) that does a deep compare for objects. Using the function could have cleaned up some code here but the reason we couldn't use it was because the mutation variables require a Date obj for `expiration` where as the the `expiration` field in the Volume gql type actually returns as a string from MyVolumes query.

PR #506, opened 2020-11-02
EVG-12880 display patch params on patch page

ablack12:
The modal comes from the Patch Parameters link on the left.

image

Evergreen PR: https://github.com/evergreen-ci/evergreen/pull/4119

My cypress test is being weird so stay tuned


tgrander:
how come it show up in da corner like that?

ablack12:
> how come it show up in da corner like that?

Oh sorry, it is in the center of my screen, I just took a partial screenshot.

cypress/integration/patch-route.ts
```diff
@@ -45,6 +45,13 @@ describe("Patch route", () => {
     cy.location("pathname").should("eq", "/commit-queue/evergreen");
   });
 
+  it("Shows patch parameters if they exist", () => {
```
tgrander, line 48: what happens if they do not exist?

package.json
```diff
@@ -34,6 +34,7 @@
     "deep-object-diff": "1.1.0",
     "env-cmd": "10.1.0",
     "graphql": "15.3.0",
+    "graphql-tag": "^2.11.0",
```
tgrander, line 37: hm?

src/pages/patch/ParametersModal.tsx
```diff
@@ -0,0 +1,41 @@
...
+  visible,
+}) => (
+  <Modal
+    visible={visible}
+    footer={null}
+    onCancel={closeModal}
+    data-cy={dataCy}
+    title="Patch Parameters"
+  >
+    <span>
```
tgrander, line 26: is this span necessary?

src/pages/patch/ParametersModal.tsx
```diff
@@ -0,0 +1,41 @@
...
+      {parameters?.map(
+        (param) =>
+          param && (
+            <StyledBadge key={`param_${param?.key}`}>
+              {param?.key}:{param?.value}
```
tgrander, line 31: is there a scenario where `parameters` will exist, as will `param`, but the `key` and `value` fields are null or undefined? i ask because this code indicates that a StyledBadge can be rendered with an undefined, i.e. blank, `key` and `value`

cypress/integration/patch-route.ts
ablack12: If they don't exist we show no module. Should I add a test for that?

package.json
ablack12: 🤷  should this not be here? Is there something else I should do? I just did `npm install`

src/pages/patch/ParametersModal.tsx
ablack12: maybe just a failure of copy/paste haha.

src/pages/patch/ParametersModal.tsx
ablack12: key has to be defined. value can be the empty string, but it's still turned into a pointer so I suppose it also can't be undefined. I assumed we used ?. for everything just to be safe but I can change them both if that's preferable

package.json
ablack12: Before it wasn't compiling because of something to do with graphql-tag but 🤷  seems fine now so I'll delete it.

src/pages/patch/ParametersModal.tsx
```diff
@@ -0,0 +1,41 @@
...
+            </StyledBadge>
+          )
```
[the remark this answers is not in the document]
ablack12: Oh gotcha!

src/pages/patch/Metadata.tsx
```diff
@@ -53,6 +56,22 @@ export const Metadata: React.FC<Props> = ({ loading, patch, error }) => {
...
+          </StyledLink>
+        </P2>
+      )}
+      <ParametersModal
+        dataCy="parameters-modal"
+        parameters={parameters}
+        visible={showModal}
+        closeModal={() => setShowModal(false)}
+      />
     </MetadataCard>
```
khelif96, line 75: I think this should be moved to its own component. Like we do for some of the other modals. Instead of embedding it in the Metadata card component.

khelif96: The state logic for the modal and the link that it shows by default should live in the ParmatersModal File.

src/types/patch.ts
```diff
@@ -11,6 +11,10 @@ export interface ModuleCodeChanges {
   rawLink: string;
   fileDiffs: FileDiff[];
 }
+export interface Parameter {
+  key: string;
+  value: string;
+}
```
khelif96, line 17: We dont need to declare this type again we can just pull the type from the generated types. See below comment

src/pages/patch/ParametersModal.tsx
```diff
@@ -0,0 +1,36 @@
+import React from "react";
+import styled from "@emotion/styled";
+import Badge from "@leafygreen-ui/badge";
+import { Modal } from "components/Modal";
+import { Parameter } from "types/patch";
```
khelif96, line 5: This is being generated and exported from our generated Types

`import {Parameter} from "gql/generated/types"`

khelif96 (changes requested):
You should also remove the changes in package-lock.json

src/types/patch.ts
ablack12: Do i also not need to declare Parameter in the Patch definition in this file?

src/pages/patch/ParametersModal.tsx
```diff
@@ -1,35 +1,50 @@
...
 interface ParametersProps {
   parameters: Parameter[];
   dataCy: string;
-  closeModal: () => void;
-  visible: boolean;
 }
+
 export const ParametersModal: React.FC<ParametersProps> = ({
   parameters,
   dataCy,
```
khelif96, line 16: nit. If we don't expect this component to be reused we can just declare the `data-cy` inline on line 36 instead of passing it in.

khelif96 (approved):
Great job! thanks for your working on this

PR #507, opened 2020-11-02
EVG-12870: Turn Trend Charts on

zamj:
Is there some place we need to add the new env variable so that this works in production?


SupaJoon:
> Is there some place we need to add the new env variable so that this works in production?

environment variables are added to `config/.cmdrc.json`. They are injected to the app when we generate the bundles locally as part of the build step.

khelif96:
> > Is there some place we need to add the new env variable so that this works in production?
> 
> environment variables are added to `config/.cmdrc.json`. They are injected to the app when we generate the bundles locally as part of the build step. https://github.com/evergreen-ci/spruce#environment-variables

There are also some other spots this needs to be added lets sync on it later

khelif96 (approved):

PR #509, opened 2020-11-02
EVG-13221: Only display the delete volume button when the volume is not mounted.

SupaJoon:
https://jira.mongodb.org/browse/EVG-13221

Deleting a volume while it is mounted doesn't return and deleting a volume moments moments after it is unmounted sometimes fails. This code changes only display the volume button when it is not mounted.


tgrander (approved):

SupaJoon:
I'm closing this PR in favor of displaying allowing users to unmount & delete with one click.

PR #510, opened 2020-11-03
EVG-13306 Enforce volume limit on spawn volume modal

khelif96:


SupaJoon (approved):

SupaJoon (changes requested):
the size selector component shows the adjusted value after creating a volume but it seems like the the reducer state is not updated when both the selector initially mounts with a value that exceeds it's limit. I can tell this is happening bc when I create a volume where the size limit - volume size > default reducer, refresh the page, and create another volume without adjusting the size selector, the created volume size will be the same as the default memory size of the modal reducer. 
weird state gif

SupaJoon (approved):
Works great!!

PR #512, opened 2020-11-03
EVG-13334: make table dropdown card caret clickable

SupaJoon:
https://jira.mongodb.org/browse/EVG-13334

### DIFF
--- a/src/components/Spawn/Layout.tsx
+++ b/src/components/Spawn/Layout.tsx
@@ -52,7 +52,10 @@ export const SpawnTable = (props: React.ComponentProps<typeof Table>) => (
               onClick={onClick}
               onKeyDown={onClick}
             >
-              <Icon glyph={expanded ? "CaretDown" : "CaretRight"} />
+              <Icon
+                data-cy={`table-caret-icon-${record.id}`}
+                glyph={expanded ? "CaretDown" : "CaretRight"}
+              />
             </span>
           );
         },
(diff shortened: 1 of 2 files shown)


src/components/Spawn/Layout.tsx
```diff
@@ -41,9 +41,24 @@ export const SpawnTable = (props: React.ComponentProps<typeof Table>) => (
...
+              onClick={onClick}
+              onKeyDown={onClick}
+            >
+              <Icon
+                data-cy={`table-caret-icon-${record.id}`}
+                glyph={expanded ? "CaretDown" : "CaretRight"}
+              />
+            </span>
+          );
+        },
```
khelif96, line 61: This can be simplified to.
```
        expandIcon: ({ expanded, onExpand, record }) => (
          <Icon
            onClick={(e: any) => {
              onExpand(record, e);
            }}
            glyph={expanded ? "CaretDown" : "CaretRight"}
          />
        ),
      }}
```

SupaJoon: I think it's important to be able to tab over over and open the caret. I was unable to achieve that with these code changes the props don't work as expected when applied to an svg. 
```
          <Icon
              tabIndex={0}
              role="button"
              onClick={onClick}
              onKeyDown={onClick}
              data-cy={`table-caret-icon-${record.id}`}
              glyph={expanded ? "CaretDown" : "CaretRight"}
            />
```

tabIndex is necessary to allow tab over the icon and onKeyDown is necessary to open the card when the user with a key press. The undesirable side affect from applying onKeyDown and and tabIndex is that the tab focus go past the svg icon.

khelif96 (approved):

PR #515, opened 2020-11-03
EVG-12443 add dropdown option to disable task

john-m-liu:
Untitled

- Add a patch/task dropdown option to disable the patch/task
- Make the patch dropdown options Sentence case to be consistent
- Enable + fix the task action tests. I'm not sure why they were disabled

Note that displaying the disabled status is a separate ticket: https://jira.mongodb.org/browse/EVG-13096

### DIFF
--- a/src/pages/patch/ActionButtons.tsx
+++ b/src/pages/patch/ActionButtons.tsx
@@ -1,6 +1,6 @@
 import React, { useRef, useState, useEffect } from "react";
 import { useParams } from "react-router-dom";
-import { ButtonDropdown } from "components/ButtonDropdown";
+import { ButtonDropdown, DropdownItem } from "components/ButtonDropdown";
 import { LinkToReconfigurePage } from "components/LinkToReconfigurePage";
 import {
   SchedulePatchTasks,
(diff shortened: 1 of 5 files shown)


src/pages/patch/ActionButtons.tsx
```diff
@@ -24,8 +32,21 @@ export const ActionButtons: React.FC<ActionButtonProps> = ({
   const wrapperRef = useRef(null);
   const [isVisible, setIsVisible] = useState(false);
   const { id: patchId } = useParams<{ id: string }>();
-
   const [isActionLoading, setIsActionLoading] = useState(false);
+  const { successBanner, errorBanner } = useBannerDispatchContext();
+  const refetchQueries = ["Patch"];
```
tgrander, line 37: i don't think this needs to be its own variable - it can be dynamically stated on like 48: `refetchQueries: ["Patch"]`. There is no perf gain from putting it as a variable outside of the mutation.

tgrander (commented):
the action text of the button in the dropdown "enable / disable" does not match the text of the success banner "priority was updated to -1". The success banner text explains to the user how disabling a task works under the hood, which is confusing and unnecessary. i think simply saying "task was disabled / endabled" is a better UX

SupaJoon:
> the action text of the button in the dropdown "enable / disable" does not match the text of the success banner "priority was updated to -1". The success banner text explains to the user how disabling a task works under the hood, which is confusing and unnecessary. i think simply saying "task was disabled / endabled" is a better UX

I think this is okay since it makes the relationship between enabling, disabling and setting priority more transparent. For example, a user is communicated that enabling will reset the assigned priority to 1 instead of keeping the priority they set previously with the priority input which could be important for them to know.

SupaJoon:
I'm unsure if this is intentional but the "Disable all tasks" button wasn't added to the patch card dropdown
Screen Shot 2020-11-05 at 11 50 20 AM

SupaJoon (approved):

src/pages/patch/ActionButtons.tsx
john-m-liu: fixed

john-m-liu (commented):
I made disable actions have a specific confirmation. Unfortunately it's more complicated for enabling a task, since changing the priority does exactly the same thing and it's something that you can do already

PR #518, opened 2020-11-06
EVG-8070 Auto populate email and slack fields in notification modals

khelif96:
Dependent on https://github.com/evergreen-ci/evergreen/pull/4145


SupaJoon (changes requested):
the slack username requires an @ prependend in the username to pass validation but auto population doesn't include it.

SupaJoon (approved):

PR #521, opened 2020-11-10
EVG-13323 Display link to JIRA search query when task annotations are populated with BB plugin results

malikchaya2:
image


src/components/Buildbaron/BuildBaronContent.tsx
```diff
@@ -31,6 +34,15 @@ const BuildBaronCore: React.FC<BuildBaronCoreProps> = ({
...
   }, [tab]); // eslint-disable-line react-hooks/exhaustive-deps
   const [createdTicketsCount, setCreatedTicketsCount] = useState<number>(0);
+
+  const { data } = useQuery<GetSpruceConfigQuery>(GET_SPRUCE_CONFIG);
+  const spruceConfig = data?.spruceConfig;
+  const jiraHost = spruceConfig?.jira?.host;
+
+  const jiraSearchString = eventData?.searchReturnInfo?.search;
+  const jqlEscaped = encodeURIComponent(jiraSearchString);
+  const jiraSearchLink = `https://${jiraHost}/secure/IssueNavigator.jspa?jql=${jqlEscaped}`;
```
khelif96, line 44: this should probably live in `constants/externalResources.ts`  https://github.com/evergreen-ci/spruce/blob/c305da71f8ef2c0aca13382efb70c3c0220c4654/src/constants/externalResources.ts#L4

malikchaya2: It's not really a constant, since it includes the jiraHost and the jira search string. I can technically move this part over: /secure/IssueNavigator.jspa?jql=$ , but I don't think I'd gain enough and it would probably feel clunky. This is not a very strong opinion though so if you prefer, I can move it over.

SupaJoon (approved):

src/components/Buildbaron/BuildBaronContent.tsx
malikchaya2: nvm, I'll model it after your getIdeUrl.

SupaJoon (changes requested):
Can we have a test to show the expected href is applied to the StyledLink in line 70?

SupaJoon (approved):

PR #522, opened 2020-11-10
EVG-8229: aborted task status badge & metadata link

tgrander:
### Tasks that were aborted will have the status badge `Aborted` in task table on patch details page

Screen Shot 2020-11-10 at 11 50 04 AM

### The details page for an aborted task will show `Aborted` status badge. The metadata panel will display the reason for the abort and a link to the failing task that caused the abort. 

Screen Shot 2020-11-10 at 11 51 13 AM

### Filter patch tasks by aborted
Screen Shot 2020-11-18 at 3 50 00 PM


src/pages/task/Metadata.tsx
```diff
@@ -186,6 +187,19 @@ export const Metadata: React.FC<Props> = ({ loading, data, error, taskId }) => {
             </StyledRouterLink>
           </P2>
         )}
+        {abortInfo && (
+          <P2>
+            {`Aborted by ${abortInfo.user} because of failing task: `}
+            <span>
```
khelif96, line 193: Is this wrapping span necessary?

khelif96 (approved):

src/components/TaskStatusBadge.tsx
```diff
@@ -72,46 +74,52 @@ const BadgeWidthMaxContent = styled(Badge)`
...
+  aborted,
 }) => {
-  // We have to do this assignment because Blocked is not an official Task
-  // status from the tasks collection but we want to represent it in the badge.
-  const adjustedTaskStatus = blocked
-    ? (TaskStatus.StatusBlocked as string)
-    : status;
+  // We have to do this assignment because Blocked and Aborted are not official Task
+  // statuses from the tasks collection but we want to represent it in the badge.
+  let taskStatus = status;
```
SupaJoon, line 86: a switch statement could make the assignment logic easier to understand

SupaJoon (approved):

john-m-liu (changes requested):
Adding a new status isn't this easy unfortunately. Since filtering/sorting of the task table is done on the server, the aborted status is not filterable and doesn't sort correctly. Also because aborted isn't actually a status, it'll display as a different status in the build section and on the waterfall when that's added. I think this needs more design work

src/components/TaskStatusBadge.tsx
tgrander: if there were more than two if statements then sure. but i think it's fine and readable as is.

tgrander:
> Adding a new status isn't this easy unfortunately. Since filtering/sorting of the task table is done on the server, the aborted status is not filterable and doesn't sort correctly. Also because aborted isn't actually a status, it'll display as a different status in the build section and on the waterfall when that's added. I think this needs more design work

I noted this while working on it as well but did not consider it a blocker. Still not sure that I do because this serves as a vanity status to differentiate failed tasks from aborted ones. I think this would be fine to add to spruce as is, especially since the users who requested this feature understand what aborted means in relation to failed. Follow up tickets to sort based on aborted and show aborted in build and on mainline commits can be additional work.

tgrander:
> Adding a new status isn't this easy unfortunately. Since filtering/sorting of the task table is done on the server, the aborted status is not filterable and doesn't sort correctly. Also because aborted isn't actually a status, it'll display as a different status in the build section and on the waterfall when that's added. I think this needs more design work

Additionally, I can easily showed failed as aborted on build section and on the mainline commits the same way I get failed to show as aborted in the task table and the task status badge

bsamek:
Maria is @mvankeulen94.

I think we all agree that the message on the metadata card is great, and we should keep it.

The open questions are all about the aborted status label:
1. What should the variant grid show?
2. What should the mainline commits page show?
3. Is it okay to merge a feature that doesn't allow sorting and filtering of aborted statuses separately from other failed statuses?

I think that #1 and #2 shouldn't block this PR, though they are definitely worth taking into account. For #3, I'd like @mvankeulen94 and/or design to weigh in. If we could signify somehow that aborted and failed are the same sort of category (have "failed/aborted" in the filter list), then this potentially would be fine. Since Maria has talked to a number of users about this, I'd like to hear more.

mvankeulen94:
I believe it will be necessary to allow filtering failed and aborted items separately. Users are generally more interested in the genuine failed statuses, so being able to focus on those somehow is necessary. Additionally, if these statuses currently display separately from one another in the task status grid sidebar, the expectation is that clicking on one type will filter to that one type. I think if all of these criteria are met, the sorting part is less important, as long as all failures and aborts appear in a single group.

tgrander:
> I believe it will be necessary to allow filtering failed and aborted items separately. Users are generally more interested in the genuine failed statuses, so being able to focus on those somehow is necessary. Additionally, if these statuses currently display separately from one another in the task status grid sidebar, the expectation is that clicking on one type will filter to that one type. I think if all of these criteria are met, the sorting part is less important, as long as all failures and aborts appear in a single group.

I'm not 100% sure what the requirements you lay out here mean. Could you please write up the exact feature expectations in [EVG-8229](https://jira.mongodb.org/browse/EVG-8229)? (or create a new ticket)

mvankeulen94:
I've added the following to the ticket: 'The aborted status should be filterable separately from other failed status types.'
I believe this is what users would expect from a new status type being added.
Regarding the sidebar--my understanding is that it's the current behavior as this patch stands to display aborted statuses separately in the task status grid sidebar. If that is not the case, then please disregard the separate task box part of my previous comment. If there is no separate status box for aborted tasks, then I believe sorting these as a group would become more important. But I am fine to leave sorting as a follow-up ticket as long as we permit filtering of aborted statuses separately from other types of failed statuses here.

tgrander:
> I've added the following to the ticket: 'The aborted status should be filterable separately from other failed status types.'
> I believe this is what users would expect from a new status type being added.
> Regarding the sidebar--my understanding is that it's the current behavior as this patch stands to display aborted statuses separately in the task status grid sidebar. If that is not the case, then please disregard the separate task box part of my previous comment. If there is no separate status box for aborted tasks, then I believe sorting these as a group would become more important. But I am fine to leave sorting as a follow-up ticket as long as we permit filtering of aborted statuses separately from other types of failed statuses here.

Thank you!!! Will work on it today.

tgrander:
Cannot merge until https://github.com/evergreen-ci/evergreen/pull/4167 is merged in evergreen

src/pages/task/Metadata.tsx
```diff
@@ -200,6 +202,19 @@ export const Metadata: React.FC<Props> = ({ loading, data, error, taskId }) => {
             </StyledRouterLink>
           </P2>
         )}
+        {abortInfo && (
+          <P2>
+            {`Aborted by ${abortInfo.user} because of failing task: `}
```
john-m-liu, line 207: I don't think this label is accurate. A task can be aborted by a user for any reason, or can be automatically aborted by the system for a couple of reasons. It'll only sometimes be because a different task in the same commit queue merge failed

john-m-liu (changes requested):

src/pages/task/Metadata.tsx
tgrander: Thanks for catching that. I added these two additional abort messages per the legacy UI. 
Screen Shot 2020-11-19 at 1 30 13 PM
Screen Shot 2020-11-19 at 1 30 25 PM

john-m-liu (approved):

tgrander:
evergreen retry

PR #523, opened 2020-11-10
EVG-13096 Surface priority in task metadata if it changed

khelif96:
If Priority is changed it should show up in the task metadata like in the legacy ui.

Also allowed setting a priority of -1 this lets you disable a task.

### DIFF
--- a/src/components/PatchActionButtons/SetPatchPriority.tsx
+++ b/src/components/PatchActionButtons/SetPatchPriority.tsx
@@ -59,7 +59,7 @@ export const SetPatchPriority: React.FC<SetPriorityProps> = ({
           <StyledBody>Set new priority:</StyledBody>
           <InputNumber
             size="small"
-            min={-1}
+            min={0}
             type="number"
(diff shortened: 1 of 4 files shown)


src/components/PatchActionButtons/SetPatchPriority.tsx
```diff
@@ -59,7 +59,7 @@ export const SetPatchPriority: React.FC<SetPriorityProps> = ({
           <StyledBody>Set new priority:</StyledBody>
           <InputNumber
             size="small"
-            min={0}
+            min={-1}
```
john-m-liu, line 62: The ability to disable a task was added in https://github.com/evergreen-ci/spruce/commit/b2552477b194f328e41359609be5b8243a3360f0. I don't think it makes sense to have 2 ways to do it, and I intentionally didn't go this route because the dropdown option is clearer

john-m-liu (changes requested):

john-m-liu (approved):

PR #535, opened 2020-11-16
EVG-13378 Dont allow user to edit spawn host if it is not running or started

khelif96:
This pr disables the ability to edit a spawn host if it is not in the running or started status. This is to prevent changes to the state of a spawned host while it is being initialized which might cause it to be in an unpredictable state.

### DIFF
--- a/src/components/Spawn/MountVolumeSelect.tsx
+++ b/src/components/Spawn/MountVolumeSelect.tsx
@@ -46,12 +46,12 @@ export const MountVolumeSelect = ({
   });
   useNetworkStatus(startPolling, stopPolling);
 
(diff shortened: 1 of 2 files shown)


ybrill (approved):

SupaJoon:
I think it makes sense to also disallow mounting a volume to a host that is not running or started from the Spawn Volume Modal. That would require adding an additional filter here: https://github.com/evergreen-ci/spruce/blob/5c1c90256acfb5e704e38241beccfb8f47680cb1/src/components/Spawn/MountVolumeSelect.tsx#L54

SupaJoon (approved):

PR #537, opened 2020-11-18
EVG-12541 Add total code changes and fix tabs padding

khelif96:
This PR adds the total code changes to the changes tab.
It also fixes the padding on the leafygreen tabs that was broken when we update @leafygreen-ui/tabs 

image

### DIFF
--- a/src/pages/patch/patchTabs/CodeChanges.tsx
+++ b/src/pages/patch/patchTabs/CodeChanges.tsx
@@ -27,36 +27,34 @@ export const CodeChanges: React.FC = () => {
   if (error) {
     return <div id="patch-error">{error.message}</div>;
   }
-  if (!data.patch.moduleCodeChanges.length) {
+  const { moduleCodeChanges } = data.patch;
+  if (!moduleCodeChanges.length) {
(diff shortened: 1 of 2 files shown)


src/pages/patch/patchTabs/CodeChanges.tsx
```diff
@@ -35,6 +36,19 @@ export const CodeChanges: React.FC = () => {
         const sortedFileDiffs = [...modCodeChange.fileDiffs].sort((a, b) =>
           a.fileName.localeCompare(b.fileName)
         );
+        const { fileDiffs } = modCodeChange;
+        const TotalFileDiff = {
```
SupaJoon, line 40: I think we should do away with the TotalFileDiff. The initial values are captured in the reduce function calls and it's nice if we can avoid mutating objects when we don't need to.

SupaJoon (approved):

src/pages/patch/patchTabs/CodeChanges.tsx
khelif96: Great idea.

PR #539, opened 2020-11-19
EVG-8324: Patch Restart Modal base status filter

SupaJoon:
https://jira.mongodb.org/browse/EVG-8324
The default values in the patch restart modal status filter should reflect the status filter values selected in the version page.
Screen Shot 2020-11-19 at 4 09 06 PM
[This computationally intensive code](https://github.com/evergreen-ci/spruce/pull/539/files#diff-a18d6f1203a517cb20cc5a043cde2df0a824431164490b5dbaa1aea620f98168R96-R143) determines the checkbox states of tasks as the status selectors and available tasks change. In order to prevent this code from blocking the UI thread and lag user interactions as well as html state changes, I offload the computation to a webworker. 
An issue I came across is that [CRA doesn't offer additional support for web workers](https://github.com/facebook/create-react-app/issues/3660) and prevents us from including web worker files in the `src` directory so they can be written in typescript and tested via jest. There are ways to accomplish this by installing additional node modules and/or ejecting CRA and is something we should consider when we get around to [this ticket](https://jira.mongodb.org/browse/EVG-13328). 

With worker
with worker

Without worker
without worker


src/types/task.ts
```diff
@@ -90,3 +91,102 @@ export enum TaskStatus {
   // not satisfied
   StatusBlocked = "blocked",
 }
+
+export const taskStatusesFilterTreeData: TreeDataEntry[] = [
```
tgrander, line 95: this is not a type really; it's constant data.

src/pages/patch/PatchRestartModal.tsx
```diff
@@ -122,14 +147,15 @@ export const PatchRestartModal: React.FC<PatchModalProps> = ({
...
-          <TreeSelect
-            onChange={setValidStatus}
-            state={validStatus}
-            tData={statusesTreeData}
-            inputLabel="Tasks Selected: "
-            dataCy="patch-status-filter"
-            width="50%"
-          />
+          <Row>
+            <TaskStatusFilters
```
khelif96, line 151: I think now that there are 2 filters we should make the dropdowns be a little larger.  its hard to tell whats in them. 
image

src/pages/patch/PatchRestartModal.tsx
```diff
@@ -71,10 +72,30 @@ export const PatchRestartModal: React.FC<PatchModalProps> = ({
...
+    patchStatusFilterTerm,
+    baseStatusFilterTerm,
+    { toggleSelectedTask, setPatchStatusFilterTerm, setBaseStatusFilterTerm },
   ] = usePatchStatusSelect(patchBuildVariants);
-
+  const { search } = useLocation();
+  const prevVisible = usePrevious(visible);
+  useEffect(() => {
+    // apply status filter from URL when the modal opens
+    if (visible && !prevVisible) {
```
khelif96, line 83: This is pretty cool idea to prevent unnecessary rerenders of the modal when its not in view.

khelif96 (commented):
Works great for normal sized patches. For larger patches it is a little slower compared to what we had already. We discussed this offline.

SupaJoon:
I'm working on performance improvements for larger patches

src/hooks/usePatchStatusSelect.ts
```diff
@@ -28,58 +83,96 @@ export const usePatchStatusSelect = (
...
+          selectedTasks,
         });
-      });
-      setSelectedTasks(tempSelectedTasks);
+      } else {
+        // fallback in case web workers are not available.
+        // This code reflects logic in public/web_worker/patchBuildVariantReduce.js
+        // Iterate through PatchBuildVariants and determine if a task should be
+        // selected or not based on if the task status correlates with the 2 filters.
+        // if 1 of the 2 filters is empty, ignore the empty filter
```
SupaJoon, line 118: The following logic exists in the web worker but are unable to share the code since the web worker file is being hosted from the `public` directory. I think it could make sense to include it in our `src` folder by installing additional node modules and/or by ejecting CRA if a) we increasingly find ourselves using web workers to solve problems b) we find other significant benefits from ejecting CRA

src/analytics/useAnalyticsAttributes.ts
```diff
@@ -14,7 +14,6 @@ export const useAnalyticsAttributes = () => {
 
   useEffect(() => {
     if (typeof newrelic !== "object") {
-      console.log("Setting userId: ", userId);
       return;
```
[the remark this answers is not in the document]
SupaJoon: what kind of failure does this log help debug? I'm wondering if we can report something to bugsnag or add a unit test instead of log to the dev console.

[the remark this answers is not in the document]
khelif96: This log only gets displayed when new relic is not in scope. I believe new relic is only in scope when we have a production build of spruce.
So for local development it allows us to confirm our analytics calls are being properly triggered.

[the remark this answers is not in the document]
khelif96: So if we perform some action when testing something locally we get immediate confirmation what ever we did is properly calling new relic. I've personally found it useful in the past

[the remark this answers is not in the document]
SupaJoon: I added the logging since it helpful for now and opened [a ticket](https://jira.mongodb.org/browse/EVG-13518) that will hopefully make us confident in removing the log line in the future

public/web_worker/patchBuildVariantsReduce.js
```diff
@@ -0,0 +1,37 @@
+// Iterate through PatchBuildVariants and determine if a task should be
```
khelif96, line 1: While i like this approach i think its important that we minimize code retyping.  And make it so pieces of code like this are more testable. Here are a few solutions i found to place this code within the src/ folder. 
We discussed this on slack. 

https://github.com/petersobolev/cra-worker/tree/master/src

https://github.com/educartoons/creact-react-app-typescript-web-workers

src/hooks/usePatchStatusSelect.ts
```diff
@@ -1,22 +1,77 @@
...
+export const usePatchStatusSelect = (
+  patchBuildVariants: PatchBuildVariant[]
+): HookResult => {
+  const [webWorker, setWebWorker] = useState<Worker>();
+  useEffect(() => {
+    if (window.Worker && !webWorker) {
+      const worker = new Worker(getWebWorkerURL("patchBuildVariantsReduce.js"));
+      setWebWorker(worker);
+    }
+  }, [webWorker]);
```
khelif96, line 65: Neat! i think this logic should be extracted to its own hook if possible. Especially if you think this is something we might do more of.

src/hooks/usePatchStatusSelect.ts
```diff
@@ -28,58 +83,96 @@ export const usePatchStatusSelect = (
...
+                    (patchStatusFilterTerm?.length
+                      ? statuses.has(task.status)
+                      : true) &&
+                    (baseStatusFilterTerm?.length
+                      ? baseStatuses.has(task.baseStatus)
+                      : true),
+                }),
+                accumA
+              ),
+            { ...selectedTasks }
```
khelif96, line 139: Having this repeated code is why i made the suggestion above we should minimize this if at all possible.

src/hooks/usePatchStatusSelect.ts
khelif96: I am all for ejecting CRA . I think if this is the approach we would like to take and something we plan on doing more of, CRA is limiting us at this point. There are other advantages i think ejecting CRA would provide such as giving us more ability to customize jest. And also allow us to update and customize our build tooling.

khelif96 (approved):

public/web_worker/patchBuildVariantsReduce.js
SupaJoon: I made [this ticket](https://jira.mongodb.org/browse/EVG-13522) to move webworkers into the /src directory when our build system allows.

PR #557, opened 2020-12-09
EVG-13517 Refactor user Tags component to be more generic

khelif96:
This PR refactors the UserTags component that was used to display the instance tags used by the hosts modals into a more generic `EditableTags` Component so that it can be used for other general use cases that need key value pairs such as for the patch parameters project. 

The `onChange` callback prop of the `EditableTagField` component now just returns an updated array with the changes to the array. And that can be used to transform into what ever format is necessary for submission as demonstrated in the `UserTagsField` Component. 

The overall look of the component hasn't hasn't changed just the state management. Render performance has also been improved. 

Dependent on https://github.com/evergreen-ci/evergreen/pull/4250


src/components/EditableTagField/6-EditableTagField.stories.tsx
```diff
@@ -0,0 +1,25 @@
+import React, { useState } from "react";
+import { EditableTagField } from "./index";
+import "antd/es/select/style/css";
+import "antd/es/carousel/style/css";
+
+const instanceTags = [
```
malikchaya2, line 6: Is this being used? It looks like you declare this again on line 18 in src/components/EditableTagField/EditableTagField.test.tsx

src/components/EditableTagField/EditableTagField.test.tsx
```diff
@@ -0,0 +1,163 @@
...
+  expect(queryAllByDataCy("user-tag-trash-icon")[0]).toBeVisible();
+
+  fireEvent.click(queryAllByDataCy("user-tag-trash-icon")[0]);
+
+  expect(updateData).toBeCalled();
+  expect(data).toStrictEqual([...defaultData.slice(1, 3)]);
+  expect(queryByText("keyA")).toBeNull();
+});
+
+test("Editing a tag key should remove the old tag and replace it with a newer tag with the updated key", async () => {
```
malikchaya2, line 95: Is this currently passing?

src/components/EditableTagField/EditableTagField.test.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+  const { queryAllByDataCy } = render(
+    <EditableTagField inputTags={editableTags} onChange={updateData} />
+  );
+
+  expect(data).toEqual(defaultData);
+  expect(queryAllByDataCy("user-tag-trash-icon")[0]).toBeVisible();
+
+  fireEvent.change(queryAllByDataCy("user-tag-value-field")[0], {
+    target: { value: "new value" },
```
malikchaya2, line 61: Can we add a test that edits the key? I know it is broken right now but I think it's important to have a test so that we know when it is working fully.

src/components/EditableTagField/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+import styled from "@emotion/styled";
+import { InstanceTag, ParameterInput } from "gql/generated/types";
+import { convertArrayToObject } from "utils/array";
+import { TagRow } from "./TagRow";
+
+type Tag = InstanceTag | ParameterInput;
+type EditableTagFieldProps = {
+  onChange: (data: Tag[]) => void;
+  inputTags: Tag[];
+  visible?: boolean;
```
malikchaya2, line 11: Can we set it up so that a tag is not only visible if visible=true, but also if visible=null? Aka only hide it if visible=false?

src/components/EditableTagField/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+};
+
+export const EditableTagField: React.FC<EditableTagFieldProps> = ({
+  onChange,
+  inputTags,
+  visible,
+}) => {
+  const [visibleTags, setVisibleTags] = useState(inputTags);
```
malikchaya2, line 19: Can we take in setVisibleTags as a prop so that the page using this component can access all the tags and do something with it?

malikchaya2 (changes requested):

src/components/EditableTagField/EditableTagField.test.tsx
khelif96: Yes it should be according to CI all tests passed. Are they failing for you?

src/components/EditableTagField/EditableTagField.test.tsx
khelif96: https://github.com/evergreen-ci/spruce/pull/557/files/146a2e818d9a004b3fd1efbeb33daee0607db074#diff-2e9113b270269f5aa49f0dedf5c065d8a1b7064e76cd3364827f82cf92826310R95

There is a test that does that. As far as i can tell it works as expected.

src/components/EditableTagField/index.tsx
khelif96: This prop serves to reset the component state when the modal rendering it is hidden. I can rename the prop to make it more clear. It does not affect which tags are rendered.

src/components/EditableTagField/index.tsx
khelif96: I dont think that is a good idea as that would imply an outside component would control the internal state of this component. The component already provides an interface to display and make changes to the tags rendered with in it.

tags can be passed in with the `inputTags` prop and any changes to the state of the tags will be returned with the `onChange` callback prop.

src/components/EditableTagField/6-EditableTagField.stories.tsx
khelif96: Yeah they are both being used in their own use cases.This is being used in the storybook story for this component. Which can be viewed by running `yarn storybook` and the other is being used in the jest tests for this component.

src/components/EditableTagField/EditableTagField.test.tsx
malikchaya2: I don't think so, I was just wondering

malikchaya2 (approved):

PR #558, opened 2020-12-14
EVG-13587 Conditionally render task tabs

khelif96:
With the recent update of leafygreen tabs and some clever conditionals we can now properly render conditional tabs in the task page.

This should allow us hide the tabs that aren't relevant to the task at hand. So the user facing changes would be if the build baron or trend charts plugin aren't available for a task the associated tab wont show. 

Standard Task
image

Task with build baron
image

Task with trend charts
image

### DIFF
--- a/src/components/styles/StyledTabs.tsx
+++ b/src/components/styles/StyledTabs.tsx
@@ -8,11 +8,14 @@ export const StyledTabs: React.FC<StyledTabsProps> = ({
   ...rest
 }) => (
   <Tabs {...rest}>
-    {children.map((c) => (
-      <Tab {...c.props}>
-        <PaddedContainer>{c.props.children}</PaddedContainer>
-      </Tab>
-    ))}
+    {children.map(
+      (c) =>
+        c && (
+          <Tab {...c.props}>
+            <PaddedContainer>{c.props.children}</PaddedContainer>
+          </Tab>
+        )
+    )}
   </Tabs>
 );
 
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "@leafygreen-ui/radio-box-group": "3.0.3",
     "@leafygreen-ui/radio-group": "2.0.1",
     "@leafygreen-ui/side-nav": "2.0.14",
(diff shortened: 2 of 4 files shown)


src/components/styles/StyledTabs.tsx
```diff
@@ -8,11 +8,14 @@ export const StyledTabs: React.FC<StyledTabsProps> = ({
...
   <Tabs {...rest}>
-    {children.map((c) => (
-      <Tab {...c.props}>
-        <PaddedContainer>{c.props.children}</PaddedContainer>
-      </Tab>
-    ))}
+    {children.map(
+      (c) =>
+        c && (
+          <Tab {...c.props}>
```
SupaJoon, line 14: I thought this looping was unintuitive since it utilizes `c` for props and children props to construct another Tab component in place. What do you think about these code changes to StyledTabs: 
```
export const StyledTabs = styled(Tabs)`
  div:nth-child(2) {
    margin-top: 12px;
  }
`;
```
Ideally <Tabs /> could have accepted a pageWrapper component prop since it controls so much of the page content. I was unable to utilize `styled(Tab)``` since wrapping Tab with emtion/styled prevents Tab children from rendering.

khelif96: We actually used to do something similar to that before i made this change. https://github.com/evergreen-ci/spruce/pull/537/files

It broke when we updated `@leafygreen-ui/tabs` since they changed the underlying html structure. 
I think if we dont have control of the component we shouldn't try to hack the css by drawing assumptions on how the component is laid out. Because it can easily change and we would have no control over it.

It would be neat if leafygreen supported this out of the box. 

> I was unable to utilize `styled(Tab)``` since wrapping Tab with emtion/styled prevents Tab children from rendering

Yeah some leafygreen components have some issues when styled with emotion.

SupaJoon: ah, thanks for showing me that. your changes look like the best abstraction.

SupaJoon (approved):

PR #559, opened 2020-12-15
EVG-12879 Change the patch submission page to allow users to specify values to parameters

malikchaya2:
Tests rely on [4261](https://github.com/evergreen-ci/evergreen/pull/4261)

When the patch is not active: 
patchparams

When it's active and there are existing parameters (otherwise it just shows the disclaimer).
image


john-m-liu (approved):
This was already here, but it'd be nice if the save button on the TagRow had an enter handler that does the same thing as clicking the button. That way you can add all parameters by using just the keyboard

src/pages/configurePatch/ConfigurePatchCore.tsx
```diff
@@ -150,6 +156,43 @@ const ConfigurePatch: React.FC<Props> = ({ patch }) => {
...
+                              key={`param_${param.key}`}
+                            >
+                              {param.key}:{param.value}
+                            </StyledBadge>
+                          ))}
+                        </ExistingParamsContainer>
+                      )}
+                    </>
+                  ) : (
+                    <EditableTagField
```
SupaJoon, line 188: can we fix the max width for this component? it becomes less ergonomic to read at wide screen widths
Screen Shot 2020-12-16 at 11 12 15 AM

SupaJoon (changes requested):
the parameters view clears selections when I switch tabs. I think it should persist state and/or show me a save button

src/pages/configurePatch/ConfigurePatchCore.tsx
malikchaya2: Can we do this in a different ticket? This is a component that has more uses, and Mohamed is more familiar with how it will affect all the use cases.

malikchaya2: scratch that. I added a max-width of 70% and it seems to work well.

src/components/EditableTagField/index.tsx
```diff
@@ -75,4 +79,5 @@ const FlexColumnContainer = styled.div`
   display: flex;
   flex-direction: column;
   margin-right: 10px;
+  max-width: 70%;
```
khelif96, line 82: Can you make sure this doesn't negatively impact the component on the edit spawn host modal

khelif96 (approved):

src/components/EditableTagField/index.tsx
malikchaya2: I checked, it doesn't.

malikchaya2:
I don't think the failing tests are related to my changes.

SupaJoon (approved):

PR #561, opened 2020-12-18
EVG-13627 Resolve some warnings on spruce

khelif96:
Went through some pages and cleaned up some errors and warnings


SupaJoon (approved):

src/gql/queries/get-task-files.ts
```diff
@@ -3,6 +3,7 @@ import { gql } from "@apollo/client";
 export const GET_TASK_FILES = gql`
   query TaskFiles($id: String!, $execution: Int) {
     taskFiles(taskId: $id, execution: $execution) {
+      fileCount
```
SupaJoon, line 6: why not define a merge function for this query?

khelif96: As far as I know You can't define merge functions for top level fields in queries only sub fields with nested fields

PR #563, opened 2020-12-23
EVG-12889: prevent showing LOADING... when the browser doesn't have an evg cookie

SupaJoon:
the [first PR](https://jira.mongodb.org/browse/EVG-12889) in [EVG-12889](https://jira.mongodb.org/browse/EVG-12889) exposed an issue where a user can get stuck seeing the FullPageLoad component when the browser is missing an evg cookie. 
These code changes fix that issue by: 
a) no longer relying on the initialLoad variable
b) reducing redirect logic and making the ApolloLink responsible in redirecting the user to the appropriate login page when an unauthorized query is confirmed

bug report: https://mongodb.slack.com/archives/G01EU33NG01/p1608679002001900

You can test these changes by deploying spruce to staging.


src/components/Content/index.tsx
```diff
@@ -44,48 +43,44 @@ export const Content: React.FC = () => {
...
   }
+
   return (
     <PageLayout>
       <Navbar />
       <ConnectivityBanner />
       <SiteBanner />
       <SlackNotificationBanner />
       <Switch>
-        <PrivateRoute path={routes.task} component={Task} />
```
SupaJoon, line 66: I removed the PrivateRoute component since we have an  `isAuthenticated` condition in this render function body.

malikchaya2 (approved):

cypress/integration/auth.ts
```diff
@@ -7,21 +7,16 @@ describe("Auth", () => {
     cy.url().should("include", "/login");
   });
 
-  it("Redirects user back to the route they were trying to visit after login", () => {
+  it("Redirects user to My Patches page after logging in.", () => {
```
khelif96, line 10: Why did this change? I think we should still preserve the previous route after a successful login

khelif96 (changes requested):
I dont think the behavior should have changed. Can we revert it back to its original behavior. This was likely from a recent commit.

khelif96:
Can you merge master?

khelif96:
https://user-images.githubusercontent.com/4605522/104212680-76d15f00-5403-11eb-9038-3829d5bfa3f8.mov

I merged in the latest master and pushed to staging. I think theres something wrong with the redirect logic. Starting in the graphql playground i confirmed that i was able to make requests. But when i try to load spruce on staging it ends up showing the loading screen and then forcing a redirect back to legacy evergreen

SupaJoon:
evergreen retry

khelif96 (approved):

PR #565, opened 2020-12-31
EVG-12955 edit commit message

ybrill:
When a user enqueues a patch we've been using the patch description as the commit message. Let users edit the commit message at enqueue time.

Some thoughts/misgivings: 
-  design incorporated the commit message into the existing popconfirm, but when I implemented it a modal seemed more intuitive. I've implemented it as a modal. Do I need to follow up with design?
- How do I deploy this change? I needed to change the enqueuePatch mutation to include a new commit message parameter. I think the mutation won't work when only one side of the interface is on the new version. Do I add a second duplicate mutation and remove the old one once they're both deployed? Is there another way?


ybrill:
Goes together with https://github.com/evergreen-ci/evergreen/pull/4283

khelif96:
Can you merge master in?

ybrill:
evergreen retry

ybrill:
@khelif96 I merged master. 
I think the remaining issues are because I haven't yet merged the schema changes on the backend.

src/components/PatchActionButtons/EnqueuePatch.tsx
```diff
@@ -1,78 +1,54 @@
...
-import { usePatchAnalytics } from "analytics";
 import { DropdownItem } from "components/ButtonDropdown";
-import { useBannerDispatchContext } from "context/banners";
-import {
-  EnqueuePatchMutation,
-  EnqueuePatchMutationVariables,
-} from "gql/generated/types";
-import { ENQUEUE_PATCH } from "gql/mutations";
-import { StyledBody } from "./UnschedulePatchTasks";
+import { EnqueuePatchModal } from "pages/patch/index";
```
khelif96, line 4: You can just do `import { EnqueuePatchModal } from "pages/patch";`

src/pages/patch/EnqueueModal.tsx
```diff
@@ -0,0 +1,105 @@
...
+              patchAnalytics.sendEvent({ name: "Enqueue" });
+            }}
+            variant="primary"
+          >
+            Enqueue
+          </Button>,
+        ]}
+        data-cy="enqueue-modal"
+      >
+        <>
```
khelif96, line 87: This fragment can probably be removed.

khelif96 (approved):

src/pages/patch/EnqueueModal.tsx
ybrill: done

src/components/PatchActionButtons/EnqueuePatch.tsx
ybrill: When I tried this it thought I was trying to import from `pages/Patch.tsx`

khelif96:
Tests should pass once the backend portion of this is merged. After it is you should try to merge in master and run `yarn codegen` one more time.

SupaJoon (approved):

PR #564, opened 2020-12-23
EVG-12488: Group code changes table contents by commit in /commit-queue

SupaJoon:
https://jira.mongodb.org/browse/EVG-12488
relies on https://github.com/evergreen-ci/evergreen/pull/4275
Screen Shot 2020-12-23 at 4 47 08 PM

PR #577, opened 2021-01-04
EVG-13578 Embed execution tasks within display tasks in task table

khelif96:
image

Dependent on https://github.com/evergreen-ci/evergreen/pull/4285


src/pages/patch/patchTabs/tasks/TasksTable.tsx
```diff
@@ -34,14 +36,21 @@ export const TasksTable: React.FC<Props> = ({ data, columns }) => {
     }
   };
 
+  const transformData = (tasks: Tasks) =>
+    tasks.map((task) => {
+      const t = { ...task, children: undefined };
```
SupaJoon, line 41: I think we should remove `children: undefined` and create a return interface for transformData. that way we don't have to make this type alias `as TaskResult[]` which misrepresents the result of `transformData`

SupaJoon (changes requested):

src/pages/patch/patchTabs/tasks/TasksTable.tsx
khelif96: Nice catch this is actually no longer needed

khelif96: It was from before i found the prob that lets your rename the children column name

src/gql/queries/get-code-changes.ts
```diff
@@ -13,6 +13,7 @@ export const GET_CODE_CHANGES = gql`
           additions
           deletions
           diffLink
+          description
```
SupaJoon, line 16: was this intentional?

src/gql/queries/get-commit-queue.ts
```diff
@@ -19,6 +19,7 @@ export const GET_COMMIT_QUEUE = gql`
             branchName
             htmlLink
             fileDiffs {
+              description
```
SupaJoon, line 22: was this intentional?

src/gql/queries/get-code-changes.ts
khelif96: I was getting a typescript error for mismatched types this was the quickest fix at the time. I think this will be fixed in your pr.  https://github.com/evergreen-ci/spruce/pull/564/files

PR #582, opened 2021-01-05
EVG-13172 add editable annotation note field

ablack12:
image

Save Note disabled unless the message has actually been changed, and the user is authorized.


src/components/BuildBaronAndAnnotations/AnnotationNote.tsx
```diff
@@ -0,0 +1,111 @@
...
+import { Disclaimer } from "@leafygreen-ui/typography";
+import { Input, Tooltip } from "antd";
+import {
+  EditAnnotationNoteMutation,
+  EditAnnotationNoteMutationVariables,
+  Note,
+} from "gql/generated/types";
+import { useBannerDispatchContext } from "../../context/banners";
+import { EDIT_ANNOTATION_NOTE } from "../../gql/mutations/edit-annotation-note";
+import { getDateCopy } from "../../utils/string";
```
khelif96, line 14: These can be rewritten as 
```
import { useBannerDispatchContext } from "context/banners";
import { EDIT_ANNOTATION_NOTE } from "gql/mutations";
import { getDateCopy } from "utils/string";
```

src/components/BuildBaronAndAnnotations/AnnotationNote.tsx
```diff
@@ -0,0 +1,111 @@
...
+import { ConditionalWrapper } from "../ConditionalWrapper";
```
khelif96, line 15: `import { ConditionalWrapper } from "components/ConditionalWrapper";`

src/components/BuildBaronAndAnnotations/AnnotationNote.tsx
```diff
@@ -0,0 +1,111 @@
...
+        newMessage,
+      },
+    });
+  };
+
+  return (
+    <TitleAndButtons>
+      <TicketsTitle>Note</TicketsTitle>
+      {note && (
+        <>
```
khelif96, line 66: This Fragment isn't necessary since `<MetaDataWrapper>` is already wrapping all the content here.

src/components/BuildBaronAndAnnotations/AnnotationNote.tsx
```diff
@@ -0,0 +1,111 @@
...
+              Updated: {getDateCopy(note.source.time, null, true)}{" "}
+            </Disclaimer>
+            <Disclaimer>Last Edited By: {note.source.author}</Disclaimer>
+          </MetaDataWrapper>
+        </>
+      )}
+      <TextArea
+        id="noteInput"
+        rows={2}
+        style={{ width: 600 }}
```
khelif96, line 78: Can you use emotion to style this component.

src/components/BuildBaronAndAnnotations/AnnotationNote.tsx
```diff
@@ -0,0 +1,111 @@
...
+          <Tooltip title="You are not authorized to edit task annotations">
+            <span>{children}</span>
+          </Tooltip>
+        )}
+      >
+        <ButtonWrapper>
+          <Button
+            data-cy="edit-annotation-button"
+            variant={Variant.Primary}
+            size="xsmall"
```
khelif96, line 94: There is an enum exported from `@leafygreen-ui/button` with sizes
you can use it by doing
```
import {Size} from "@leafygreen-ui/button"

Size.XSmall
```

src/components/BuildBaronAndAnnotations/AnnotationNote.tsx
```diff
@@ -0,0 +1,111 @@
...
+      </ConditionalWrapper>
+    </TitleAndButtons>
+  );
+};
+
+export const ButtonWrapper = styled.div`
+  margin-right: 8px;
+  padding-top: 5px;
+`;
+const { TextArea } = Input;
```
khelif96, line 111: We usually do this at the top under the imports like 
https://github.com/evergreen-ci/spruce/blob/7f35ea9ee8d6230fee42a492ba52a2d7d2302733/src/components/BuildBaronAndAnnotations/AddIssueModal.tsx#L16

src/components/BuildBaronAndAnnotations/AnnotationTickets.tsx
```diff
@@ -64,7 +64,7 @@ export const AnnotationTickets: React.FC<Props> = ({
   ) : null;
 };
 
-const StyledButton = styled(PlusButton)`
+export const StyledButton = styled(PlusButton)`
```
khelif96, line 67: Is this used anywhere else?

src/components/BuildBaronAndAnnotations/BBFIleTicket.tsx
```diff
@@ -79,13 +78,15 @@ export const FileTicket: React.FC<FileTicketProps> = ({
...
-        variant={Variant.Primary}
-        size="xsmall"
-      >
-        {buttonText}
-      </Button>
+      <ButtonWrapper>
+        <Button
+          data-cy="file-ticket-button"
+          variant={Variant.Primary}
+          size="xsmall"
```
khelif96, line 85: Same comment as above of using `Size.XSmall`

src/components/BuildBaronAndAnnotations/AnnotationNote.tsx
ablack12: Gotcha, thanks! Separately, I got the information about how to define TextArea from the internet, but I'm confused about how it works like this. I would think that I could add "rows" to input and have it translate to area, or that I would import TextArea instead of importing Input, but that's not the case. It seems instead like TextArea is a subclass of Input? Can you explain how this works? (or help me figure out what to google haha)

src/components/BuildBaronAndAnnotations/AnnotationTickets.tsx
ablack12: Ah I was using it in AnnotationNote for a bit but it didn't do what I wanted so I changed it, I'll remove the export.

src/components/BuildBaronAndAnnotations/AnnotationNote.tsx
khelif96: Your exactly right antd has TextArea declared as a subclass within `Input`
image

src/components/BuildBaronAndAnnotations/AnnotationNote.tsx
```diff
@@ -0,0 +1,113 @@
...
+          >
+            Save Note
+          </Button>
+        </ButtonWrapper>
+      </ConditionalWrapper>
+    </TitleAndButtons>
+  );
+};
+
+export const ButtonWrapper = styled.div`
```
khelif96, line 106: I think `BBComponents.tsx` would be a better home for this since it also includes other general components for the build baron

src/components/BuildBaronAndAnnotations/BBFIleTicket.tsx
```diff
@@ -79,13 +78,15 @@ export const FileTicket: React.FC<FileTicketProps> = ({
...
     >
-      <Button
-        data-cy="file-ticket-button"
-        variant={Variant.Primary}
-        size="xsmall"
-      >
-        {buttonText}
-      </Button>
+      <ButtonWrapper>
+        <Button
```
khelif96, line 82: You can alternatively just apply the styles from ButtonWrapper directly to the button if you dont need it to be wrapped in a div.

src/components/BuildBaronAndAnnotations/AnnotationNote.tsx
khelif96: You can cmd+click on almost any variable or imported variable in vscode and jump to its definition if you're curious.

src/components/BuildBaronAndAnnotations/BBFIleTicket.tsx
ablack12: Ah, I used a wrapper because otherwise the button was showing up directly next to the thing above it; Chaya mentioned that a div would fix that.

src/components/BuildBaronAndAnnotations/AnnotationNote.tsx
ablack12: Thank you!

khelif96 (approved):

src/components/BuildBaronAndAnnotations/AnnotationNote.tsx
```diff
@@ -0,0 +1,113 @@
...
+          </Disclaimer>
+          <Disclaimer>Last Edited By: {note.source.author}</Disclaimer>
+        </MetaDataWrapper>
+      )}
+      <StyledTextArea
+        id="noteInput"
+        rows={2}
+        value={newMessage}
+        onChange={(e) => setMessage(e.target.value)}
+      />
```
malikchaya2, line 85: Nit: would it be possible to add more space between the metadata (author, time...) and the note, as well as between the note and the button? If you click on the inspect button on the design, it lets you click on individual items and view how much space there is between items (in pixels).

src/components/BuildBaronAndAnnotations/BBFIleTicket.tsx
```diff
@@ -23,7 +23,6 @@ export const BBFileTicket: React.FC<BBFileTicketProps> = ({
   createdTicketsCount,
 }) => (
   <>
-    <TicketsTitle margin>Create a new ticket in Jira </TicketsTitle>
```
malikchaya2, line 26: I'm a bit concerned about the design choice of removing this title. If there are already filed tickets, then the user will see a title, a list of tickets, and a "file another ticket" button making the use case clear. However, if there are no tickets yet, all the user will see is the button with no explanation. I think we need another design or screenshot with just the button and no filed tickets to make sure that things are still clear.

malikchaya2 (changes requested):

SupaJoon (approved):

src/components/BuildBaronAndAnnotations/BBFIleTicket.tsx
ablack12: Fair--I removed this during earlier talks with Lara! I'm going to switch the title to what was in the most recent design

ablack12:
image
Here is an image with the updated title and styling.
As for why the Edit Note button is grayed out, I'm not sure why this task has tickets from the jira search but no filed tickets; it seems like the page is visible because this exists, even though there is no annotation. Because there is no annotation, "UserCanModify" is undefined, so the note button isn't actually usable. I think this case is wrapped up in EVG-13167 so I'm not concerned but I did want to explain it.

malikchaya2 (approved):

PR #588, opened 2021-01-08
EVG-13701 Add analytics for display task dropdown

khelif96:
I added back the console.logs for analytics since they are useful when testing analytics locally. And they wont show up in production.

### DIFF
--- a/src/analytics/addPageAction.ts
+++ b/src/analytics/addPageAction.ts
@@ -52,9 +52,7 @@ export const addPageAction = <A extends ActionType, P extends Properties>(
 
   if (typeof newrelic !== "object") {
     // These will only print when new relic is not available such as during local development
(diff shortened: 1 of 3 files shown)


src/analytics/addPageAction.ts
```diff
@@ -51,6 +51,10 @@ export const addPageAction = <A extends ActionType, P extends Properties>(
   };
 
   if (typeof newrelic !== "object") {
+    // These will only print when new relic is not available such as during local development
+    console.log("____ ANALYTICS EVENT ____");
```
SupaJoon, line 55: can we put these all on one line? console.log("ANALYTICS EVENT:", {name, attributesToSend})

SupaJoon (approved):

PR #591, opened 2021-01-08
EVG-13365 allow multiple sorts in tasks table, default new bugs

john-m-liu:
- change the sorting in the patchTasks query from the old one to the new one added in https://github.com/evergreen-ci/evergreen/pull/4299
- enable multi-column sorting on the tasks table. Since the column headers have 3 states, this actually makes it kind of annoying to switch from exclusive sorting on 1 column to another, but I don't have a graceful solution for this at the moment (maybe a button to clear all sorts)
- default sorting new bugs to the top (patch status then base status)
- fix an existing bug where an initial sort stored in the URL does not set the initial state of the sort indicators on the column headers


src/pages/patch/patchTabs/Tasks.tsx
```diff
@@ -49,13 +56,21 @@ export const Tasks: React.FC<Props> = ({ taskCount }) => {
 
   const queryVariables = getQueryVariables(search, resourceId);
 
-  const { sortBy, sortDir, limit, page } = queryVariables;
+  let { sorts } = queryVariables;
```
khelif96, line 59: This should be be a const.

src/pages/patch/patchTabs/Tasks.tsx
```diff
@@ -49,13 +56,21 @@ export const Tasks: React.FC<Props> = ({ taskCount }) => {
...
+  const { limit, page } = queryVariables;
 
-  const columns = useSetColumnDefaultSortOrder<TaskResult>(
-    columnsTemplate,
-    sortBy,
-    sortDir
-  );
+  if (sorts.length === 0) {
+    sorts = [
```
khelif96, line 63: > This should be be a const.

sorts represents a value that exists in the url. by over writing it here without also updating it within the url we are invalidating the truthiness to this value. It no longer represents the state of the url. If you want to apply default sort parameters you should instead update the query params within the url.

khelif96 (changes requested):

src/pages/patch/patchTabs/Tasks.tsx
```diff
@@ -193,6 +270,32 @@ const getQueryVariables = (
   };
 };
 
+const parseSortString = (sortQuery: string | string[]): SortOrder[] => {
```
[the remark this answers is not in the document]
john-m-liu: done

src/pages/patch/patchTabs/tasks/TasksTable.tsx
```diff
@@ -55,3 +52,25 @@ export const TasksTable: React.FC<Props> = ({ data, columns }) => {
 };
 
 const rowKey = ({ id }: { id: string }): string => id;
+
+const toSortString = (
```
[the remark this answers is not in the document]
john-m-liu: done, they aren't inverses of each other so I added some comments explaining it

src/pages/patch/patchTabs/Tasks.tsx
```diff
@@ -71,6 +74,82 @@ export const Tasks: React.FC<Props> = ({ taskCount }) => {
...
+      },
+      className: "cy-task-table-col-STATUS",
+      render: renderStatusBadge,
+    },
+    {
+      title: "Base Status",
+      dataIndex: "baseStatus",
+      key: TaskSortCategory.BaseStatus,
+      sorter: {
+        multiple: 4,
```
khelif96, line 106: Why are all these the same multiple? According to the docs this field sets the priority for each column when sorting but if the are all the same priority why even set a priority? https://ant.design/components/table/#components-table-demo-multiple-sorter

src/pages/patch/patchTabs/Tasks.tsx
```diff
@@ -71,6 +74,82 @@ export const Tasks: React.FC<Props> = ({ taskCount }) => {
...
+        }
+        case TaskSortCategory.BaseStatus: {
+          columnsTemplate[2].defaultSortOrder = direction;
+          break;
+        }
+        case TaskSortCategory.Variant: {
+          columnsTemplate[3].defaultSortOrder = direction;
+          break;
+        }
+        default:
```
khelif96, line 147: I think this logic could be broken out into its own util function. The reason your probably experience some weird errors with the defaultSorts not being reflected in the table is likely a result of 2 things.

khelif96, line 147: 2. Updating these fields by iterating through the array sequentially instead of performing the update in an atomic way.

src/pages/patch/patchTabs/Tasks.tsx
```diff
@@ -71,6 +74,82 @@ export const Tasks: React.FC<Props> = ({ taskCount }) => {
...
+        multiple: 4,
+      },
+      className: "cy-task-table-col-VARIANT",
+    },
+  ];
+  if (sorts.length === 0) {
+    // works around a quirk/bug in the antd table where default sorts don't
+    // get updated on subsequent renders
+    columnsTemplate[1].defaultSortOrder = "ascend";
+    columnsTemplate[2].defaultSortOrder = "ascend";
```
khelif96, line 125: 1. This bit of logic is not necessary. It should be accomplished by you just updating the query params here. https://github.com/evergreen-ci/spruce/pull/591/files#diff-9dac9e5fbcce9bdaff0aa2c7996ad847f32fdcca0c789b9dd1286cf659cef383R57
And then letting the sortOrder be inferred from the query params instead of manually setting them

src/pages/patch/patchTabs/Tasks.tsx
```diff
@@ -71,6 +74,82 @@ export const Tasks: React.FC<Props> = ({ taskCount }) => {
...
+        }
+        case TaskSortCategory.Variant: {
+          columnsTemplate[3].defaultSortOrder = direction;
+          break;
+        }
+        default:
+          break;
+      }
+    });
+  }
```
khelif96, line 151: Its probably best to split this out to a utility function that just returns the columnsTemplate after the sorts have been applied.

It should look something like this.
`const sortedColumnTemplate = getSortedColumns(columns, sorts);`

```
const getSortedColumns = (columns: ColumnProps<TaskResult>[], sorts: <TypeForSorts>) => {
... // Logic to set defaultSortOrder and return the updatedColumnsTemplate
}
```

khelif96 (changes requested):

src/pages/patch/patchTabs/Util.test.tsx
```diff
@@ -0,0 +1,30 @@
+import { SorterResult } from "antd/es/table/interface";
```
SupaJoon, line 1: this file extension should be .ts

john-m-liu: fixed

src/pages/patch/patchTabs/Tasks.tsx
john-m-liu: I actually did this previously, but switched to what you see now because it didn't work. I tried again with your suggestions (https://gist.github.com/john-m-liu/3cffc77e817b4f0eb80462c185558381) and see the same problem

src/pages/patch/patchTabs/Tasks.tsx
```diff
@@ -114,7 +193,10 @@ export const Tasks: React.FC<Props> = ({ taskCount }) => {
         </TableControlInnerRow>
       </TableControlOuterRow>
       <TableContainer hide={showSkeleton}>
-        <TasksTable columns={columns} data={get(data, "patchTasks", [])} />
+        <TasksTable
+          columns={columnsTemplate}
+          data={get(data, "patchTasks", [])}
```
SupaJoon, line 198: this can be `data?.patchTasks ?? []`

src/pages/patch/patchTabs/Util.tsx
```diff
@@ -0,0 +1,62 @@
...
+    sortStrings = sortStrings.concat(
+      `${sorts.columnKey},${shortenSortOrder(sorts.order)}`
+    );
+  }
+
+  return sortStrings.join(";");
+};
+
+// takes a sort query string and parses it into valid GQL params
+export const parseSortString = (sortQuery: string | string[]): SortOrder[] => {
```
SupaJoon, line 33: it would be neat to pass the result of this into useSetDefaultColumnSortOrder

SupaJoon (approved):

khelif96:
I figured out what was going on. 
The [memo behavior](https://github.com/evergreen-ci/spruce/pull/591#discussion_r556758439) of the `defaultSortOrder` that you mentioned happens because the wrong prop is being used .  `sortOrder` should be used instead of `defaultSortOrder`. 
I was able to build out a working example of the solution i suggested [here](https://gist.github.com/khelif96/9996c896f0662c4192ac932ab21cbce7)


https://user-images.githubusercontent.com/4605522/104521325-aed6ce80-55ca-11eb-967c-f1d1969e3078.mov

src/pages/patch/patchTabs/Tasks.tsx
john-m-liu: I'd have to assert that the empty array is an empty array of PatchTasks which I think is messier than what's here

src/pages/patch/patchTabs/Util.tsx
john-m-liu: I don't see anything named that. Are you suggesting to make this a hook instead?

john-m-liu:
Changing it from defaultSortOrder worked, thanks for finding that!

src/pages/patch/patchTabs/Util.tsx
SupaJoon: @john-m-liu we originally had a [useSetColumnDefaultSortOrder](https://github.com/evergreen-ci/spruce/blob/master/src/hooks/useSetColumnDefaultSortOrder.ts#L6) that helped us set the defaultSortOrder key in the columns array. Since relying sortOrder is a cleaner solution, you can actually ignore my original comment.

src/pages/patch/patchTabs/Util.tsx
```diff
@@ -0,0 +1,62 @@
...
+  let sorts: SortOrder[] = [];
+  let sortArray: string[] = [];
+  if (typeof sortQuery === "string") {
+    sortArray = sortQuery.split(";");
+  } else {
+    sortArray = sortQuery;
+  }
+  if (sortArray?.length > 0) {
+    sortArray.forEach((singleSort) => {
+      const parts = singleSort.split(",");
```
khelif96, line 43: Out of curiousity why use `,` as the identifier for key value pairs? 
why not `:` since thats what is typically used.

john-m-liu: No special reason, I just picked some separator. Since this will be url encoded and not really human-readable, I don't feel that one is better than the other

src/pages/patch/patchTabs/Util.tsx
khelif96, line 33: In what case would this take in an array of sortStrings?

khelif96:
Could you also merge master into this pr.

src/pages/patch/patchTabs/Util.tsx
```diff
@@ -0,0 +1,62 @@
+import { SorterResult } from "antd/es/table/interface";
```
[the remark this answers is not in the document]
john-m-liu: fixed

src/pages/patch/patchTabs/Util.tsx
john-m-liu: I'm not sure when we would actually get an array, but this is the type returned by queryString.parse which I pass directly into this

src/pages/patch/patchTabs/Util.tsx
john-m-liu: changed to :

src/pages/patch/patchTabs/Util.test.ts
```diff
@@ -4,7 +4,7 @@ import {
   TaskSortCategory,
```
khelif96, line 4: you forgot to change the tests. 
test file name and assertions

khelif96 (approved):

PR #594, opened 2021-01-13
EVG-13570 fix tests

john-m-liu:
https://github.com/evergreen-ci/evergreen/commit/cdccf100650cba4409a2444d63711d37cc998464 added a task which broke the counts. We should probably make this test more resilient, but for now this should make things green

### DIFF
--- a/cypress/integration/patch/task-table.ts
+++ b/cypress/integration/patch/task-table.ts
@@ -61,7 +61,7 @@ describe("Task table", () => {
   it("Task count displays total tasks", () => {
     cy.visit(pathTasks);
     cy.waitForGQL("PatchTasks");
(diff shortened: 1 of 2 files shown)


cypress/integration/patch/task-table.ts
```diff
@@ -61,7 +61,7 @@ describe("Task table", () => {
   it("Task count displays total tasks", () => {
     cy.visit(pathTasks);
     cy.waitForGQL("PatchTasks");
-    cy.get("[data-cy=total-task-count]").contains("49");
+    cy.get("[data-cy=total-task-count]").contains("50");
```
khelif96, line 64: `cy.dataCy("total-task-count")`

SupaJoon (approved):

khelif96 (approved):

PR #595, opened 2021-01-13
EVG-13167 move annotation permission to task

ablack12:
Combined with [the EVG PR](https://github.com/evergreen-ci/evergreen/pull/4310)
image

### DIFF
--- a/src/components/BuildBaronAndAnnotations/AnnotationTickets.tsx
+++ b/src/components/BuildBaronAndAnnotations/AnnotationTickets.tsx
@@ -30,7 +30,7 @@ export const AnnotationTickets: React.FC<Props> = ({
     isAddAnnotationModalVisible,
     setIsAddAnnotationModalVisible,
   ] = useState<boolean>(false);
-  return tickets?.length ? (
+  return (
     <>
       <TitleAndButtons>
         <TicketsTitle>{title} </TicketsTitle>
@@ -51,17 +51,21 @@ export const AnnotationTickets: React.FC<Props> = ({
           </StyledButton>
         </ConditionalWrapper>
       </TitleAndButtons>
-      <AnnotationTicketsTable jiraIssues={tickets} />{" "}
-      <AddIssueModal
-        dataCy="addIssueModal"
-        visible={isAddAnnotationModalVisible}
-        closeModal={() => setIsAddAnnotationModalVisible(false)}
-        taskId={taskId}
-        execution={execution}
-        isIssue={isIssue}
-      />
+      {tickets?.length > 0 && (
+        <>
+          <AnnotationTicketsTable jiraIssues={tickets} />{" "}
+          <AddIssueModal
+            dataCy="addIssueModal"
+            visible={isAddAnnotationModalVisible}
+            closeModal={() => setIsAddAnnotationModalVisible(false)}
+            taskId={taskId}
+            execution={execution}
+            isIssue={isIssue}
+          />
+        </>
+      )}
     </>
-  ) : null;
+  );
 };
 
 const StyledButton = styled(PlusButton)`
--- a/src/components/BuildBaronAndAnnotations/AnnotationTickets.tsx
+++ b/src/components/BuildBaronAndAnnotations/AnnotationTickets.tsx
@@ -55,25 +55,23 @@ export const AnnotationTickets: React.FC<Props> = ({
         </ConditionalWrapper>
       </TitleAndButtons>
       {tickets?.length > 0 && (
-        <>
-          <AnnotationTicketsTable
-            jiraIssues={tickets}
-            annotationId={annotationId}
-            taskId={taskId}
-            execution={execution}
-            isIssue={isIssue}
-            userCanModify={userCanModify}
-          />
-          <AddIssueModal
-            dataCy="addIssueModal"
-            visible={isAddAnnotationModalVisible}
-            closeModal={() => setIsAddAnnotationModalVisible(false)}
-            taskId={taskId}
-            execution={execution}
-            isIssue={isIssue}
-          />
-        </>
+        <AnnotationTicketsTable
+          jiraIssues={tickets}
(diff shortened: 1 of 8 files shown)


malikchaya2 (approved):

src/pages/Task.tsx
```diff
@@ -114,6 +114,7 @@ const TaskCore: React.FC = () => {
   const fileCount = get(data, "taskFiles.fileCount");
   const logLinks = get(task, "logs");
   const isPerfPluginEnabled = get(task, "isPerfPluginEnabled");
+  const canModifyAnnotation = get(task, "canModifyAnnotation");
```
SupaJoon, line 117: this is better written as `task?.canModifyAnnotation ?? false` so we can get closer to removing the `get` module

ablack12:
Referencing the ticket you mentioned in the evergreen PR:

> As part of this ticket, we want to edit the logic for displaying the tab, so that we show the tab to everyone who has modify permissions so they can go to the tab and add annotations, even if there are no annotations.

> Since after this we will be showing this tab to people who don't have the build baron enabled, many people will end up seeing and empty "Related Tickets From Jira" table. We want to edit that to only show it if we actually have related tickets to show.

> Also, since we will now show this tab to anyone with modify permissions, we want to show the AddIssue and AddSuspectedIssue buttons even if there are no issues or suspected issues in the list.

The first bullet point was already done, except for the issues / suspected issues part which is now added. I've hidden the build baron tab if there are no issues.
image
(The page looked the same without an annotation, except that there was no hello; I just wanted to check that it didn't error.)

src/pages/Task.tsx
ablack12: Why do we not like the get module?

src/components/BuildBaronAndAnnotations/BuildBaronContent.tsx
```diff
@@ -100,17 +100,20 @@ const BuildBaronCore: React.FC<BuildBaronCoreProps> = ({
               />
             </>
           )}
-
```
malikchaya2, line 103: Right now the File Ticket Button is only configured for teams who use the build baron (to be configured for other teams in other tickets). We don't want someone from other teams to attempt to use it. Since we will now show this tab to them, we should disable that button [here](https://github.com/evergreen-ci/spruce/blob/master/src/components/BuildBaronAndAnnotations/BBFIleTicket.tsx#L82) with a tooltip giving them some information.

malikchaya2 (changes requested):

src/pages/Task.tsx
SupaJoon: [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) fulfills our use case for `get` and doesn't require an additional installation

src/components/BuildBaronAndAnnotations/BuildBaronContent.tsx
ablack12: image
I think the tooltip might be confusing, since this will be eventually folded into task annotations. I think just hiding it for the moment will do the trick.

ablack12:
evergreen retry

SupaJoon:
@ablack12 the e2e test failure in patch-modal is not related to your code changes

src/components/BuildBaronAndAnnotations/BuildBaronContent.tsx
```diff
@@ -68,11 +68,13 @@ const BuildBaronCore: React.FC<BuildBaronCoreProps> = ({
             removeBanner={dispatchBanner.removeBanner}
           />
 
-          <CreatedTickets
-            taskId={taskId}
-            setCreatedTicketsCount={setCreatedTicketsCount}
-            createdTicketsCount={createdTicketsCount}
-          />
+          {bbData.buildBaronConfigured && (
```
malikchaya2, line 71: I don't know if we should be hiding this here because the component already checks if `length > 0`, and this will eventually be used outside of the build baron.

src/components/BuildBaronAndAnnotations/AnnotationTickets.tsx
```diff
@@ -54,22 +54,26 @@ export const AnnotationTickets: React.FC<Props> = ({
...
+        <>
+          <AnnotationTicketsTable
+            jiraIssues={tickets}
+            annotationId={annotationId}
+            taskId={taskId}
+            execution={execution}
+            isIssue={isIssue}
+            userCanModify={userCanModify}
+          />
+          <AddIssueModal
```
malikchaya2, line 67: Did you test it to make sure that the AddIssueModal shows up when the button is clicked? I'm not sure how this being hidden behind a `tickets?.length > 0` conditional affects it.

src/components/BuildBaronAndAnnotations/BuildBaronContent.tsx
ablack12: @malikchaya2 I addressed this in the other thread; I disagree that a tooltip is the right move, since this is currently not a feature that's available for projects that aren't configured. When it _is_ a task annotation feature, we can just make it visible at that point. Lara has agreed that this design is good, and in slack you said either a tooltip or hiding it was fine, so I'd like to move forward with this.

src/components/BuildBaronAndAnnotations/AnnotationTickets.tsx
ablack12: Good point, this should just cover the table. It's fixed now.

src/components/BuildBaronAndAnnotations/BuildBaronContent.tsx
malikchaya2: Discussed over slack- moving the check into the BBCreatedTickets component to only hide the button and title and not the list.

malikchaya2 (approved):

PR #597, opened 2021-01-14
EVG-13683 Add highlight to newly added issues

malikchaya2:
rowHighlight

### DIFF
--- a/src/components/BuildBaronAndAnnotations/AddIssueModal.tsx
+++ b/src/components/BuildBaronAndAnnotations/AddIssueModal.tsx
@@ -1,4 +1,4 @@
-import React, { useReducer } from "react";
+import React, { Dispatch, SetStateAction, useReducer } from "react";
 import { useMutation } from "@apollo/client";
 import styled from "@emotion/styled";
 import { Body } from "@leafygreen-ui/typography";
@@ -18,6 +18,7 @@ interface Props {
   visible: boolean;
   dataCy: string;
   closeModal: () => void;
+  setSelectedRowKeys: Dispatch<SetStateAction<string>>;
   taskId: string;
   execution: number;
   isIssue: boolean;
@@ -27,6 +28,7 @@ export const AddIssueModal: React.FC<Props> = ({
   visible,
   dataCy,
   closeModal,
+  setSelectedRowKeys,
   taskId,
   execution,
   isIssue,
@@ -95,6 +97,7 @@ export const AddIssueModal: React.FC<Props> = ({
     dispatch({
       type: "reset",
     });
+    setSelectedRowKeys(addIssueModalState.issueKey);
     closeModal();
   };
 
--- a/src/components/BuildBaronAndAnnotations/AddIssueModal.tsx
+++ b/src/components/BuildBaronAndAnnotations/AddIssueModal.tsx
@@ -1,4 +1,4 @@
-import React, { Dispatch, SetStateAction, useReducer } from "react";
+import React, { useReducer } from "react";
 import { useMutation } from "@apollo/client";
 import styled from "@emotion/styled";
 import { Body } from "@leafygreen-ui/typography";
@@ -18,7 +18,7 @@ interface Props {
   visible: boolean;
   dataCy: string;
   closeModal: () => void;
-  setSelectedRowKey: Dispatch<SetStateAction<string>>;
+  setSelectedRowKey: React.Dispatch<React.SetStateAction<string>>;
   taskId: string;
   execution: number;
   isIssue: boolean;
--- a/src/components/BuildBaronAndAnnotations/AddIssueModal.tsx
+++ b/src/components/BuildBaronAndAnnotations/AddIssueModal.tsx
@@ -18,7 +18,7 @@ interface Props {
   visible: boolean;
   dataCy: string;
   closeModal: () => void;
-  setSelectedRowKeys: Dispatch<SetStateAction<string>>;
+  setSelectedRowKey: Dispatch<SetStateAction<string>>;
   taskId: string;
   execution: number;
   isIssue: boolean;
@@ -28,7 +28,7 @@ export const AddIssueModal: React.FC<Props> = ({
   visible,
   dataCy,
   closeModal,
-  setSelectedRowKeys,
+  setSelectedRowKey,
   taskId,
   execution,
   isIssue,
@@ -83,6 +83,7 @@ export const AddIssueModal: React.FC<Props> = ({
       dispatchBanner.errorBanner(
         `There was an error adding the issue: ${error.message}`
       );
+      window.scrollTo(0, 0);
     },
     refetchQueries: ["GetTask"],
   });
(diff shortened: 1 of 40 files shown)


src/components/BuildBaronAndAnnotations/AnnotationTicketsTable.tsx
```diff
@@ -165,16 +167,38 @@ export const AnnotationTicketsTable: React.FC<Props> = ({
     moveAnnotation({ variables: { annotationId, apiIssue, isIssue } });
   };
 
+  // SCROLL TO added Issue
+  const rowRef = useRef<HTMLSpanElement>(null);
+
+  useEffect(() => {
+    if (selectedRowKeys && rowRef.current) {
+      rowRef.current.scrollIntoView({
```
SupaJoon, line 175: we should adjust when this function is called so we do not scroll when the mutation fails 

https://user-images.githubusercontent.com/10734386/104774683-1ca70580-5745-11eb-9c32-89c2d3d47534.mov

src/components/BuildBaronAndAnnotations/AnnotationTicketsTable.tsx
```diff
@@ -165,16 +167,38 @@ export const AnnotationTicketsTable: React.FC<Props> = ({
...
+        behavior: "smooth",
+        block: "start",
+      });
+    }
+  });
+
+  console.log(selectedRowKeys);
```
SupaJoon, line 182: leftover log

src/components/BuildBaronAndAnnotations/AnnotationTicketsTable.tsx
```diff
@@ -165,16 +167,38 @@ export const AnnotationTicketsTable: React.FC<Props> = ({
...
           columns={columns}
           pagination={false}
           showHeader={false}
+          rowSelection={{
+            hideSelectAll: true,
+            selectedRowKeys: [selectedRowKeys],
+            columnWidth: 0,
+            renderCell: () => <span ref={rowRef} />,
+          }}
+          rowClassName={(record, index) => (index === 0 ? "baseStyle" : "")}
```
SupaJoon, line 201: I don't think this prop is required

SupaJoon (changes requested):

src/components/BuildBaronAndAnnotations/AnnotationTicketsTable.tsx
```diff
@@ -27,6 +27,7 @@ interface Props {
   execution: number;
   isIssue: boolean;
   userCanModify: boolean;
+  selectedRowKeys: string;
```
SupaJoon, line 30: nit: can we make this variable name singular since it represents 1 thing

src/components/BuildBaronAndAnnotations/AnnotationTicketsTable.tsx
malikchaya2: Thank you for finding this!

malikchaya2: I added window.scrollTo(0, 0); to onError.

src/components/BuildBaronAndAnnotations/AddIssueModal.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { useReducer } from "react";
+import React, { Dispatch, SetStateAction, useReducer } from "react";
```
khelif96, line 1: We dont usually import these types directly 
We normal use them as such `React.Dispatch` `React.SetStateAction` https://github.com/evergreen-ci/spruce/blob/ea864962358c85134c190b7ee9951e1b18cef51f/src/pages/spawn/spawnHost/spawnHostModal/PublicKeyForm.tsx#L21

src/components/BuildBaronAndAnnotations/AnnotationTicketsTable.tsx
khelif96: I dont think we should do that. Since all thats doing is adding an additional scroll to mask the original scroll. It shouldn't scroll if there isn't anything to scroll to

khelif96 (changes requested):

src/components/BuildBaronAndAnnotations/AnnotationTicketsTable.tsx
SupaJoon: I think adding apollo error variable to the condition in this useEffect could tell us when it's appropriate to scroll

SupaJoon (changes requested):

src/components/BuildBaronAndAnnotations/AnnotationTicketsTable.tsx
malikchaya2: I tried nulling selectedRowKey when the mutation fails, it doesn't work. I'll try using the apollo error.

src/components/BuildBaronAndAnnotations/AnnotationTicketsTable.tsx
```diff
@@ -165,16 +167,35 @@ export const AnnotationTicketsTable: React.FC<Props> = ({
...
+  const rowRef = useRef<HTMLSpanElement>(null);
+
+  useEffect(() => {
+    if (selectedRowKey && rowRef.current) {
+      rowRef.current.scrollIntoView({
+        behavior: "smooth",
+        block: "start",
+      });
+    }
+  });
```
khelif96, line 180:
```suggestion
  // SCROLL TO added Issue
  // Will add a span with a ref to the row that matches the selectedRowKey
  // And will scroll to that ref.
  const rowRef = useRef<HTMLSpanElement>(null);

  useEffect(() => {
    if (selectedRowKey && rowRef.current) {
      rowRef.current.scrollIntoView({
        behavior: "smooth",
        block: "start",
      });
    }
  }, [selectedRowKey]);
```

src/components/BuildBaronAndAnnotations/AnnotationTicketsTable.tsx
```diff
@@ -165,16 +167,35 @@ export const AnnotationTicketsTable: React.FC<Props> = ({
...
           rowKey={({ issueKey }) => issueKey}
           columns={columns}
           pagination={false}
           showHeader={false}
+          rowSelection={{
+            hideSelectAll: true,
+            selectedRowKeys: [selectedRowKey],
+            columnWidth: 0,
+            renderCell: () => <span ref={rowRef} />,
+          }}
```
khelif96, line 198:
```suggestion
          rowSelection={{
            renderCell: (checked, record) =>
              record.issueKey === selectedRowKey && <span ref={rowRef} />,
            selectedRowKeys: [selectedRowKey],
          }}
```

src/components/BuildBaronAndAnnotations/AddIssueModal.tsx
```diff
@@ -81,6 +83,7 @@ export const AddIssueModal: React.FC<Props> = ({
       dispatchBanner.errorBanner(
         `There was an error adding the issue: ${error.message}`
       );
+      window.scrollTo(0, 0);
```
khelif96, line 86:
```suggestion
```

khelif96 (commented):
Heres how you can fix that scroll

khelif96 (approved):

SupaJoon (approved):

PR #601, opened 2021-01-15
EVG-13246 Task in Spruce host page links to the wrong URL

malikchaya2:


src/pages/host/getHostEventString.tsx
```diff
@@ -212,7 +212,7 @@ export const getHostEventString = (
           Assigned to run task{" "}
           <StyledLink
             data-cy="host-running-task-set-link"
-            href={`${getUiUrl()}/task${data.taskId}/${data.execution}`}
+            href={`${getUiUrl()}/task/${data.taskId}/${data.execution}`}
```
SupaJoon, line 215: are these links meant to point to spruce or the legacy ui?

malikchaya2: They currently link to legacy, but there is probably no reason for that. Should I switch them to Spruce?

malikchaya2: fixed

src/pages/host/Metadata.tsx
```diff
@@ -26,7 +26,7 @@ export const Metadata: React.FC<{
   const runningTaskId = runningTask?.id;
   const runningTaskName = runningTask?.name;
 
-  const taskLink = `${getUiUrl()}/task/${runningTaskId}`;
+  const taskLink = `/task/${runningTaskId}`;
```
khelif96, line 29: We have a utility function for this
https://github.com/evergreen-ci/spruce/blob/82684a550fe376bf2661be5fb74c4cd911941707/src/constants/routes.ts#L94

src/pages/host/getHostEventString.tsx
```diff
@@ -225,7 +224,7 @@ export const getHostEventString = (
           Current running task cleared (was:
           <StyledLink
             data-cy="host-running-task-cleared-link"
-            href={`${getUiUrl()}/task${data.taskId}/${data.execution}`}
+            href={`/task/${data.taskId}/${data.execution}`}
```
khelif96, line 227: Could you use the utility function i mentioned above for these use cases. You will need to modify it to also take in execution as an optional paramater

khelif96 (changes requested):

src/pages/host/getHostEventString.tsx
```diff
@@ -212,7 +211,7 @@ export const getHostEventString = (
           Assigned to run task{" "}
           <StyledLink
             data-cy="host-running-task-set-link"
-            href={`${getUiUrl()}/task${data.taskId}/${data.execution}`}
+            href={`/task/${data.taskId}/${data.execution}`}
```
khelif96, line 214: Also this url pattern isn't correct for tasks.
Task urls follow the following pattern
`/task/:id/:tab?url_params`
url_params being `execution` in this case

src/pages/host/getHostEventString.tsx
malikchaya2: I actually won't use execution because it's null

khelif96 (approved):

PR #602, opened 2021-01-19
EVG-13596: Project defined startup script box

SupaJoon:
https://jira.mongodb.org/browse/EVG-13596
these code changes hook up the "project defined startup script" and "load from task sync". 
These code changes also set the "load data onto host on startup" checkbox to checked since that happens in EVG whenever a task id is passed to the spawn host gql mutation. 
Test these changes locally at: localhost:3000/spawn/host?spawnHost=True&distroId=rhel71-power8-large&taskId=evergreen_ubuntu1604_dist_patch_33016573166a36bd5f46b4111151899d5c4e95b1_5ecedafb562343215a7ff297_20_05_27_21_39_46
Screen Shot 2021-01-19 at 10 51 12 AM


ablack12:
It looks like load data from task is now required if the host is spawned from a task. I thought we discussed not making this the case, because users may want to do "start any hosts this task started" without having to load data from the task (I may be remembering the wrong thing, though). If we _do_ decide to make load data from task required, then I don't think this should be a checkbox -- it's confusing that I can't unclick it (I'm not sure what it should be instead though), and this checkbox wouldn't be visible if we weren't spawning from a task, so it has no use really.

src/pages/spawn/spawnHost/spawnHostModal/SetupScriptForm.tsx
```diff
@@ -25,31 +30,37 @@ export const SetupScriptForm: React.FC<SetupScriptFormProps> = ({
...
+        type: "ingestQueryParams",
         taskId: getString(taskId),
-        useProjectSetupScript: true,
         distroId: getString(distroId),
       });
     }
   }, [taskId, distroId, getTask, onChange]);
 
-  const { displayName, buildVariant, revision } = taskData?.task || {};
+  console.log(taskData);
```
john-m-liu, line 61: left some debug code

john-m-liu (approved):
I took a look and the project-specific setup script, task sync, and "also start hosts..." checkboxes do depend on checking the load task data checkbox. I think it's not too hard to fix, so if doing that would make your life easier let me know

SupaJoon:
I adjusted the checkbox behavior so project-specific setup script, task sync, and "also start hosts..." rely on the load task data checkbox. Task Id error checking takes place in the **[resolver](https://github.com/evergreen-ci/evergreen/pull/4317/files#diff-a3ce50f20b746b196f29c6d414e685b0110c24b768bcf83756de8d842c98f6dcR435-R441).**  
https://user-images.githubusercontent.com/10734386/105219707-c19a5780-5b24-11eb-8f17-636c489905ba.mov

ablack12 (approved):

src/pages/spawn/spawnHost/spawnHostModal/SetupScriptForm.tsx
```diff
@@ -25,31 +30,37 @@ export const SetupScriptForm: React.FC<SetupScriptFormProps> = ({
...
+    useProjectSetupScript,
+    spawnHostsStartedByTask,
+    taskSync,
+  } = data;
 
   const { search } = useLocation();
   const queryParams = parseQueryString(search);
-  const { taskId, distroId } = queryParams;
+  const { distroId } = queryParams;
+  const taskIdQueryParam = getString(queryParams.taskId);
```
khelif96, line 43: why was this renamed?

khelif96 (changes requested):
i'm seeing this error in the console when ever i click on the distro input 
image

src/pages/spawn/spawnHost/SpawnHostModal.tsx
```diff
@@ -141,22 +140,38 @@ export const SpawnHostModal: React.FC<SpawnHostModalProps> = ({
...
+    [distrosData?.distros, distroInput]
+  );
+  const notVirtualWorkstationDistros = useMemo(
+    () =>
+      (distrosData?.distros ?? [])
+        .filter((d) => d.name.includes(distroInput))
+        .filter((d) => !d.isVirtualWorkStation)
+        .sort((a, b) => a.name.localeCompare(b.name)),
+    [distrosData?.distros, distroInput]
+  );
```
khelif96, line 158: We actually dont need to recalculate this every time. The input element automatically handles filtering the inputs. This is likely also the reason the error above also shows. Since we are changing the elements in the list every-time we enter a new character.

khelif96: This should be reverted to what we were doing before.

src/pages/spawn/spawnHost/SpawnHostModal.tsx
```diff
@@ -141,22 +140,38 @@ export const SpawnHostModal: React.FC<SpawnHostModalProps> = ({
...
+  // recalculate isVirtualWorkstation whenever distro changes
+  // initial distroId can be changed from URL
+  useEffect(() => {
+    dispatch({
+      type: "editDistroEffect",
+      isVirtualWorkstation: !!virtualWorkstationDistros.find(
+        (vd) => distroId === vd.name
+      ),
+    });
+  }, [distroId, dispatch, virtualWorkstationDistros]);
```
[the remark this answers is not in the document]
SupaJoon: I also update the distro from the `ingestQueryParams` action. If I follow the pattern you describe, the component that dispatches the `ingestQueryParams` will also need access to `virtualWorkstationDistros` from useQuery (or readQuery) and will also need to apply this logic again: 
```
    isVirtualWorkstation: !!virtualWorkstationDistros.find(
        (vd) => distroId === vd.name
      ),
```
If I adjust the props to PublicKeyForm to only accept a subset of values for spawnHostModalState I think we should expect very similar dom updates with only 1 additional call to render and we will get the benefit of not having to maintain isVirtualWorkstation across multiple actions. what do you think?

src/pages/spawn/spawnHost/spawnHostModal/SetupScriptForm.tsx
SupaJoon: I was getting confused between the taskId from from reducer state and the taskId from the URL params

src/pages/spawn/spawnHost/spawnHostModal/useSpawnHostModalState.ts
```diff
@@ -61,15 +85,18 @@ const reducer = (state: spawnHostState, action: Action) => {
...
-        noExpiration: isVirtualWorkstation,
       };
     }
+    case "editDistroEffect":
+      return {
+        ...state,
+        isVirtualWorkStation: action.isVirtualWorkstation,
+        homeVolumeSize: action.isVirtualWorkstation ? 500 : null,
+        noExpiration: action.isVirtualWorkstation,
+      };
```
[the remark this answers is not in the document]
SupaJoon: let me know what you think about this comment: https://github.com/evergreen-ci/spruce/pull/602/files#r567001955

SupaJoon:
evergreen-retry

SupaJoon:
evergreen retry

src/pages/spawn/spawnHost/SpawnHostModal.tsx
```diff
@@ -141,30 +140,45 @@ export const SpawnHostModal: React.FC<SpawnHostModalProps> = ({
...
+    .filter((d) => d.name.includes(distroInput))
     .sort((a, b) => a.name.localeCompare(b.name));
 
+  const virtualWorkstationDistrosMinusCurrentSelection = allDistroOptionsMinusCurrentSelection.filter(
+    (d) => d.isVirtualWorkStation
+  );
+
+  const notVirtualWorkstationDistrosMinusCurrentSelection = allDistroOptionsMinusCurrentSelection.filter(
+    (d) => !d.isVirtualWorkStation
+  );
```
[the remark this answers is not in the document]
SupaJoon: what I have actually makes the most sense for me otherwise it is easy to miss that the lists derived from `fetchedDistros.filter((d) => d.name.includes(distroInput))` excludes the current selection

src/pages/spawn/spawnHost/SpawnHostModal.tsx
```diff
@@ -189,9 +203,9 @@ export const SpawnHostModal: React.FC<SpawnHostModalProps> = ({
     dispatch({
       type: "editDistro",
       distroId: d,
-      isVirtualWorkstation: !!virtualWorkstationDistros.find(
-        (vd) => d === vd.name
-      ),
+      isVirtualWorkstation: !!distrosData?.distros.find(
+        (vd) => distroId === vd.name
+      )?.isVirtualWorkStation,
```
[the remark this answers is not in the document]
SupaJoon: I'm okay with this actually. The subset you are referring to includes all virtual workstations aside from the current selection. This could seem like a stretch, but I'd like to avoid an edgecases that could result in the same virtual workstation distro being selected twice in a row and on the second time doesn't register as a virtual workstation bc we filtered out the current selection.

src/pages/spawn/spawnHost/SpawnHostModal.tsx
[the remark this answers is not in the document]
SupaJoon: @khelif96 I realized I was wrong about the `excludes the current selection` bit. I'll change the name to "filteredDistros"

src/pages/spawn/spawnHost/SpawnHostModal.tsx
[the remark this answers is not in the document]
SupaJoon: @khelif96 I realized that the smaller list is filtered to match the current distroInput. that means if the "editDistro" will often set isVirtualWorkstation to false if this function was called twice in a row with differing distroIds. 
However, I realized that the useEffect introduced in this file to recalc isVirtualWorkstation would correct the form state

src/pages/spawn/spawnHost/spawnHostModal/SetupScriptForm.tsx
SupaJoon: I wasn't able to access queryParams.taskId as a string without checking it's type. queryParams is a ParsedString https://github.com/evergreen-ci/spruce/blob/master/src/utils/queryString/parseQueryString.ts#L4 
so taskId is a `string | string[]`

cypress/integration/spawn/host.ts
```diff
@@ -130,8 +130,57 @@ describe("Navigating to Spawn Host page", () => {
...
+
+      it("If 'Load data for dist' is unchecked, selecting one of it's children will check it.'", () => {
+        cy.visit(
+          `spawn/host?spawnHost=True&distroId=rhel71-power8-large&taskId=${taskId}`
+        );
+        cy.wait(200);
+        cy.dataCy("parent-checkbox").then((chk) => {
+          if (chk[0].checked) {
+            cy.dataCy("parent-checkbox").click({ force: true });
+          }
```
khelif96, line 143: Why is there a conditional in the test? Shouldn't there be an assumption that the form should be in the correct state?

src/pages/spawn/spawnHost/SpawnHostModal.tsx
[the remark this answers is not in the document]
SupaJoon: @khelif96 I decided not to adjust public key form props since the refactor may require additional testing to ensure we do not run into undefined value errors

cypress/integration/spawn/host.ts
SupaJoon: i wanted to decouple this test from relying on initial form state, since that could change the outcome of this test

khelif96 (approved):
After making the changes we spoke about

PR #604, opened 2021-01-20
EVG-13523 Task annotations analytics

malikchaya2:


src/analytics/task/useTaskAnalytics.ts
```diff
@@ -40,7 +40,17 @@ type Action =
...
+  | { name: "Save Annotation Note" }
+  | { name: "Click Annotation Ticket Link" }
+  | { name: "Remove Annotation Issue" }
+  | { name: "Remove Annotation Suspected Issue" }
+  | { name: "Move Annotation Issue" }
+  | { name: "Move Annotation Suspected Issue" }
+  | { name: "Click Add Annotation Issue Button" }
+  | { name: "Click Add Annotation Suspected Issue Button" }
+  | { name: "Add Task Annotation Issue" }
+  | { name: "Add Task Annotation Suspected Issue" };
```
khelif96, line 53: Maybe these should live in a separate build baron analytics hook?

src/gql/generated/types.ts
```diff
@@ -1212,12 +1211,76 @@ export type GetTaskEventDataQueryVariables = Exact<{
 export type GetTaskEventDataQuery = {
   task?: Maybe<{
     id: string;
-    execution: number;
+    execution?: Maybe<number>;
```
khelif96, line 1214: i think the `schema.graphql` you generated this with is out of date this field is always there. Pull the latest evergreen master and run `yarn codegen` again

src/analytics/task/query.ts
```diff
@@ -10,3 +10,74 @@ export const GET_TASK_EVENT_DATA = gql`
     }
   }
 `;
+
+export const GET_ANNOTATION_EVENT_DATA = gql`
```
khelif96, line 14: Are all these fields necessary for the data needed for the annotation analytics?

src/analytics/task/useTaskAnalytics.ts
khelif96, line 53: Thse should be removed from here since they already exist in the useAnnnotationAnalytics hook

src/components/BuildBaronAndAnnotations/AddIssueModal.tsx
```diff
@@ -33,6 +34,7 @@ export const AddIssueModal: React.FC<Props> = ({
   execution,
   isIssue,
 }) => {
+  const taskAnalytics = useTaskAnalytics();
```
khelif96, line 37: This should be replaced with the useAnnotationAnalytics hook you created.

khelif96 (changes requested):

src/analytics/task/query.ts
malikchaya2: I spoke to Maria and we figure out which fields can be removed. Thanks!

malikchaya2:
@khelif96  I accidentally combined my original changes with the pr changes, my apologies. Hopefully it's not too hard to follow.

malikchaya2:
The save note field event can't be tested currently, but @ablack12  will test it after completing [EVG-13815](https://jira.mongodb.org/browse/EVG-13815)

src/analytics/task/useAnnotationAnalytics.ts
```diff
@@ -0,0 +1,77 @@
...
+      variables: { taskId: id, execution },
+      fetchPolicy: "cache-first",
+    }
+  );
+
+  const annotation = get(eventData, "task.annotation", undefined);
+  const bbConfigured = get(
+    bbData,
+    "buildBaron.buildBaronConfigured",
+    undefined
```
khelif96, line 64: could we use `eventData?.task?.annotation` instead we wanted to move away from _.get

khelif96 (approved):

PR #606, opened 2021-01-20
EVG-13571: Configure Page: Prevent text selection when using hot keys

SupaJoon:
https://jira.mongodb.org/browse/EVG-13571
Before & After

https://user-images.githubusercontent.com/10734386/105244709-e69bc400-5b3e-11eb-98ba-331e9a6e9509.mov

https://user-images.githubusercontent.com/10734386/105244730-ee5b6880-5b3e-11eb-9ccc-ba47dd801101.mov

### DIFF
--- a/src/pages/configurePatch/configurePatchCore/ConfigureBuildVariants.tsx
+++ b/src/pages/configurePatch/configurePatchCore/ConfigureBuildVariants.tsx
@@ -81,42 +81,44 @@ export const ConfigureBuildVariants: React.FC<Props> = ({
     }
   };
   return (
-    <StyledSiderCard isHotKeyPressed={state.numButtonsPressed !== 0}>
-      <Container>
-        <Body weight="medium">Select Build Variants and Tasks</Body>
-        <Divider />
-      </Container>
-      {variants.map(({ displayName, name }) => {
-        const taskCount = selectedVariantTasks[name]
-          ? Object.values(selectedVariantTasks[name]).filter((v) => v).length
-          : null;
-        const isSelected = selectedBuildVariant.includes(name);
-        return (
-          <BuildVariant
-            data-cy="configurePatch-buildVariantListItem"
-            data-cy-name={name}
-            data-cy-selected={isSelected}
-            key={name}
-            isSelected={isSelected}
-            onClick={getClickVariantHandler(name)}
-          >
-            <VariantName>
-              <Body weight={isSelected ? "medium" : "regular"}>
-                {displayName}
-              </Body>
-            </VariantName>
-            {taskCount > 0 && (
-              <StyledBadge
-                data-cy={`configurePatch-taskCountBadge-${name}`}
-                variant={isSelected ? Variant.DarkGray : Variant.LightGray}
-              >
-                {taskCount}
-              </StyledBadge>
-            )}
-          </BuildVariant>
-        );
-      })}
-    </StyledSiderCard>
+    <UserSelectWrapper isHotKeyPressed={state.numButtonsPressed !== 0}>
+      <StyledSiderCard>
+        <Container>
+          <Body weight="medium">Select Build Variants and Tasks</Body>
+          <Divider />
+        </Container>
+        {variants.map(({ displayName, name }) => {
+          const taskCount = selectedVariantTasks[name]
+            ? Object.values(selectedVariantTasks[name]).filter((v) => v).length
+            : null;
+          const isSelected = selectedBuildVariant.includes(name);
+          return (
+            <BuildVariant
+              data-cy="configurePatch-buildVariantListItem"
+              data-cy-name={name}
+              data-cy-selected={isSelected}
+              key={name}
+              isSelected={isSelected}
+              onClick={getClickVariantHandler(name)}
+            >
+              <VariantName>
+                <Body weight={isSelected ? "medium" : "regular"}>
+                  {displayName}
+                </Body>
+              </VariantName>
+              {taskCount > 0 && (
+                <StyledBadge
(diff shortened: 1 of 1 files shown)


src/pages/configurePatch/configurePatchCore/ConfigureBuildVariants.tsx
```diff
@@ -56,7 +81,7 @@ export const ConfigureBuildVariants: React.FC<Props> = ({
     }
   };
   return (
-    <StyledSiderCard>
+    <StyledSiderCard isHotKeyPressed={state.numButtonsPressed !== 0}>
```
SupaJoon, line 84: I get a react console warning because isHotKeyPressed gets propagated to the dom. Do ya'll know a workaround for this?

khelif96 (changes requested):
This can be done with one line of css 
Add `user-select: none;` to The styles for `VariantName`
https://user-images.githubusercontent.com/4605522/105525195-bc691400-5cae-11eb-8585-d7d6281ff2df.mp4

SupaJoon:
> This can be done with one line of css
> Add `user-select: none;` to The styles for `VariantName`
> https://user-images.githubusercontent.com/4605522/105525195-bc691400-5cae-11eb-8585-d7d6281ff2df.mp4

We tried a similar solution the past and we reverted it after processing user feedback: https://mongodb.slack.com/archives/C014R0J5XM4/p1596214127255600

src/pages/configurePatch/configurePatchCore/ConfigureBuildVariants.tsx
```diff
@@ -23,6 +23,31 @@ export const ConfigureBuildVariants: React.FC<Props> = ({
...
+    [dispatch]
+  );
+  useEffect(() => {
+    window.addEventListener("keydown", keyDownCb);
+    window.addEventListener("keyup", keyUpCb);
+    return () => {
+      window.removeEventListener("keyDown", keyDownCb);
+      window.removeEventListener("keyup", keyUpCb);
+    };
+  }, [keyUpCb, keyDownCb]);
```
khelif96, line 50: I still think this use reducer is a little overkill. Using a useState would be simpler and just as effective.

```suggestion
 const [isHotKeyPressed, setIsHotKeyPressed] = useState(false);

  useEffect(() => {
    const keyDownCb = (e: KeyboardEvent) => {
      if (hotKeys.has(e.key)) {
        setIsHotKeyPressed(true);
      }
    };
    const keyUpCb = (e: KeyboardEvent) => {
      if (hotKeys.has(e.key)) {
        setIsHotKeyPressed(false);
      }
    };

    window.addEventListener("keydown", keyDownCb);
    window.addEventListener("keyup", keyUpCb);
    return () => {
      window.removeEventListener("keyDown", keyDownCb);
      window.removeEventListener("keyup", keyUpCb);
    };
  });
```

src/pages/configurePatch/configurePatchCore/ConfigureBuildVariants.tsx
```diff
@@ -95,9 +120,32 @@ export const ConfigureBuildVariants: React.FC<Props> = ({
...
+        numButtonsPressed: state.numButtonsPressed + 1,
+      };
+    case "decrement":
+      return {
+        numButtonsPressed: state.numButtonsPressed - 1,
+      };
+    default:
+      throw new Error();
+  }
+};
```
khelif96, line 141:
```suggestion
```

src/pages/configurePatch/configurePatchCore/ConfigureBuildVariants.tsx
```diff
@@ -106,11 +154,13 @@ export const cardSidePadding = css`
 const Container = styled.div`
   ${cardSidePadding}
 `;
-const StyledSiderCard = styled(SiderCard)`
+const StyledSiderCard = styled(SiderCard)<StyledSiderCardProps>`
```
khelif96, line 157: The issue you mentioned about the error with the props is due to how leafygreen uses emotion that messes with how props are propagated. They provide a custom version of emotion to fix this. 
https://github.com/mongodb/leafygreen-ui/tree/main/packages/emotion

An alternative is to just move the props to another non leafygreen element

src/pages/configurePatch/configurePatchCore/ConfigureBuildVariants.tsx
khelif96: See my comment below

src/pages/configurePatch/configurePatchCore/ConfigureBuildVariants.tsx
SupaJoon: we have to increment a decrement a number because it is possible the user is pressing multiple hotkeys at once. holding shift, then holding command, then releasing shift will result in `isHotKeyPressed` to be false

khelif96 (approved):
Pending the changes we discussed

src/pages/configurePatch/configurePatchCore/ConfigureBuildVariants.tsx
SupaJoon: @khelif96 I'm going to keep the reducer as is. it's possible that we will need to implement a counter state in another component. In that case, the useReducer implementation is readily typed and reusable

src/pages/configurePatch/configurePatchCore/ConfigureBuildVariants.tsx
SupaJoon: went with the alternative

PR #646, opened 2021-02-08
EVG-13759 Eject Spruce and customize build process

khelif96:
Key Changes:
Spruce is now ejected and we have control of our build pipeline

Added native support for `*.graphql` files
Now we can write our queries in a `*.graphql` file and utilize more advanced features like fragments more effectively. 

Added [babel-plugin-import](https://www.npmjs.com/package/babel-plugin-import) which helps optimize bundling of antd to reduce package sizes. Also auto imports the css files for antd components we use so its no longer a manual process. 

Added better emotion support so styled components should now build better source maps which should allow for easier debugging in the browser. As well as providing better support for some emotion features like the css prop

Added better support for  web workers. They can live with in the src/ folder and can thus be more thoroughly tested. They should also support dependencies and typescript. (I haven't tested this out) 

One change is the location of `.cmdrc.json` it was moved to the `/env` folder since the `/config` folder was needed by webpack. so you will have to manually move your `.cmdrc.json` to `/env`


package.json
```diff
@@ -36,6 +90,7 @@
     "deep-object-diff": "1.1.0",
     "env-cmd": "10.1.0",
     "graphql": "15.5.0",
+    "graphql-tag": "^2.11.0",
```
SupaJoon, line 93: why is this needed

SupaJoon: I see that graphql-tag/loader is utilized in webpack.config. maybe this can be a dev dep?

khelif96: This is used to load graphql documents. We use the `loader` function in our webpack build

yarn.lock
```diff
@@ -1225,6 +1269,17 @@
...
+  resolved "https://registry.yarnpkg.com/@babel/preset-react/-/preset-react-7.12.13.tgz#5f911b2eb24277fa686820d5bd81cad9a0602a0a"
+  integrity sha512-TYM0V9z6Abb6dj1K7i5NrEhA13oS5ujUYQYDfqIBXYHOc2c2VkFgc+q9kyssIyUfy4/hEwqrgSlJ/Qgv8zJLsA==
+  dependencies:
+    "@babel/helper-plugin-utils" "^7.12.13"
+    "@babel/plugin-transform-react-display-name" "^7.12.13"
+    "@babel/plugin-transform-react-jsx" "^7.12.13"
+    "@babel/plugin-transform-react-jsx-development" "^7.12.12"
+    "@babel/plugin-transform-react-pure-annotations" "^7.12.1"
+
 "@babel/preset-typescript@7.12.1":
```
SupaJoon, line 1283: should this have been included in the .babelrc preset list?

SupaJoon: I see that [this line](https://github.com/evergreen-ci/spruce/pull/646/files#diff-c49a22733230fd83e34ff3bc78014ee352ad46ce10c7c502dd2a7c3982adb0a5R6) imports a utility that allows us to to transpile ts

SupaJoon (approved):

PR #656, opened 2021-02-16
EVG-13991 Convert all graphql queries and mutations to `*.graphql` files

khelif96:
This is a super long pr sorry. 
I tried to organize each commit based off of what it does to make it a little easier to review.
This pr does a few key things

640b0dd Support dynamic imports in `graphql` files. Unfortunately only relative imports are supported for now.

Convert all our gueries to graphql files
Convert all our mutations to graphql files
Extract common fields and turn them into fragments

de7487b Utilize the apollo cache to update some fields and remove query refetches for the tasks page


SupaJoon (approved):

package.json
```diff
@@ -27,7 +27,7 @@
     "prettier": "prettier --write",
     "prod": "env-cmd -e devProd -r env/.cmdrc.json yarn start",
     "staging": "env-cmd -e devStaging -r env/.cmdrc.json yarn start",
-    "start": "node scripts/start.js",
+    "start": "rm -rf ./node_modules/.cache/babel-loader && node scripts/start.js",
```
SupaJoon, line 30: what happens if we don't do `rm -rf ./node_modules/.cache/babel-loader `

src/components/PatchesPage/patchCard/DropdownMenu.tsx
```diff
@@ -57,4 +57,4 @@ export const DropdownMenu: React.FC<Props> = ({
   );
 };
 
-const refetchQueries = ["PatchBuildVariantsAndStatus"];
+const refetchQueries = ["patchBuildVariants"];
```
SupaJoon, line 60: I think this should be PatchBuildVariants

package.json
khelif96: Its recommended in the docs for the graphql loader plugin. I think it wont always catch some changes to queries when hot reloading.
https://www.npmjs.com/package/babel-plugin-import-graphql

PR #659, opened 2021-02-17
EVG-13990 Fix storybook and setup useful addons

khelif96:
Fix storybook setup.

Colocate our stories with their components

Adds a storybook task which builds and uploads storybook to s3 to help with easy code reviews.

also adds a few useful plugins

a11y- which helps with building accessible components.
knobs - lets you interact with storybook components by providing an easy interface to change props
storyshots-  capture screenshots of stories to help catch unintentional changes.
actions- Like knobs but lets you know if component callbacks were called.
queryparams - allow mocking queryparams in stories


SupaJoon:
do you think we should make a follow up ticket to capture screenshots of our stories in jest.
im also wondering how you imagine we begin using actions. do you think we should begin unit testing our shared components in storybook?

SupaJoon (approved):

PR #661, opened 2021-02-22
EVG-14073 fix annotation note metadata styling

ablack12:
image

### DIFF
--- a/cypress/integration/project/patches.ts
+++ b/cypress/integration/project/patches.ts
@@ -21,7 +21,6 @@ describe("Project Patches Page", () => {
     cy.visit(route);
     cy.dataCy("patches-page-title").contains("evergreen smoke test Patches");
     const patchDisplayNames = [
-      "Commit Queue Merge: No Commits Added",
       "dist",
       "dist",
       "test meee",
--- a/cypress/integration/project/patches.ts
+++ b/cypress/integration/project/patches.ts
@@ -21,6 +21,7 @@ describe("Project Patches Page", () => {
     cy.visit(route);
     cy.dataCy("patches-page-title").contains("evergreen smoke test Patches");
     const patchDisplayNames = [
+      "main: EVG-7823 add a commit queue message (#4048)",
       "dist",
       "dist",
       "test meee",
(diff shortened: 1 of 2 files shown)


khelif96 (approved):

khelif96:
The tests are failing can you please update this test along side this pr. https://github.com/evergreen-ci/spruce/blob/82684a550fe376bf2661be5fb74c4cd911941707/cypress/integration/project/patches.ts#L24
I think the patch name just needs to be updated

malikchaya2 (changes requested):
I don't know how easy it is to accomplish this, but I think ideally we should try to leave less space between the two things displayed, and not wrap the line (closer to how it's shown on the design). I also noticed that the the time and author are flipped from what the design shows. Was that done on purpose (either seem fine to me)?
image

ablack12:
> I don't know how easy it is to accomplish this, but I think ideally we should try to leave less space between the two things displayed, and not wrap the line (closer to how it's shown on the design). 

I tweaked it a little bit (and tweaked the ticket metadata, which also wraps pretty frequently). If the name was even _longer_ than this it could still wrap, but for the regular case this will look better (also if the name is any longer than this I'll be really impressed haha).
image

> I also noticed that the the time and author are flipped from what the design shows. Was that done on purpose (either seem fine to me)?

This was done on purpose, in order to more closely match the jira ticket metadata, and was discussed at the time.

ablack12:
> I think the patch name just needs to be updated

@khelif96 it looks like the patch isn't on that page; I'm not sure what would've changed but? I removed it.

malikchaya2 (approved):

ablack12:
https://github.com/evergreen-ci/evergreen/pull/4422 this should fix the spruce test; @khelif96 can I merge?

khelif96:
> [evergreen-ci/evergreen#4422](https://github.com/evergreen-ci/evergreen/pull/4422) this should fix the cypress test; @khelif96 can I merge?

Can you merge the test data fixes and try to rerun the tests?

cypress/integration/project/patches.ts
```diff
@@ -21,7 +21,6 @@ describe("Project Patches Page", () => {
     cy.visit(route);
     cy.dataCy("patches-page-title").contains("evergreen smoke test Patches");
     const patchDisplayNames = [
-      "Commit Queue Merge: No Commits Added",
       "dist",
```
khelif96, line 24:
```suggestion
      "Commit Queue Merge:"
      "dist",
```

PR #678, opened 2021-03-01
EVG-13633 Refactor slack notification banner and remove dead banner code.

khelif96:


src/components/Banners/SlackNotificationBanner.tsx
```diff
@@ -103,114 +80,67 @@ export const SlackNotificationBannerCore = () => {
...
-const Wrapper = styled.div`
-  display: flex;
-  align-items: center;
-  font-size: 14px;
-`;
-
-const StyledIcon = styled(Icon)`
-  margin-right: 16px;
+const isNotificationSet = (field: string) =>
+  field !== "" && field !== undefined;
```
SupaJoon, line 131: this is the same as `!!field`

src/components/Banners/SlackNotificationBanner.tsx
```diff
@@ -77,18 +33,39 @@ export const SlackNotificationBannerCore = () => {
...
+  const { data: userSettingsData } = useQuery<GetUserSettingsQuery>(
+    GET_USER_SETTINGS
+  );
+  const { userSettings } = userSettingsData || {};
+  const { slackUsername: defaultSlackUsername, notifications } =
+    userSettings || {};
+  const { patchFinish, patchFirstFailure } = notifications || {};
+
+  const [slackUsername, setSlackUsername] = useState(
+    () => defaultSlackUsername
```
SupaJoon, line 56: `() =>` seems like a typo

SupaJoon (approved):

src/components/Banners/SlackNotificationBanner.tsx
khelif96: It is not a typo passing a function into useState allows you to lazily init the state. Doing this basically says set the initial state when defaultSlackUsername becomes defined. Allows us to avoid a useEffect to update the state when `defaultSlackUsername` becomes defined.

khelif96: https://kentcdodds.com/blog/use-state-lazy-initialization-and-function-updates/

PR #691, opened 2021-03-18
EVG-14020 Sort failed test logs to top

khelif96:


src/pages/task/taskTabs/testsTable/TestsTableCore.tsx
```diff
@@ -44,6 +44,12 @@ export const TestsTableCore: React.FC = () => {
   const queryVariables = getQueryVariables(search, resourceId);
   const { cat, dir, pageNum, limitNum } = queryVariables;
 
+  if (cat === undefined) {
```
SupaJoon, line 47: we should check this condition and call updateQueryParams in a useEffect. that lets us be more intentional about calling updateQueryParams since the useEffect will run only when the specified value change

SupaJoon (changes requested):

src/pages/task/taskTabs/testsTable/TestsTableCore.tsx
khelif96: Good catch

src/pages/task/taskTabs/testsTable/TestsTableCore.tsx
```diff
@@ -44,6 +44,14 @@ export const TestsTableCore: React.FC = () => {
...
   const { cat, dir, pageNum, limitNum } = queryVariables;
 
+  useEffect(() => {
+    if (cat === undefined) {
+      updateQueryParams({
+        [RequiredQueryParams.Category]: TestSortCategory.Status,
+        [RequiredQueryParams.Sort]: SortDirection.Asc,
+      });
+    }
+  }, []); // eslint-disable-line react-hooks/exhaustive-deps
```
SupaJoon, line 54: I think it's okay to add cat to the dep array since it should never be omitted

SupaJoon (approved):

src/pages/task/taskTabs/testsTable/TestsTableCore.tsx
khelif96: I purposefully omitted it so that we only default sort status if its the first time the component renders. We dont want to force the user to sort by status if they want to remove all sorts intentionally.

PR #695, opened 2021-03-29
EVG-14215 check for valid pr number

john-m-liu:


src/pages/commitqueue/CommitQueueCard.tsx
```diff
@@ -73,7 +73,7 @@ export const CommitQueueCard: React.FC<Props> = ({
       <CommitQueueCardGrid>
         {patchId ? (
           <CommitInfo>
-            {versionId !== "" || issue === "" ? (
+            {versionId !== "" || issue === "" || Number.isNaN(Number(issue)) ? (
```
khelif96, line 76: I don't get the point of this? converting an issue to a number and then checking if its not a real number?

john-m-liu: it should return false if issue contains a value that can convert to a number like "123", but true otherwise like "abc"

khelif96 (approved):

PR #696, opened 2021-03-29
EVG-14181 Create JSON Schema generated forms POC 

khelif96:
The newly introduced `<SpruceForm />` components allows you to dynamically create forms which automatically handle managing their own state. 

I opted to use the `react-jsonschema-form` [library](https://github.com/rjsf-team/react-jsonschema-form) as the basis for this implementation. I chose this library for its extensive community support as well as detailed [documentation](https://react-jsonschema-form.readthedocs.io/en/latest/) and [playground](https://rjsf-team.github.io/react-jsonschema-form/).

We can now define a structured json schema that will will output a form in react consisting of leafygreen components and our own custom components. I've only added a few for now and more can be added as needed.
This has the advantage of allowing us to easily create and modify forms without having to worry about any messy state management. There are also useful features such as adding automatic validation, and creating field dependencies. 

There are [2 storybook stories](https://mciuploads.s3.amazonaws.com/spruce/spruce_ubuntu1604_storybook_patch_788731f40431670f4137acde3f0e8322d28cbeae_60620b6f61837d55d05dbbb0_21_03_29_17_16_32/storybook/index.html?path=/story/spruce-form--example-1) created to demo this component.

I've tagged everyone that i think might be interested or has expressed interest in this feel free to untag yourself.


john-m-liu:
Is the main place where you expect this to be useful in the admin/distro workflows? I imagine for workflows that end users interact with, we want more control over the layout

khelif96:
> Is the main place where you expect this to be useful in the admin/distro workflows? I imagine for workflows that end users interact with, we want more control over the layout

Yeah this was the main motivation behind this. There were dozens of forms that would be pretty tedious to manually write. 
The library has some pretty powerful features  that allow us to conditionally render fields based off of certain conditions as well as defining relationships between fields.
 (For example if Field A is populated then Field B is now required or if Field C is Checked then Display Field D)

src/utils/array/index.ts
```diff
@@ -16,6 +16,7 @@ export const convertArrayToObject = <T>(
   key: string
 ): { [key: string]: T } => {
   const initialValue = {};
+  if (array === undefined) return initialValue;
```
SupaJoon, line 19: when is this condition satisfied from an example in our codebase

SupaJoon (approved):

malikchaya2 (commented):
I see that the form is inside of a card with a title above it. Is the greater goal to use a few of these together on a page? I wonder if there would be any merit to maybe making it more generic with no card (or with the card being optional) so that it can be used in more designs, including in modals? I'm thinking it might make it more portable since adding styles is easier than removing styles.  I'm also wondering if it might be useful to add save and cancel buttons?

khelif96:
> I see that the form is inside of a card with a title above it. Is the greater goal to use a few of these together on a page? I wonder if there would be any merit to maybe making it more generic with no card (or with the card being optional) so that it can be used in more designs, including in modals? I'm thinking it might make it more portable since adding styles is easier than removing styles. I'm also wondering if it might be useful to add save and cancel buttons?

The goal of this PR was more so to demonstrate a form built from json and demonstrate how it manages its own state, It can be styled how ever we want. I tried to mimic the designs from the distro pages and admin pages which is why they utilize the card with title pattern. But we can definitely use this in other forms such as the modals. I omitted the save and cancel buttons since there would likely be an outside component that decides what to do with the data returned from these forms but it can be added if we need to.

malikchaya2:
> The goal of this PR was more so to demonstrate a form built from json and demonstrate how it manages its own state, It can be styled how ever we want. I tried to mimic the designs from the distro pages and admin pages which is why they utilize the card with title pattern. But we can definitely use this in other forms such as the modals. I omitted the save and cancel buttons since there would likely be an outside component that decides what to do with the data returned from these forms but it can be added if we need to.

ah that makes sense. Will we not be committing the SpruceForm component to Master?

khelif96:
> > The goal of this PR was more so to demonstrate a form built from json and demonstrate how it manages its own state, It can be styled how ever we want. I tried to mimic the designs from the distro pages and admin pages which is why they utilize the card with title pattern. But we can definitely use this in other forms such as the modals. I omitted the save and cancel buttons since there would likely be an outside component that decides what to do with the data returned from these forms but it can be added if we need to.
> 
> ah that makes sense. Will we not be committing the SpruceForm component or any of this to Master?

We can! This can def be committed and used in its current state or updated in future PRs. I was waiting for your feedback since you will most likely be the first person to use this in the  Project Settings Project.

malikchaya2:
> We can! This can def be committed and used in its current state or updated in future PRs. I was waiting for your feedback since you will most likely be the first person to use this in the Project Settings Project.

This looks great! Thanks for putting it together! It looks like it will be fun to use :)

PR #700, opened 2021-04-01
April 1st LOLiday special

SupaJoon:


src/components/Prank/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+const END_PRANK = new Date("April 2, 2021 00:00:00");
+const COOKIE_NAME = "PRANK_EXPERIENCED";
+
+export const Prank = () => {
+  const [isVisibleModal, setIsVisibleModal] = useState(false);
+  const { info } = useToastContext();
+  const now = new Date();
+  const isTodayAprilFoolsDay = now >= START_PRANK && now < END_PRANK;
+  const prankExperienced = Boolean(Cookies.get(COOKIE_NAME));
+  const shouldRunPrank = !prankExperienced && isTodayAprilFoolsDay;
```
SupaJoon, line 18: set this to true to see modal before april 1 11am

src/components/Prank/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  useEffect(() => {
+    if (shouldRunPrank) {
+      setTimeout(
+        () =>
+          info(
+            <>
+              <StyledLink
+                onClick={() => {
+                  //   Cookies.set(COOKIE_NAME, "true");
```
khelif96, line 28: Accidentally left this commented. But i think its fine if people want to see it again its easy to do so and its easily dismissible.

src/components/Prank/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+import { StyledLink } from "components/styles";
+import { useToastContext } from "context/toast";
+
+const START_PRANK = new Date("April 1, 2021 11:00:00");
+const END_PRANK = new Date("April 2, 2021 00:00:00");
+const COOKIE_NAME = "PRANK_EXPERIENCED";
+
+export const Prank = () => {
+  const [isVisibleModal, setIsVisibleModal] = useState(false);
+  const { info } = useToastContext();
```
khelif96, line 14: nit. I think this is one of the few cases i prefer not to use destructuring `info` is such a generic term. `dispatchToast.info` makes it so much clearer as to the intent of the function.

src/components/Prank/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+        }}
+      >
+        <div
+          style={{
+            width: "100%",
+            height: 0,
+            paddingBottom: "56%",
+            position: "relative",
+          }}
+        >
```
khelif96, line 70: emotion 😔

src/context/toast.tsx
```diff
@@ -1,10 +1,19 @@
-import React, { useState, useCallback, useEffect } from "react";
+import React, { useState, useCallback, useEffect, ReactNode } from "react";
```
khelif96, line 1: nit. ReactNode import is not necessary.
Since react 17 React is always in scope so this can be 
`import { useState, useCallback, useEffect } from "react";`
and you can access the `ReactNode` type using `React.ReactNode` from anywhere.

khelif96 (approved):
Had a few nits. Since this commit will only be temporary for april fools its ok to ignore them since we will revert after april fools

khelif96:
Merged before tests completed since this commit is temporary.

PR #711, opened 2021-04-01
EVG-13855 Create TupleSelect Component

khelif96:


src/components/TupleSelect/TupleSelect.test.tsx
```diff
@@ -0,0 +1,176 @@
...
+    key: "Enter",
+    keyCode: 13,
+  });
+  const { location } = history;
+  expect(location.search).toBe(
+    `?${ProjectFilterOptions.BuildVariant}=some-filter`
+  );
+});
+
+// test("Should allow multiple input filters for different keys as query params", () => {
```
khelif96, line 142: I'm still trying to write this test but theres an issue with the antd select component and react-testing-library. 
https://stackoverflow.com/questions/61080116/ant-design-v4-breaks-react-testing-library-tests-for-select-and-autocomplete

src/components/TupleSelect/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+      </Select>
+      <Input
+        aria-label="Select Text Input"
+        data-cy="tuple-select-input"
+        value={input}
+        onChange={(e) => setInput(e.target.value)}
+        style={{ width: "70%" }}
+        placeholder={selectedOption.placeHolderText}
+        suffix={
+          <Icon
```
SupaJoon, line 54: can we give this glyph a pointer: 'cursor'

src/components/TupleSelect/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+    updateQueryParams({ [selected]: updatedParams });
+    setInput("");
+  };
+  const selectedOption = options.find((o) => o.value === selected);
+  return (
+    <Input.Group compact>
+      <Select
+        style={{ width: "30%" }}
+        value={selected}
+        onChange={(v) => setSelected(v)}
```
SupaJoon, line 36: I think this can be `setSelected`

src/components/TupleSelect/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+      />
+    </Input.Group>
+  );
+};
+
+const upsertQueryParam = (params: string[] | string, value: string) => {
+  if (params === undefined) {
+    return [value];
+  }
+  if (typeof params === "string" && params === value) {
```
SupaJoon, line 71: this function sometimes returns a string and sometimes returns an array. is that intentional?

src/components/TupleSelect/index.tsx
khelif96: I tried this but ran into some type errors with the antd select component exporting a number type while setSelected expects a string

src/components/TupleSelect/index.tsx
khelif96: Updated so its more consistent.

SupaJoon (approved):

PR #712, opened 2021-04-01
EVG-7872 show failing command and timeout in task metadata

john-m-liu:
relies on https://github.com/evergreen-ci/evergreen/pull/4559


src/pages/task/Metadata.tsx
```diff
@@ -167,6 +167,12 @@ export const Metadata: React.FC<Props> = ({ loading, task, error, taskId }) => {
             </StyledLink>
           </P2>
         )}
+        {details?.status === TaskStatus.Failed && (
+          <P2>Failing command: {details?.description}</P2>
+        )}
+        {details?.timeoutType && details?.timeoutType !== "" && (
```
khelif96, line 173: https://github.com/evergreen-ci/evergreen/pull/4559/files#diff-5424116e8713d835fc94a4a0b222638eebf7c61231381c46bee66d6e549855f4R568
This field is optional is there still a chance of it being an empty string instead of null?

john-m-liu: Actually I think it'll always be a blank string just because that's what the utility function we use to convert between the internal and public data types does

SupaJoon: "" == false evaluates to true so I think you can just check `!!details?.timeoutType`

SupaJoon (approved):

PR #718, opened 2021-04-20
EVG-14064 Change Aborted background color

khelif96:


malikchaya2 (commented):
Are we sure we want to do this? Don't aborted tasks fail the parent patch?

khelif96:
> Are we sure we want to do this? Don't aborted tasks fail the parent patch?

Theres some background in the ticket for the reasoning. 
https://jira.mongodb.org/browse/EVG-14064

malikchaya2:
> > Are we sure we want to do this? Don't aborted tasks fail the parent patch?
> 
> Theres some background in the ticket for the reasoning.
> https://jira.mongodb.org/browse/EVG-14064

If we want to do this, we should probably reconsider how we handle patches with aborted tasks. I'll approve this PR and leave a comment in the ticket.

malikchaya2 (approved):

khelif96:
> > > Are we sure we want to do this? Don't aborted tasks fail the parent patch?
> > 
> > 
> > Theres some background in the ticket for the reasoning.
> > https://jira.mongodb.org/browse/EVG-14064
> 
> If we want to do this, we should probably reconsider how we handle patches with aborted tasks. I'll approve this PR and leave a comment in the ticket.

Thanks i will leave this PR open until theres some clarification on if this is the right approach/ if we need to do anything else

SupaJoon (approved):

malikchaya2 (changes requested):
I have some more comments, on this, sorry

src/components/StatusSquare/index.tsx
```diff
@@ -24,6 +24,7 @@ export const mapVariantTaskStatusToColor = {
   [TaskStatus.SystemUnresponsive]: failurePurple,
   [TaskStatus.SetupFailed]: failureLavender,
   [TaskStatus.StatusBlocked]: gray.dark1,
+  [TaskStatus.Aborted]: gray.light1,
```
malikchaya2, line 27: Do we want to add this to mapVariantTaskStatusToDarkColor as well?

khelif96: Thanks forgot about that just updated!

khelif96:
> Do we want to consider it a failed status or no? For example, should someone see the task annotations tab on a task with the aborted status?
> 
> https://spruce.mongodb.com/task/evergreen_ubuntu1604_test_agent_patch_7db2224cf93e03a385d88ab501e5c15b13873bfa_60808e600ae606147fd2b80d_21_04_21_20_43_13/annotations?execution=1
> 
> (this is actually out of scope for this ticket since the status existed before as a task status)

I dont think so. I think the original motivation of this ticket was that Aborted should be more in line with an undispatched task/ thus have a similar color scheme.  It isn't a failure as a result of something going wrong but more so just a task that was unscheduled by a user.

malikchaya2 (approved):

PR #719, opened 2021-04-20
EVG-13459 Add vertical scroll to configure patch page variant selector

khelif96:
https://user-images.githubusercontent.com/4605522/115468422-11122180-a201-11eb-914f-d555cd31ae8e.mov

### DIFF
--- a/src/pages/configurePatch/configurePatchCore/ConfigureBuildVariants.tsx
+++ b/src/pages/configurePatch/configurePatchCore/ConfigureBuildVariants.tsx
@@ -10,6 +10,8 @@ import { ProjectBuildVariant } from "gql/generated/types";
 import { VariantTasksState } from "pages/configurePatch/ConfigurePatchCore";
 import { toggleArray } from "utils/array";
 
+const { green } = uiColors;
+
 interface Props {
   variants: ProjectBuildVariant[];
   selectedVariantTasks: VariantTasksState;
@@ -176,10 +178,10 @@ const BuildVariant = styled.div<VariantProps>`
   padding: 8px 0;
   ${cardSidePadding}
   background-color: ${(props: VariantProps): string =>
-    props.isSelected ? uiColors.green.light3 : "none"};
+    props.isSelected ? green.light3 : "none"};
   border-left: 3px solid white;
   border-left-color: ${(props: VariantProps): string =>
-    props.isSelected ? uiColors.green.base : "none"};
+    props.isSelected ? green.base : "none"};
 `;
 const VariantName = styled.div`
   word-break: break-all;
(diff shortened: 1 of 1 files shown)


mvankeulen94:
This left scrollbar looks great! I'm wondering if it might make sense to slightly reduce the font of the variants section now, since users may need to scroll more than they used to in order to select all required builders in this example. 
To double-check, scrolling on the tasks list to the right will not change the view of the left sidebar, correct?

malikchaya2 (commented):
This looks great! 
However, since the scrollbar symbol is hidden, I wonder if users will know to scroll to find the rest of the build-variants. I worry if this might make someone frustrated when some build-variants disappear and they don't know to scroll to "find" them.

SupaJoon:
the shift+click multiselect feature becomes more tedious to use when we hide buildvariants because the user now has to scroll to reveal more build variants.  
hiding selected build variants makes the checkbox view less clear since the user cannot easily glance and see which build variants are being selected. 
here is a video to show what a populated left list could look like with hotkey actions applied and many checkboxes:

https://user-images.githubusercontent.com/10734386/115579463-31d78700-a294-11eb-94e9-b6de0b7e7d54.mov

SupaJoon (commented):
I think the 2 concerns I pointed out could be resolved by extending the height of the scrollable build variant column to the remaining view height

khelif96:
I dont see the shift+ click issue that you have in your video it seems to correctly select the correct elements maybe theres something im not seeing?
https://user-images.githubusercontent.com/4605522/115580352-fee1c300-a294-11eb-9bac-1287daf53071.mov



> the shift+click multiselect feature becomes more tedious to use when we hide buildvariants because the user now has to scroll to reveal more build variants.
> hiding selected build variants makes the checkbox view less clear since the user cannot easily glance and see which build variants are being selected.
> here is a video to show what a populated left list could look like with hotkey actions applied and many checkboxes:
> 
>  hiding.data.mov

khelif96:
> I think the 2 concerns I pointed out could be resolved by extending the height of the scrollable build variant column to the remaining view height

Already ahead of you :) 
image

khelif96:
> This looks great!
> However, since the scrollbar symbol is hidden, I wonder if users will know to scroll to find the rest of the build-variants. I worry if this might make someone frustrated when some build-variants disappear and they don't know to scroll to "find" them.

Good idea i've made the scrollbar always visible
image

malikchaya2 (approved):

SupaJoon:
> I dont see the shift+ click issue that you have in your video it seems to correctly select the correct elements maybe theres something im not seeing?
> https://user-images.githubusercontent.com/4605522/115580352-fee1c300-a294-11eb-9bac-1287daf53071.mov
> 
> > the shift+click multiselect feature becomes more tedious to use when we hide buildvariants because the user now has to scroll to reveal more build variants.
> > hiding selected build variants makes the checkbox view less clear since the user cannot easily glance and see which build variants are being selected.
> > here is a video to show what a populated left list could look like with hotkey actions applied and many checkboxes:
> > hiding.data.mov

I didn't find a JS bug with the shift+ click feature. I meant that it is less ergonomic to use the feature when the scrollable box is short and displays less visible list items than before.

PR #724, opened 2021-04-27
DAG-765: Replace trend-charts-ui component with iframe linking perfomance app for tasks.

ancostas:


bsamek:
@ancostas @mvankeulen94 My impression was that we were discussing whether this was the right approach. I see some comments in the ticket. Could we have this discussion first?

ancostas:
Sure

src/components/PerfPlugin/TrendChartsPlugin.tsx
```diff
@@ -1,12 +1,12 @@
 import React from "react";
-import TrendCharts from "@mongodb-dev-prod/trend-charts-ui";
+import {getSignalProcessingUrl} from "../../utils/environmentalVariables";
```
khelif96, line 3:
```suggestion
import { environmentalVariables } from "utils";

const { getSignalProcessingUrl } = environmentalVariables;

```

src/components/PerfPlugin/TrendChartsPlugin.tsx
```diff
@@ -1,12 +1,12 @@
...
 
 interface Props {
   taskId: string;
 }
 
 const TrendChartsPlugin: React.FC<Props> = ({ taskId }) => (
-  <TrendCharts taskId={taskId} />
+    <iframe style={{width:"100%", minHeight: "1000px"}} src={`${getSignalProcessingUrl()}/task/${taskId}/performanceData`} title="Task Performance Data"/>
```
khelif96, line 9: Can you use emotion to style this element. 

```

const StyledIframe = styled(iframe)`
width: 100%;
height: 1000px;
`
```

khelif96: Why 1000px for the height?

khelif96 (changes requested):

src/components/PerfPlugin/TrendChartsPlugin.tsx
ancostas: I just set that min height since it was defaulting to something quite small like 50px. Should I just set height to 100% and let spruce handle the size of the wrapping component?

src/components/PerfPlugin/TrendChartsPlugin.tsx
```diff
@@ -1,12 +1,20 @@
 import React from "react";
-import TrendCharts from "@mongodb-dev-prod/trend-charts-ui";
+import styled from "@emotion/styled";
+import { environmentalVariables } from "utils";
+
+const { getSignalProcessingUrl } = environmentalVariables;
+
+const StyledIframe = styled.iframe`
```
khelif96, line 7: nit. styles at the bottom of the file and before the default export

khelif96 (approved):
G2g when tests pass

PR #725, opened 2021-04-29
EVG-13265 use expected duration instead of base duration for ETA

john-m-liu:


src/pages/task/Metadata.tsx
```diff
@@ -125,8 +126,8 @@ export const Metadata: React.FC<Props> = ({ loading, task, error, taskId }) => {
         )}
         {/* Can only show the time running and eta if the task is running and 
       it has a baseTaskDuration to calculate the eta with */}
```
khelif96, line 128: Update comment.

khelif96 (approved):
TIL there was an expectedDuration field

src/pages/task/Metadata.tsx
john-m-liu: I deleted the comment instead because I think the code basically says the same thing

PR #726, opened 2021-04-30
EVG-14550 Should use strict eslint check

khelif96:


malikchaya2 (commented):
What's the motivation behind this change?

khelif96:
> What's the motivation behind this change?

I noticed the lint task was passing but the compile task was failing. The compile task was failing because it requires validating with the strict linter rules. The lint task should also require those same rules since that is its purpose.

malikchaya2 (approved):

PR #727, opened 2021-04-30
EVG-13307 Add commit queue confirmation modal

khelif96:
image


malikchaya2 (commented):
Can you please include some screenshots of the visible changes?

malikchaya2 (commented):
The ticket also mentions 

> On a related note, we could also consider not showing the 'remove patch from commit queue' button unless it belongs to a user's own patch.

This seems like it would be an improvement. Do we want to include that?

khelif96:
> The ticket also mentions
> 
> > On a related note, we could also consider not showing the 'remove patch from commit queue' button unless it belongs to a user's own patch.
> 
> This seems like it would be an improvement. Do we want to include that?

I think there should be more of a discussion since that would also prevent project admins from easily removing patches.

khelif96:
> Can you please include some screenshots of the visible changes?

Just included some in the description :)

malikchaya2 (approved):
> > What's the motivation behind this change?
> 
> I noticed the lint task was passing but the compile task was failing. The compile task was failing because it requires validating with the stricter set of linter rules. The lint task should also require those same rules since that is its purpose.
> 
> It should have been using the stricter set of rules from the beginning.

Can we please either create a ticket or start that discussion somewhere so that it doesn't get lost?

PR #740, opened 2021-05-03
EVG-14074 add display task test

john-m-liu:
relies on https://github.com/evergreen-ci/evergreen/pull/4635
there aren't really any changes here, I just cleaned up a query so that we can remove an unneeded field and also added a test


khelif96 (approved):

src/pages/configurePatch/ConfigurePatchCore.tsx
```diff
@@ -52,7 +52,7 @@ export const ConfigurePatchCore: React.FC<Props> = ({ patch }) => {
   const { tab: urlTab } = useParams<{ tab: PatchTab | null }>();
 
   const { project, variantsTasks, id } = patch;
-  const { variants, tasks } = project;
+  const { variants } = project;
```
khelif96, line 55: We should remove the tasks field from the query that fetches this data. If we deploy this pr first with those changes, We wont risk breaking anything on the backend deploy.

PR #752, opened 2021-05-20
EVG-14632 Fix Failing cypress tests and Refactor Configure patch page

khelif96:
I tracked down some of the failures to some changes that were made in Johns PR due to some data changing.
Some of the failures were a result of evergreen returning a host query slower then usual which caused an element to not be clickable. I resolved this by asserting the element is clickable.

I also did some refactoring on the configure patch page to make the state management more readable and maintainable. 
I also noticed some broken behavior with the indeterminate state handling for partially selected tasks, caused by the leafygreen checkbox component. I ommited that for now but will be fixing them in a follow up ticket once leafygreen makes the changes they need to. That work is tracked in https://jira.mongodb.org/browse/EVG-14652 and https://jira.mongodb.org/browse/PD-1386 According to the leafygreen folks it should be fixed by next week.

### DIFF
--- a/cypress/integration/hosts/update_status_modal.ts
+++ b/cypress/integration/hosts/update_status_modal.ts
@@ -11,8 +11,9 @@ describe("Update Status Modal", () => {
   });
 
   it("Update status for selected hosts", () => {
-    cy.get(".ant-checkbox-input").should("not.be.disabled");
-    cy.get(".ant-checkbox-input").check({ force: true });
+    cy.get(".ant-checkbox-input").first().should("be.visible");
+    cy.get(".ant-checkbox-input").first().should("not.be.disabled");
+    cy.get(".ant-checkbox-input").first().check({ force: true });
 
     cy.dataCy("update-status-button").click();
 
--- a/cypress/integration/hosts/update_status_modal.ts
+++ b/cypress/integration/hosts/update_status_modal.ts
@@ -3,17 +3,16 @@ const hostsRoute = "/hosts";
 describe("Update Status Modal", () => {
   before(() => {
     cy.login();
+    cy.preserveCookies();
   });
 
   beforeEach(() => {
-    cy.preserveCookies();
     cy.visit(`${hostsRoute}?limit=100&page=0`);
   });
 
   it("Update status for selected hosts", () => {
-    cy.get(".ant-table-selection-column").within(() => {
-      cy.get(".ant-checkbox-input").check({ force: true });
-    });
+    cy.get(".ant-checkbox-input").should("not.be.disabled");
+    cy.get(".ant-checkbox-input").check({ force: true });
 
     cy.dataCy("update-status-button").click();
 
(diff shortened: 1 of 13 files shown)


cypress/integration/hosts/update_status_modal.ts
```diff
@@ -36,28 +35,4 @@ describe("Update Status Modal", () => {
 
     cy.dataCy("host-status-notes").invoke("val").should("eq", "");
   });
-
-  it("Clears form values when modal is closed", () => {
```
malikchaya2, line 40: Is this being tested in some other way?

cypress/integration/project/patches.ts
```diff
@@ -16,20 +13,4 @@ describe("Project Patches Page", () => {
     cy.dataCy("project-patches-link").first().click();
     cy.location("pathname").should("eq", route);
   });
-
-  it("Page title and data rows are displayed.", () => {
```
malikchaya2, line 20: Should we replace this with something more generic?

cypress/integration/hosts/update_status_modal.ts
khelif96: Yes the previous test already tests for this https://github.com/evergreen-ci/spruce/pull/752/files#diff-0eaef7d1d21a157b987896b82ee405a1b16bf7d9a75732be6305649d51e305f7R30-R36 
All this is doing is is repeating what already happens in the previous test with a different test name

src/pages/configurePatch/configurePatchCore/ConfigureBuildVariants.tsx
```diff
@@ -7,24 +7,26 @@ import { Body } from "@leafygreen-ui/typography";
...
 
 const { toggleArray } = array;
 const { green } = uiColors;
 
 interface Props {
   variants: ProjectBuildVariant[];
   selectedVariantTasks: VariantTasksState;
-  selectedBuildVariant: string[];
-  setSelectedBuildVariant: React.Dispatch<React.SetStateAction<string[]>>;
+  selectedBuildVariants: string[];
```
SupaJoon, line 19: I noticed that selectedBuildVariants only is ever length 1 regardless of the number of build variants selected on the left panel. should this variable type `string` instead of `string[]`

src/utils/array/index.ts
```diff
@@ -50,3 +50,16 @@ export const convertObjectToArray = <T>(object: {
   });
   return result;
 };
+
+// Takes a deduplicated string array and returns an object of key value pairs where the array keys  are the object key mapped to a value
+export const mapStringArrayToObject = <T>(
+  array: string[],
+  v: T
+): { [key: string]: T } =>
```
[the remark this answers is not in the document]
khelif96: Is there a better typing wouldn't the returned value be what ever is returned from the generic T ?
if it is a function

SupaJoon (approved):

cypress/integration/project/patches.ts
khelif96: I capture its functionality in the previous test

PR #753, opened 2021-05-21
EVG-14659 Add Known-Issue status to Spruce And Turn off Cache

malikchaya2:
[evergreen changes ](https://github.com/evergreen-ci/evergreen/pull/4673)

image

image

image


src/pages/Task.tsx
```diff
@@ -43,6 +43,7 @@ export const Task: React.FC = () => {
     GetTaskQueryVariables
   >(GET_TASK, {
     variables: { taskId: id, execution: selectedExecution },
+    fetchPolicy: "network-only",
```
SupaJoon, line 46: why change this from cache-first ([default option](https://www.apollographql.com/docs/react/data/queries/#supported-fetch-policies)) to network-only?

malikchaya2: We want to show the status as "known-failure" on patch details, but show the real failure on the task page. If it pulls it from the cache, we can't show the user the real failure since it's overwritten by "known-issue".

SupaJoon (approved):

src/types/task.ts
```diff
@@ -93,5 +93,5 @@ export enum TaskStatus {
   StatusBlocked = "blocked",
   Aborted = "aborted",
 
-  Known = "known-failure",
+  Known = "known-issue",
```
khelif96, line 96: nit. Can we change this name to `KnownIssue` i think it reads better as `TaskStatus.KnownIssue` as opposed to `TaskStatus.Known`

khelif96 (commented):
Can you merge main in first to ensure tests pass

src/types/task.ts
malikchaya2: done

khelif96 (approved):

khelif96:
Looks great 😄  Don't forget to fix the failing test before merging

khelif96 (approved):
Thank You!

PR #755, opened 2021-05-24
EVG-14418 restore ability to remove patch from commit queue cards

ablack12:
Reverting the change to remove the commit queue remove patch button, since non-existent PRs on the commit queue should no longer exist.


khelif96 (approved):
Thanks can we merge this after. https://github.com/evergreen-ci/spruce/pull/752

PR #756, opened 2021-05-24
EVG-14445 Rename 'Task Annotations' tab to Failure Details

malikchaya2:
image

### DIFF
--- a/src/pages/task/TaskTabs.tsx
+++ b/src/pages/task/TaskTabs.tsx
@@ -127,7 +127,7 @@ export const TaskTabs: React.FC<TaskTabProps> = ({ task, taskFiles }) => {
     ),
     [TaskTab.Annotations]: (
       <Tab
-        name="Task Annotations"
+        name="Failure Details"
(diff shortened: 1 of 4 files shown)


src/pages/Task.tsx
```diff
@@ -43,6 +43,7 @@ export const Task: React.FC = () => {
     GetTaskQueryVariables
   >(GET_TASK, {
     variables: { taskId: id, execution: selectedExecution },
+    fetchPolicy: "network-only",
```
ablack12, line 46: what does this do?

malikchaya2: I accidentally branched off of my previous branch and not master, this is from my previous branch. 
It gets all the data from actually performing the query even if there is data cached from a previous query (like from the Tasks page).

malikchaya2: I fixed the branch issue, it should be easier to review now.

SupaJoon (approved):

ablack12 (approved):

PR #757, opened 2021-05-27
EVG-13205: Move filters to table column headers

SupaJoon:
[EVG-13205](https://jira.mongodb.org/browse/EVG-13205)

### Description 
move table filters into columns for tests, tasks table. 
doesn't affect exec task table or files table

### Testing 
2 new jest hook tests
cypress test updates for tests table and task table


src/components/Table/Filters.tsx
```diff
@@ -58,89 +27,86 @@ export const getColumnSearchFilterProps = ({
...
+          size="small"
+          variant="primary"
+          onClick={updateUrlParam}
+        >
+          Search
+        </Button>
+      </ButtonsWrapper>
+    </FilterWrapper>
   ),
+  filterIcon: <SearchOutlined data-cy={dataCy} style={getStyle(value)} />,
```
khelif96, line 56: We should use the leafygreen Icon Magnifying glass component

khelif96 (changes requested):
This is a good start but im gonna split it up into a few smaller reviews, Given how large this PR is.

This seems to work. but i had a few concerns.
The tree select does not match the designs
image
vs
image

You left the original filters they are no longer needed.

Render performance has decreased considerably compared to the old app. 
There is allot of dead code. 
Applying something as simple as a single status filter takes nearly 100 renders.

khelif96:
@arjunpatel100  does #867 supersede this?

PR #758, opened 2021-05-27
EVG-14156 Replace Undispatched Status badges with Task Will[not] run badges

khelif96:
[EVG-14156](https://jira.mongodb.org/browse/EVG-1456)

### Description 
Refactor badge component to just take in a status and display it instead of making adjustments to a status (Removing the blocked prop) This is because `Blocked` is now an official display status instead of one we just make on the ui. 
Adds support for 2 additional statuses `Task Will Run` and `Unscheduled` These will replace the Undispatched status

### Screenshots
image
image

### Testing 
Storybook tests alongside storyshots for the badges,
Did some manual testing with some patches. Will update any e2e tests when i see the results

### Evergreen PR 
https://github.com/evergreen-ci/evergreen/pull/4705


src/components/Table/TasksTable.tsx
```diff
@@ -143,14 +143,11 @@ const getColumnDefsControlled = (
   }));
 };
 
-const renderStatusBadge = (
-  status: string,
-  { blocked }: Task
-): null | JSX.Element => {
+const renderStatusBadge = (status: string): null | JSX.Element => {
   if (status === "" || !status) {
```
SupaJoon, line 147: we could probably add this logic to the TaskStatus badge and replace renderStatusBadge usage with TaskStatusBadge

SupaJoon (approved):

ablack12 (commented):
I worry that "Task Will Not Run" might be misleading, like there's something preventing the task from being scheduled. Was this wording discussed in design / with Maria? What about Queued and Not Queued?

khelif96:
> I worry that "Task Will Not Run" might be misleading, like there's something preventing the task from being scheduled. Was this wording discussed in design / with Maria? What about Queued and Not Queued?
> 
> Also that dark color for the not scheduled state is really eye catching, more so than the green and red. Is there something else we can do?

Yeah the wording and design was discussed and is also documented in the designs attached to the ticket. 
https://jira.mongodb.org/browse/EVG-14156
image

ablack12 (approved):

PR #775, opened 2021-06-08
EVG-14694 Auto select required tasks in configure patch page

khelif96:
[EVG-14694](https://jira.mongodb.org/browse/EVG-14694)

### Description 
This auto selects required tasks on the configure patch page. This used to work but was removed in a past change.
It depended on the [variantsTasks patch field](https://github.com/evergreen-ci/evergreen/blob/4071b3c5984744e3a71589c95445d89f6db446be/graphql/schema.graphql#L490) which only returned required tasks.
It was replaced  by the [Patch.Project.variants](https://github.com/evergreen-ci/evergreen/blob/4071b3c5984744e3a71589c95445d89f6db446be/graphql/schema.graphql#L532) field which returns all the tasks for a patch. 
There was the assumption that these fields were the same and only one was needed but they aren't and they both are.
### Testing 
Local testing and added cypress tests

### Evergreen PR 
Related to/ Dependent on https://github.com/evergreen-ci/evergreen/pull/4734

### DIFF
--- a/src/pages/configurePatch/ConfigurePatchCore.tsx
+++ b/src/pages/configurePatch/ConfigurePatchCore.tsx
@@ -25,7 +25,7 @@ import { CodeChanges } from "pages/patch/patchTabs/CodeChanges";
 import { ParametersContent } from "pages/patch/patchTabs/ParametersContent";
 import { PatchTab } from "types/patch";
 import { queryString, string } from "utils";
-import { mapStringArrayToObject } from "utils/array";
+import { convertArrayToObject, mapStringArrayToObject } from "utils/array";
 import { ConfigureBuildVariants } from "./configurePatchCore/ConfigureBuildVariants";
 import { ConfigureTasks } from "./configurePatchCore/ConfigureTasks";
 import { VariantTasksState } from "./configurePatchCore/state";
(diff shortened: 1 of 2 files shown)


src/pages/configurePatch/ConfigurePatchCore.tsx
```diff
@@ -172,7 +172,7 @@ export const ConfigurePatchCore: React.FC<Props> = ({ patch }) => {
         description: patch.description,
         buildVariants: [variants[0]?.name],
         params: patch.parameters,
-        variantTasks: convertPatchVariantTasksToStateShape(variants),
+        variantTasks: initializeTaskState(variants, patch.variantsTasks),
```
SupaJoon, line 175: This assignment is executed when the user clicks "Schedule". If we didn't have [this redirect](https://github.com/evergreen-ci/spruce/blob/main/src/pages/configurePatch/ConfigurePatchCore.tsx#L139) after successful mutation, I think we will actually see the users selected options joined with the contents of variantsTasks. 
I don't think there is a bug here, but it could be helpful to adjust the useEffect so it doesn't run upon a successful mutation.

SupaJoon (approved):

src/pages/configurePatch/ConfigurePatchCore.tsx
khelif96: Yeah i can see that being executed because patch is updated. But i think as long as we redirect the user we don't have to worry about it.

PR #776, opened 2021-06-08
EVG-14618: Sort execution tasks

sophstad:
[EVG-14618](https://jira.mongodb.org/browse/EVG-14618)

### Description (updated)
- Fixed bug in `toSortString()` util that prevented query params from being removed when a sort was unapplied
- Handle sorting of execution tasks on client side (add `controlled` prop to `TasksTable` to indicate that the table props are sorted by GQL resolver)

### Screenshots
Default sort is applied:
image

### Testing 
- Added Cypress test to ensure that sort order is set by default and that clicking on the sort buttons resets query params


SupaJoon (changes requested):

src/pages/task/TaskTabs.tsx
```diff
@@ -10,7 +10,9 @@ import { getTaskRoute } from "constants/routes";
 import { GetTaskQuery } from "gql/generated/types";
 import { usePrevious } from "hooks";
 import { useBuildBaronVariables } from "hooks/useBuildBaronVariables";
-import { TaskTab } from "types/task";
+import { useUpdateURLQueryParams } from "hooks/useUpdateURLQueryParams";
+import { toSortString } from "pages/patch/patchTabs/util";
```
khelif96, line 14: Can you please move this to a generic util folder. If we need something from another page it should be moved to a more generic folder.

src/pages/task/TaskTabs.tsx
```diff
@@ -96,6 +99,12 @@ export const TaskTabs: React.FC<TaskTabProps> = ({ task, taskFiles }) => {
         key="execution-tasks-tab"
       >
         <TasksTable
+          tableChangeHandler={(...[, , sorter]) => {
+            updateQueryParams({
+              sorts: toSortString(sorter),
+              [RequiredQueryParams.Execution]: "0",
```
khelif96, line 105: We can't hardcode execution since tasks can have multiple executions this will cause it to always show the first one.

khelif96 (changes requested):

khelif96 (approved):

SupaJoon (changes requested):
I think we need to add weights/sort instructions for the antd status column. 
We do this for the PatchTasks query for the other Tasks table here: https://github.com/evergreen-ci/evergreen/blob/92438f2eac2677e38a8738e9be5cc914aba3b0c6/model/task/task.go#L2938-L2980
I think currently the antd table will use alphabetical sorting by default and the status order ascending will look like this: ```["aborted", "blocked", "dispatched", "failed", "inactive", "known-failure", "pending", "setup-failed", "started", "success", "system-failed", "system-timed-out", "system-unresponsive", "task-timed-out", "test-timed-out", "undispatched", "unstarted"]```

Alphabetical sorting is really close to the weights we use in the db query except for statuses like 'system-failed' which should be alongside other "failed" statuses. 

I think it's okay if we hardcode the weights client side to resemble what we are doing in the PatchTasks resolver to avoid additional resolver changes or add sort params to the executionTasksFull resolver.

SupaJoon (approved):

PR #777, opened 2021-06-11
EVG-13887,EVG-14762: Update Navigation

sophstad:
[EVG-13887](https://jira.mongodb.org/browse/EVG-13887)
[EVG-14762](https://jira.mongodb.org/browse/EVG-14762)
cc @mvankeulen94 

### Description 
- Add new "More" dropdown and reorganize content
- Use LeafyGreen Menu component -- this component doesn't support opening the menu on hover, but requiring a click to open is better for accessibility and more similar to patterns used in Atlas. If we want to preserve hover I think we should keep the `antd` component in use.

### Screenshots
image
image
image

### Testing
- Lemme know if you think this should be tested on staging!


src/components/Navbar/Dropdown.tsx
```diff
@@ -0,0 +1,36 @@
...
+  menuItems: JSX.Element;
+  title: string;
+}
+
+export const Dropdown: React.FC<DropdownProps> = ({
+  dataCy,
+  menuItems,
+  title,
+}) => (
+  <AntdDropdown overlay={menuItems}>
```
malikchaya2, line 19: This might be out of scope for this ticket if we are already using this in other parts of the header, but I know that we are trying to move away from using antd components.

malikchaya2 (commented):
Adding a comment but I'm not a blocking reviewer.

src/components/Navbar/Dropdown.tsx
khelif96: Yeah could you replace this with a leafygreen menu https://www.mongodb.design/component/menu/example/

sophstad: Good idea, I'll take a stab at this!

khelif96:
try merging in main

src/components/Navbar/Dropdown.tsx
```diff
@@ -0,0 +1,82 @@
...
+      to={to}
+      href={href}
+      data-cy={itemDataCy}
+      onClick={() => {
+        setOpenMenu(false);
+      }}
+    >
+      {text}
+    </MenuItem>
+  );
```
khelif96, line 51: Can you declare this outside the component. Its a bit of an antipattern since this gets redeclared every render.

src/components/Navbar/Dropdown.tsx
```diff
@@ -0,0 +1,82 @@
...
+      }
+    >
+      {menuItems.map((menuItem) => (
+        <DropdownItem key={`dropdown_${menuItem.text}`} {...menuItem} />
+      ))}
+    </Menu>
+  );
+};
+
+export const DropdownItem = (props) => {
```
khelif96, line 72: nit can you add a type

khelif96: also its not being used anywhere else so we can omit the export

khelif96: What is the difference between this DropdownItem and the one declared above?

khelif96 (commented):
Looks good so far just had a few comments

src/components/Navbar/Dropdown.tsx
sophstad: Oops yeah, meant to remove this in favor of the above! Thanks for catching!

khelif96 (approved):

PR #785, opened 2021-06-23
EVG-13600 Mainline Commit Graph Component

ziyielenachen:
[EVG-13600](https://jira.mongodb.org/browse/EVG-13600)

### Description 
* Created components called CommitGraphWrapper and CommitGraph
* Commit.tsx queries mainline commits
* CommitGraphWrapper displays (at most 7) commit graphs in a row
* CommitGraphWrapper has a prop "pageType" to determine whether the bar charts should be in absolute value or percentage view (absolute value means that the graphs are relative to each other in bar heights. Percentage means that only the bars within a graph are relative to each other in height.)
* CommitGraph takes in a mapping between task statuses and task counts under that status and displays one bar chart, either of absolute value or percentage
* Note that CommitGraphWrapper gets a list of versions, and each version contains a list of tasks, while CommitGraph is supposed to take in a map. We will calculate the task counts on the backend for better performance, and this will be implemented in another ticket. For now, data fed into CommitGraph are hard-coded.

### Screenshots
Screen Shot 2021-06-23 at 11 08 29 AM


src/components/CommitsPage/CommitGraph.tsx
```diff
@@ -0,0 +1,120 @@
...
+  taskCounts: TaskCounts;
+  max?: number;
+  graphType: "percentage" | "absolute";
+}
+
+function calculateHeight(
+  value: number,
+  max: number,
+  total: number,
+  graphType: string
```
SupaJoon, line 25: i think graphType is actually "percentage" | "absolute"

src/components/CommitsPage/CommitGraphWrapper.tsx
```diff
@@ -0,0 +1,125 @@
...
+export const CommitGraphWrapper: React.FC<{
+  versions: MainlineCommitsQuery["mainlineCommits"]["versions"];
+  error?: ApolloError;
+  graphType: "percentage" | "absolute";
+}> = ({ versions, error, graphType }) => {
+  let max = -1;
+  if (error) {
+    return <PageWrapper>ERROR</PageWrapper>;
+  }
+  if (!versions) {
```
SupaJoon, line 19: CommitGraphWrapper could also accept an `isLoading` prop that comes from the `useQuery` hook that also supplies the ApolloError. I think this would be more accurate view since it's possible for versions to be empty after network resources are done loading

src/components/CommitsPage/CommitGraphWrapper.tsx
```diff
@@ -0,0 +1,125 @@
...
+  }
+  if (graphType === "absolute") {
+    max = findMax(data);
+  }
+  if (versions?.length !== 0) {
+    return (
+      <>
+        <FlexRowContainer>
+          {data.map((value) => (
+            <CommitGraph taskCounts={value} max={max} graphType={graphType} />
```
SupaJoon, line 30: you can replace `max={max}` with `max={graphType === "absolute" ? findMax(data) : -1}`

src/components/CommitsPage/CommitGraphWrapper.tsx
```diff
@@ -0,0 +1,125 @@
...
+          ))}
+        </FlexRowContainer>
+      </>
+    );
+  }
+  return <NoResults data-cy="no-patches-found">No commits found</NoResults>;
```
SupaJoon, line 36: we should only show this when we are done loading resources

src/components/CommitsPage/CommitGraphWrapper.tsx
```diff
@@ -0,0 +1,125 @@
...
+  align-items: flex-end;
+  height: 222px;
+  width: 100%;
+`;
+
+const NoResults = styled.div`
+  margin-top: 12px;
+`;
+
+function findMax(data: Array<Object>) {
```
SupaJoon, line 56: i think something like this works also: `const findMax = (data: [{[key: string]: number}]) => data.length ? Math.max(...data.reduce((accum: [number], curr: {[key: string]: number}) => [...accum, ...Object.values(curr)], []) : -1`

SupaJoon (changes requested):

src/components/CommitsPage/CommitGraphWrapper.tsx
```diff
@@ -0,0 +1,125 @@
...
+  let max = -1;
+  data.forEach((item) => {
+    max = Math.max(
+      max,
+      Math.max.apply(
+        null,
+        Object.keys(item).map((x) => item[x])
```
SupaJoon, line 63: we can use Object.values(item) instead

src/components/CommitsPage/CommitGraphWrapper.tsx
```diff
@@ -0,0 +1,125 @@
...
+const NoResults = styled.div`
+  margin-top: 12px;
+`;
+
+function findMax(data: Array<Object>) {
+  let max = -1;
+  data.forEach((item) => {
+    max = Math.max(
+      max,
+      Math.max.apply(
```
SupaJoon, line 61: you can replace this with Math.max(...Object.values(item))

khelif96 (changes requested):
Looks good i had a few comments. I think this could use a storybook story instead of hardcoding data in the page

src/components/CommitsPage/CommitGraph.tsx
```diff
@@ -0,0 +1,118 @@
...
+};
+
+const barColors = {
+  systemFailure: "#633F70",
+  setupFailure: "#A075AF",
+};
+
+const GraphContainer = styled.div`
+  height: 100%;
+  width: ${(1 / 7) * 100}%;
```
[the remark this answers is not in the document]
ziyielenachen: Can we leave it like this for now and modify this once the design for inactive commits is updated?

src/components/CommitsPage/CommitGraphWrapper.tsx
```diff
@@ -0,0 +1,126 @@
...
+  error?: ApolloError;
+  graphType: "percentage" | "absolute";
+  isLoading: boolean;
+}> = ({ versions, isLoading, error, graphType }) => {
+  if (error) {
+    return <PageWrapper>ERROR</PageWrapper>;
+  }
+  if (isLoading) {
+    return <StyledSkeleton active title={false} paragraph={{ rows: 6 }} />;
+  }
```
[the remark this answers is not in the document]
ziyielenachen: I will put this in the agenda for next design meeting. In the meantime, can I keep the horizontal loading bars for now?

src/pages/commits/commitChart/CommitChart.tsx
```diff
@@ -0,0 +1,76 @@
...
+};
+
+const barColors = {
+  success: green.base,
+  failure: red.base,
+  dispatched: yellow.base,
+  scheduled: gray.dark1,
+  unscheduled: gray.light1,
+  systemFailure: "#633F70",
+  setupFailure: "#A075AF",
```
khelif96, line 56: Can you use this enum for the TaskStatuses
https://github.com/evergreen-ci/spruce/blob/d2e0865ab48c5f8f83462a98ec12b20eaaa59f96/src/types/task.ts#L51

We also have this constants file with color mappings can we use those?
https://github.com/evergreen-ci/spruce/blob/d2e0865ab48c5f8f83462a98ec12b20eaaa59f96/src/constants/task.ts#L137

khelif96: Although the colors are slightly different we plan on consolidating them soon so they will eventually be the same.

src/pages/commits/commitChart/CommitChart.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+export type TaskStats = {
+  success?: number;
+  failure?: number;
+  dispatched?: number;
+  scheduled?: number;
+  unscheduled?: number;
+  systemFailure?: number;
+  setupFailure?: number;
+  total: number;
```
khelif96, line 15: did we confirm these are the only statuses we care about? 
Also `scheduled` is not an official task status this should be `will-run`

src/pages/commits/CommitsWrapper.tsx
```diff
@@ -0,0 +1,49 @@
...
+  isLoading: boolean;
+}> = ({ versions, isLoading, error }) => {
+  if (error) {
+    return <PageWrapper>ERROR</PageWrapper>;
+  }
+  if (isLoading) {
+    return <StyledSkeleton active title={false} paragraph={{ rows: 6 }} />;
+  }
+  if (!isLoading && versions?.length !== 0) {
+    return <FlexRowContainer />;
```
khelif96, line 21: Can we just render a few empty `<CommitChart />` elements  here with version descriptions  under them as placeholders?

khelif96 (changes requested):
Looks good thanks for making the changes i had a few more comments

src/components/CommitsPage/CommitGraphWrapper.tsx
```diff
@@ -0,0 +1,126 @@
+import React from "react";
+import { ApolloError } from "@apollo/client";
+import styled from "@emotion/styled";
+import { Array } from "@ungap/global-this";
```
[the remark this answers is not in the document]
SupaJoon: I think Elena's editor auto-imported a random library from node_modules

src/pages/commits/commitChart/CommitChart.stories.tsx
```diff
@@ -0,0 +1,92 @@
...
+      { status: "Dispatched", count: 5 },
+      { status: "Unscheduled", count: 2 },
+    ],
+  },
+];
+
+function findMaxGroupedTaskStats(groupedTaskData: GroupedResult[]) {
+  const maxes = groupedTaskData.map((data) => data.max);
+  return Math.max(...maxes);
+}
```
khelif96, line 87: This seems like it will be used elsewhere should this live somewhere else?

src/pages/commits/commitChart/CommitChart.stories.tsx
```diff
@@ -0,0 +1,92 @@
...
+  </FlexRowContainer>
+);
+
+export const PercentChart = () => (
+  <FlexRowContainer>
+    {groupedTaskData.map((item) => (
+      <CommitChart
+        groupedTaskStats={item.stats}
+        total={item.total}
+        max={-1}
```
khelif96, line 32: should pass max in see how it behaves with it.

src/pages/commits/commitChart/utils.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export type GroupedResult = {
+  stats: {
+    [key: string]: { count: number; statuses: string[] };
+  };
+  max: number;
+  total: number;
+};
+
+export const groupTasksByColor = (
```
khelif96, line 12: can you write some tests for this.

src/pages/Commits.tsx
```diff
@@ -13,6 +13,7 @@ import {
 import { GET_MAINLINE_COMMITS } from "gql/queries";
 import { usePageTitle, useNetworkStatus } from "hooks";
 import { PageDoesNotExist } from "pages/404";
+import { CommitsWrapper } from "pages/commits/CommitsWrapper";
```
khelif96, line 16:
```suggestion
import { CommitsWrapper } from "./commits/CommitsWrapper";
```

khelif96 (commented):
Looks great! Almost done

src/pages/commits/commitChart/CommitChart.stories.tsx
ziyielenachen: This max is -1 because chart type is percentage.

src/pages/commits/commitChart/utils.ts
ziyielenachen: Thank you for all the comments!! They are super helpful!

khelif96 (approved):
Thanks for building this!

SupaJoon (approved):

PR #803, opened 2021-07-06
EVG-13849: Chart Toggle for Mainline Commits Charts

ziyielenachen:
[EVG-13849](https://jira.mongodb.org/browse/EVG-13849)

### Description 
* Added view options to toggle between absolute-value charts and percentage charts
* For absolute-value chart type, charts are relative to each other in terms of bar heights
* For percentage chart type, bar heights are only relative to each other within the same chart (i.e. if you add all bars together within a chart, it should equal to 100%)
* chart type is a URL query param
* default chart type is absolute-value

### Screenshots
Default:
Screen Shot 2021-07-06 at 4 29 07 PM

Absolute:
Screen Shot 2021-07-06 at 4 29 36 PM

Percentage:
Screen Shot 2021-07-06 at 4 29 28 PM

Design:
Screen Shot 2021-07-06 at 4 50 01 PM

### Testing 
Added cypress test chart_toggle to test that:
1. the default chart type is absolute when not indicated in URL query param
2. updates chart type query param to absolute when absolute radio button checked
3. updates chart type query param to percentage when percentage radio button checked
4. absolute radio button is checked when chart type query param is absolute
5. percentage radio button is checked when chart type query param is percentage
6. absolute radio button is checked when chart type query param is invalid


ablack12:
Removing myself since it doesn't look like there are backend changes, but this looks great! Mohamed would probably be interested in taking a look as well. My only comment is that (probably in a future ticket) we'll likely want to add a tooltip or something to explain the different chart options.

src/pages/Commits.tsx
```diff
@@ -1,6 +1,8 @@
+import { useEffect, useState } from "react";
 import { useQuery } from "@apollo/client";
 import styled from "@emotion/styled";
-import { useParams } from "react-router-dom";
+import queryString from "query-string";
```
sophstad, line 4: [nit] I believe it's preferred to use Spruce's query-string wrapper util, aka what you're doing [here](https://github.com/evergreen-ci/spruce/blob/b91133995a3b7af0676f1f609a5a0b4815298902/src/pages/commits/commitChart/ChartToggle.tsx#L8) :)

src/pages/commits/commitChart/ChartToggle.tsx
```diff
@@ -0,0 +1,90 @@
...
+          onChange={onChangeChartType}
+          value={currentChartType}
+          name="chart-select"
+        >
+          <Radio
+            data-cy="chart-absolute-radio"
+            id="cy-chart-absolute-radio"
+            value={ChartTypes.Absolute}
+          >
+            <Label htmlFor="chart-radio-absolute">Absolute Number</Label>
```
sophstad, line 45: `htmlFor` is an accessibility-related prop used to identify which input element a label is associated with. This means that the contents of `htmlFor` should match the input's `id`, so here, the ID on line 42 should be `chart-radio-absolute`.

src/pages/commits/commitChart/ChartToggle.tsx
```diff
@@ -0,0 +1,90 @@
...
+          </Radio>
+          <Radio
+            data-cy="chart-percent-radio"
+            id="cy-chart-percent-radio"
+            value={ChartTypes.Percentage}
+          >
+            <Label htmlFor="chart-radio-percent">Percentage</Label>
```
sophstad, line 52: Same as above!

sophstad (commented):
Nice work, this is looking super cool!

cypress/integration/commit/chart_toggle.ts
```diff
@@ -0,0 +1,45 @@
+const COMMITS_ROUTE = "/commits/evergreen";
```
SupaJoon, line 1: we should move this variable inside of a function. otherwise, i think this variable could be available to other cypress tests.

src/components/FilterBadges/index.tsx
```diff
@@ -14,7 +15,9 @@ export const FilterBadges: React.FC = () => {
   const location = useLocation();
   const { search } = location;
   const queryParams = parseQueryString(search);
-  const queryParamsList = convertObjectToArray(queryParams);
+  const queryParamsList = convertObjectToArray(queryParams).filter(
+    ({ key }) => key !== ChartToggleQueryParams.chartType
```
SupaJoon, line 19: [EVG-14940]( https://jira.mongodb.org/browse/EVG-14940) will help improve this filter solution

src/pages/Commits.tsx
```diff
@@ -13,14 +15,40 @@ import {
...
 import { ProjectSelect } from "./commits/projectSelect";
 
+const DEFAULT_CHART_TYPE = ChartTypes.Absolute;
+
 export const Commits = () => {
   const { projectId } = useParams<{ projectId: string }>();
   const options = { projectID: projectId };
   const dispatchToast = useToastContext();
+  const { search } = useLocation();
+  const [currentChartType, setcurrentChartType] = useState<ChartTypes>(
```
SupaJoon, line 33: [nit] adjust casing to `setCurrentChartType`

src/pages/commits/commitChart/ChartToggle.tsx
```diff
@@ -0,0 +1,90 @@
...
+export const ChartToggle: React.FC<{
+  currentChartType: ChartTypes;
+}> = ({ currentChartType }) => {
+  const { pathname } = useLocation();
+  const { replace } = useHistory();
+
+  const onChangeChartType = (
+    event: React.ChangeEvent<HTMLInputElement>
+  ): void => {
+    const nextChartType = event.target.value as ChartTypes;
```
SupaJoon, line 22: I think we can omit `as ChartTypes`. Do you know why we might need it?

SupaJoon (changes requested):

src/pages/Commits.tsx
```diff
@@ -13,14 +15,40 @@ import {
...
+  useEffect(() => {
+    if (
+      chartTypeParam === ChartTypes.Absolute ||
+      chartTypeParam === ChartTypes.Percentage
+    ) {
+      setcurrentChartType(chartTypeParam);
+    } else {
+      setcurrentChartType(DEFAULT_CHART_TYPE);
+    }
+  }, [chartTypeParam]);
```
SupaJoon, line 51: we should add `setcurrentChartType` to the dep array

SupaJoon (changes requested):

src/pages/Commits.tsx
ziyielenachen: nice catch

src/pages/commits/commitChart/ChartToggle.tsx
ziyielenachen: Just removed it. I'm not sure why exactly. I followed the example of `onChangeLogType()` in `LogTypes.tsx`.

SupaJoon (approved):

PR #806, opened 2021-07-08
EVG-14944: Remove references to childPatch

sophstad:
EVG-14944

### Description 
- Remove all references to `childPatches` so that this field can be safely converted to `Patch` type
- Not sure if I should include the `groupId` and `StartTime` fields that got added to the generated `types.ts` file -- any thoughts?

### Testing 
- Downstream Task tab no longer renders (cleanly) when running local dev server


malikchaya2 (approved):
GroupId and StartTime are probably there from other commits. I would maybe pull main, merge it in, and run yarn codegen again to make sure it's the most recent version. If it's still there I'd think it's fine. 

In a future PR, once the backend changes are done, let's edit the local data to fit the new type so that we can render it again. 

Thank you for doing this work and dealing with how complex it ended up getting!

sophstad:
Confirmed that `generated/types.ts` is up to date, I'll keep those fields around.

PR #814, opened 2021-07-19
EVG-8289 Support mainline commits in spruce

khelif96:
[EVG-8289](https://jira.mongodb.org/browse/EVG-8289)

### Description 
Support mainline commits on spruce

- [ x ] Create version resolver in graphql
- [ x ] Utilize version resolver when viewing versions
- [ x ] Utilize patch resolver only when viewing unconfigured patches that aren't versions yet
- [ x ] Only allow viewing unconfigured patches in the configure page.
- [ x ] Create a middle layer that first determines the state of a version, If its a mainline commit or a activated patch or an unactivated patch and routes the user accordingly
- [ x ] Dynamically render tabs on version page depending on type of version (mainline commits don't have changes or child patches)
- [ x ] Support commit queue patches,
- -  [ x ]  Patches that are not yet versions and cannot be manually configured
- -  [ x ]  Route users to the commit queue page for a project if they try to view an unactivated commit queue patch

### Evergreen PR 
https://github.com/evergreen-ci/evergreen/pull/4840

### DIFF
--- a/cypress/integration/version/version_route.ts
+++ b/cypress/integration/version/version_route.ts
@@ -83,7 +83,7 @@ describe("Version route", () => {
 
     it("Navigates to task tab and applies filters when clicking on task square", () => {
       cy.dataCy("task-square").first().click();
-      cy.location("search").should("include", "statuses=success");
+      cy.location("search").should("include", "statuses=will-run");
     });
   });
 });
--- a/cypress/integration/version/task_filters.ts
+++ b/cypress/integration/version/task_filters.ts
@@ -173,14 +173,7 @@ describe("Tasks filters", () => {
       expect(postFilterCount).to.not.eq(multiFilterCount);
     });
     it("Clicking on 'All' checkbox adds all the base statuses and clicking again removes them", () => {
-      const taskStatuses = [
-        "All",
-        "Failed",
-        "Success",
-        "Dispatched",
-        "Running",
-        "Unscheduled",
-      ];
+      const taskStatuses = ["All", "Failed", "Success", "Running", "Will Run"];
       cy.getInputByLabel("All").check({ force: true });
 
       taskStatuses.forEach((status) => {
@@ -189,7 +182,7 @@ describe("Tasks filters", () => {
       urlSearchParamsAreUpdated({
         pathname: pathTasks,
         paramName: urlParam,
-        search: "all,failed,success,dispatched,started,unscheduled",
+        search: "all",
       });
 
       cy.getInputByLabel("All").uncheck({ force: true });


src/pages/version/Metadata.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const { makespan, timeTaken } = versionTiming || {};
+  return (
+    <MetadataCard
+      loading={loading}
+      error={null}
+      title={isPatch ? "Patch Metadata" : "Version Metadata"}
+    >
+      <P2>
+        Project:{" "}
+        <StyledLink href={getProjectPatchesRoute(project)}>
```
arjunpatel100, line 60: this should be a StyledRouterLink since the link is a spruce route

src/pages/version/Metadata.tsx
```diff
@@ -0,0 +1,93 @@
...
+      </P2>
+      <P2>Makespan: {makespan && msToDuration(makespan)}</P2>
+      <P2>Time taken: {timeTaken && msToDuration(timeTaken)}</P2>
+      <P2>Submitted at: {getDateCopy(createTime)}</P2>
+      <P2>Started: {startTime && getDateCopy(startTime)}</P2>
+      <P2>Finished: {finishTime && getDateCopy(finishTime)}</P2>
+      <P2>{`Submitted by: ${author}`}</P2>
+      {baseVersionID && revision && (
+        <P2>
+          <StyledLink
```
arjunpatel100, line 72: this could also be a StyledRouterLink

src/pages/version/Metadata.tsx
```diff
@@ -0,0 +1,93 @@
...
+            data-cy="patch-base-commit"
+            href={`${getUiUrl()}/version/${baseVersionID}`}
+          >
+            Base commit: {revision.slice(0, 10)}
+          </StyledLink>
+        </P2>
+      )}
+      {isPatch && commitQueuePosition !== undefined && (
+        <P2>
+          <StyledLink
```
arjunpatel100, line 82: this could also be a StyledRouterLink

arjunpatel100 (changes requested):

src/components/PatchActionButtons/RestartPatch.tsx
```diff
@@ -1,7 +1,7 @@
 import React, { useState } from "react";
 import { Button } from "components/Button";
 import { DropdownItem } from "components/ButtonDropdown";
-import { PatchRestartModal } from "pages/patch/index";
+import { PatchRestartModal } from "pages/version/PatchRestartModal";
```
ablack12, line 4: if this restart modal is now going to apply to both patches and versions, maybe it should just be called RestartModal (same with other things that use this naming convention; I'm also not sure if that's what's happening here though). Or VersionRestartModal. Patches can't actually technically be restarted since they are the unfinalized representation of a version.

src/pages/PatchRedirect.tsx
```diff
@@ -1,13 +1,60 @@
...
 const { parseQueryString } = queryString;
 
 export const PatchRedirect = () => {
   const { id, tab } = useParams<{ id: string; tab: PatchTab }>();
   const { search } = useLocation();
+  const [shouldRedirect, setShouldRedirect] = useState(false);
+  const [getPatch, { data }] = useLazyQuery<
+    IsPatchConfigurableQuery,
+    IsPatchConfigurableQueryVariables
+  >(GET_IS_PATCH_CONFIGURED, {
```
ablack12, line 29: "is patch configured" and "is patch configurable" seem to be saying two different things. the first says "is the patch finalized" and the second says "can i modify the tasks/variants for the patch". We should standardize this language.

src/pages/ConfigurePatch.tsx
```diff
@@ -31,6 +32,11 @@ export const ConfigurePatch: React.FC = () => {
   const { patch } = data || {};
   usePageTitle(`Configure Patch`);
 
+  // Cant configure a mainline version so should redirect to the version page
```
ablack12, line 35: nit: can't

src/pages/PatchRedirect.tsx
```diff
@@ -1,13 +1,60 @@
...
+    } else {
+      setShouldRedirect(true);
+    }
+  }, [getPatch, id]);
+
+  if (!shouldRedirect) {
+    return <PatchAndTaskFullPageLoad />;
+  }
+  const { activated, alias, projectID } = data?.patch;
+  // If we are viewing an unconfigured patch thats not on the commit queue we should be taken to the configure view
```
ablack12, line 50: nit: that's. Also with comments linebreaks shouldn't denote sentence breaks, so 50 should have a comma or etc so it's clear 51 isn't running on

src/pages/Version.tsx
```diff
@@ -0,0 +1,154 @@
...
+  });
+
+  const [getVersion, { data, error, startPolling, stopPolling }] = useLazyQuery<
+    VersionQuery,
+    VersionQueryVariables
+  >(GET_VERSION, {
+    variables: { id },
+    pollInterval,
+    onError: (e) =>
+      dispatchToast.error(`There was an error loading the patch: ${e.message}`),
```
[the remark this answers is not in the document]
ablack12: whoa this is cool

src/utils/validators/index.ts
```diff
@@ -23,3 +23,13 @@ export const validatePercentage = (percent: any) => {
...
 export const validateSlack = (v: string): boolean =>
   v.match("(#|@).+") !== null;
+
+/**
+ *  validatePatchId tests if a provided id is a mongo objectId indicating that it likely belongs to a patch and not a version
+ *
+ */
+//
+export const validatePatchId = (id: string): boolean => {
+  const mgobsonRegex = /^[a-f\d]{24}$/i;
```
ablack12, line 33: What rules is this using to determine if this is an ObjectId? Can we link them?

src/utils/validators/index.ts
```diff
@@ -23,3 +23,13 @@ export const validatePercentage = (percent: any) => {
 
 export const validateSlack = (v: string): boolean =>
   v.match("(#|@).+") !== null;
+
+/**
+ *  validatePatchId tests if a provided id is a mongo objectId indicating that it likely belongs to a patch and not a version
```
ablack12, line 28: This comment is formatted oddly, do we get to blame autopilot?

src/pages/Version.tsx
```diff
@@ -0,0 +1,154 @@
...
+    onCompleted() {
+      setIsLoadingData(false);
+    },
+  });
+
+  useNetworkStatus(startPolling, stopPolling);
+
+  // If we are viewing a patch we should first check if its configured before trying to check
```
ablack12, line 66: I feel like this comment isn't quite accurate; Doesn't validatePatchId just check if it's a patch or not a patch? I don't see where we're checking whether or not it's a configured patch. (Edit: does this happen at line 90? In which case can we make this comment more helpful?)

src/pages/Version.tsx
```diff
@@ -0,0 +1,154 @@
...
+  // for a version
+  useEffect(() => {
+    if (validatePatchId(id)) {
+      getPatch();
+    } else {
+      getVersion();
+    }
+  }, [getPatch, getVersion, id]);
+
+  // If we are viewing an unconfigured patch thats not on the commit queue we should be taken to the configure view
```
ablack12, line 76: same comment about comments :)

src/pages/Version.tsx
```diff
@@ -0,0 +1,157 @@
...
+  VersionQueryVariables,
+  IsPatchConfigurableQuery,
+  IsPatchConfigurableQueryVariables,
+} from "gql/generated/types";
+import { GET_VERSION, GET_IS_PATCH_CONFIGURED } from "gql/queries";
+import { usePageTitle, useNetworkStatus } from "hooks";
+import { PageDoesNotExist } from "pages/404";
+import { validatePatchId } from "utils/validators";
+import { BuildVariants } from "./version/BuildVariants";
+import { ActionButtons } from "./version/index";
```
arjunpatel100, line 29: Does it still make sense to show the "Reconfigure tasks/variants" link for mainline commit versions or should it only be meant for Patches?

ablack12: +1, should only be for patches

khelif96: Nice catch ive disabled the option to select it for non mainline commits

src/pages/PatchRedirect.tsx
```diff
@@ -1,13 +1,60 @@
+import { useState, useEffect } from "react";
```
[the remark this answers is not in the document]
khelif96: nice catch it is.

src/pages/version/Metadata.tsx
khelif96: This links to a legacy ui page so needs to be a normal link

src/utils/validators/index.ts
khelif96: This comment is using [jsdoc](https://jsdoc.app/) syntax 
image

src/utils/validators/index.ts
khelif96: This is the official regex the mongodb bson spec uses for objectids i've linked them in a comment

src/pages/version/Metadata.tsx
```diff
@@ -0,0 +1,93 @@
...
+          </StyledLink>
+        </P2>
+      )}
+      {isPatch && commitQueuePosition !== undefined && (
+        <P2>
+          <StyledRouterLink
+            data-cy="commit-queue-position"
+            to={getCommitQueueRoute(project)}
+          >
+            Commit queue position: {commitQueuePosition}
```
ablack12, line 86: preference for "Commit Queue Position: " although idk if that's consistent with how we do things

src/utils/validators/index.ts
```diff
@@ -23,3 +23,15 @@ export const validatePercentage = (percent: any) => {
...
+ *  @param  {string} id - the id to test
+ * @return {boolean} - true if it is a mongo objectId, false otherwise
+ */
+//
+export const validatePatchId = (id: string): boolean => {
```
ablack12, line 33: nit: can we call this isPatchId or something? validatePatchId makes it sound like it should do something or error but you're just checking if it's patch or not

ablack12 (approved):
Great work!

arjunpatel100 (changes requested):
I think we need to update the breadcrumb. This is what I see when I point to prod and navigate to this url: http://localhost:3000/version/evergreen_10968647c1917709f5290d684136314fe5897814/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC
Screen Shot 2021-07-30 at 10 54 10 AM

src/pages/version/Metadata.tsx
khelif96: This file was just renamed so most of the changes are not new.  But this seems like an easy change.

src/pages/version/Metadata.tsx
arjunpatel100: ah right, I think it could be possible to link this to the spruce version page now since it goes live in this PR.

khelif96:
> I think we need to update the breadcrumb. This is what I see when I point to prod and navigate to this url: http://localhost:3000/version/evergreen_10968647c1917709f5290d684136314fe5897814/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC
> Screen Shot 2021-07-30 at 10 54 10 AM

This will be addressed in a follow up or

arjunpatel100 (changes requested):
We show a loading skeleton and don't make any network requests when we start on a patch page like http://localhost:3000/version/61041d0661837d6603beac3d/tasks?page=0&sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC%3BNAME%3AASC, click on a task, and then click on the Patch 434 bread crumb. 
Landing on the task page from a URL entry and then clicking the "Patch 434" breadcrumb doesn't show this bug

src/constants/routes.ts
```diff
@@ -172,3 +172,6 @@ export const getCommitQueueRoute = (projectId: string) =>
 
 export const getCommitRoute = (projectIdentifier: string) =>
   `${paths.commits}/${projectIdentifier}`;
+
+export const getCommitsRoute = (projectId: string) =>
```
arjunpatel100, line 176: this functions looks the same as `getCommitRoute` aside from variable naming

src/pages/version/Tabs.tsx
```diff
@@ -0,0 +1,105 @@
...
+      selected={selectedTab}
+      setSelected={setSelectedTab}
+      aria-label="Patch Tabs"
+    >
+      {activeTabs.map((t: string) => tabMap({ taskCount, childPatches })[t])}
+    </StyledTabs>
+  );
+};
+
+// const ActiveTabs = ({ taskCount, childPatches, isPatch }) => {
```
arjunpatel100, line 104: leftover comments

src/constants/routes.ts
khelif96: oops

cypress/integration/version/version_route.ts
```diff
@@ -0,0 +1,91 @@
...
+      cy.visit(versionRoute(versions[0]));
+    });
+
+    it("Lists the patch's build variants", () => {
+      cy.dataCy("patch-build-variant").within(
+        ($variants) => Array.from($variants).length > 0
+      );
+    });
+
+    it("Shows tooltip with task's name on hover", () => {
```
arjunpatel100, line 80: this test verifies whether the tooltip has text but it doesn't test whether it displays the task's name

src/analytics/breadcrumb/useBreadcrumbAnalytics.ts
```diff
@@ -5,10 +5,13 @@ import {
...
 
-type Action = { name: "Click Link"; link: "myPatches" | "patch" };
+type Action = {
+  name: "Click Link";
+  link: "myPatches" | "patch" | "version" | "waterfall";
+};
 
 interface P extends Properties {}
-interface Analytics extends A<Action> {}
+export interface Analytics extends A<Action> {}
```
arjunpatel100, line 14: I don't see a corresponding import for Analytics

src/components/VersionRestartModal/VersionRestartModal.tsx
```diff
@@ -68,7 +66,7 @@ export const PatchRestartModal: React.FC<PatchModalProps> = ({
     PatchBuildVariantsQuery,
```
arjunpatel100, line 66: should we use BuildVariantsQuery here instead?

cypress/integration/version/version_route.ts
khelif96: nice catch i just renamed this file didn't really touch that test but i updated it to make it actually do something.

sophstad (approved):

arjunpatel100 (changes requested):
The task status filters on this page are not populated with statuses when pointing to prod: http://localhost:3000/version/evergreen_65cdf890e49b342699e7e56e331a4917b4ae0773/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

[The query that populates the dropdown](https://github.com/evergreen-ci/spruce/blob/main/src/components/TaskStatusFilters/index.tsx#L36) will return `400 (Bad Request): id 'evergreen_65cdf890e49b342699e7e56e331a4917b4ae0773' is not a valid patch id`

arjunpatel100:
I think we can disable the "Schedule failing base tasks" button for versions since base task status and the version status will always be the same and the condition: `Are you sure you want to schedule all the undispatched base tasks for this patch's failing tasks?` will always yield 0 tasks:

src/pages/version/Tabs.tsx
```diff
@@ -0,0 +1,106 @@
+import { useState, useEffect, useMemo } from "react";
+import { Tab } from "@leafygreen-ui/tabs";
+import { useParams, useHistory, useLocation } from "react-router-dom";
+import { usePatchAnalytics } from "analytics";
+import { CodeChanges } from "components/PatchTabs/CodeChanges";
+import { DownstreamTasks } from "components/PatchTabs/DownstreamTasks";
```
arjunpatel100, line 6: DownstreamTasks asks for `childPatches` which is currently a property of Patch but not Version. We should most likely omit rendering this tab if there is no synonym for Version

khelif96: ChildPatches gets imported from the patch field in the version query https://github.com/khelif96/spruce/blob/07d0d313e05e437ad97b448652d51f2c4c5baff7/src/pages/Version.tsx#L106 

And is passed in that same file.
 It will only be available if viewing a version that has a corresponding patch. Since it is only relevant to those types of versions it made sense to keep it as a field belonging to  a patch.

arjunpatel100 (approved):

PR #820, opened 2021-07-26
EVG-14597: Patch through childPatches

malikchaya2:
[EVG-14597](https://jira.mongodb.org/browse/EVG-14597)

### Description 
I am breaking out this work so that Sophie can re-use it when working on EVG-14974 so that we can avoid merge conflicts and duplicating work.

### DIFF
--- a/src/pages/patch/PatchRestartModal.tsx
+++ b/src/pages/patch/PatchRestartModal.tsx
@@ -11,6 +11,7 @@ import { Modal } from "components/Modal";
 import { TaskStatusFilters } from "components/TaskStatusFilters";
 import { useToastContext } from "context/toast";
 import {
+  Patch,
   PatchBuildVariantsQuery,
   PatchBuildVariantsQueryVariables,
   RestartPatchMutation,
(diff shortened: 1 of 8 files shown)


src/pages/patch/PatchRestartModal.tsx
```diff
@@ -90,6 +93,7 @@ export const PatchRestartModal: React.FC<PatchModalProps> = ({
 
   const patchAnalytics = usePatchAnalytics();
   const handlePatchRestart = async (e): Promise<void> => {
+    console.log(childPatches);
```
sophstad, line 96:
```suggestion
```

sophstad (approved):

src/pages/patch/PatchRestartModal.tsx
malikchaya2: I have to keep it there because otherwise it errors because childPatches is not used. Implementing it to actually use it will take some time.

khelif96: I would have advocated for removing it from the component. It will log undefined for every user who uses the patchRestartModal

malikchaya2: It's temporary. It'll be removed once EVG-14597 is done.

sophstad: Ah understood! Going forward is this maybe a good use case for `@ts-ignore`?

PR #824, opened 2021-07-30
EVG-14995: Grouped Task Status Badge Component

ziyielenachen:
[EVG-14995](https://jira.mongodb.org/browse/EVG-14995)

### Description 
Created a reusable grouped task status badge component that displays task status name and total count.

### Screenshots
Mock:
Screen Shot 2021-07-30 at 4 07 57 PM

Design:
Screen Shot 2021-07-30 at 4 04 43 PM

### Testing 
Storybook called "Grouped Task Status Badge"


src/gql/generated/types.ts
```diff
@@ -13,339 +13,159 @@ export type Scalars = {
   Boolean: boolean;
   Int: number;
   Float: number;
+  Time: Date;
```
arjunpatel100, line 16: You can remove this file since your PR didn't require adjusting the gql schema or any queries

src/components/GroupedTaskStatusBadge/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+interface Props {
+  status: TaskStatus;
+  count: number;
+}
+export const GroupedTaskStatusBadge: React.FC<Props> = ({ count, status }) => (
+  <BadgeContainer css={statusToColorVariant[status]}>
+    <Number css={statusToColorVariant[status]}>{count}</Number>
+    <Status css={statusToColorVariant[status]}>
+      {status === TaskStatus.Succeeded || status === TaskStatus.SetupFailed
+        ? [
```
arjunpatel100, line 17: we should remove the bracket since so this line evaluates to `string` instaed of `string[]`

arjunpatel100 (approved):
LGTM after removing the generated file and array bracket

src/components/GroupedTaskStatusBadge/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+interface Props {
+  status: TaskStatus;
+  count: number;
+}
+export const GroupedTaskStatusBadge: React.FC<Props> = ({ count, status }) => (
+  <BadgeContainer css={statusToColorVariant[status]}>
+    <Number css={statusToColorVariant[status]}>{count}</Number>
+    <Status css={statusToColorVariant[status]}>
+      {status === TaskStatus.Succeeded || status === TaskStatus.SetupFailed
```
khelif96, line 16: I think this component should be able to accept any status not only succeeded and failed so that it can be reused in other places. (Build Variant squares, My Patches table)

khelif96 (commented):
What changed in types.ts?

src/components/GroupedTaskStatusBadge/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+  `,
+  [TaskStatus.Undispatched]: css`
+    background-color: ${uiColors.gray.light3};
+    border-color: ${uiColors.gray.light2};
+    color: ${uiColors.gray.dark1};
+  `,
+  [TaskStatus.WillRun]: css`
+    background-color: ${uiColors.gray.dark1};
+    border-color: ${uiColors.gray.dark2};
+    color: ${uiColors.gray.light3};
```
khelif96, line 88: can we destructure the colors.

src/components/GroupedTaskStatusBadge/index.tsx
ziyielenachen: Line 16 is for conditional rendering of plural modifier for "Success" => "Successes" and "Setup Failure" => "Setup Failure". Other statuses such as "Running" and "Failed" don't need the plural modifier. Should I take this out and do it before the return statement? The nested conditional rendering may be confusing. :)

ziyielenachen: Just pushed another method of doing this! :)

src/components/GroupedTaskStatusBadge/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+      <Status css={statusToColorVariant[status]}>{statusDisplayName}</Status>
+    </BadgeContainer>
+  );
+};
+
+const BadgeContainer = styled.div`
+  height: 27px;
+  width: 57px;
+  border-radius: 3px;
+  border: 0.75px solid;
```
arjunpatel100, line 32: can we round this to 1

src/components/GroupedTaskStatusBadge/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+  font-size: 11px;
+  letter-spacing: 0.18px;
+  line-height: 10px;
+`;
+
+const Status = styled(Body)`
+  font-size: 8px;
+  letter-spacing: 0.13px;
+  line-height: 10px;
+`;
```
khelif96, line 52: Is this styling necessary? I would like to avoid magic numbers and would prefer if we were more consistent/deliberate about the styles we apply. 
Could we update some of these values to more closely follow the leafygreen guidelines.  https://www.mongodb.design/component/typography/guidelines/ 
Specifically surrounding the typography elements.

src/components/GroupedTaskStatusBadge/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+  `,
+  [TaskStatus.Undispatched]: css`
+    background-color: ${uiColors.gray.light3};
+    border-color: ${uiColors.gray.light2};
+    color: ${uiColors.gray.dark1};
+  `,
+  [TaskStatus.WillRun]: css`
+    background-color: ${uiColors.gray.dark1};
+    border-color: ${uiColors.gray.dark2};
+    color: ${uiColors.gray.light3};
```
khelif96, line 88: These styles look very familiar. Do you think we could consolidate the colors so we have one source of truth for status/color mappings? 

 https://github.com/evergreen-ci/spruce/blob/d2e0865ab48c5f8f83462a98ec12b20eaaa59f96/src/components/TaskStatusBadge/index.tsx

khelif96 (changes requested):

src/components/GroupedTaskStatusBadge/index.tsx
ziyielenachen: yes yes

src/components/GroupedTaskStatusBadge/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+export const GroupedTaskStatusBadge: React.FC<Props> = ({ count, status }) => {
+  let statusDisplayName = taskStatusToCopy[status];
+  if (statusDisplayName.slice(-1) === "s" && count !== 1) {
+    statusDisplayName += "es";
+  } else if (statusDisplayName.slice(-1) === "e" && count !== 1) {
+    statusDisplayName += "s";
+  }
+
+  return (
+    <BadgeContainer css={statusToColorVariant[status]}>
```
khelif96, line 21: This component should also be a link. Since it should link to the version page with the filters applied. 
I think it would make sense for this to accept two optional props.

`onClick` and `to` 
`to` enabling this to become a custom react-router <Link /> element.

src/components/GroupedTaskStatusBadge/index.tsx
ziyielenachen: For `Number` and `Status`, I think line-height is necessary because otherwise the words will be out of the box like this:
Screen Shot 2021-07-30 at 5 37 32 PM

Same thing applies to font-size, they would spell outside the box.
Other stylings were also copied from Figma, but I think we can removed the letter-spacing.

src/components/GroupedTaskStatusBadge/index.tsx
ziyielenachen: Sounds good!

src/components/GroupedTaskStatusBadge/index.tsx
ziyielenachen: Looks like the two mappings only share one status (WillRun) with the same colors. Others such as SetupFailed and SystemFailed inside TaskStatusBadge have different background, border and text colors than the SetupFailed and SystemFailed inside GroupedTaskStatusBadge, so I don't think we can consolidate the colors. I wonder if this is because the colors for TaskStatusBadge are supposed to be updated?

ziyielenachen: For `Success` `Failed` `Running` and `Undispatched`, they seem to share the same colors with `badgeVariants` inside leafygreen's `Badge` component. Should we use leafygreen's mapping for these statuses instead?

src/components/GroupedTaskStatusBadge/index.tsx
sophstad: I think Mohamed makes a good point here! One approach could be to not use LG's `Body` component at all in favor of a more neutral `<p>` or `<span>` tag. IMO this badge doesn't really represent body text, as it has a pretty different purpose and style from a normal paragraph with text, which is what `Body` is designed for.

src/components/GroupedTaskStatusBadge/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+    statusDisplayName += "es";
+  } else if (statusDisplayName.slice(-1) === "e" && count !== 1) {
+    statusDisplayName += "s";
+  }
+
+  return (
+    <BadgeContainer
+      css={statusToColorVariant[status]}
+      onClick={onClick}
+      to={href}
```
khelif96, line 33: This to prop doesn't seem to do anything

khelif96, line 33: For my original comment i intended for this component to serve as both a clickable button as well as a link. Similar to how the leafygreen button does it https://github.com/mongodb/leafygreen-ui/blob/main/packages/button/src/Button.tsx

ziyielenachen: Cool I will take a look

PR #835, opened 2021-08-02
EVG-14826: Render build variant cards without icons

ziyielenachen:
[EVG-14826](https://jira.mongodb.org/browse/EVG-14826)

### Description 
* Created buildVariantCard component
* Rendered the component on waterfall page according to expected default and filtering behaviors
* Default: show buildVariantCards with failed tasks
* When filterings are applied, only show buildVariantCards that meet the filtering criteria.
* Created ActiveCommit component

### Screenshots
Design:
Screen Shot 2021-08-02 at 12 11 40 PM

Without filter, only show build variants with failed taskd:
Screen Shot 2021-08-10 at 3 58 48 PM

With filters:
Screen Shot 2021-08-10 at 3 58 39 PM
Screen Shot 2021-08-10 at 3 59 02 PM
Screen Shot 2021-08-10 at 3 59 49 PM

### Testing 
Added component to project health storybook


src/pages/commits/BuildVariantAccordion.tsx
```diff
@@ -0,0 +1,36 @@
...
+  width: 124px;
+  display: flex;
+  justify-content: flex-start;
+  word-break: break-word;
+`;
+
+export const ColumnContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+  justify-content: flex-start;
```
khelif96, line 34: This is the default so we can omit this style.

src/pages/commits/BuildVariantAccordion.tsx
```diff
@@ -0,0 +1,36 @@
...
+}
+export const BuildVariantAccordion: React.FC<Props> = ({ buildVariants }) => (
+  <ColumnContainer>
+    {buildVariants.map(({ displayName }) => (
+      <AccordionContainer key={displayName}>{displayName}</AccordionContainer>
+    ))}
+  </ColumnContainer>
+);
+
+const AccordionContainer = styled(Disclaimer)`
```
khelif96, line 21: Maybe Body would work better here? Since you are overriding the disclaimers font size.

src/pages/commits/BuildVariantAccordion.tsx
```diff
@@ -0,0 +1,36 @@
...
+  margin-top: 20px;
+  color: ${gray.dark2};
+  font-size: 14px;
+  width: 124px;
```
khelif96, line 25: Is this width needed?

src/pages/commits/BuildVariantAccordion.tsx
```diff
@@ -0,0 +1,36 @@
...
+  display: flex;
+  justify-content: flex-start;
+  word-break: break-word;
```
khelif96, line 28: Also are these flex properties necessary. They dont seem to affect the styling much when they are removed.

src/pages/commits/BuildVariantAccordion.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { Disclaimer } from "@leafygreen-ui/typography";
+
+const { gray } = uiColors;
+
+interface Props {
+  buildVariants: {
+    displayName: string;
+  }[];
+}
+export const BuildVariantAccordion: React.FC<Props> = ({ buildVariants }) => (
```
khelif96, line 13: nit. I think this should be renamed to something more representative of what it is. BuildVariantAccordion sounds like it should be a singular element when in reality this component represents a container for a bunch of buildVariantAccordions. Maybe BuildVariantAccordionContainer or something better?

src/pages/commits/BuildVariantAccordion.tsx
```diff
@@ -0,0 +1,36 @@
...
+  <ColumnContainer>
+    {buildVariants.map(({ displayName }) => (
+      <AccordionContainer key={displayName}>{displayName}</AccordionContainer>
```
khelif96, line 16: We have an [Accordion component](https://github.com/evergreen-ci/spruce/blob/ee5c3c4fe6d8af169696475e0e168249703cc5c9/src/components/Accordion/index.tsx) that would be useful here. I know you aren't rendering any tasks yet but this will likely come in handy when you do.

src/pages/commits/CommitsWrapper.tsx
```diff
@@ -57,6 +58,7 @@ export const CommitsWrapper: React.FC<Props> = ({
                   author={version.author}
                   message={version.message}
                 />
+                <BuildVariantAccordion buildVariants={version.buildVariants} />
               </ColumnContainer>
```
khelif96, line 62: I can't select the entire block of code but this might better suited as its own component. To help group all of the components that represents an active version on the waterfall.

khelif96 (changes requested):

src/pages/commits/BuildVariantAccordion.tsx
ziyielenachen: Yes good idea! Thanks!

src/pages/commits/BuildVariantAccordion.tsx
ziyielenachen: Grabbed this from the figma. I think it looks better when it is slightly skinnier than the commit chart column? So there is more white space. But not necessary at all, just a design choice.

src/pages/commits/BuildVariantAccordion.tsx
ziyielenachen: Good idea!

src/pages/commits/BuildVariantAccordion.tsx
ziyielenachen: Here is the difference:
124px:
image

100%:
image

src/pages/commits/CommitsWrapper.tsx
ziyielenachen: Good idea. I will create `ActiveCommit`. Since this is no longer "accorion" I will also rename couple of files. I also think it makes more sense now if we have `buildVariantCard` as its own component instead of having `buildVariantCardContainer`

src/pages/commits/BuildVariantAccordionContainer.tsx
```diff
@@ -0,0 +1,39 @@
...
+  }[];
+}
+export const BuildVariantAccordionContainer: React.FC<Props> = ({
+  buildVariants,
+}) => (
+  <ColumnContainer>
+    {buildVariants.map(({ displayName }) => (
+      <div>
+        <AccordionContainer key={displayName}>{displayName}</AccordionContainer>
+        <Accordion contents={displayName} title="" />
```
arjunpatel100, line 21: we can replace this with a div since we found out the icons are not collapsible

src/pages/commits/BuildVariantAccordionContainer.tsx
```diff
@@ -0,0 +1,39 @@
...
+    displayName: string;
+  }[];
+}
+export const BuildVariantAccordionContainer: React.FC<Props> = ({
+  buildVariants,
+}) => (
+  <ColumnContainer>
+    {buildVariants.map(({ displayName }) => (
+      <div>
+        <AccordionContainer key={displayName}>{displayName}</AccordionContainer>
```
arjunpatel100, line 20: nit: since this component renders a title, can we call it <BuildVariantName /> instead?

arjunpatel100 (changes requested):

src/gql/queries/get-mainline-commits.graphql
```diff
@@ -1,5 +1,6 @@
 query MainlineCommits(
   $mainlineCommitsOptions: MainlineCommitsOptions!
+  $buildVariantOptionsForTask: BuildVariantOptions!
   $buildVariantOptions: BuildVariantOptions!
```
khelif96, line 4: Why are these different? Is this so we can show the overall task stats but only show build variant cards that match the filter criteria for the default view?

src/pages/Commits.tsx
```diff
@@ -60,16 +68,31 @@ export const Commits = () => {
...
     statuses: filterStatuses,
     variants: filterVariants,
     tasks: filterTasks,
   };
+  const defaultFilterByFailed = !(
+    filterStatuses.length ||
+    filterVariants.length ||
+    filterTasks.length
+  );
+  console.log(defaultFilterByFailed);
```
khelif96, line 81: console.log

src/pages/commits/ActiveCommits/BuildVariantCard.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { Body } from "@leafygreen-ui/typography";
+
+const { gray } = uiColors;
+
+interface Props {
+  buildVariantDisplayName: string;
+}
+export const BuildVariantCard: React.FC<Props> = ({
+  buildVariantDisplayName,
+}) => <Card key={buildVariantDisplayName}>{buildVariantDisplayName}</Card>;
```
khelif96, line 13: key isn't needed here since this is not rendering a list. It would be passed into the component when its rendered in the list.

src/pages/commits/ActiveCommits/BuildVariantCard.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+const Card = styled(Body)`
```
khelif96, line 15: nit. Could we name this something else since this is a Text element and not really a [Card](https://www.mongodb.design/component/card/example/)

src/pages/commits/ActiveCommits/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+      <CommitChartLabel
+        githash={version.revision.substring(0, 5)}
+        createTime={version.createTime}
+        author={version.author}
+        message={version.message}
+      />
+    </ColumnContainer>
+    <ColumnContainer>
+      {version.buildVariants.map(({ displayName }) => (
+        <BuildVariantCard buildVariantDisplayName={displayName} />
```
khelif96, line 54: We should pass key in over here. 
```suggestion
        <BuildVariantCard buildVariantDisplayName={displayName} key={`${version.id}_${displayName}`}/>
```

src/pages/commits/CommitsWrapper.tsx
```diff
@@ -42,22 +41,13 @@ export const CommitsWrapper: React.FC<Props> = ({
...
-                  githash={version.revision.substring(0, 5)}
-                  createTime={version.createTime}
-                  author={version.author}
-                  message={version.message}
-                />
-              </ColumnContainer>
+              <ActiveCommit
+                version={version}
+                chartType={chartType}
+                total={idToTaskStatsGroupedByColor[version.id].total}
```
khelif96, line 47: nit: renamed this to `versionToTaskStatColorMap`?

khelif96 (commented):
Great work just a few more comments

src/pages/Commits.tsx
ziyielenachen: Oopsssss

src/gql/queries/get-mainline-commits.graphql
ziyielenachen: They are different because for the default`taskStatusCounts` (the thing that renders active commit charts) aren't filtered by any status but by default build variant cards are filtered by all the failure statuses.

src/pages/commits/ActiveCommits/BuildVariantCard.tsx
ziyielenachen: Good idea! Does Label work?

src/pages/commits/CommitsWrapper.tsx
ziyielenachen: renamed to `versionToGroupedTaskStatsMap` since it is `{ [id: string]: GroupedResult;}` :)

khelif96 (approved):

src/pages/commits/ActiveCommits/BuildVariantCard.tsx
```diff
@@ -0,0 +1,21 @@
+import React from "react";
```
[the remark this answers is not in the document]
ziyielenachen: Nice catch thanks!!

src/pages/commits/ActiveCommits/ProjectHealth.stories.tsx
```diff
@@ -23,20 +22,13 @@ export const WaterfallAbsolute = () => (
...
-              chartType={ChartTypes.Absolute}
-            />
-            <CommitChartLabel
-              githash={version.revision.substring(0, 5)}
-              createTime={version.createTime}
-              author={version.author}
-              message={version.message}
-            />
-          </ColumnContainer>
+          <ActiveCommit
```
arjunpatel100, line 25: nit: this component should have a key prop

src/pages/commits/ActiveCommits/ProjectHealth.stories.tsx
```diff
@@ -54,20 +46,13 @@ export const WaterfallPercentage = () => (
...
-              chartType={ChartTypes.Percentage}
-            />
-            <CommitChartLabel
-              githash={version.revision.substring(0, 5)}
-              createTime={version.createTime}
-              author={version.author}
-              message={version.message}
-            />
-          </ColumnContainer>
+          <ActiveCommit
```
arjunpatel100, line 49: nit: this component should have a key prop

arjunpatel100 (approved):

PR #837, opened 2021-08-04
EVG-14404: Upgrade LeafyGreen Button

sophstad:
EVG-14404

### Description 
- Update `@leafygreen-ui/button` to v12.0.5 and follow [upgrade guide](https://github.com/mongodb/leafygreen-ui/blob/54daf9a46f78af967773bf7b1c74f72386ab0733/packages/button/UPGRADE.md)
- Upgrade other dependencies in order to avoid mysterious bug introduced in button v12.0.4. I tried to upgrade as few as possible, and couldn't identify any API changes introduced by these new versions in LG's changelog.


src/pages/task/taskTabs/logs/LogTypes.tsx
```diff
@@ -286,10 +286,11 @@ const useRenderBody: React.FC<{
 
 const ButtonContainer = styled.div`
   display: flex;
-  a:first-of-type {
+  > :not(:last-child) {
     margin-right: 8px;
   }
   margin-right: 16px;
+  padding-left: 1px;
```
arjunpatel100, line 293: did you add this style so we can see left side shadow better?

arjunpatel100 (changes requested):

src/pages/task/taskTabs/logs/LogTypes.tsx
sophstad: Exactly!

khelif96:
image
The ellipses button doesn't look centered.

sophstad:
> image
> The ellipses button doesn't look centered.

Yeah, this is an existing problem 😵 
image

The issue is that the LG button component will remove the icon padding iff the icon is the only child of the button component ([here](https://github.com/mongodb/leafygreen-ui/blob/8df11bf4d76813ecaf9a9e339c072cf5ba8f386b/packages/button/src/Button.tsx#L90)). However, the `Menu` component [injects an additional child](https://github.com/mongodb/leafygreen-ui/blob/8df11bf4d76813ecaf9a9e339c072cf5ba8f386b/packages/menu/src/Menu.tsx#L382-L389) to mange the menu content, thus messing with the Button component logic. I'll flag this to the LeafyGreen team!

src/components/Button/index.tsx
```diff
@@ -51,8 +52,7 @@ export const Button: React.FC<Props> = ({
 
 const mapVariantToLeafyGreenVariant: { [key: string]: ButtonTypeKeys } = {
```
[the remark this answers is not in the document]
khelif96: tbh. I wouldn't be opposed to getting rid of this component everywhere it doesn't seem necessary and we aren't really consistent with how we use it.

khelif96 (approved):

sophstad:
Closing, too many errors like the following:

```
[Warning] You are loading @emotion/react when it is already loaded. Running multiple instances may cause problems. This can happen if multiple versions are used, or if multiple builds of the same version are used. (0.chunk.js, line 25246)
```

PR #838, opened 2021-08-04
EVG-15146 Commit message length limit and tooltip

ziyielenachen:
[EVG-15146](https://jira.mongodb.org/browse/EVG-15146)

### Description 
* Commit message now only displays two lines
* If a commit message is longer than two lines, a "more" tooltip will appear

### Screenshots
Screen Shot 2021-08-04 at 3 20 51 PM
Screen Shot 2021-08-03 at 5 11 34 PM
Screen Shot 2021-08-03 at 5 11 44 PM

### Testing 
* Changes reflected in Project Health Page storybook
* Added test `CommitChartLabel.test.tsx`


src/pages/commits/ActiveCommits/CommitChartLabel.test.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  test("Displays entire commit message if it does not break length limit", async () => {
+    const { queryByDataCy } = render(RenderCommitChartLabel(versionShort));
+    expect(queryByDataCy("commit-label")).toHaveTextContent(
+      "SERVER-57332 Create skeleton Internal"
+    );
+  });
+
+  test(
+    "Clicking over the 'more' button should open a tooltip that displays" +
```
khelif96, line 46: nit. `Clicking on`

src/pages/commits/ActiveCommits/CommitChartLabel.tsx
```diff
@@ -19,13 +20,34 @@ export const CommitChartLabel: React.FC<Props> = ({
   message,
 }) => {
   const createDate = new Date(createTime);
+  const maxChars = 42;
```
khelif96, line 23: nit. move this out of the component body since its a static value. It doesn't need to be initialized every render.

src/pages/commits/ActiveCommits/CommitChartLabel.tsx
```diff
@@ -36,11 +58,31 @@ const LabelContainer = styled.div`
...
+`;
+
+const TooltipContainer = styled(Body)`
+  width: 200px;
+  border-radius: 3px;
+  background-color: ${gray.light3};
+  color: ${gray.dark3};
+  word-break: break-word;
+  text-align: left;
 `;
```
khelif96, line 88: Is this styling necessary can we just use the default tooltip styles?

src/pages/commits/ActiveCommits/CommitChartLabel.test.tsx
ziyielenachen: Nice catch! Thanks!

src/pages/commits/ActiveCommits/CommitChartLabel.tsx
ziyielenachen: You're right! Removed most of the styling.

src/pages/commits/ActiveCommits/CommitChartLabel.test.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  test("Displays entire commit message if it does not break length limit", async () => {
+    const { queryByDataCy } = render(RenderCommitChartLabel(versionShort));
+    expect(queryByDataCy("commit-label")).toHaveTextContent(
+      "SERVER-57332 Create skeleton Internal"
+    );
+  });
+
+  test(
+    "Clicking on the 'more' button should open a tooltip that displays" +
```
arjunpatel100, line 46: nit: can use one string for the test description

src/pages/commits/ActiveCommits/CommitChartLabel.test.tsx
```diff
@@ -0,0 +1,81 @@
...
+    const { queryByDataCy } = render(RenderCommitChartLabel(versionShort));
+    expect(queryByDataCy("commit-label")).toHaveTextContent(
+      "4137c 6/16/21 11:38 PMMohamed Khelif"
+    );
+  });
+
+  test(
+    "Displays shortened commit message and the 'more' button if commit message" +
+      "is too long",
+    async () => {
```
arjunpatel100, line 29: do we get the same result after removing `async`. I don't think we need it here since we don't use `await` in the function body

src/pages/commits/ActiveCommits/CommitChartLabel.test.tsx
```diff
@@ -0,0 +1,81 @@
...
+      const { queryByDataCy } = render(RenderCommitChartLabel(versionLong));
+      expect(queryByDataCy("tooltip-button")).toBeInTheDocument();
+      expect(queryByDataCy("commit-label")).toHaveTextContent(
+        "4137c 6/16/21 11:38 PMMohamed Khelif -SERVER-57332 Create skeleton InternalDo...more"
+      );
+    }
+  );
+
+  test("Displays entire commit message if it does not break length limit", async () => {
```
arjunpatel100, line 38: do we get the same result after removing `async`. I don't think we need it here since we don't use `await` in the function body

src/pages/commits/ActiveCommits/CommitChartLabel.test.tsx
```diff
@@ -0,0 +1,81 @@
...
+    const { queryByDataCy } = render(RenderCommitChartLabel(versionShort));
+    expect(queryByDataCy("commit-label")).toHaveTextContent(
+      "SERVER-57332 Create skeleton Internal"
+    );
+  });
+
+  test(
+    "Clicking on the 'more' button should open a tooltip that displays" +
+      "the long commit message",
+    async () => {
```
arjunpatel100, line 48: do we get the same result after removing async. I don't think we need it here since we don't use await in the function body

arjunpatel100 (changes requested):

arjunpatel100 (approved):

PR #839, opened 2021-08-04
EVG-13851: History Table Icon

arjunpatel100:
[EVG-13851](https://jira.mongodb.org/browse/EVG-13851)

### Description 
Renders the history table icon to be used in the task and build variant history pages
### Screenshots
figma: https://www.figma.com/file/TxlYD1vlJPgZ8S2Z8sZfe9/Mainline-Commit-UI?node-id=0%3A2707
storybook:
Screen Shot 2021-08-04 at 6 20 44 PM


src/pages/commits/HistoryTableIcon/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+  align-items: center;
+`;
+
+interface IconContainerProps {
+  inactive?: boolean;
+}
+const IconContainer = styled.div<IconContainerProps>`
+  ${({ inactive }) => inactive && "opacity: .4;"}
+  height: 30px;
+  width: 30px;
```
arjunpatel100, line 59: Currently some of the icons are different heights ranging between 28-30px. We should be able to remove this height and width after I upload updated svgs.

src/components/TaskStatusIcon/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { SystemFailure } from "./SystemFailure";
+import { TimedOut } from "./TimedOut";
+import { WillNotRun } from "./WillNotRun";
+import { WillRun } from "./WillRun";
+
+interface TaskStatusIconProps {
+  status: TaskStatus;
+}
+
+export const TaskStatusIcon: React.FC<TaskStatusIconProps> = ({ status }) => {
```
arjunpatel100, line 17: We still need to confirm some of these status/icon mappings. I added it to the design meeting agenda for 8/5

src/pages/commits/HistoryTableIcon/index.tsx
ziyielenachen: Would it be possible for us to choose a specific width when using the TaskStatusIcons? Asking because there is [this comment](https://www.figma.com/file/TxlYD1vlJPgZ8S2Z8sZfe9/Mainline-Commit-UI?node-id=0%3A1) on the Figma.

src/components/TaskStatusIcon/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+      return <Icon glyph="TimedOut" size={30} />;
+    case TaskStatus.Aborted:
+    case TaskStatus.Blocked:
+    case TaskStatus.Unscheduled:
+      return <Icon glyph="WillNotRun" size={30} />;
+    case TaskStatus.WillRun:
+    case TaskStatus.Pending:
+      return <Icon glyph="Calendar" fill={gray.dark3} size={30} />;
+    default:
+      return <>{status}</>;
```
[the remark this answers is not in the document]
arjunpatel100: I don't think so. This condition is only met when a task status from evg is unsupported in Spruce. I removed the text and opted for a bugsnag log instead. This is how we do it in the TaskStatusBadge: https://github.com/evergreen-ci/spruce/blob/main/src/components/TaskStatusBadge/index.tsx#L78-L83

src/pages/commits/HistoryTableIcon/index.tsx
arjunpatel100: We can choose the size of the icon after switching over to using the leafygreen Icon component instead of rendering the SVG directly
https://github.com/evergreen-ci/spruce/pull/839/files#diff-99f280eb2b949701a956ff9ebb2408253a12a15d650f42b0ba7bba675ba80b02R16

src/pages/commits/HistoryTableIcon/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+  flex-direction: column;
+  align-items: center;
+  ${({ onClick }) => onClick && "cursor: pointer;"}
+`;
+
+interface IconContainerProps {
+  inactive?: boolean;
+}
+const IconContainer = styled.div<IconContainerProps>`
+  ${({ inactive }) => inactive && "opacity: .4;"}
```
sophstad, line 64: [nit] It seems like this same inactive styling appears [elsewhere](https://github.com/evergreen-ci/spruce/blob/cf94d556699ba8d6f862c3cc5be3b50dd7b5c619/src/pages/commits/ActiveCommits/CommitChartTooltip.tsx#L98), could be nice to make it a reusable style

src/components/TaskStatusIcon/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+      return <Icon glyph="Checkmark" fill={green.base} size={30} />;
+    case TaskStatus.Failed:
+      return <Icon glyph="X" fill={red.base} size={30} />;
+    case TaskStatus.KnownIssue:
+      return <Icon glyph="KnownFailureIcon" fill={red.base} size={30} />;
+    case TaskStatus.Dispatched:
+    case TaskStatus.Started:
+      return <Icon glyph="Refresh" fill={yellow.dark2} size={30} />;
+    case TaskStatus.SetupFailed:
+      return <Icon glyph="SetupFailure" size={30} />;
```
sophstad, line 25: [nit] It may be preferable to have consistent usage of the `fill` prop among all icons (e.g. have Spruce custom icons accept `fill` in the same manner as LG), but definitely not a necessary change.

sophstad (approved):

src/components/TaskStatusIcon/index.tsx
arjunpatel100: nice catch. this will be a useful change to follow updates in @leafygreen-ui/palette

khelif96 (approved):

src/components/Icon/icons/index.tsx
```diff
@@ -45,30 +45,6 @@ export const FailedIcon = () => (
   </svg>
 );
 
-export const RunningIcon = () => (
```
khelif96, line 48: This is depended on by a [component](https://github.com/evergreen-ci/spruce/blob/main/src/components/PatchesPage/patchCard/BuildStatusIcon.tsx#L37) We still need this Icon or need to replace it where we use it.

khelif96: This is why the tests  are failing.

khelif96:
Completed in https://github.com/evergreen-ci/spruce/pull/846

PR #840, opened 2021-08-06
EVG-15187  Evergreen Use project display name instead of id for downstream tasks

malikchaya2:
[EVG-15187](https://jira.mongodb.org/browse/EVG-15187)

### Description 
We want the downstream task to show the name of the project instead of object id. 

### Screenshots
before: 
image

after: 
image

### Testing 
tested manually

### DIFF
--- a/src/gql/queries/get-patch.graphql
+++ b/src/gql/queries/get-patch.graphql
@@ -7,7 +7,7 @@ query Patch($id: String!) {
       baseVersionID
       githash
       id
-      projectID
+      projectIdentifier
       taskCount
       status
     }
--- a/src/gql/queries/get-patch.graphql
+++ b/src/gql/queries/get-patch.graphql
@@ -7,7 +7,7 @@ query Patch($id: String!) {
       baseVersionID
       githash
       id
-      projectIdentifier
+      projectID
       taskCount
       status
     }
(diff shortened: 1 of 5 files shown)


src/gql/queries/get-patch.graphql
```diff
@@ -7,7 +7,7 @@ query Patch($id: String!) {
       baseVersionID
       githash
       id
-      projectID
```
arjunpatel100, line 10: why do we need to change this field? will DownstreamTasks read this field?

src/gql/queries/get-version.graphql
```diff
@@ -35,7 +35,7 @@ query Version($id: String!) {
         baseVersionID
         githash
         id
-        projectID
```
arjunpatel100, line 38: why do we need to change this field? will DownstreamTasks read this field?

arjunpatel100 (changes requested):
we need to run `yarn run codegen` after making changes to graphql files

src/gql/queries/get-version.graphql
malikchaya2: turns out this one is the only one we need. I removed the others.

src/gql/queries/get-patch.graphql
malikchaya2: removed

khelif96 (approved):

sophstad (approved):

arjunpatel100 (approved):

PR #842, opened 2021-08-09
EVG-13902 Create manifest link

khelif96:
[EVG-13902](https://jira.mongodb.org/browse/EVG-13902)

### Description 
Creates a link in the version metadata that opens the manifest in a new tab

### Screenshots
image

opens this
image

### Testing 
Tested manually. I couldn't figure out a way to write any tests for this
CreateObjectURL isn't supported in jest and can't really be mocked since it does some stuff behind the scenes to generate the url.
Can't test in cypress because `createObjectURL` changes the domain name by prepending a `blob` to the link and cypress wont allow cross origin requests.
### Evergreen PR 
https://github.com/evergreen-ci/evergreen/pull/4904

### DIFF
--- a/src/pages/version/ManifestBlob.tsx
+++ b/src/pages/version/ManifestBlob.tsx
@@ -4,7 +4,7 @@ import { Manifest } from "gql/generated/types";
 import { string } from "utils";
 
 const { omitTypename } = string;
-// typescript interface with manifest that is an object of key-value pairs
+
 interface Props {
   manifest: Manifest;
 }
--- a/src/pages/version/ManifestBlob.tsx
+++ b/src/pages/version/ManifestBlob.tsx
@@ -1,25 +1,12 @@
 import { StyledLink } from "components/styles";
 import { P2 } from "components/Typography";
+import { Manifest } from "gql/generated/types";
 import { string } from "utils";
 
 const { omitTypename } = string;
 // typescript interface with manifest that is an object of key-value pairs
 interface Props {
-  manifest: {
-    id: string;
-    revision: string;
-    project: string;
-    branch: string;
-    isBase: boolean;
-    moduleOverrides?: {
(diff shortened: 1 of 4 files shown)


ablack12 (approved):

src/pages/version/Metadata.tsx
```diff
@@ -69,11 +86,12 @@ export const Metadata: React.FC<Props> = ({ loading, version }) => {
       <P2>{`Submitted by: ${author}`}</P2>
       {baseVersionID && revision && (
         <P2>
+          Base commit:{" "}
           <StyledLink
             data-cy="patch-base-commit"
             href={`${getUiUrl()}/version/${baseVersionID}`}
           >
-            Base commit: {revision.slice(0, 10)}
+            {revision.slice(0, 10)}
```
sophstad, line 94: Nice, this was a lil pet peeve of mine 😁

src/pages/version/ManifestBlob.tsx
```diff
@@ -0,0 +1,43 @@
...
+    branch: string;
+    isBase: boolean;
+    moduleOverrides?: {
+      [key: string]: string;
+    };
+    modules?: {
+      [key: string]: {
+        [key: string]: string;
+      };
+    };
```
sophstad, line 21: Did you enumerate this instead of importing from `gql/generated/types` in order to have more precision over the `modules` and `moduleOverrides` type?

sophstad (approved):

src/pages/version/ManifestBlob.tsx
khelif96: Thats a good point. That was my reason for defining the type, but you are correct we don't actually do anything with it apart from reexporting that data so it doesn't make sense to redeclare the type. I updated it to just use the generated type

PR #847, opened 2021-08-17
EVG-15257 Create a reusable multi select  component with search

khelif96:
[EVG-15257](https://jira.mongodb.org/browse/EVG-15257)

[Check the storybook build when the storybook task completes for a live demo.](https://mciuploads.s3.amazonaws.com/spruce/spruce_ubuntu1604_storybook_patch_cf94d556699ba8d6f862c3cc5be3b50dd7b5c619_611d308c2fbabe101d249f8c_21_08_18_16_08_45/storybook/index.html?path=/story/searchable-dropdown--story)

### Testing 
Unit tests, and manual storybook tests

### DIFF
--- a/src/components/SearchableDropdown/index.tsx
+++ b/src/components/SearchableDropdown/index.tsx
@@ -1,0 +1,170 @@
+import { useState, PropsWithChildren, useRef } from "react";
+import styled from "@emotion/styled";
+import IconButton from "@leafygreen-ui/icon-button";
+import { uiColors } from "@leafygreen-ui/palette";
+import { Body, Label } from "@leafygreen-ui/typography";
+import { Input } from "antd";
+import Icon from "components/Icon";
+import { useOnClickOutside } from "hooks";
+
+const { Search } = Input;
+const { gray, white, blue } = uiColors;
+
+interface SearchableDropdownProps<ObjectType> {
+  label: string | React.ReactNode;
+  value: string | ObjectType[];
+  onChange: (value: string | ObjectType) => void;
+  searchPlaceholder?: string;
+  options: string[] | ObjectType[];
+  optionRenderer?: (option: string | ObjectType) => React.ReactNode;
+  allowMultiselect?: boolean;
+}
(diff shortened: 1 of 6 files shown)


src/components/SearchableDropdown/index.tsx
```diff
@@ -0,0 +1,237 @@
...
+      <Option
+        onClick={() => onClick(v)}
+        key={`select_${v}`}
+        data-cy="searchable-dropdown-option"
+      >
+        <CheckmarkContainer>
+          {isChecked(v) && (
+            <Icon glyph="Checkmark" height={12} width={12} fill={blue.base} />
+          )}
+        </CheckmarkContainer>
```
sophstad, line 88: Right now multi-select is slightly hidden from the user imo -- there's no visual difference between the single and multi select dropdown when it's initially opened. Maybe including empty checkmark boxes when `isChecked` is false would make this more apparent?

src/components/SearchableDropdown/index.tsx
```diff
@@ -0,0 +1,237 @@
...
+        <StyledButton
+          ref={menuButtonRef} // @ts-expect-error
+          onClick={() => setisOpen((curr) => !curr)}
+          data-cy="searchable-dropdown"
+          id="searchable-dropdown"
+          value={value}
+        >
+          <ButtonContent>
+            <LabelWrapper>
+              <Body data-cy="dropdown-value">{value}</Body>
```
sophstad, line 140: When multi-select is used and `value` is an array, it'd be nice to display the values here as a comma-separated list rather than just concatenating them
image

sophstad (changes requested):
this is awesome!

src/components/SearchableDropdown/index.tsx
khelif96: image
I based it off of the designs used here.

src/components/SearchableDropdown/index.tsx
khelif96: done

sophstad (approved):

PR #648, opened 2021-02-09
EVG-13992 Replace banners with toasts throughout the app

khelif96:
Sorry about all these large PRs! 
https://user-images.githubusercontent.com/4605522/107401639-14867f80-6ad1-11eb-9f35-9709f8350196.mp4

PR #848, opened 2021-08-19
EVG-15074: Add project settings page structure

sophstad:
EVG-15074
[Figma](https://www.figma.com/file/Z0rfsPjVKglRq9baDhsSux/Project-Settings-Sections)

### Description
- Add project settings pages to `/project/<project>/settings`
- Nested pages (e.g. "General Configuration" under "General Settings") will be added in a future PR since they represent anchor links to content that doesn't exist yet. Nesting also isn't available in `@leafygreen-ui/side-nav@5.0.2` ([see changelog](https://github.com/mongodb/leafygreen-ui/blob/main/packages/side-nav/CHANGELOG.md#710))
- We need to upgrade `@leafygreen-ui/side-nav` in order to match the designs and use nested pages; this work is captured in EVG-15289

### Screenshots
image

### DIFF
--- a/src/constants/routes.ts
+++ b/src/constants/routes.ts
@@ -54,6 +54,22 @@ const paths = {
   jobLogs: "/job-logs",
 };
 
+const projectSettingsSlug = `${paths.project}/:id/${PageNames.Settings}`;
+
+const projectSettingsRoutes = {
+  projectSettings: `${projectSettingsSlug}/:tab?`,
+  projectSettingsAccess: `${projectSettingsSlug}/${ProjectSettingsTabRoutes.Access}`,
+  projectSettingsGeneral: `${projectSettingsSlug}/${ProjectSettingsTabRoutes.General}`,
+  projectSettingsGitHubCommitQueue: `${projectSettingsSlug}/${ProjectSettingsTabRoutes.GitHubCommitQueue}`,
+  projectSettingsEventLog: `${projectSettingsSlug}/${ProjectSettingsTabRoutes.EventLog}`,
+  projectSettingsNotifications: `${projectSettingsSlug}/${ProjectSettingsTabRoutes.Notifications}`,
+  projectSettingsPatchAliases: `${projectSettingsSlug}/${ProjectSettingsTabRoutes.PatchAliases}`,
+  projectSettingsPeriodicBuilds: `${projectSettingsSlug}/${ProjectSettingsTabRoutes.PeriodicBuilds}`,
+  projectSettingsProjectTriggers: `${projectSettingsSlug}/${ProjectSettingsTabRoutes.ProjectTriggers}`,
+  projectSettingsVariables: `${projectSettingsSlug}/${ProjectSettingsTabRoutes.Variables}`,
+  projectSettingsVirtualWorkstation: `${projectSettingsSlug}/${ProjectSettingsTabRoutes.VirtualWorkstation}`,
+};
+
 export const routes = {
   cliPreferences: `${paths.preferences}/${PreferencesTabRoutes.CLI}`,
   commitQueue: `${paths.commitQueue}/:id`,
@@ -68,17 +84,6 @@ export const routes = {
   preferences: `${paths.preferences}/:tab?`,
   profilePreferences: [`${paths.preferences}/${PreferencesTabRoutes.Profile}`],
   projectPatches: `${paths.project}/:id/${PageNames.Patches}`,
-  projectSettings: `${paths.project}/:id/${PageNames.Settings}/:tab?`,
-  projectSettingsAccess: `${paths.project}/:id/${PageNames.Settings}/${ProjectSettingsTabRoutes.Access}`,
-  projectSettingsGeneral: `${paths.project}/:id/${PageNames.Settings}/${ProjectSettingsTabRoutes.General}`,
-  projectSettingsGitHubCommitQueue: `${paths.project}/:id/${PageNames.Settings}/${ProjectSettingsTabRoutes.GitHubCommitQueue}`,
-  projectSettingsEventLog: `${paths.project}/:id/${PageNames.Settings}/${ProjectSettingsTabRoutes.EventLog}`,
-  projectSettingsNotifications: `${paths.project}/:id/${PageNames.Settings}/${ProjectSettingsTabRoutes.Notifications}`,
-  projectSettingsPatchAliases: `${paths.project}/:id/${PageNames.Settings}/${ProjectSettingsTabRoutes.PatchAliases}`,
-  projectSettingsPeriodicBuilds: `${paths.project}/:id/${PageNames.Settings}/${ProjectSettingsTabRoutes.PeriodicBuilds}`,
-  projectSettingsProjectTriggers: `${paths.project}/:id/${PageNames.Settings}/${ProjectSettingsTabRoutes.ProjectTriggers}`,
-  projectSettingsVariables: `${paths.project}/:id/${PageNames.Settings}/${ProjectSettingsTabRoutes.Variables}`,
-  projectSettingsVirtualWorkstation: `${paths.project}/:id/${PageNames.Settings}/${ProjectSettingsTabRoutes.VirtualWorkstation}`,
   publicKeysPreferences: `${paths.preferences}/${PreferencesTabRoutes.PublicKeys}`,
   spawn: `${paths.spawn}/:tab?`,
   spawnHost: `${paths.spawn}/${SpawnTab.Host}`,
@@ -92,6 +97,7 @@ export const routes = {
   variantHistory: `${paths.variantHistory}/:projectId/:variantId`,
   taskHistory: `${paths.taskHistory}/:projectId/:taskId`,
   jobLogs: `${paths.jobLogs}/:taskId/:execution/:groupId?`,
+  ...projectSettingsRoutes,
 };
 
 export const DEFAULT_PATCH_TAB = PatchTab.Tasks;
--- a/src/constants/routes.ts
+++ b/src/constants/routes.ts
@@ -6,6 +6,7 @@ const { stringifyQuery } = queryString;
 
 export enum PageNames {
   Patches = "patches",
+  Settings = "settings",
 }
 
 export enum SpawnTab {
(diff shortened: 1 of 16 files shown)


malikchaya2 (commented):
This looks great! 
How should we go about hiding it from the user until it's ready to use? Should we just rely on the fact that it's not on the menu? 
cc @ablack12

ablack12:
Wow this is exciting! 
We may want to do more than relying on it not being in the menu, since users can be smart about hacking URLs and as we add things to the page it might be in a halfway state; I'm not sure how this can be done though; could restrict this to superusers for now (not sure if there's a way to do that though).

malikchaya2:
> Wow this is exciting!
> We may want to do more than relying on it not being in the menu, since users can be smart about hacking URLs and as we add things to the page it might be in a halfway state; I'm not sure how this can be done though; could restrict this to superusers for now (not sure if there's a way to do that though).

I don't think that that is very doable right now, especially before https://jira.mongodb.org/browse/EVG-15038. It would probably be easier to add a boolean somewhere at a top level that just hides the whole page (and maybe adds a "coming soon" message for the crafty url hackers, lol).

src/constants/routes.ts
```diff
@@ -54,6 +68,17 @@ export const routes = {
...
+  projectSettingsAccess: `${paths.project}/:id/${PageNames.Settings}/${ProjectSettingsTabRoutes.Access}`,
+  projectSettingsGeneral: `${paths.project}/:id/${PageNames.Settings}/${ProjectSettingsTabRoutes.General}`,
+  projectSettingsGitHubCommitQueue: `${paths.project}/:id/${PageNames.Settings}/${ProjectSettingsTabRoutes.GitHubCommitQueue}`,
+  projectSettingsEventLog: `${paths.project}/:id/${PageNames.Settings}/${ProjectSettingsTabRoutes.EventLog}`,
+  projectSettingsNotifications: `${paths.project}/:id/${PageNames.Settings}/${ProjectSettingsTabRoutes.Notifications}`,
+  projectSettingsPatchAliases: `${paths.project}/:id/${PageNames.Settings}/${ProjectSettingsTabRoutes.PatchAliases}`,
+  projectSettingsPeriodicBuilds: `${paths.project}/:id/${PageNames.Settings}/${ProjectSettingsTabRoutes.PeriodicBuilds}`,
+  projectSettingsProjectTriggers: `${paths.project}/:id/${PageNames.Settings}/${ProjectSettingsTabRoutes.ProjectTriggers}`,
+  projectSettingsVariables: `${paths.project}/:id/${PageNames.Settings}/${ProjectSettingsTabRoutes.Variables}`,
+  projectSettingsVirtualWorkstation: `${paths.project}/:id/${PageNames.Settings}/${ProjectSettingsTabRoutes.VirtualWorkstation}`,
```
khelif96, line 81: Woah, So many routes. 
Maybe we could clean this up and extract these into a separate object or separate the route slugs into their own variable so we can reuse it or even a combination of the two.

```
const projectSettingsRoutes = {
  projectSettingsAccess: `${paths.project}/:id/${PageNames.Settings}/${ProjectSettingsTabRoutes.Access}`,
  projectSettingsGeneral: `${paths.project}/:id/${PageNames.Settings}/${ProjectSettingsTabRoutes.General}`,
  projectSettingsGitHubCommitQueue: `${paths.project}/:id/${PageNames.Settings}/${ProjectSettingsTabRoutes.GitHubCommitQueue}`,
  ...
}

export const routes = {
  preferences: `${paths.preferences}/:tab?`,
  profilePreferences: [`${paths.preferences}/${PreferencesTabRoutes.Profile}`],
  projectPatches: `${paths.project}/:id/${PageNames.Patches}`,
  ...projectSettingsRoutes,
}
```
or
```
const projectSettingsSlug = `${paths.project}/:id/${PageNames.Settings}`
...
projectSettingsAccess: `${projectSettingsSlug}/${ProjectSettingsTabRoutes.Access}`,
projectSettingsGeneral: `${projectSettingsSlug}/${ProjectSettingsTabRoutes.General}`
```

What do you think?

src/pages/projectSettings/ProjectSettingsTabs.tsx
```diff
@@ -0,0 +1,116 @@
...
+  PeriodicBuildsTab,
+  ProjectTriggersTab,
+  VariablesTab,
+  VirtualWorkstationTab,
+} from "./projectSettingsTabs/index";
+
+export const ProjectSettingsTabs: React.FC = () => {
+  const { tab } = useParams<{ tab: string }>();
+
+  useEffect(() => {}, [tab]); // eslint-disable-line react-hooks/exhaustive-deps
```
khelif96, line 22: What is this for?

src/pages/projectSettings/ProjectSettingsTabs.tsx
```diff
@@ -0,0 +1,116 @@
...
+  EventLogTab,
+  GeneralTab,
+  GitHubCommitQueueTab,
+  NotificationsTab,
+  PatchAliasesTab,
+  PeriodicBuildsTab,
+  ProjectTriggersTab,
+  VariablesTab,
+  VirtualWorkstationTab,
+} from "./projectSettingsTabs/index";
```
khelif96, line 17: You shouldn't need to import from `/index` directly is the file not being resolved automatically?

src/pages/projectSettings/ProjectSettingsTabs.tsx
```diff
@@ -0,0 +1,116 @@
...
+      <Route
+        path={routes.projectSettingsPeriodicBuilds}
+        component={PeriodicBuildsTab}
+      />
+      <Route path={routes.projectSettingsEventLog} component={EventLogTab} />
+    </Container>
+  );
+};
+
+export const getTitle = (
```
khelif96, line 65: Do you think we might need this in other pages apart from the `ProjectSettings` Page itself and the `ProjectSettingsTabs` component above? If so it might make sense to move this to a more shared constants or utils file in this same directory

khelif96:
> Wow this is exciting!
> We may want to do more than relying on it not being in the menu, since users can be smart about hacking URLs and as we add things to the page it might be in a halfway state; I'm not sure how this can be done though; could restrict this to superusers for now (not sure if there's a way to do that though).

@ablack12  you have a good point this will be one of the first permission gated routes we have on spruce. We haven't really given much thought to how we should handle user permissions on spruce. It would be a good time to come up with a strategy.

khelif96:
> https://jira.mongodb.org/browse/EVG-15038



> > Wow this is exciting!
> > We may want to do more than relying on it not being in the menu, since users can be smart about hacking URLs and as we add things to the page it might be in a halfway state; I'm not sure how this can be done though; could restrict this to superusers for now (not sure if there's a way to do that though).
> 
> I don't think that that is very doable right now, especially before https://jira.mongodb.org/browse/EVG-15038. It would probably be easier to add a boolean somewhere at a top level that just hides the whole page (and maybe adds a "coming soon" message for the crafty url hackers, lol).

Good point there is likely some middleware we can add to gqlgen to handle permission checks automagically. https://gqlgen.com/reference/directives/  

We would need to do this on the resolver level. As well as have a separate mechanism for the UI

src/pages/projectSettings/ProjectSettingsTabs.tsx
sophstad: No, I actually noticed this happens elsewhere in the codebase (e.g. [constants/index](https://github.com/evergreen-ci/spruce/search?q=%22constants+index%22)). Maybe worth looking into separately?

src/pages/projectSettings/ProjectSettingsTabs.tsx
sophstad: Ha, this is a holdover from `PreferencesTabs.tsx` but looks like it [should have been removed there too](https://github.com/evergreen-ci/spruce/pull/648#discussion_r573096670). I'll delete both!

src/pages/projectSettings/ProjectSettingsTabs.tsx
sophstad: Just had another look through the Figma and I don't really think so — but will keep this in mind if that does come up!

src/pages/ProjectSettings.tsx
```diff
@@ -0,0 +1,127 @@
...
+  getProjectSettingsRoute,
+} from "constants/routes";
+import { usePageTitle } from "hooks";
+import { environmentalVariables } from "utils";
+import {
+  ProjectSettingsTabs,
+  getTitle,
+} from "./projectSettings/ProjectSettingsTabs";
+
+const { isProduction } = environmentalVariables;
```
malikchaya2, line 16: This checks if it's "production", but at what point does this page get set to "development"?

sophstad: The [`isProduction()`](https://github.com/evergreen-ci/spruce/blob/main/src/utils/environmentalVariables.ts#L21-L22) util uses the `process.env.NODE_ENV` environment variable, which is set [here](https://github.com/evergreen-ci/spruce/blob/main/scripts/build.js#L5) for prod and [here](https://github.com/evergreen-ci/spruce/blob/main/scripts/start.js#L5) for dev.

malikchaya2: Ah, I was misreading what this is doing. It makes sense, thanks!

malikchaya2 (approved):

src/pages/ProjectSettings.tsx
khelif96: This happens in the build step for deploys this variable gets set to true automatically.
When developing locally this variable will be false.

src/pages/ProjectSettings.tsx
```diff
@@ -0,0 +1,127 @@
...
+      <Redirect
+        to={getProjectSettingsRoute(
+          projectId,
+          ProjectSettingsTabRoutes.General
+        )}
+      />
+    );
+  }
+
+  if (disablePage) {
```
khelif96, line 34: nit. maybe we should do this earlier before redirecting?

khelif96 (approved):
Looks great!

PR #850, opened 2021-08-20
EVG-13862 mainline commits pagination buttons

khelif96:
[EVG-13862](https://jira.mongodb.org/browse/EVG-13862)

### Description 
Create pagination buttons

### Screenshots

https://user-images.githubusercontent.com/4605522/130266521-b5674ce5-79d0-4f6b-993b-ebb8a86cfad1.mov

### Testing 
On staging
### Evergreen PR 
https://github.com/evergreen-ci/evergreen/pull/4960

### DIFF
--- a/src/components/FilterBadges/index.tsx
+++ b/src/components/FilterBadges/index.tsx
@@ -2,7 +2,10 @@ import styled from "@emotion/styled";
 import Button, { Variant, Size } from "@leafygreen-ui/button";
 import { useLocation } from "react-router";
 import { useUpdateURLQueryParams } from "hooks/useUpdateURLQueryParams";
-import { ChartToggleQueryParams } from "types/commits";
+import {
+  ChartToggleQueryParams,
+  MainlineCommitQueryParams,
+} from "types/commits";
 import { PatchTasksQueryParams } from "types/task";
 import { queryString, array } from "utils";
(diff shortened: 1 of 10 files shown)


src/components/FilterBadges/index.tsx
```diff
@@ -11,15 +14,18 @@ import { SeeMoreModal } from "./SeeMoreModal";
 const { convertObjectToArray } = array;
 const { parseQueryString } = queryString;
 
+const hiddenQueryParams = [
```
sophstad, line 17: Great use case for a [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set)!

src/pages/Commits.tsx
```diff
@@ -99,41 +109,51 @@ export const Commits = () => {
...
+          isLoading={loading}
+          chartType={currentChartType}
+          hasTaskFilter={hasTaskFilter}
+        />
+      </PageContainer>
     </PageWrapper>
   );
 };
 
+const PageContainer = styled.div`
```
sophstad, line 153: I feel like we include something like this on every page 😆 wonder if we should think about standardizing page containers at some point. but no need to address now!

sophstad (commented):
Looks good, do you anticipate merging in the backend changes will fix those e2e tests?

PR #855, opened 2021-08-31
EVG-15287: Add trigger aliases on configure page

sophstad:
EVG-15287

### Description 
- Add "Downstream Tasks" section on patch configure page, allowing aliases to be scheduled from the UI _only_ if the patch has not yet run
- Also show child patches that have been invoked via `--trigger-alias` on the CLI; these patches are read-only on the UI and interaction is disabled
- Upcoming work
  - Show name of trigger used to invoke child patch (EVG-15385)
  - Attach tasks/variants to aliases so that users can see which tasks will be run

### Screenshots
image
image
image

### Testing 
- Staging has some good data for testing this page; `yarn run staging` and visit http://localhost:3000/patch/61378d50b237365d9b57e1aa/configure/tasks
- Add integration tests

### DIFF
--- a/src/hooks/useConfigurePatch.ts
+++ b/src/hooks/useConfigurePatch.ts
@@ -73,7 +73,7 @@ const reducer = (state: ConfigurePatchState, action: Action) => {
     case "setSelectedAliases":
       return {
         ...state,
-        setSelectedAliases: action.aliases,
+        selectedAliases: action.aliases,
       };
     case "setPatchParams":
       return {
--- a/src/hooks/useConfigurePatch.ts
+++ b/src/hooks/useConfigurePatch.ts
@@ -16,12 +16,12 @@ const { omitTypename } = string;
 
 type ConfigurePatchState = {
   description: string;
+  selectedAliases: AliasState;
   selectedBuildVariants: string[];
   selectedBuildVariantTasks: VariantTasksState;
   patchParams: ParameterInput[];
   selectedTab: number;
   disableBuildVariantSelect: boolean;
-  selectedDownstreamPatches: DownstreamPatchState;
 };
 
 type Action =
@@ -36,21 +36,21 @@ type Action =
       buildVariants: string[];
       params: ParameterInput[];
       variantTasks: VariantTasksState;
-      downstreamPatches: DownstreamPatchState;
+      aliases: AliasState;
     }
   | {
-      type: "setSelectedDownstreamPatches";
-      downstreamPatches: DownstreamPatchState;
+      type: "setSelectedAliases";
+      aliases: AliasState;
     };
 
 const initialState = ({ selectedTab = 0 }: { selectedTab: number }) => ({
   description: "",
+  selectedAliases: {},
   selectedBuildVariants: [],
   selectedBuildVariantTasks: {},
   patchParams: null,
   selectedTab,
   disableBuildVariantSelect: tabToIndexMap[selectedTab] === PatchTab.Tasks,
-  selectedDownstreamPatches: {},
 });
 
 const reducer = (state: ConfigurePatchState, action: Action) => {
@@ -70,10 +70,10 @@ const reducer = (state: ConfigurePatchState, action: Action) => {
         ...state,
         selectedBuildVariantTasks: action.variantTasks,
       };
-    case "setSelectedDownstreamPatches":
+    case "setSelectedAliases":
       return {
         ...state,
-        selectedDownstreamPatches: action.downstreamPatches,
+        setSelectedAliases: action.aliases,
       };
     case "setPatchParams":
       return {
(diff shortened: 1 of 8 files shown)


src/pages/configurePatch/configurePatchCore/ConfigureTasks.tsx
```diff
@@ -39,40 +57,73 @@ export const ConfigureTasks: React.FC<Props> = ({
...
-    selectedBuildVariants.length > 1 ? "these variants" : "this variant"
-  }`;
+  const selectAllCheckboxState = getSelectAllCheckboxState(
+    currentTasks,
+    currentDownstreamPatches,
+    enumerateChildPatches
+  );
+  const selectAllCheckboxCopy =
+    Object.entries(currentTasks).length === 0
+      ? `Select all tasks in ${
```
malikchaya2, line 114: I'm personally a bigger fan of wording it "Add alias to patch" as opposed to "select all tasks in ___", because the user is not able to have any control over the tasks in the alias. But if you prefer to keep it like this and worry about it in the ticket that adds the tasks, that works too.

src/pages/configurePatch/configurePatchCore/ConfigureTasks.tsx
```diff
@@ -39,40 +57,73 @@ export const ConfigureTasks: React.FC<Props> = ({
...
+        }`
+      : `Select all tasks in ${
+          selectedBuildVariants.length > 1 ? "these variants" : "this variant"
+        }`;
   const selectedTaskDisclaimerCopy = `${taskCount} task${
     taskCount !== 1 ? "s" : ""
   } across ${buildVariantCount} build variant${
     buildVariantCount !== 1 ? "s" : ""
-  }`;
+  }, ${aliasCount} trigger alias${aliasCount !== 1 ? "es" : ""}`;
```
malikchaya2, line 126: I don't know if I'm a fan of displaying  "X tasks across X build variants, X trigger aliases", because it will always be either tasks selected, or trigger aliases selected. I'd personally either only show "X trigger aliases" on the trigger aliases tabs and "X tasks across X build variants" on the main tab, or show "X tasks across X build variants" on the main tab but nothing on the trigger alias tab. I'm open to other opinions though, and I'm not against this being discussed/tackled in a separate ticket.

malikchaya2 (approved):
Added some general nits, but I'm fine with this being merged as is and these things being addressed/discussed in the ticket that adds the tasks.

src/hooks/useConfigurePatch.ts
```diff
@@ -0,0 +1,246 @@
...
+        ...state,
+        description: action.description,
+        selectedBuildVariants: action.buildVariants,
+        patchParams: omitTypename(action.params),
+        selectedBuildVariantTasks: action.variantTasks,
+        selectedDownstreamPatches: action.downstreamPatches,
+      };
+
+    default:
+      throw new Error();
```
khelif96, line 106: Can we throw a more detailed error here? So its easier to trace down errors 
`Unrecognized reducer action ${action.type}`

khelif96 (changes requested):
Great work on this Refactor it works well! Just had a few comments

src/pages/configurePatch/configurePatchCore/ConfigureTasks.tsx
```diff
@@ -113,20 +169,92 @@ export const ConfigureTasks: React.FC<Props> = ({
...
+                disabled
+                checked
+              />
+            ))}
+          </Tasks>
+        </>
+      )}
+      {enumerateChildPatches && (
+        <>
+          {currentChildPatches[0].variantsTasks.map(
```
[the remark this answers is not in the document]
sophstad: No, we only want to show the variant/task details if there is only one patch selected. That logic is handled by `enumerateChildPatches`, where we check if `currentChildPatches.length === 1`.

src/pages/configurePatch/configurePatchCore/ConfigureTasks.tsx
```diff
@@ -39,40 +57,73 @@ export const ConfigureTasks: React.FC<Props> = ({
...
+  const enumerateChildPatches =
+    currentChildPatches.length === 1 && selectedBuildVariants.length === 1;
 
   const onClickCheckbox = (taskName: string) => (e) => {
     const selectedBuildVariantsCopy = { ...selectedBuildVariantTasks };
-    selectedBuildVariants.forEach((bv) => {
-      if (selectedBuildVariantsCopy[bv][taskName] !== undefined) {
-        selectedBuildVariantsCopy[bv][taskName] = e.target.checked;
+    const selectedDownstreamPatchesCopy = { ...selectedDownstreamPatches };
+    selectedBuildVariants.forEach((v) => {
```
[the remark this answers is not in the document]
sophstad: Good question, this is [existing code](https://github.com/evergreen-ci/spruce/blob/32ce975c828926b398d9ba0cac1b287b2d6aaa5e/src/pages/configurePatch/configurePatchCore/ConfigureTasks.tsx#L49) that I just moved into the hook. I would imagine that the goal here is to iterate over an unchanging object and avoiding the messiness of making updates to the same object that's being looped. Since it's working fine I'm inclined to keep as-is!

arjunpatel100 (approved):

src/hooks/useConfigurePatch.ts
khelif96: bump.

khelif96 (approved):

PR #869, opened 2021-09-08
EVG-13852 Task History table 

khelif96:
[EVG-13852](https://jira.mongodb.org/browse/EVG-13852)

### Description 
Create task history table component.
This is an infinitely scrollable table. It relies on the same pagination principles as the other mainline commits pages.
I Utilize a reducer to manage state and handle keeping track of commits and processing them.
Processing commits involves classifying each commit into 1 of 2 types. 
1) a Normal commit
2) A folded up commit

And then separating commits and placing date dividers between them.

The list virtualization library relies on the row index to determine what it should render which is why we need to preprocess data as we ingest it. 

processing also handles deduplicating data in the scenarios where a commit is inadvertently passed in twice.

If a component looks out of place or has a "Wonky" color its most likely just a placeholder component.
### Screenshots

https://user-images.githubusercontent.com/4605522/132557316-0bf0d73c-ab46-426a-a286-624e8e2be719.mov

### Testing 
Wrote unit tests for state management functions,
Tested on prod db

The best place to test this is by connecting to the prod backend and testing there. Since that is one of the only places with enough data to test this feature


src/components/HistoryTable/Row.tsx
```diff
@@ -0,0 +1,103 @@
...
+`;
+
+const Cell = styled.div`
+  display: flex;
+  height: 100%;
+  width: 140px;
+  justify-content: center;
+  align-items: center;
+  :hover {
+    background-color: red;
```
sophstad, line 83: This color looks pretty harsh/off-brand to me, would any of the `@leafygreen-ui/palette` reds work instead?

src/components/HistoryTable/Row.tsx
```diff
@@ -0,0 +1,103 @@
...
+    cursor: pointer;
+  }
+`;
+
+const DateSeperator = styled.div`
+  width: 100%;
+  padding-right: 40px;
+  background-color: lightblue;
```
sophstad, line 91: Same as above

src/components/HistoryTable/Row.tsx
```diff
@@ -0,0 +1,103 @@
...
+  display: flex;
+  align-items: center;
+`;
+
+const FoldedCommitContainer = styled.div`
+  width: 100%;
+  padding-right: 40px;
+  background-color: yellow;
```
sophstad, line 99: Same as above

src/components/HistoryTable/HistoryTable.stories.tsx
```diff
@@ -0,0 +1,25 @@
+import { HistoryTableProvider } from "./HistoryTableContext";
+import HistoryTable from "./index";
+import { mainlineCommitData } from "./testData";
+
+export default {
+  title: "History Table",
+};
+
+export const Default = () => (
```
sophstad, line 9: Could we consider moving the History Table Icon stories into this file so that they're nested together?

src/components/HistoryTable/HistoryTableContext.tsx
```diff
@@ -0,0 +1,92 @@
...
+  });
+
+  const itemHeight = (index) => {
+    // TODO: Fix bug causing itemHeight to be 0 on intial render
+    // There is a race condition where it tries to calculate element heights before the element content is loaded in
+    if (processedCommits[index]) {
+      switch (processedCommits[index].type) {
+        case rowType.COMMIT:
+          return 120;
+        case rowType.DATE_SEPERATOR:
```
sophstad, line 43:
```suggestion
        case rowType.DATE_SEPARATOR:
```

src/components/HistoryTable/Row.tsx
```diff
@@ -0,0 +1,103 @@
...
+  width: 140px;
+  justify-content: center;
+  align-items: center;
+  :hover {
+    background-color: red;
+    cursor: pointer;
+  }
+`;
+
+const DateSeperator = styled.div`
```
sophstad, line 88:
```suggestion
const DateSeparator = styled.div`
```

src/components/HistoryTable/HistoryTableContext.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+const HistoryTableProvider: React.FC = ({ children }) => {
+  const [
+    { processedCommits, processedCommitCount, visibleColumns },
+    dispatch,
+  ] = useReducer(reducer, {
+    loadedCommits: [],
+    processedCommits: [],
+    processedCommitCount: 0,
+    commitCache: {},
```
sophstad, line 28: This is probably a good use case for a map! The [MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#objects_vs._maps) suggest maps could yield some performance gains:

> Performs better in scenarios involving frequent additions and removals of key-value pairs.

This would perhaps also eliminate the need for the `cacheSize` field because of its `size` property.

src/components/HistoryTable/index.tsx
```diff
@@ -0,0 +1,52 @@
+import { useEffect } from "react";
+import { VariableSizeList as List } from "react-window";
+import InfiniteLoader from "react-window-infinite-loader";
+import { MainlineCommitsForHistoryQuery } from "gql/generated/types";
+import { useHistoryTable } from "./HistoryTableContext";
+import Row from "./Row";
+
+const itemStatusMap = {};
```
sophstad, line 8: AFAICT this object is never updated? But maybe I'm missing something :)

sophstad (changes requested):

src/components/HistoryTable/Row.tsx
khelif96: Sorry i should have clarified these colors were just placeholders used for testing and will eventually be replaced by other components.

src/components/HistoryTable/Row.tsx
khelif96: Same comment as above.

src/components/HistoryTable/Row.tsx
khelif96: Same comment as above

src/components/HistoryTable/Row.tsx
sophstad: Ah makes sense 👍 thanks!

src/components/HistoryTable/historyTableContextReducer.ts
```diff
@@ -0,0 +1,122 @@
...
+        );
+        return {
+          ...state,
+          commitCache: updatedObjectCache,
+          cacheSize: updatedCache.length,
+          processedCommits,
+          processedCommitCount: processedCommits.length,
+        };
+      }
+      return {
```
arjunpatel100, line 50: we can improve performance by not creating a new reference for state when there are no changes to it. i wonder if this can fix the useEffect involving fetchNewCommit [here](https://github.com/evergreen-ci/spruce/pull/869/files#diff-9b06585f88cd125563a91d3134e8296f02b8e553b1873edf086247fcbbcddb41R23-R28) so we can pass in all deps

src/components/HistoryTable/historyTableContextReducer.ts
```diff
@@ -0,0 +1,122 @@
...
+        ...state,
+        columns: action.columns,
+        visibleColumns: action.columns.slice(0, 8),
+        currentPage: 0,
+      };
+    case "nextPageColumns": {
+      const pageCount = Math.ceil(state.columns.length / 8);
+      if (pageCount <= state.currentPage + 1) {
+        return {
```
arjunpatel100, line 64: same comment as above

src/components/HistoryTable/historyTableContextReducer.ts
```diff
@@ -0,0 +1,122 @@
...
+      );
+      return {
+        ...state,
+        currentPage: state.currentPage + 1,
+        visibleColumns: nextPageColumns,
+      };
+    }
+    case "prevPageColumns": {
+      if (state.currentPage === 0) {
+        return {
```
arjunpatel100, line 81: same comment as above

arjunpatel100 (changes requested):

src/components/HistoryTable/HistoryTableContext.tsx
```diff
@@ -0,0 +1,92 @@
...
+      isItemLoaded,
+      getItem,
+      processedCommitCount,
+      processedCommits,
+      visibleColumns,
+      addColumns: (columns) => dispatch({ type: "addColumns", columns }),
+      nextPage: () => dispatch({ type: "nextPageColumns" }),
+      previousPage: () => dispatch({ type: "prevPageColumns" }),
+    }),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
khelif96: dispatch doesn't need to be included / shouldn't be included in the useMemo since it should remain stable through out different executions. I only included these 3 fields because those are the only portions of the state that can change and would impact the data present in the table.

arjunpatel100 (approved):

package.json
```diff
@@ -107,7 +107,9 @@
     "react": "17.0.1",
     "react-dom": "17.0.1",
     "react-router-dom": "5.2.0",
-    "react-window": "1.8.6"
+    "react-virtualized-auto-sizer": "^1.0.6",
```
arjunpatel100, line 110: some of new deps have a caret in their version number

src/components/HistoryTable/Row.tsx
sophstad: Bump

sophstad (approved):

PR #877, opened 2021-09-14
EVG-14404: Update LeafyGreen packages

sophstad:
EVG-14404

### Description 
- Update all LG packages so that we are consistently using Emotion 11

### Testing 
- Manual testing, updated snapshots


src/pages/spawn/spawnHost/SpawnHostTableActions.tsx
```diff
@@ -59,8 +59,8 @@ export const CopySSHCommandButton: React.FC<{
             size={Size.XSmall}
             data-cy="copy-ssh-button"
           >
-            <Label>Copy SSH command</Label>
-          </PaddedButton>
+            <Label>Copy SSH Command</Label>
```
khelif96, line 62: The Button is so large compared to the other ones. Can we check with design to see if this is ok.
image

khelif96 (approved):
Looks Great thanks for all the work!
https://user-images.githubusercontent.com/4605522/133487524-d973daa5-9564-4090-ae51-19b0d1eae3de.mp4

Looks like the button width changes depending on the disabled state. Its most likely a bug with the LG button itself. We should bring it up with them.

src/pages/spawn/spawnHost/SpawnHostTableActions.tsx
sophstad: Agreed, maybe we could change the copy to something shorter. I'll add it to tomorrow's agenda!

PR #878, opened 2021-09-15
EVG-6677: Link to GitHub and Jira from Patch page title

arjunpatel100:
[EVG-6677](https://jira.mongodb.org/browse/EVG-6677)

### Screenshots
Screen Shot 2021-09-15 at 6 23 34 PM


src/utils/string/index.ts
```diff
@@ -1,6 +1,8 @@
 import { format, utcToZonedTime } from "date-fns-tz";
 import get from "lodash/get";
 
+export { githubPRLinkify } from "./githubPRLinkify";
```
sophstad, line 4: [nit] We could export `jiraLinkify` here too for consistency

src/utils/string/githubPRLinkify.tsx
```diff
@@ -0,0 +1,11 @@
+import ReactDOMServer from "react-dom/server";
+import { StyledLink } from "components/styles/StyledLink";
+
+export const githubPRLinkify = (unlinkified: string) =>
+  unlinkified.replace(
+    /https:\/\/github.com\/[a-zA-Z0-9-]+\/[a-zA-Z0-9-]+\/pull\/\d+/gi,
+    (match) =>
+      ReactDOMServer.renderToString(
+        <StyledLink href={match}>{match}</StyledLink>
```
sophstad, line 9: It seems a little redundant to include the entire URL both written out and as a link destination. Something like this could include the same information and be significantly less noisy:

image

I wonder if at some point we should look into making app changes to update the `message`. Another option within Spruce could be to replace the `match` text here with some other text, like "(permalink)" or "(GitHub)" or something.

However, I think this is a great improvement for now, no need to to block merging!

sophstad (approved):

src/components/Banners/SiteBanner.tsx
```diff
@@ -4,9 +4,9 @@ import Icon from "@leafygreen-ui/icon";
 import IconButton from "@leafygreen-ui/icon-button";
 import parse from "html-react-parser";
 import Cookies from "js-cookie";
-import { getJiraTicketUrl } from "constants/externalResources";
 import { GetSpruceConfigQuery } from "gql/generated/types";
 import { GET_SPRUCE_CONFIG } from "gql/queries";
+import { jiraLinkify } from "utils/string/jiraLinkify";
```
khelif96, line 9: nit. for consistency can we use 
```
import { string } from "utils";

const { jiraLinkify} = "string";
```

src/utils/string/githubPRLinkify.tsx
```diff
@@ -0,0 +1,11 @@
+import ReactDOMServer from "react-dom/server";
+import { StyledLink } from "components/styles/StyledLink";
+
+export const githubPRLinkify = (unlinkified: string) =>
+  unlinkified.replace(
+    /https:\/\/github.com\/[a-zA-Z0-9-]+\/[a-zA-Z0-9-]+\/pull\/\d+/gi,
+    (match) =>
+      ReactDOMServer.renderToString(
```
khelif96, line 8: This seems redundant. Why would we render this to a string and then convert it back to html here https://github.com/evergreen-ci/spruce/pull/878/files#diff-6ac1cd60dde9a2d56f1196499e04205d1e42f8ac7c96c7992b8b5763f814a700R138

khelif96: We should just return the jsx element

src/utils/string/githubPRLinkify.tsx
khelif96: Great Idea! We should bring this up in the design sync tho since this would mean changing the commit message contents. We would also need to keep this uniform in other places we show commit messages such as the waterfall and history pages to avoid confusion. 

But i agree this might be out of scope so no need to block merging on it.

src/utils/string/jiraLinkify.tsx
```diff
@@ -0,0 +1,10 @@
+import ReactDOMServer from "react-dom/server";
+import { StyledLink } from "components/styles/StyledLink";
+import { getJiraTicketUrl } from "constants/externalResources";
+
+export const jiraLinkify = (unlinkified: string, jiraHost: string) =>
+  unlinkified.replace(/[A-Z]{1,10}-\d{1,6}/gi, (match) =>
+    ReactDOMServer.renderToString(
```
khelif96, line 7: Same comment about renderToString

khelif96 (changes requested):

src/utils/string/index.ts
arjunpatel100: jiraLinkify actually can't be exported from index without running into a dep cycle. jiraLinkify imports externalResources and the contents of string and externalResources are exported from utils

src/utils/string/githubPRLinkify.tsx
arjunpatel100: I think this may be the best we can do with the `.replace` solution since the callback can only return a string. I utilized the renderToString in order to reuse the <StyledLink /> component. 
The alternate solution would be to step through the unlinkified string based on start/end indices for matched items and append JSX elements to an array to render. I prefer the current solution for simplicity.

src/components/Banners/SiteBanner.tsx
arjunpatel100: jiraLinkify actually can't be exported from index without running into a dep cycle. jiraLinkify imports externalResources and the contents of string and externalResources are exported from utils

khelif96 (changes requested):
I think we should explore a safer approach for doing this. I was able to inject arbitrary html into the page by just creating a patch with html elements in the description
image
image

Here are some examples. (Use prod backend)
http://localhost:3000/version/6144c03d2fbabe28a1d25fde/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC
http://localhost:3000/version/6144b9f857e85a45025931b0/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

cypress/integration/version/version_route.ts
```diff
@@ -86,4 +87,26 @@ describe("Version route", () => {
...
+
+  describe("Page title", () => {
+    before(() => {
+      cy.preserveCookies();
+      cy.visit(versionRoute(versions[6]));
+    });
+    it("Should include a link to Jira", () => {
+      cy.dataCy("page-title")
+        .contains("a", "EVG-7425")
+        .should("have.attr", "href", "https:///browse/EVG-7425");
```
khelif96, line 99: nit. Maybe Should we update the backend to provide the jira base url so we can have a semi usable link in testing.
But not necessary for this ticket.

src/utils/string/githubPRLinkify.tsx
```diff
@@ -0,0 +1,13 @@
+import reactStringReplace from "react-string-replace";
+import { StyledLink } from "components/styles/StyledLink";
```
khelif96, line 2:
```suggestion
import { StyledLink } from "components/styles";
```

src/utils/string/jiraLinkify.tsx
```diff
@@ -0,0 +1,13 @@
+import reactStringReplace from "react-string-replace";
+import { StyledLink } from "components/styles/StyledLink";
```
khelif96, line 2:
```suggestion
import { StyledLink } from "components/styles";
```

src/components/Banners/SiteBanner.tsx
khelif96: Woah, This is the first time i've seen a dependency cycle in javascript.

khelif96 (approved):
LGTM Thanks for making these changes. Just had a few nits.

PR #881, opened 2021-09-16
EVG-14333 Save last used notification type and subscription method

khelif96:
[EVG-14333](https://jira.mongodb.org/browse/EVG-143333)

### Description 
Save last used notification type and subscription method in a cookie and retrieve them next time the modal is opened


src/hooks/useNotificationModal.ts
```diff
@@ -31,17 +34,20 @@ export const useNotificationModal = ({
...
   // target represents the input value for a subscription method
   const [target, setTarget] = useState<Target>({});
   const [isFormValid, setIsFormValid] = useState(false);
   const [extraFieldErrorMessages, setExtraFieldErrorMessages] = useState<
     string[]
   >([]);
 
-  const [selectedTriggerIndex, setSelectedTriggerIndex] = useState<number>();
+  const [selectedTriggerIndex, setSelectedTriggerIndex] = useState<number>(() =>
+    parseInt(Cookies.get(`${type}-notification-trigger`), 10)
```
sophstad, line 49: I'm seeing `NaN` on initial load of this modal, I think since there's no cookie found and then we try and run `parseInt()` on it.

image

src/hooks/tests/useNotificationModal.test.tsx
```diff
@@ -120,6 +121,7 @@ it("Should have correctly formatted request payload after selecting options, add
         triggers: patchTriggers,
         subscriptionMethodControls: patchSubscriptionMethodControls,
         resourceId: "a patch id",
+        type: "patch",
```
sophstad, line 124:
```suggestion
        type: "version",
```

src/hooks/tests/useNotificationModal.test.tsx
```diff
@@ -263,6 +265,7 @@ it("Should auto populate email and slack fields if they exist", async () => {
...
```
sophstad, line 268:
```suggestion
        type: "version",
```

sophstad (changes requested):

src/hooks/useNotificationModal.ts
khelif96: Thanks for catching i added a default case

src/hooks/useNotificationModal.ts
```diff
@@ -220,13 +227,22 @@ export const useNotificationModal = ({
...
     regexSelectorPropsTemplate.length >= (regexSelectors?.length ?? 0);
 
+  const onChangeSubscriptionMethod = (v: string) => {
+    setSelectedSubscriptionMethod(v);
+    Cookies.set("subscription-method", v, { expires: 365 });
+  };
+
+  const onChangeTrigger = (v: number) => {
+    setSelectedTriggerIndex(v);
+    Cookies.set(`${type}-notification-trigger`, `${v}`, { expires: 365 });
```
sophstad, line 244: Just a thought that is maybe out of scope — setting a cookie based on the trigger index could mess up the cookies if we ever make any updates to the dropdown by adding or reordering (which should theoretically be a low-risk update). There are also lots of duplicate key errors on this dropdown because the key is based on the `payloadResourceIdKey`, which is a duplicate value for many entries [here](https://github.com/evergreen-ci/spruce/blob/main/src/components/PatchActionButtons/addNotification/PatchNotificationModal.tsx#L91). It would be nice to use the [`trigger`](https://github.com/evergreen-ci/spruce/blob/bc200b7a93d9e311e0e24603b718c8da02e4b821/src/hooks/useNotificationModal.ts#L279) field instead for the key and value for each Option, and then use that value to store the cookie here.

But that is a bigger refactor, and I think this would work for the foreseeable future! We could make a tech debt ticket instead.

khelif96: Yeah def I actually tried to take a crack at refactoring the state management in this hook to support this originally. It ended up becoming a larger undertaking than i wanted to pursue. I opened a ticket to come back to this in the future. https://jira.mongodb.org/browse/EVG-15458 
i wasn't aware about the duplicate key errors. I will fix those though.

src/components/NotificationModal/index.tsx
```diff
@@ -136,13 +136,13 @@ export const NotificationModal: React.FC<NotificationModalProps> = ({
           onChange={(v: number) => {
             setSelectedTriggerIndex(v);
           }}
-          data-test-id="when-select"
+          data-cy="when-select"
         >
           {triggers.map((t, i) => (
             <Option
-              key={`trigger_${t.payloadResourceIdKey}`}
+              key={`trigger_${t.resourceType}`}
```
sophstad, line 143: `resourceType` is also repeated a lot [here](https://github.com/evergreen-ci/spruce/blob/main/src/components/PatchActionButtons/addNotification/PatchNotificationModal.tsx#L95) so I'm still seeing key errors. I think `trigger` would be the appropriate field to use?

sophstad (changes requested):

src/components/NotificationModal/index.tsx
khelif96: Thanks! I realized none of these keys individually are unique enough to be used so i created a key that combines a few of them.

sophstad (commented):
I just noticed that we're not caching any of the secondary options that appear. For example, if I select "the runtime for this version exceeds some duration" and then change the timeout to 90 seconds, the modal resets to 10 seconds on the next use.
image
I feel like it's not super useful to save part of the configuration but not the whole thing; it may even be more misleading if a user doesn't notice that the second value wasn't preserved. What do you think would be a good approach here? Should we just make a ticket for more followup work?

khelif96:
> I just noticed that we're not caching any of the secondary options that appear. For example, if I select "the runtime for this version exceeds some duration" and then change the timeout to 90 seconds, the modal resets to 10 seconds on the next use.
> image
> I feel like it's not super useful to save part of the configuration but not the whole thing; it may even be more misleading if a user doesn't notice that the second value wasn't preserved. What do you think would be a good approach here? Should we just make a ticket for more followup work?

Great point. tbh i can't really see a clear cut way to cache those values, with the way the state management currently is. I think it would be allot easier after we clean up some of this state management in https://jira.mongodb.org/browse/EVG-15458

sophstad (approved):
LGTM, let's just make sure to ticket the secondary caching work and link it to EVG-15458. Thanks for all the iterating on this!

khelif96:
Done https://jira.mongodb.org/browse/EVG-15486 @sophstad 

> LGTM, let's just make sure to ticket the secondary caching work and link it to [EVG-15458](https://jira.mongodb.org/browse/EVG-15458). Thanks for all the iterating on this!

PR #886, opened 2021-09-20
EVG-15347: Link task and grouped icons

arjunpatel100:
[EVG-15347](https://jira.mongodb.org/browse/EVG-15347)

### Description 
Links tasks status icons in the water and history table as well as the grouped status icons in the waterfall page. 
Also updates task status categories and their colors.


src/pages/commits/ActiveCommits/BuildVariantCard.tsx
```diff
@@ -67,34 +89,52 @@ const RenderGroupedIcons = ({ tasks }) => {
...
-            count={count}
-          />
-        </GroupedTaskStatusBadgeWrapper>
-      ))}
+      {otherTasks.map(({ count, umbrellaStatus }) => {
+        const onClick = () =>
+          push(
+            getVersionRoute(versionId, {
+              statuses: mapUmbrellaStatusToQueryParam[umbrellaStatus],
+              variant: buildVariantDisplayName,
```
khelif96, line 97: This generated a broken link http://localhost:3000/version/mongodb_mongo_master_4a43e817011e7cde84c2dce87619d81921f19c9a/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC&statuses=running-umbrella,started,dispatched&variant=%2A%20Windows%20DEBUG 
Which did not return a usable page since the variant did not match the input. 
image

We should use a strict regex and match against the variant name instead of the display name since the variant name is guaranteed to be unique. 
Something similar to what we do here https://github.com/evergreen-ci/spruce/blob/ae1f4d36bb82735479366507f625761386dd46e2/src/pages/version/buildVariants/GroupedTaskSquare.tsx#L28-L32

src/pages/commits/ActiveCommits/BuildVariantCard.tsx
```diff
@@ -67,34 +89,52 @@ const RenderGroupedIcons = ({ tasks }) => {
...
+            })
+          );
+        return (
+          <GroupedTaskStatusBadgeWrapper>
+            <GroupedTaskStatusBadge
+              status={umbrellaStatus}
+              key={`${umbrellaStatus}_groupedBadge`}
```
khelif96, line 104: We should also include versionId and buildVariant in this key to ensure uniqueness. Since there is a chance there can be multiple badges with the same umbrella status

src/pages/commits/ActiveCommits/BuildVariantCard.tsx
```diff
@@ -67,34 +89,52 @@ const RenderGroupedIcons = ({ tasks }) => {
...
+  return (
+    <>
+      {tasks.map(({ id, status }) => (
+        <IconButton
+          key={`task_${id}`}
+          aria-label="task icon"
+          onClick={() => push(getTaskRoute(id))}
+        >
+          <TaskStatusIcon status={status} size={16} />
+        </IconButton>
```
khelif96, line 133: We should wrap this with a Router link here instead of pushing to the history. 
Using a link allows us to be more declarative with our links and makes the links more accessible. Since they will be a part of the dom as opposed to something that programmatically changes the browser history. 
It also allows for easy usability use cases like being able to right click on a task icon and copy its link.

src/pages/commits/ActiveCommits/BuildVariantCard.tsx
```diff
@@ -67,34 +89,52 @@ const RenderGroupedIcons = ({ tasks }) => {
...
+              variant: buildVariantDisplayName,
+            })
+          );
+        return (
+          <GroupedTaskStatusBadgeWrapper>
+            <GroupedTaskStatusBadge
+              status={umbrellaStatus}
+              key={`${umbrellaStatus}_groupedBadge`}
+              count={count}
+              onClick={onClick}
```
khelif96, line 106: Same comment as above about wrapping this in a React router link instead of manipulating the browser history.

khelif96 (changes requested):

src/pages/commits/ActiveCommits/BuildVariantCard.tsx
```diff
@@ -2,24 +2,32 @@ import styled from "@emotion/styled";
...
 import { TaskStatusIcon } from "components/TaskStatusIcon";
+import { getVersionRoute, getTaskRoute } from "constants/routes";
+import { mapUmbrellaStatusToQueryParam } from "constants/task";
 import { groupStatusesByColor, isFailedTaskStatus } from "utils/statuses";
+import { applyStrictRegex } from "utils/string";
 
 const { gray } = uiColors;
 
 interface Props {
+  buildVariantId: string;
```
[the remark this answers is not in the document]
khelif96: This isn't a unique id its a name for build variant.

src/pages/commits/ActiveCommits/BuildVariantCard.tsx
```diff
@@ -50,13 +62,25 @@ export const BuildVariantCard: React.FC<Props> = ({
   }
   return (
     <Container>
-      <Label key={buildVariantDisplayName}>{buildVariantDisplayName}</Label>
+      <Label key={buildVariantId}>{buildVariantDisplayName}</Label>
```
khelif96, line 65: This also isn't unique since multiple versions will have the same the buildVariant.  
```suggestion
      <Label key={`${buildVariant}_${versionId}`}>{buildVariantDisplayName}</Label>
```

src/pages/commits/ActiveCommits/BuildVariantCard.tsx
```diff
@@ -50,13 +62,25 @@ export const BuildVariantCard: React.FC<Props> = ({
...
 };
 
-const RenderGroupedIcons = ({ tasks }) => {
+interface RenderGroupedIconsProps {
+  tasks: {
+    id: string;
+    status: string;
+  }[];
+  versionId: string;
+  buildVariantId: string;
```
khelif96, line 77: same comment about this fields name.

src/pages/commits/ActiveCommits/BuildVariantCard.tsx
```diff
@@ -68,30 +92,40 @@ const RenderGroupedIcons = ({ tasks }) => {
...
+        <GroupedTaskStatusBadgeWrapper data-cy="grouped-task-status-badge">
+          <Link
+            to={getVersionRoute(versionId, {
+              statuses: mapUmbrellaStatusToQueryParam[umbrellaStatus],
+              variant: applyStrictRegex(buildVariantId),
+            })}
+          >
+            <GroupedTaskStatusBadge
+              status={umbrellaStatus}
+              key={`${umbrellaStatus}_${versionId}_${buildVariantId}_groupedBadge`}
```
khelif96, line 104: Nice

src/utils/string/index.ts
```diff
@@ -155,3 +155,5 @@ export const sortFunctionDate = (a, b, key) => {
   }
   return dateA - dateB;
 };
+
+export const applyStrictRegex = (str: string) => `^${str}$`;
```
khelif96, line 159: Great idea bringing it out!

src/utils/string/string.test.ts
```diff
@@ -260,3 +261,9 @@ describe("getDateCopy", () => {
     );
   });
 });
+
+describe("applyStrictRegex", () => {
+  test("converts string to strict regex", () => {
+    expect(applyStrictRegex("dog")).toEqual("^dog$");
+  });
```
khelif96, line 268: We should also add some test cases to check if this regex works.
```suggestion
  test("converts string to strict regex expression", () => {
    expect(applyStrictRegex("dog")).toEqual("^dog$");
  });
  test("returned regex should have a strict match", () => {
  // run the regex against a few test cases and make sure it only matches on values we expect
  })
```

src/constants/task.ts
```diff
@@ -111,23 +129,49 @@ export const taskStatusesFilterTreeData: TreeDataEntry[] = [
...
+    ...accum,
+    ...(children
+      ? children.reduce(
+          (cAccum, child) => ({ ...cAccum, [child.value]: value }),
+          {}
+        )
+      : { [value]: value }),
+  }),
+  {}
+);
```
khelif96, line 174: This is pretty difficult to read and understand. The mix of the nested reduce's and ternaries is difficult to understand. Could we make this a little more verbose to make it easier to understand whats going on.

khelif96 (changes requested):

src/pages/commits/ActiveCommits/BuildVariantCard.tsx
SupaJoon: We could actually omit this key since it's not being included in an array

src/pages/commits/ActiveCommits/BuildVariantCard.tsx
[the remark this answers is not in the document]
SupaJoon: Sure thing! I named this buildVariantId to distinguish the two fields in GroupedBuildVariant: https://github.com/evergreen-ci/evergreen/blob/main/graphql/schema.graphql#L497-L499
I can rename buildVariantId to variant to resemble our schema more closely.

src/pages/commits/ActiveCommits/BuildVariantCard.tsx
SupaJoon: Since `versionId` and `buildVariantId` are fixed for all elements of this array and [keys must only be unique amongst siblings](https://reactjs.org/docs/lists-and-keys.html#keys-must-only-be-unique-among-siblings), we could actually omit `versionId` and `buildVariantId` since key stability will have to be guaranteed by `umbrellaStatus` either way.

src/utils/string/string.test.ts
SupaJoon: Sure, I can include some tests that the regex work in JS but I don't think it will be exhaustive since the regular expression is being processed in our mongodb query and this test will utilize a js lib util instead. 
I will include a comment that a precondition for the param is that it should not contain any regex operators. This precondition should be satisfied by the [variant field in GroupedBuildVariant](https://github.com/evergreen-ci/evergreen/blob/main/graphql/schema.graphql#L497)

src/constants/task.ts
SupaJoon: Sure, I just added some comments

cypress/integration/commits/task_status_icons.ts
```diff
@@ -0,0 +1,36 @@
...
+      "/task/evergreen_ubuntu1604_test_cloud_patch_5e823e1f28baeaa22ae00823d83e03082cd148ab_5e4ff3abe3c3317e352062e4_20_02_21_15_13_48",
+      "/task/evergreen_ubuntu1604_test_service_patch_5e823e1f28baeaa22ae00823d83e03082cd148ab_5e4ff3abe3c3317e352062e4_20_02_21_15_13_48",
+    ];
+    cy.dataCy("task-status-icon").each(($el, index) => {
+      cy.wrap($el).should("have.attr", "href").and("equals", hrefs[index]);
+    });
+  });
+
+  const route =
+    "/commits/evergreen?chartType=soeiantsrein&statuses=all,failed-umbrella,failed,task-timed-out,test-timed-out,known-issue,success,running-umbrella,started,dispatched,scheduled-umbrella,will-run,pending,unstarted,system-failure-umbrella,system-failed,system-timed-out,system-unresponsive,setup-failed,inactive";
```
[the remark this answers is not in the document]
SupaJoon: I must have copied the query param from this test: https://github.com/evergreen-ci/spruce/blob/main/cypress/integration/commits/chart_toggle.ts#L42

src/constants/task.ts
```diff
@@ -271,3 +279,57 @@ export const mapBadgeColors = {
     text: gray.light3,
   },
 };
+
+export const mapUmbrellaStatusColors = {
```
SupaJoon, line 283: This object will represent all task status color mappings. We still need to confirm with design if we need different colors for the waterfall barchart.

src/utils/statuses/groupStatusesByUmbrellaStatus.ts
```diff
@@ -0,0 +1,51 @@
...
+    if (counts[umbrellaStatus]) {
+      counts[umbrellaStatus].count += stat.count;
+      if (!counts[umbrellaStatus].statuses.includes(stat.status)) {
+        counts[umbrellaStatus].statuses.push(taskStatusToCopy[stat.status]);
+      }
+    } else {
+      counts[umbrellaStatus] = {
+        count: stat.count,
+        statuses: [taskStatusToCopy[stat.status]],
+        color: mapUmbrellaStatusColors[umbrellaStatus].border,
```
SupaJoon, line 31: This color resembles commit chart tooltip icon fill and barchart fill. We may need to expand a different value from mapUmbrellaStatusColors in here depending on design feedback.

src/pages/commits/ActiveCommits/CommitChartTooltip.tsx
```diff
@@ -32,19 +32,24 @@ export const CommitChartTooltip: React.FC<Props> = ({
...
         ))}
-        {zeroCountStatus.map((status) => (
+        {zeroCountStatus.map((umbrellaStatus) => (
           <TotalCount
-            key={status}
-            status={status}
-            color={mapTaskStatusToColor[status]}
+            key={umbrellaStatus}
+            status={umbrellaStatus}
+            color={mapUmbrellaStatusColors[umbrellaStatus].border}
```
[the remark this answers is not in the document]
SupaJoon: Nice catch. The color field actually maps to `barChart` instead of `border` https://github.com/evergreen-ci/spruce/pull/886/files#diff-fb5d09bb6b48aa77ae57c5ce3398f59dc7200b1c8b1ad3c2ede6161096198e28R31

We won't need to access `mapUmbrellaStatusColors` here once the taskStatusCounts resolver includes zero counts ([ticket](https://jira.mongodb.org/browse/EVG-15459))

khelif96 (approved):

PR #890, opened 2021-09-24
EVG-14405 Update project select component to use searchable dropdown component

khelif96:
[EVG-14405](https://jira.mongodb.org/browse/EVG-14405)

### Description 
This updates the Project select component to reflect the latest designs as well as utilize the SearchableDropdown component
### Screenshots
image

### Testing 
Unit tests and locally and storybook


src/components/SearchableDropdown/index.tsx
```diff
@@ -73,13 +72,13 @@ const SearchableDropdown = <T extends {}>({
     }
   }, [options]);
 
-  const onClick = (v: string | T) => {
+  const onClick = (v: T) => {
     if (allowMultiselect) {
       if (Array.isArray(value)) {
-        const newValue = toggleArray(v, value) as any[];
+        const newValue = toggleArray(v, value) as T[];
```
SupaJoon, line 78:
```suggestion
        const newValue = toggleArray(v, value);
```

src/components/SearchableDropdown/index.tsx
```diff
@@ -73,13 +72,13 @@ const SearchableDropdown = <T extends {}>({
...
         onChange(newValue);
       } else {
-        onChange([v as any]);
+        onChange([v as T]);
```
SupaJoon, line 81:
```suggestion
        onChange([v]);
```

src/components/SearchableDropdown/index.tsx
```diff
@@ -12,22 +12,20 @@ const { gray, white, blue } = uiColors;
...
-  searchFunc?: (
-    value: string | T | string[] | T[],
-    match: string | T
-  ) => boolean;
+  value: T | T[];
+  onChange: (value: T | T[]) => void;
+  searchFunc?: (options: T[], match: string) => T[];
   searchPlaceholder?: string;
   valuePlaceholder?: string;
   options: string[] | Array<T>;
```
SupaJoon, line 20: If we omit string[], we may be able to clean up the if statement block on line 119

SupaJoon (approved):
I think we can simplify the props further if we change

src/components/SearchableDropdown/index.tsx
khelif96: `string[]` needs to be supplied in order to support the default filtering behavior of the component. Otherwise `searchFunc` would need to be required for this component to work. Since the default `filter` function is not defined on the `T[]` type, typescript will throw an error. So we need to explicitly specify that theres the possibility for `options` to be a `string[]` I added some error handling for that edge case.

PR #891, opened 2021-09-27
EVG-8326 Create button to override dependencies

khelif96:
[EVG-8326](https://jira.mongodb.org/browse/EVG-8326)

Dependent on https://github.com/evergreen-ci/evergreen/pull/5058

### DIFF
--- a/src/pages/task/ActionButtons.tsx
+++ b/src/pages/task/ActionButtons.tsx
@@ -21,15 +21,19 @@ import {
   ScheduleTaskMutationVariables,
   UnscheduleTaskMutation,
   UnscheduleTaskMutationVariables,
+  OverrideTaskDependenciesMutation,
+  OverrideTaskDependenciesMutationVariables,
 } from "gql/generated/types";
 import {
   ABORT_TASK,
+  OVERRIDE_TASK_DEPENDENCIES,
   RESTART_TASK,
   SCHEDULE_TASK,
   SET_TASK_PRIORTY,
   UNSCHEDULE_TASK,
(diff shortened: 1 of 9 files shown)


src/pages/task/ActionButtons.tsx
```diff
@@ -130,13 +136,31 @@ export const ActionButtons = ({
...
     loadingAbortTask ||
     loadingRestartTask ||
     loadingSetPriority ||
     loadingUnscheduleTask ||
-    loadingScheduleTask;
+    loadingScheduleTask ||
+    loadingOverrideTaskDependencies;
 
+  const canOverrideDependencies =
+    status === TaskStatus.Blocked || status === TaskStatus.WillRun;
```
SupaJoon, line 163: The condition to enable this button is slightly different in the legacy ui: https://github.com/evergreen-ci/evergreen/blob/main/service/templates/task.html#L145. Do you think we need to reutilize any part of the legacy condition here?

The other buttons' enabled state is fetched from evergreen to make it easier to leverage EVG constants https://github.com/evergreen-ci/evergreen/blob/main/graphql/schema.graphql#L757-L761.
It could be a good idea to create a canOverrideDependencies resolver in case we need compare a value against a permission constant like we do in the legacy UI.

SupaJoon (changes requested):

src/pages/task/ActionButtons.tsx
khelif96: Great Idea! Just implemented

SupaJoon (approved):

PR #893, opened 2021-09-28
EVG-15443: Render variant tasks for aliases

sophstad:
EVG-15443

### Description 
- Display variants/tasks that would be run by scheduling an alias. Use these tasks to update the sidebar's badge count.

### Screenshots
image
image

### Testing 
- Add integration test to ensure that the one task associated with the "logkeeper" alias renders correctly
- I set up staging so that you can view a patch with both a child patch and an alias [here](https://spruce-staging.corp.mongodb.com/patch/6155cb4997b1d3345c40835f/configure/tasks). Or run your own on staging:
  ```
  evergreen -c ~/.evergreen-staging.yml patch -p mci --trigger-alias e2e
  ```

### Evergreen PR 
https://github.com/evergreen-ci/evergreen/pull/5047
https://github.com/evergreen-ci/evergreen/pull/5072

### DIFF
--- a/src/hooks/useConfigurePatch.ts
+++ b/src/hooks/useConfigurePatch.ts
@@ -115,10 +115,11 @@ const tabToIndexMap = {
   [PatchTab.Parameters]: 2,
 };
 
-// Extract the type of a child patch from array of child patches in this query so we can add an alias field
+// Helper to extract type T from an array of type T
 type Unpacked<T> = T extends (infer U)[] ? U : T;
 
-export interface ChildPatchComplete
+// Extract the type of a child patch and append alias field
+export interface ChildPatchAliased
   extends Unpacked<ConfigurePatchQuery["patch"]["childPatches"]> {
   alias: string;
 }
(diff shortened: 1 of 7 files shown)


malikchaya2 (commented):
image
I haven't looked at the code yet, but looking at this screenshot, we probably want to use the project identifier when there is one instead of the project id so that the project name is more readable.

malikchaya2:
image
For this use case, where an alias was specified using the CLI, can we please add a note (maybe similar to [this](https://github.com/evergreen-ci/spruce/blob/main/src/components/PatchTabs/ParametersContent.tsx#L24), or whatever style you think makes sense ) saying that aliases that we specified in the CLI cannot be edited, so that users know that it's not a bug? (This can also be done in a separate ticket if you prefer that).

malikchaya2 (approved):

src/hooks/useConfigurePatch.ts
```diff
@@ -115,6 +115,15 @@ const tabToIndexMap = {
   [PatchTab.Parameters]: 2,
 };
 
+// Helper to extract type T from an array of type T
+type Unpacked<T> = T extends (infer U)[] ? U : T;
```
SupaJoon, line 119: Can we export this from a util file? There are plenty of places in our codebase where we should use this helper instead of taking the zero index of an array type

SupaJoon (approved):

PR #895, opened 2021-10-01
EVG-15515 Fix Downstream Tasks cutting off after the 25th task

malikchaya2:
[EVG-15515](https://jira.mongodb.org/browse/EVG-15515)

### Description 
The accordion had a max height of 1500px which cut off the downstream tasks after the 25th task. 

### Screenshots
Before:
image

After: 
image

### Testing 
Tested on staging

### DIFF
--- a/src/components/Accordion/index.tsx
+++ b/src/components/Accordion/index.tsx
@@ -57,7 +57,7 @@ const AnimatedAccordion = styled.div`
   max-height: 0;
   /* This is used to calculate a fixed height for the Accordion since height
      transitions require a fixed height for their end height */
-  max-height: ${(props: { hide: boolean }): string => !props.hide && "1500px"};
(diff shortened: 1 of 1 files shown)


src/components/Accordion/index.tsx
```diff
@@ -57,7 +57,7 @@ const AnimatedAccordion = styled.div`
   max-height: 0;
   /* This is used to calculate a fixed height for the Accordion since height
      transitions require a fixed height for their end height */
-  max-height: ${(props: { hide: boolean }): string => !props.hide && "1500px"};
+  max-height: ${(props: { hide: boolean }): string => !props.hide && "6000px"};
```
khelif96, line 60: This seems more like a bandaid fix over the real problem. Since theres a chance the table grows to the point where 6000px is not enough. Maybe we should explore paginating the table?

khelif96 (approved):

PR #907, opened 2021-10-05
EVG-15062: Cache project settings form changes

sophstad:
EVG-15062

### Description 
- Use context to hoist form state for project settings tabs, allowing us to preserve unsaved forms when navigating within the section and identify unsaved pages
- Reorganize: move tabs into `components/` directory

### Screenshots
This is a sample form I added for testing purposes, it will be removed soon as part of EVG-15076
image

### Testing 
- Add unit test for context


src/context/project-settings.tsx
```diff
@@ -0,0 +1,175 @@
...
+        [action.tab]: action.formData,
+      };
+    default:
+      return state;
+  }
+};
+
+const ProjectSettingsContext = createContext<ProjectSettingsState | null>(null);
+const SaveTabDispatchContext = createContext<SaveTabDispatchState | null>(null);
+const FormContext = createContext<FormContextState | null>(null);
```
khelif96, line 63: I think it would make sense to combine these 3 contexts into 1 context. That manages the state of all interactions on the page. 
1. Keeping track of initial data. 
2. Keeping track of changes in form data. 
3. And using that to manage the save state of each tab. 

The  state changes that are happening could be handled by one centralized state change instead of 
happening in different places and would reduce the amount of side effects that occur as the result of them. https://github.com/evergreen-ci/spruce/pull/907/files#diff-7baaa44e7e71b372734d844c07ce31e5ecf2f902baf3da6c38be5ec3205a5fa2R85-R89

A state transition could handle all the related side effects in one action. As opposed to making those in separate updates.

Something like
```
    case "updateForm": {
          if(state.originalFormData !== formData) {
             return {
                ...state,
                [action.tab]: action.formData, // represents the form data for the tab
                hasChanges: true
              };
          }else{
             return {
             ...state,
             hasChanges: false
             }
       }
   }
```
How do you feel about that?

khelif96 (commented):
Looks great! I had some thoughts about the state management.

src/context/project-settings.tsx
```diff
@@ -0,0 +1,175 @@
+import { createContext, useContext, useEffect, useReducer } from "react";
+import { FormProps } from "@rjsf/core";
+import { ProjectSettingsTabRoutes } from "constants/routes";
+
+interface ProjectSettingsState
+  extends Record<ProjectSettingsTabRoutes, boolean> {}
```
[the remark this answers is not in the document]
sophstad: Yeah exactly! Sorry I think this was one of the first things I named, I will change it to something more descriptive now :)

src/context/project-settings.tsx
```diff
@@ -0,0 +1,151 @@
...
+  if (context === undefined) {
+    throw new Error(
+      "useProjectSettingsContext must be used within a ProjectSettingsProvider"
+    );
+  }
+  return context;
+};
+
+const usePopulateForm = (
+  initialState: any,
```
SupaJoon, line 109: should this be type `FormProps`?

SupaJoon (approved):

src/context/project-settings.tsx
```diff
@@ -0,0 +1,151 @@
...
+  tab: ProjectSettingsTabRoutes
+): void => {
+  const { saveTab, updateForm } = useProjectSettingsContext();
+  const isSaved = useIsTabSaved(tab);
+
+  useEffect(() => {
+    // Ensure form does not have unsaved changes before writing.
+    // This preserves the unsaved form state when switching between project settings tabs.
+    if (isSaved) {
```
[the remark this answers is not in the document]
sophstad: Will address these concerns in EVG-15076 when we start making network calls!

src/context/project-settings.tsx
```diff
@@ -0,0 +1,151 @@
+import { createContext, useContext, useEffect, useReducer } from "react";
+import { FormProps as SpruceFormProps } from "@rjsf/core";
```
khelif96, line 2: nit. We can probably reexport this from SpruceForm by default. No preference tho.

khelif96 (approved):
Great Job on this!

sophstad:
Merged since base branch is failing

PR #913, opened 2021-10-07
EVG-13953: History Table Date Separator 

SupaJoon:
[EVG-13953](https://jira.mongodb.org/browse/EVG-13953)

Actual: 
Screen Shot 2021-10-07 at 1 26 32 PM

Design:
Screen Shot 2021-10-07 at 1 23 32 PM

### DIFF
--- a/src/components/HistoryTable/row/DateSeparator.tsx
+++ b/src/components/HistoryTable/row/DateSeparator.tsx
@@ -45,5 +45,5 @@ const DateWrapper = styled(Body)`
   white-space: nowrap;
   padding-right: 28px;
   text-transform: uppercase;
-  color: ${gray.dark2}; ;
+  color: ${gray.dark2};
 `;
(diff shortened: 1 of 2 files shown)


src/components/HistoryTable/row/DateSeparator.tsx
```diff
@@ -0,0 +1,50 @@
...
+interface DateSeparatorProps {
+  style: CSSProperties;
+  date: Date;
+}
+
+export const DateSeparator: React.FC<DateSeparatorProps> = ({
+  style,
+  date,
+}) => {
+  const tz = useUserTimeZone();
```
khelif96, line 18: Can you look at the network activity as we scroll. Im worried this  might get called allot as the user scrolls. Since react-window will continuously rerender this component.

khelif96: If it does we should either pass in a flag to use a cached value. Or pass this value from a higher level component that doesn't rerender often.

PR #915, opened 2021-10-12
EVG-15520: Linkify mainline commits pages

SupaJoon:
[EVG-15520](https://jira.mongodb.org/browse/EVG-15520)

### Description 
Linkifies version breadcrumb, waterfall build variant card, see history button, last passing version button and base commit link in patch metadata

### Testing 
Jest test for Last Passing Version button

### Evergreen PR 
Last Passing Version button relies on https://github.com/evergreen-ci/evergreen/pull/5090
Links to build variant history page from tasks table relies on https://github.com/evergreen-ci/evergreen/pull/5110


src/analytics/index.ts
```diff
@@ -1,9 +1,11 @@
 export { usePatchAnalytics } from "./patch/usePatchAnalytics";
 export { useTaskAnalytics } from "./task/useTaskAnalytics";
+export type { Analytics as TaskAnalytics } from "./task/useTaskAnalytics";
 export { useAnnotationAnalytics } from "./task/useAnnotationAnalytics";
 export { useUserPatchesAnalytics } from "./patches/useUserPatchesAnalytics";
 export { useNavbarAnalytics } from "./navbar/useNavbarAnalytics";
 export { useBreadcrumbAnalytics } from "./breadcrumb/useBreadcrumbAnalytics";
+export type { Analytics as BreadcrumbAnalytics } from "./breadcrumb/useBreadcrumbAnalytics";
```
sophstad, line 8: I didn't know about this syntax for named exports of types! Nice 😎

src/pages/task/ActionButtons.tsx
```diff
@@ -1,14 +1,16 @@
 import React, { useState } from "react";
 import { useMutation } from "@apollo/client";
 import styled from "@emotion/styled";
+import LeafyGreenButton from "@leafygreen-ui/button";
```
sophstad, line 4: I think we should be able to use the Button imported on line 8 since it's just a wrapper over `@leafygreen-ui/button` (maybe along with `@ts-expect-error`)

SupaJoon: our Button component doesn't accept a `to` or `as` prop. I created this ticket to simplify the component interface so it's easier to maintain. https://jira.mongodb.org/browse/EVG-15602. I made an attempt in c6b393f but it fails to apply the extra props, loading & dataCy.

src/constants/routes.ts
```diff
@@ -208,3 +208,11 @@ export const getCommitQueueRoute = (projectId: string) =>
 
 export const getCommitsRoute = (projectId: string) =>
   `${paths.commits}/${projectId}`;
+
+export const getVariantHistoryRoute = (
+  projectId: string,
```
ablack12, line 213: I think we should actually label this projectIdentifier since that's what we're actually passing through; in the future we may otherwise rely on this being ID when it's actually the mutable identifier.

ablack12 (commented):
LGTM; really I think we should make this naming change for all of these links but understood if that's out of scope

SupaJoon:
evergreen retry

SupaJoon:
I removed the Last Passing Version button. I will be included in this ticket with a different design: https://jira.mongodb.org/browse/EVG-13178

PR #917, opened 2021-10-13
EVG-15518 Support variant history pages

khelif96:
[EVG-15518](https://jira.mongodb.org/browse/EVG-15518)

### Description 
Refactor row component to make it reusable.
Add support for variant history 
Add empty state for tasks/variants that have no matches
### Screenshots
image
image

### Testing 
Connected to prod or staging backend.

### DIFF
--- a/src/pages/variantHistory/ColumnHeaders.tsx
+++ b/src/pages/variantHistory/ColumnHeaders.tsx
@@ -11,7 +11,7 @@ const ColumnHeaders: React.FC<ColumnHeadersProps> = ({ columns, loading }) => {
   const { visibleColumns, addColumns } = useHistoryTable();
   useEffect(() => {
     if (columns) {
-      addColumns(columns.map((c) => c));
+      addColumns(columns);
     }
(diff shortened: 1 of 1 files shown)


khelif96:
> This is awesome!!! I noticed this bug we could address in another ticket: Scrolling down with the scroll bar quickly shows the row loading state even when all network requests have succeeded: https://user-images.githubusercontent.com/10734386/137744617-148ab232-4dfc-49bd-9c3f-c7531dff0010.mov

Nice Catch! thats actually a symptom of us passing in a fixed itemCount for the infiniteLoader component https://github.com/evergreen-ci/spruce/blob/6e0138500580e264466f77c6721f156190652304/src/components/HistoryTable/HistoryTable.tsx#L41 

If you look at the network requests there aren't actually 10000 mainline commits in the spruce waterfall so it eventually hits a point where it expects there to be more commits but there aren't any. 

We should probably do some sort of stats query to populate this value with the actual mainline commit count for a project.

PR #930, opened 2021-10-20
Add navigation warning modal

sophstad:
EVG-15064

### Description 
- Add modal warning users when they are navigating out of project settings with unsaved changes
- Note: this warning doesn't currently work when navigating to the waterfall page (which is still legacy), as browsers implement security restrictions when navigating to a new site. Since mainline commits should be released before this work, it shouldn't be a problem when project settings launches. Clicking on a different navbar link will allow you to test this feature.

### Screenshots
image

### Testing 
- Update unit and integration tests


SupaJoon (approved):
The design shows the red button to say "Leave" instead of "Delete"
https://www.figma.com/file/dglsaa2OMnyyFeQQLUGXta/Project-Settings-Sections-(Copy)?node-id=70%3A4920

PR #931, opened 2021-10-21
EVG-14563 Apply default sort on Patch page when clearing filters 

minnakt:
[EVG-14563](https://jira.mongodb.org/browse/EVG-14563)

### Description 
This PR makes it so that when a user presses the "Clear All Filters" button, the default sorting is applied to the Patch page. (Previously, clicking the button would cause all sorting to be cleared.)

### Screenshots
https://user-images.githubusercontent.com/47064971/138209639-8c210678-7198-41f3-be44-ac370893800a.mov

### Testing 
Edited `task_filters.ts` test to pass with frontend changes.

### DIFF
--- a/src/components/PatchTabs/Tasks.tsx
+++ b/src/components/PatchTabs/Tasks.tsx
@@ -4,14 +4,15 @@ import styled from "@emotion/styled";
 import Button from "@leafygreen-ui/button";
 import { Skeleton } from "antd";
 import every from "lodash.every";
-import { useParams, useLocation, useHistory } from "react-router-dom";
+import { useParams, useLocation } from "react-router-dom";
+// import { useParams, useLocation, useHistory } from "react-router-dom";
 import { usePatchAnalytics } from "analytics";
 import { PageSizeSelector } from "components/PageSizeSelector";
 import { Pagination } from "components/Pagination";
 import { ResultCountLabel } from "components/ResultCountLabel";
 import { TableControlOuterRow, TableControlInnerRow } from "components/styles";
 import { pollInterval } from "constants/index";
-import { getVersionRoute } from "constants/routes";
+// import { getVersionRoute } from "constants/routes";
 import { useToastContext } from "context/toast";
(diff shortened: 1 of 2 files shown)


src/components/PatchTabs/Tasks.tsx
```diff
@@ -68,6 +66,18 @@ export const Tasks: React.FC<Props> = ({ taskCount }) => {
   useNetworkStatus(startPolling, stopPolling);
   const { patchTasks } = data || {};
 
+  const onClearAll = () => {
+    patchAnalytics.sendEvent({ name: "Clear all filter" });
+    updateQueryParams({
+      statuses: [],
+      baseStatuses: [],
```
khelif96, line 73: Can we also set these to undefined?

src/components/PatchTabs/Tasks.tsx
```diff
@@ -80,10 +90,7 @@ export const Tasks: React.FC<Props> = ({ taskCount }) => {
             denominator={taskCount}
           />
           <PaddedButton // @ts-expect-error
-            onClick={() => {
-              patchAnalytics.sendEvent({ name: "Clear all filter" });
-              router.push(getVersionRoute(versionId));
-            }}
+            onClick={() => onClearAll()}
```
khelif96, line 93:
```suggestion
            onClick={onClearAll}
```

src/components/PatchTabs/Tasks.tsx
```diff
@@ -68,6 +66,18 @@ export const Tasks: React.FC<Props> = ({ taskCount }) => {
...
 
+  const onClearAll = () => {
+    patchAnalytics.sendEvent({ name: "Clear all filter" });
+    updateQueryParams({
+      statuses: [],
+      baseStatuses: [],
+      taskName: undefined,
+      variant: undefined,
+      page: undefined,
+      sorts: "STATUS:ASC;BASE_STATUS:DESC",
```
khelif96, line 77: Can we extract this into a shared variable for L45 and here.

PR #943, opened 2021-11-01
EVG-13832: Improve spawn host page on narrow screens

sophstad:
EVG-13832

### Description 
- Fix modal and table layout on narrow screens

### Screenshots
image


SupaJoon (approved):
This looks great! We could improve the table by allowing the action buttons to stack. All it takes is a `flex-wrap: wrap;`

khelif96:
I wonder if we should instead increase the modal width on smaller screens and apply a minimum width instead of squishing the content of the modal to fit the smaller width? What do you think?

This is also using an antd modal underneath do you think we could port this to leafygreens modal instead Instead ? 

I also had a question for @SupaJoon  @sophstad  how do y'all feel about imposing a minimum supported width? I read a thread that cloud supports a minimum of  1024px i think we should probably lean to something like that since spruce wasn't originally designed to be responsive or mobile friendly.

sophstad:
> how do y'all feel about imposing a minimum supported width? I read a thread that cloud supports a minimum of 1024px i think we should probably lean to something like that since spruce wasn't originally designed to be responsive or mobile friendly.

Makes sense to me! This ticket linked to our design sync agenda [here](https://docs.google.com/document/d/1iVWEBz0Y1Np22sYye9k_mF9bJgTb6rnThuJcrQdnwGg/edit) in which it was decided to support a minimum of 800px, so I think we should stick with that going forward. I'll take another look at this ticket with that in mind!

sophstad:
> do you think we could port this to leafygreens modal instead Instead ?

It turns out antd's searchable select component does not play nice with LeafyGreen's confirmation modal (not clickable). This should be tackled once LG's new searchable select is in use. I just added a min width to the modal since we don't need to support very small screens.

PR #957, opened 2021-11-02
EVG-13626 Remove spawnhostbutton test

khelif96:
[EVG-13626](https://jira.mongodb.org/browse/EVG-13626)

### Description 
Spent a few hours trying to debug this test with little luck.  Something is causing the test runner to crash. Since this test has been suppressed for a few months i think its safe to get rid of it instead of investing more time trying to resolve the crashes. All it does anyways is just assert that a button is disabled based off the results of a query. 
### Screenshots
image


minnakt (approved):

minnakt:
I could get this to pass if I added `await` in front of `waitFor` (removed the Promise on line 309 too, and added `async` definition to function on line 274), but maybe that's the wrong way to do it... 🤔 Just wanted to leave a comment about it.

Screen Shot 2021-11-02 at 1 46 16 PM

PR #961, opened 2021-11-02
EVG-15457 Support periodic builds in spruce

khelif96:
[EVG-15457](https://jira.mongodb.org/browse/EVG-15457)

### Description and background
This adds support for periodic builds in spruce. 
Periodic builds are similar to mainline commits in that they do not have a corresponding patch. But their id's resemble those of a patch. 
We had a function called `validatePatchId()` on the UI that would take a given id and return a boolean indicating that it resembled an id typically associated with a patch. 
It was used to [determine](https://github.com/evergreen-ci/spruce/blob/c4a1226d7b96f55a05a58fea261551c97c96d1d0/src/pages/Version.tsx#L76-L82) if we should kick off a request to fetch a patch or if we should kick off a request for the version directly. (The reason we do that is because we need to determine if a [patch is configured](https://github.com/evergreen-ci/spruce/blob/c4a1226d7b96f55a05a58fea261551c97c96d1d0/src/pages/Version.tsx#L86-L100)  to decide if we should redirect the user)

The reason this didn't work is because a periodic build id resembles a patch id so the initial inquiry request fails with a 404 since there is no patch associated with a periodic build id. 

The remedy is to kick off the request for a version anyways just in case the id actually is a periodic build instead of just throwing an error when the initial request fails.
### Screenshots

image
You may notice that there is no version description/title that behavior is normal for periodic builds. 
### Testing 
Against a periodic build on prod
Heres one for you to test `613fdf03d1fe07352d5821c0`


.eslintrc.js
```diff
@@ -125,6 +125,7 @@ module.exports = {
       { some: ["nesting", "id"] },
     ],
     "no-console": OFF,
+    "no-debugger": errorIfStrict,
```
khelif96, line 128: Enabled this so we can use the debugger locally :)

src/pages/Version.tsx
```diff
@@ -97,7 +99,12 @@ export const VersionPage: React.FC = () => {
         getVersion();
       }
     }
-  }, [patchData, getVersion, id]);
+    // if there was an error fetching the patch and the patch id was a real id it could be a periodic patch build
```
ablack12, line 102: nit: periodic build, not periodic patch build

src/pages/Version.tsx
```diff
@@ -117,9 +124,11 @@ export const VersionPage: React.FC = () => {
   } = patch || {};
   const isPatchOnCommitQueue = commitQueuePosition !== null;
 
-  const title = isPatch
-    ? `Patch - ${patchNumber}`
-    : `Version - ${revision?.substr(0, 6)}`;
+  // If a revision exists
+  const versionText = revision?.length
+    ? revision?.substr(0, 7)
```
SupaJoon, line 129: we should use substring instead of substr: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring#the_difference_between_substring_and_substr

src/components/Breadcrumb/index.tsx
```diff
@@ -60,7 +60,7 @@ export const BreadCrumb: React.FC<Props> = ({
 
 interface PatchBreadcrumbProps {
   patchAuthor: string;
-  analytics: any;
+  analytics: ReturnType<typeof useBreadcrumbAnalytics>;
```
SupaJoon, line 63: The type for useBreadCrumbAnalytics is established in it's file. Could we import that here instead using ReturnType and typeof?

src/pages/Version.tsx
```diff
@@ -97,7 +99,12 @@ export const VersionPage: React.FC = () => {
         getVersion();
       }
     }
-  }, [patchData, getVersion, id]);
+    // if there was an error fetching the patch and the patch id was a real id it could be a periodic build
+    // in which case we should check if theres a corresponding version associated with it.
+    if (validatePatchId(id) && patchError) {
```
sophstad, line 104: The console is logging lots of (expected) errors, which we should probably suppress. But moreover, I feel like this approach is not ideal because we'd start logging all of these GraphQL errors to Splunk, and since they're not really errors, it would be noisy and misleading. Is there a way that the resolvers could be updated, or a new one added, to avoid relying on an error?

Also interesting because this is really similar to the repo/project issue we spoke about the other day! This may be a pattern we start seeing more often.

src/pages/Version.tsx
```diff
@@ -48,39 +50,63 @@ export const VersionPage: React.FC = () => {
   const [redirectURL, setRedirectURL] = useState(undefined);
   const [isLoadingData, setIsLoadingData] = useState(true);
 
-  const [getPatch, { data: patchData }] = useLazyQuery<
+  const { error: isPatchOrVersionError } = useQuery<
```
ablack12, line 53: nit: can this error be called something more related to the new query name

sophstad (approved):

SupaJoon (approved):

src/pages/Version.tsx
khelif96: Nice catch!

PR #962, opened 2021-11-03
EVG-15731: Render repo settings pages

sophstad:
EVG-15731

### Description 
- Fetch repo data when either:
  - On a repo settings page (i.e. path includes an Object ID), or
  - Project's `useRepoSettings` field is true
- Propagate repoData and projectData to form
- Correctly rendering repo settings when a project is defaulting to repo will be handled in EVG-15643

### Testing 
- To preview with local data, visit http://localhost:3000/project/602d70a2b2373672ee493184/settings/general
- Will add integration tests as part of EVG-15643, in which the general settings page will be completed


src/components/ProjectSettingsTabs/GeneralTab.tsx
```diff
@@ -11,21 +11,26 @@ import { GeneralTabProps } from "./types";
...
+    repoData,
+  ]);
   usePopulateForm(initialFormState, tab);
 
   const onChange = useCallback(({ formData }) => updateForm(tab, formData), [
     updateForm,
   ]);
 
-  const { validDefaultLoggers } = data;
+  const validDefaultLoggers =
```
malikchaya2, line 32: should this be the reverse? first try to use the repodata and only use the project one if repo is empty?

sophstad: No, we want to use the project's value if it has been set and, if not, use the repo value. Ultimately though it doesn't matter for this field–it's very static data that is the [same in both places](https://github.com/evergreen-ci/evergreen/blob/main/graphql/resolvers.go#L414-L420).

malikchaya2: ah true and true.

malikchaya2 (approved):

src/pages/ProjectSettings.tsx
```diff
@@ -34,13 +37,30 @@ export const ProjectSettings: React.FC = () => {
...
     ProjectSettingsQueryVariables
   >(GET_PROJECT_SETTINGS, {
+    skip: isRepo,
     variables: { identifier },
   });
 
+  const repoRefId =
+    projectData?.projectSettings?.projectRef?.repoRefId || identifier;
+  const useRepoSettings =
+    projectData?.projectSettings?.projectRef?.useRepoSettings;
```
khelif96, line 54: nit. destructure the common portions of these 2 imports

src/pages/ProjectSettings.tsx
```diff
@@ -34,13 +37,30 @@ export const ProjectSettings: React.FC = () => {
...
+
+  const { data: repoData } = useQuery<
+    RepoSettingsQuery,
+    RepoSettingsQueryVariables
+  >(GET_REPO_SETTINGS, {
+    skip: useRepoSettings === false,
```
khelif96, line 60: does [useLazyQuery](https://www.apollographql.com/docs/react/data/queries/#manual-execution-with-uselazyquery) work here instead?

SupaJoon: I think this query will always be dispatched since the initial render value of useRepoSettings will be undefined

SupaJoon: adjusting line 53 to  `const useRepoSettings = projectData?.projectSettings?.projectRef?.useRepoSettings ?? false` could prevent this

SupaJoon (approved):

src/pages/ProjectSettings.tsx
sophstad: I think since both of these variables require optional chaining, any attempt to use destructuring ends up being messier than this.

khelif96 (approved):

PR #963, opened 2021-11-03
EVG-15735 Create Loading Row component for History Table

minnakt:
[EVG-15735](https://jira.mongodb.org/browse/EVG-15735)

### Description 
This PR adds a visual loading state for rows of the history table that haven't loaded yet. There are no Figma designs for this component, so any requests for visual changes are appreciated.

### Screenshots
https://user-images.githubusercontent.com/47064971/140196186-4f0326db-c0f6-4147-b3d7-d25391f42a6c.mov

### Testing 
 - Ran existing tests


minnakt:
There are no Figma designs for this loading state, so any comments/requests to change the design are appreciated! I also played around with something like this, but given that there don't seem to be spinners in any other part of Spruce, it didn't seem to fit.

https://user-images.githubusercontent.com/47064971/140195794-e525606f-dd84-4595-98bc-d4929a46c337.mov

src/components/HistoryTable/row/LoadingRow.tsx
```diff
@@ -0,0 +1,23 @@
...
+interface LoadingRowProps {
+  numVisibleCols: number;
+}
+export const LoadingRow: React.FC<LoadingRowProps> = ({ numVisibleCols }) => (
+  <>
+    <LabelCellContainer>
+      <Skeleton active title={false} paragraph={{ rows: 3 }} />
+    </LabelCellContainer>
+    {Array.from(Array(numVisibleCols)).map(() => (
+      <LoadingCell key={`loading_row_${uuid()}`} />
```
minnakt, line 15: I added `uuid` to the key here because I was concerned that there would be multiple duplicate keys for these cells
Maybe there is a better way to do this 🤔

khelif96: uuid will change the key on every render and isn't useful as a key. You will be better off using the array index as the key in this case. Since we are rendering the same element every time it would be ok.

khelif96 (approved):
Just had 1 comment this looks awesome! 🔥

PR #967, opened 2021-11-05
EVG-13226: Remove logo from navbar

SupaJoon:
[EVG-13226](https://jira.mongodb.org/browse/EVG-13226)

It's not really easy to see the alignment improvement from a screenshot since it's just a few pixels so it's best if ya'll run it locally.  

Update: We decided to remove the logo in the last design sync.
Screen Shot 2021-11-11 at 3 14 47 PM


src/components/Header/Navbar.tsx
```diff
@@ -89,6 +89,7 @@ const StyledNav = styled.nav`
 const Logo = styled.div`
   display: flex;
   align-items: center;
+  margin-top: -3px;
```
khelif96, line 92: now the logo isn't centered in the nav. :/

khelif96 (changes requested):
I would push back on the need to do this ticket.
Both the logo and the nav items technically are centered relative to the nav. They just  don't appear centered relative to each other because they have different font-sizes and are positioned near each other. 

If you give them both the same font weight and size they are centered perfectly.
image

There is a hack that will allow you to center the nav  items with the logo. Even with different font sizes. You can give the nav items the same `line-height` as the logo and they will be centered automatically with the same positioning as the logo. But like i said i feel  that it's a bit of a hack. But i think its better than manually fine tuning the margins of the elements to get them to align perfectly.

SupaJoon:
I moved this ticket into waiting to get feedback from design on whether we need the logo text at all.

khelif96 (approved):
Can we add some analytics to the other nav items. Just to understand what people are clicking?

SupaJoon:
> Can we add some analytics to the other nav items. Just to understand what people are clicking?

I created another ticket for this since it will require some modifications to Header/Dropdown.tsx
https://jira.mongodb.org/browse/EVG-15850

PR #969, opened 2021-11-08
EVG-14385 Prevent restarting commit queue tasks from the UI

minnakt:
[EVG-14385](https://jira.mongodb.org/browse/EVG-14385)

### Description 
This PR makes it so that a user cannot `Schedule` or `Restart` tasks if the patch is on the commit queue by disabling those buttons.

### Screenshots
https://user-images.githubusercontent.com/47064971/140812011-8f1a40d5-b175-4624-8147-a5fd49e42077.mov

### Testing 
- Modified e2e tests that no longer pass due to disabling the `Schedule` and `Restart` buttons

### Evergreen PR 
- evergreen-ci/evergreen/pull/5171


SupaJoon:
I just restarted the e2e_test task since it was aborted

src/gql/generated/types.ts
```diff
@@ -1318,6 +1318,7 @@ export type Task = {
   /** @deprecated version is deprecated. Use versionMetadata instead. */
   version: Scalars["String"];
   versionMetadata: Version;
+  patch?: Maybe<Patch>;
 };
```
minnakt, line 1322: Fix; `patch` field doesn't obey alphabetical order

khelif96 (changes requested):

src/pages/Task.tsx
```diff
@@ -66,10 +66,14 @@ export const Task: React.FC = () => {
     annotation,
     latestExecution,
     versionMetadata,
+    patch,
```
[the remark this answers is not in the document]
minnakt: Sounds good! It won't work locally because I think the requester fields in the database are incorrect for the one patch on the commit queue, but it should work for deployed version (I'll check once with `yarn prod` once I can find a patch on a commit queue...)

minnakt:
Tested with `yarn prod` on a patch that was on commit queue for `mongodb-mongo-master`

Screen Shot 2021-11-10 at 3 29 20 PM
:

src/pages/Task.tsx
```diff
@@ -103,6 +107,7 @@ export const Task: React.FC = () => {
         }
         buttons={
           <ActionButtons
+            isPatchOnCommitQueue={isPatchOnCommitQueue}
             canAbort={canAbort}
             canRestart={canRestart}
```
khelif96, line 112: Instead of adding another prop could we just update the existing props? 
`canRestart={!isPatchOnCommitQueue && canRestart}`

minnakt: yes! done! Checked with `yarn prod` that it still works.

khelif96 (approved):

src/types/commits.ts
```diff
@@ -17,3 +17,5 @@ export enum ChartTypes {
   Absolute = "absolute",
   Percentage = "percentage",
 }
+
+export const commitQueueRequester = "merge_test";
```
khelif96, line 21: we have a constants directory. I think this should live in one of the files there.

khelif96 (approved):
Apart from my comment LGTM

PR #970, opened 2021-11-08
EVG-15809: Utilize mci-project-cookie and SpruceConfig to load default Project Health project

SupaJoon:
[EVG-15809](https://jira.mongodb.org/browse/EVG-15809)

### Description 
Utilize mci-project-cookie and then the GetSpruceConfig query to load the appropriate waterfall project when navigating to /commits

### DIFF
--- a/src/constants/cookies.ts
+++ b/src/constants/cookies.ts
@@ -4,3 +4,4 @@ export const SUBSCRIPTION_METHOD = "subscription-method";
 export const getNotificationTriggerCookie = (type: string) =>
   `${type}-notification-trigger`;
 export const SLACK_NOTIFICATION_BANNER = "has-closed-slack-banner";
+export const CURRENT_PROJECT = "mci-project-cookie";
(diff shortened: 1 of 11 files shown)


src/constants/cookies.ts
```diff
@@ -0,0 +1,7 @@
+export const HIDE_FEEDBACK = "HIDE_FEEDBACK";
```
khelif96, line 1: Great IDEA

khelif96 (approved):

src/constants/cookies.ts
```diff
@@ -0,0 +1,7 @@
...
+export const ANNOUNCEMENT_TOAST = "announcement-toast";
+export const SUBSCRIPTION_METHOD = "subscription-method";
+export const getNotificationTriggerCookie = (type: string) =>
+  `${type}-notification-trigger`;
+export const SLACK_NOTIFICATION_BANNER = "has-closed-slack-banner";
+export const CURRENT_PROJECT = "mci-project-cookie";
```
sophstad, line 7: Tiny nit but probably can drop the "-cookie" since we don't use that pattern in the others
```suggestion
export const CURRENT_PROJECT = "mci-project";
```

sophstad (approved):

src/constants/cookies.ts
SupaJoon: I actually meant to reuse the cookie that we use on the legacy waterfall: https://github.com/evergreen-ci/evergreen/blob/main/service/middleware.go#L26

PR #976, opened 2021-11-09
EVG-15796: Linkify CommitChartLabel

SupaJoon:
[EVG-15796](https://jira.mongodb.org/browse/EVG-15796)

### Description 
The githash links to the version page and the Jira ticket links to Jira
Screen Shot 2021-11-09 at 5 21 43 PM
Screen Shot 2021-11-09 at 5 22 23 PM


src/components/CommitChartLabel/index.tsx
```diff
@@ -12,25 +21,37 @@ interface Props {
...
+    GetSpruceConfigQueryVariables
+  >(GET_SPRUCE_CONFIG);
+  const jiraHost = configData?.spruceConfig?.jira?.host;
 
   return (
     <LabelContainer data-cy="commit-label">
       <LabelText>
-        {githash.substr(0, 6)} {shortDate(createDate)}
+        <StyledRouterLink to={getVersionRoute(versionId)}>
+          {githash.substring(0, 6)}
```
sophstad, line 47: Based on a [quick search](https://github.com/evergreen-ci/spruce/search?q=%22substring%22) it looks like we're shortening git hashes somewhat inconsistently across Spruce. Is there a particular reason to use a 5-character hash here? If not, it may be nice to standardize (the [default is 7 characters](https://git-scm.com/book/en/v2/Git-Tools-Revision-Selection#_short_sha_1), fwiw). This may just make sense as a followup ticket!

sophstad (approved):

src/components/CommitChartLabel/index.tsx
SupaJoon: Good point. It seems like we show hash lengths are between 5-7 characters in various parts of our app. I will ask in the design sync whether we should just stick with one length.

SupaJoon: I will change this to 5 for now since that's what we currently have in the legacy ui labels.

PR #977, opened 2021-11-10
EVG-15643: Project inherits repo for General Settings

sophstad:
EVG-15643

### Description
- Add helper functions for form fields that conditionally show "Default to repo option" and update all general settings so that they can default to repo
- Update "File patterns to ignore" section to only allow users to update the array of patterns if they are not defaulting to repo settings
- Improve typing to facilitate save functionality in EVG-15642

### Screenshots
Project attached to repo:
image

Project not attached to repo:
image

Repo:
image

### Testing
Added unit and integration tests. To preview locally when running a dev server using the latest `main` branch of evergreen:
- http://localhost:3000/project/evergreen/settings/general to view a project that inherits from a repo
- http://localhost:3000/project/spruce/settings/general to view a project that does not inherit from a repo
- http://localhost:3000/project/602d70a2b2373672ee493184/settings/general to view a repository's settings


src/pages/projectSettings/Tabs.tsx
```diff
@@ -60,7 +60,9 @@ export const ProjectSettingsTabs: React.FC<Props> = ({
         >
           Save changes on page
         </Button>
-        {!useRepoSettings && <Button>Default to Repo on page</Button>}
+        {projectData && useRepoSettings && (
```
malikchaya2, line 63: should this be !useRepoSettings?

src/components/ProjectSettingsTabs/utils.ts
```diff
@@ -1,5 +1,45 @@
...
 }
+
+const insertIf = (condition, ...elements) => (condition ? elements : []);
+
+/**
+ * Generate the options for a radio box group that conditionally includes a third option to "Default to repo".
+ *
+ * @param {options} string array representing the primary two radio box labels.
+ * @param {field} value upon which the default box should be conditionally shown.
+ * @param {invert} specify whether or not the field represents an "inverted" value; i.e. the option indicates a field is disabled rather than enabled.
```
malikchaya2, line 15: the term invert confuses me a bit, can you please explain a bit more? how do we know what value it holds?

malikchaya2 (commented):
This looks great! I don't know how form states work though, so another reviewer is probably a good idea:)

src/components/ProjectSettingsTabs/GeneralTab.tsx
```diff
@@ -74,27 +85,21 @@ export const GeneralTab: React.FC<GeneralTabProps> = ({
...
-  defaultLogger = "",
-  cedarTestResultsEnabled = false,
-  patchingDisabled = false,
+  defaultLogger,
   taskSync,
-  disabledStatsCache = false,
-  filesIgnoredFromCache = [],
-}) => ({
-  enabled: enabled ? "enabled" : "disabled",
+  ...data
```
SupaJoon, line 98: nit: can we either keep all of these variables spread and assign them to the schema like we do for owner and branch or access them from a data object?

src/pages/projectSettings/Tabs.tsx
sophstad: No -- this was confusing me at first too, but this button actually relates to "unsetting"/nullifying all the project values on this page, not opting in to use repo settings. A user can only do so if they have already opted to use repo settings.

SupaJoon (approved):

src/pages/projectSettings/Tabs.tsx
```diff
@@ -60,7 +60,9 @@ export const ProjectSettingsTabs: React.FC<Props> = ({
...
+          <Button data-cy="default-to-repo">Default to Repo on page</Button>
```
ablack12, line 64: Can we use capitalization for buttons/headers (Default to Repo on Page)? sorry i didn't see this when we did "save changes on page" but I think this should extend to that (move to new repo, etc).

src/pages/projectSettings/Tabs.tsx
ablack12: If "Default to Repo on Page" is confusing, we could do something like "Set Page to Repo Defaults" ? idk if that's better

PR #979, opened 2021-11-10
EVG-15826: update evergreen YAML for Go module changes

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-15826

### Description 
Spruce has a dependency on how Evergreen builds its local server for CI testing, Since Evergreen uses modules, it's no longer tied to a `GOPATH`. I updated the evergreen YAML to reflect this. I put it in the `build` directory instead of `gopath/src/github.com/evergreen-ci`, but if you prefer it to be somewhere else, let me know. The location doesn't really matter that much.

* Clean up the usages of `GOPATH` (no longer necessary) and `GO_BIN_PATH` (no longer used for Evergreen Go builds).
* Update the README slightly since Evergreen no longer depends on `GOPATH`.

### Screenshots
N/A

### Testing 
N/A

### Evergreen PR 
N/A

### DIFF
--- a/.evergreen.yml
+++ b/.evergreen.yml
@@ -8,7 +8,6 @@ modules:
 - name: evergreen
   repo: git@github.com:evergreen-ci/evergreen.git
   branch: main
-  prefix: build
 
 #######################################
 #              Functions              #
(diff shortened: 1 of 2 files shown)


khelif96 (approved):
LGTM as long as tests pass.

khelif96:
The compile task relies on the `schema.graphql` file that exists in the evergreen repo. It creates a symlink from that file to the spruce repo. Its failing because the symlink isn't being created properly. 

The sym-link is created here https://github.com/evergreen-ci/spruce/blob/487c9d6892e054f67fc0430883afeb23f078ae6c/.evergreen.yml#L88 So you would need to update that as well. 

Feel free to reach out if it does not work after updating that.

Kimchelly:
The compile failure is a bit mysterious because if I delete the `sym-link` command entirely, both `compile` and `lint` fail. I'm not sure why `lint` succeeded if the symlink is invalid, but I'll keep looking.

Kimchelly:
I fixed compile. Apparently, yarn deletes everything in the build directory before during `yarn build`.

PR #980, opened 2021-11-11
Evg-15507: Replace task status squares with status icons

SupaJoon:
[Evg-15507](https://jira.mongodb.org/browse/EVG-<number>)

### Description 
Replaces status squares with status icons in the version restart modal and execution selector

### Screenshots
Screen Shot 2021-11-11 at 5 09 32 PM
Screen Shot 2021-11-11 at 5 10 14 PM


src/components/TaskStatusIcon/index.tsx
```diff
@@ -10,39 +10,45 @@ const failurePurple = "#36367F";
 interface TaskStatusIconProps {
   status: string;
   size?: Size | number;
+  style?: React.CSSProperties;
```
SupaJoon, line 13: Unfortunately wrapping `<Icon />` with `styled` wouldn't apply the generated className to the component even though wrapping `Icon` does.

src/components/TaskStatusIcon/index.tsx
```diff
@@ -10,39 +10,47 @@ const failurePurple = "#36367F";
 interface TaskStatusIconProps {
   status: string;
   size?: Size | number;
+  rightMargin?: boolean;
 }
 
 export const TaskStatusIcon: React.FC<TaskStatusIconProps> = ({
   status,
   size = Size.Default,
+  rightMargin,
```
SupaJoon, line 19: Unfortunately wrapping <Icon /> with styled wouldn't apply the generated className to the component even though wrapping Icon does.

sophstad: Interesting, would this not be a matter of capturing the `className` prop that's passed to `TaskStatusIcon` and then including it in the props applied to the returned `<Icon />`

SupaJoon: Great idea 😅

src/components/VersionRestartModal/TaskStatusCheckbox.tsx
```diff
@@ -28,11 +27,9 @@ const CheckboxComponent: React.FC<TaskStatusCheckboxProps> = ({
     name={taskId}
     label={
       <StateItemWrapper>
-        <PaddedSquare color={mapTaskStatusToColor[status]} />
-        {baseStatus && (
-          <PaddedSquare color={mapTaskStatusToColor[baseStatus]} />
-        )}
-        {displayName}
+        <StyledTaskStatusIcon status={status} />
```
khelif96, line 30: We could give these icons a fixed width so if baseStatus doesn't exist the displayName text doesn't get shifted to the side.

src/components/TaskStatusIcon/index.tsx
```diff
@@ -10,39 +10,45 @@ const failurePurple = "#36367F";
...
 
 export const TaskStatusIcon: React.FC<TaskStatusIconProps> = ({
   status,
   size = Size.Default,
+  className,
 }) => {
+  const props = {
+    size,
+    ...(className && { className }),
+  };
```
khelif96, line 24:
```suggestion
interface TaskStatusIconProps extends React.ComponentProps<typeof Icon> {
  status: string;
}

export const TaskStatusIcon: React.FC<TaskStatusIconProps> = ({
  status,
  size = Size.Default,
  ...rest,
}) => {

```

src/components/TaskStatusIcon/index.tsx
```diff
@@ -10,39 +10,45 @@ const failurePurple = "#36367F";
...
   switch (status) {
     case TaskStatus.Succeeded:
-      return <Icon glyph="Checkmark" fill={green.base} size={size} />;
+      return <Icon glyph="Checkmark" fill={green.base} {...props} />;
```
khelif96, line 27:
```suggestion
      return <Icon glyph="Checkmark" fill={green.base} size={size} {...rest} />;
```

src/components/TaskStatusIcon/index.tsx
khelif96: You can simplify the types for this and inherit the types of Icon.

src/components/VersionRestartModal/TaskStatusCheckbox.tsx
```diff
@@ -28,11 +27,9 @@ const CheckboxComponent: React.FC<TaskStatusCheckboxProps> = ({
...
+        {baseStatus && <StyledTaskStatusIcon status={baseStatus} />}
+        <div>{displayName}</div>
```
khelif96, line 32: If you wrap <StyledTaskStatusIcon /> with a fixed width wrapper it can handle positioning of the icons and prevent the `displayName` text from being pushed to the side when `baseStatus` doesn't exist for a task. 

So we can avoid situations where the list is not aligned.

khelif96 (commented):
LGTM pending the comments below

PR #981, opened 2021-11-12
EVG-14680 Resolve incorrect display of TaskLogs between executions

minnakt:
[EVG-14680](https://jira.mongodb.org/browse/EVG-14680)

### Description 
This is a very specific bug that seems to only occur in one situation.

When you restart a task, the execution for that task will show `No logs` (as the task has not run yet). Switching to another execution whose logs are _already in the cache_ will cause `No logs` to persist, which is not correct. However, this behavior will only persist for **one** execution. Here is an example:

https://user-images.githubusercontent.com/47064971/141480405-07e85eac-e76d-428b-84a6-584ff2835b7c.mov

The root cause appears to be that we determine a `noLogs` status based on if an error occurs. For some reason, switching to another execution whose logs are already in the cache will return both an error and the data that we want to display. This behavior is shown below:

https://user-images.githubusercontent.com/47064971/141481714-ccb277ce-9993-45c8-8c59-2326035d9ffd.mov

I don't believe an error should be returned if Apollo is just looking up the cache, so I can't explain why the error persists. Since returning `undefined` data and an error should be equivalent in this situation, this PR modifies how we determine the `noLogs` status.

### Screenshots
https://user-images.githubusercontent.com/47064971/141482156-d1853053-c8ac-4940-861a-58aad559a8f1.mov

### Testing 
- Tested manually with `yarn prod`

### DIFF
--- a/src/gql/GQLWrapper.tsx
+++ b/src/gql/GQLWrapper.tsx
@@ -44,6 +44,9 @@ const cache = new InMemoryCache({
     User: {
       keyFields: ["userId"],
     },
+    TaskLogs: {
+      keyFields: ["execution", "taskId"],
+    },
     Task: {
       keyFields: ["execution", "id"],
       fields: {
--- a/src/gql/generated/types.ts
+++ b/src/gql/generated/types.ts
@@ -2419,7 +2419,11 @@ export type AgentLogsQueryVariables = Exact<{
 }>;
 
 export type AgentLogsQuery = {
-  taskLogs: { agentLogs: Array<LogMessageFragment> };
+  taskLogs: {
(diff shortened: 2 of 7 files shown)


minnakt:
I changed some GQL cache configurations so that the `TaskLogs` properly appear in the Apollo cache.

Screen Shot 2021-11-11 at 10 03 27 PM

Unfortunately, this change did not affect the current behavior of the task logs, which is why I altered the `noLogs` conditional.

SupaJoon (approved):
Thanks for digging into this!

khelif96 (approved):

PR #983, opened 2021-11-12
EVG-13586 Add Reprovision button to hosts pages

minnakt:
[EVG-13586](https://jira.mongodb.org/browse/EVG-13586)

### Description 
This PR adds a new button to the `/hosts` and `/host/:id` pages which can be used to reprovision a host. This button is useful to have for static hosts, as it can essentially reset them when they get into a bad state (from what I understand from a conversation with Kim).

### Screenshots
On `/hosts` page:

https://user-images.githubusercontent.com/47064971/141543590-de0cd873-c652-4466-bb2c-d41468c50b20.mov

On `/host/:id` page:

https://user-images.githubusercontent.com/47064971/141540918-7bae2543-98db-4f00-96e9-d770f516630a.mov

### Testing 
- Tested on staging
- Added e2e tests that interact with new `Reprovision` button

### Evergreen PR 
- evergreen-ci/evergreen/pull/5195


cypress/integration/host/host_buttons.ts
```diff
@@ -14,6 +14,14 @@ describe("Host page restart jasper and update host status buttons", () => {
     cy.dataCy("restart-jasper-button").click();
     cy.get(popconfirmYesClassName).click();
     cy.dataCy("toast").should("exist");
+    cy.get(`[aria-label="Close Message"]`).click();
+  });
+
+  it("Should show a toast when host is reprovisioned", () => {
+    cy.dataCy("reprovision-button").click();
+    cy.get(popconfirmYesClassName).last().click();
```
minnakt, line 22: I'm not sure why I need this extra `last()` call for the reprovision test but not the restart jasper test. The e2e tests won't pass without it.

I also tried replacing both of these calls with `cy.contains("button", "Yes").click()`, but similarly, this passes for the first call and fails for the second call.

src/pages/Host.tsx
```diff
@@ -77,7 +78,7 @@ export const Host: React.FC = () => {
 
   usePageTitle(`Host${hostId ? ` - ${hostId}` : ""}`);
 
-  const canRestartJasper =
+  const canRestartJasperOrReprovision =
     host?.status === "running" &&
     (bootstrapMethod === "ssh" || bootstrapMethod === "user-data");
```
minnakt, line 83: I assumed the condition for reprovisioning from [here](https://github.com/evergreen-ci/evergreen/blob/dd69abc6b54990e27968dc5488aa2c839adc6647/public/static/js/host_admin.js#L15-L17)

cypress/integration/host/host_buttons.ts
khelif96: Wanna dive down the rabbit hole of components mounting on to the dom when you first open the tooltip but not unmounting? 
The tooltip is mounting on to the dom when you click on the button in the previous test. 
When the tooltip closes it still exists on the dom. Since you are selecting it based off of a classname selector it is finding the first tooltip you clicked. That still exists but just isn't visible.

src/components/Hosts/Reprovision.tsx
```diff
@@ -0,0 +1,89 @@
+import React from "react";
+import { useMutation } from "@apollo/client";
+import { Popconfirm, Tooltip } from "antd";
```
khelif96, line 3: Can you replace these with leafygreen components?

src/components/Hosts/Reprovision.tsx
```diff
@@ -0,0 +1,101 @@
...
+        </Button>
+      </div>
+    </ConditionalWrapper>
+  );
+};
+
+// @ts-expect-error
+const StyledTooltip = styled(Tooltip)`
+  width: 300px;
+`;
```
minnakt, line 101: For leafygreen `Tooltip` & `Popover`, there is a bug where you have to set the width to prevent misalignment when the trigger element is near the right side of a page. You can see the ticket here: [PD-1542](https://jira.mongodb.org/browse/PD-1542)

src/components/Hosts/Reprovision.tsx
minnakt: Yes, here are the updated screenshots.

With leafygreen `Tooltip`:

Screen Shot 2021-11-16 at 9 35 19 AM

With leafygreen `Popover`:

Screen Shot 2021-11-16 at 9 35 55 AM

Replaced for both `Restart Jasper` & `Reprovision` button.

cypress/integration/host/host_buttons.ts
```diff
@@ -12,8 +10,16 @@ describe("Host page restart jasper and update host status buttons", () => {
...
+    cy.contains("button", "Yes").click();
+    cy.dataCy("toast").should("exist");
+    cy.get(`[aria-label="Close Message"]`).click();
+  });
+
+  it("Should show a toast when host is reprovisioned", () => {
+    cy.dataCy("reprovision-button").click();
+    cy.contains("button", "Yes").click();
     cy.dataCy("toast").should("exist");
+    cy.get(`[aria-label="Close Message"]`).click();
```
minnakt, line 22: Will these tests with toasts have to change given the refactor to toast component?

khelif96 (changes requested):

src/components/Hosts/HostPopover.tsx
```diff
@@ -0,0 +1,99 @@
...
+  width: 300px;
+`;
+
+const PopoverContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+  background-color: white;
+  padding: 16px;
+  box-shadow: 0 3px 6px -4px rgb(0 0 0 / 12%), 0 6px 16px 0 rgb(0 0 0 / 8%),
+    0 9px 28px 8px rgb(0 0 0 / 5%);
```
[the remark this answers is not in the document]
minnakt: Sorry, I took a little creative freedom because the default `Popover` is pretty hard to see (no border, no shading):

Screen Shot 2021-11-18 at 8 31 54 PM

I can make this line neater by using the official MongoDB colors from `uiColors.ts`.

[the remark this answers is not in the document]
khelif96: haha all good. Might also be a good idea to bring it up with leafygreen maybe they have some suggestions in mind.

src/components/Hosts/HostPopover.tsx
```diff
@@ -0,0 +1,118 @@
...
+        popoverRef.current!.contains(event.target as Node) ||
+        buttonRef.current!.contains(event.target as Node);
+      setActive(stillFocused);
+    };
+
+    document.addEventListener("mousedown", onClickOutside);
+    return () => {
+      document.removeEventListener("mousedown", onClickOutside);
+    };
+  }, [buttonRef, popoverRef, active]);
```
minnakt, line 46: I was going to use [the hook](https://github.com/evergreen-ci/spruce/blob/92f2dbe3f6c8104e0bc2313b76489e28a6d0d647/src/hooks/useOnClickOutside.ts) but then I found that I needed to use two `refs` simultaneously.

It's because clicking the button when the popover is already open should close the popover, but the `popoverRef` interferes with that behavior. The `popoverRef` will detect that click as an outside click and close the button, so the button click will just open the popover again (i.e. it's not possible to close the popover). The behavior is shown in the video below:

https://user-images.githubusercontent.com/47064971/142654379-8c6fafc2-5f0a-4d23-83f9-b46f770340d1.mov

src/components/Hosts/HostPopover.tsx
```diff
@@ -0,0 +1,99 @@
...
+      justify="middle"
+      active={active}
+      data-cy={dataCy}
+    >
+      <PopoverContainer ref={popoverRef}>
+        {titleText}
+
+        <ButtonContainer>
+          <ButtonSpacer>
+            <Button size="xsmall" disabled={loading}>
```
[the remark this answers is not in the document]
minnakt: fixed

src/components/Hosts/HostPopover.tsx
khelif96: Could we extend that hook to maybe allow passing in many refs?  It would also be useful here. https://github.com/evergreen-ci/spruce/blob/main/src/components/SearchableDropdown/index.tsx#L51-L66

minnakt: I'll look into it 🥸

src/hooks/useOnClickOutside.ts
```diff
@@ -1,18 +1,24 @@
...
+      );
+      cb(isFocused);
     }
     document.addEventListener("mousedown", handleClickOutside);
     return (): void => {
       document.removeEventListener("mousedown", handleClickOutside);
     };
-  }, [cb, ref]);
+  }, [cb, refs, isOpen]);
 };
```
minnakt, line 24: I pretty much just moved the existing code into the hook... I tested with one ref and two refs, and it works as I would expect

src/hooks/useOnClickOutside.ts
```diff
@@ -1,18 +1,24 @@
 import { useEffect } from "react";
 
 export const useOnClickOutside = (
-  ref: React.RefObject<HTMLElement>,
-  cb: () => void
+  refs: Array<React.RefObject<HTMLElement>>,
+  cb: (isFocused: boolean) => void,
+  isOpen: boolean
```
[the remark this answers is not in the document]
khelif96: I think as a generic hook it shouldn't case on specific use cases. This implies it can only be used for some openable element.  

A suggestion would be the implementer should make their callback case on this condition.

khelif96 (commented):
Last set of suggestions

src/hooks/tests/useOnClickOutside.test.tsx
```diff
@@ -0,0 +1,58 @@
+import React from "react";
```
minnakt, line 1: Let me know if I did these wrong 😭

khelif96 (approved):

PR #985, opened 2021-11-15
Spruce beta

SupaJoon:
contains [EVG-15520](https://jira.mongodb.org/browse/EVG-15520)

### DIFF
--- a/src/components/Header/Navbar.tsx
+++ b/src/components/Header/Navbar.tsx
@@ -6,23 +6,19 @@ import { Link } from "react-router-dom";
 import { useNavbarAnalytics } from "analytics";
 import Icon from "components/Icon";
 import { StyledLink } from "components/styles";
-import { getUserPatchesRoute, routes } from "constants/routes";
+import { getCommitsRoute, getUserPatchesRoute, routes } from "constants/routes";
 import { useAuthStateContext } from "context/auth";
 import { GetUserQuery } from "gql/generated/types";
 import { GET_USER } from "gql/queries";
 import { useLegacyUIURL } from "hooks";
-import { environmentalVariables } from "utils";
 import { AuxiliaryDropdown } from "./AuxiliaryDropdown";
 import { UserDropdown } from "./UserDropdown";
-
-const { getUiUrl } = environmentalVariables;
 
 const { white, blue, gray } = uiColors;
 
 export const Navbar: React.FC = () => {
   const { isAuthenticated } = useAuthStateContext();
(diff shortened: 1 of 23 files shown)


src/components/styles/Layout.tsx
```diff
@@ -64,10 +64,12 @@ export const TableControlOuterRow = styled(TableControlInnerRow)`
...
 export const PageButtonRow = styled.div`
   display: flex;
-  button {
-    margin-left: 24px;
+  justify-content: flex-end;
+  > button,
+  > a {
+    margin-left: 16px;
+    white-space: nowrap;
   }
```
khelif96, line 72: This can be more generic and not rely on the implementation details of the child components.
```suggestion
  > * {
    margin-right: 16px;
  }
  > *:last-child {
    margin-right: 0;
  }
```

src/pages/task/ActionButtons.tsx
```diff
@@ -250,6 +256,18 @@ export const ActionButtons: React.FC<Props> = ({
   return (
     <>
       <PageButtonRow>
+        <LeafyGreenButton
```
khelif96, line 259: Shouldn't need to import a separate button component component/button should implement all the features of a lg button.

khelif96:
@sophstad @malikchaya2  RFAL added a beta feature flag so this can be merged safely.

src/components/Header/Navbar.tsx
```diff
@@ -97,6 +105,10 @@ const PrimaryLink = styled(Link)`
   color: ${white};
 `;
 
+const PrimaryA = styled.a`
+  color: ${white};
```
sophstad, line 109: [nit] Let's define a shared `primaryStyle` using the `css` prop as is done for secondary links

src/utils/environmentalVariables.ts
```diff
@@ -21,6 +21,8 @@ export const isDevelopment: () => boolean = () =>
 export const isProduction = (): boolean =>
   process.env.NODE_ENV === "production";
 
+export const isBeta = (): boolean => process.env.REACT_ENV_BETA === "true";
```
sophstad, line 24: Assuming it's a limitation of environment variables that they're all represented as strings? Since this would make more sense as a boolean.

sophstad (approved):
LGTM assuming my assumption about env vars is correct!

src/utils/environmentalVariables.ts
khelif96: yup env variables are always strings.

PR #986, opened 2021-11-16
EVG-15857 Refactor Toast context and Testing 

khelif96:
[EVG-15857](https://jira.mongodb.org/browse/EVG-EVG-15857)

### Description 
While working on another ticket i noticed that the implementations of the Toast Context functions have moved away from the typescript types assigned to them. Some Features did not work or where mistyped. It was also difficult to implement certain features of the useToastContext methods. 
This pr rectifies the typescript types. 
Refactors the call signature for some of the methods hopefully making it easier to use.
Implements test coverage around this hook.
Provides a method to mock this logic out in tests.
For both component tests and hook tests.

### Testing 
Thats all i did.


src/context/toast.tsx
```diff
@@ -4,16 +4,26 @@ import Toast, { Variant } from "@leafygreen-ui/toast";
...
+  onClose?: () => void;
+  shouldTimeout?: boolean;
+  title?: string;
+};
+
+type AddToast = (
+  message: string,
+  closable?: boolean,
+  options?: ToastOptions
+) => void;
```
khelif96, line 26: This is the function signature that will be exposed to the user.

src/context/toast.tsx
```diff
@@ -4,16 +4,26 @@ import Toast, { Variant } from "@leafygreen-ui/toast";
 import { WordBreak } from "components/Typography";
 import { TOAST_TIMEOUT } from "constants/index";
 
-export type ToastProps = {
+type ToastProps = {
```
khelif96, line 7: No need to expose this since it was used internally. It also wasn't used anywhere.

src/types/utils.ts
```diff
@@ -1,2 +1,26 @@
-// Helper to extract type T from an array of type T
+/** Helper to extract the type of a single element in an array of elements */
 export type Unpacked<T> = T extends (infer U)[] ? U : T;
+
+/** Helper that takes in an object of key/value pairs and returns an object of key/value pairs where only one value is required */
+export type RequireOnlyOne<T, Keys extends keyof T = keyof T> = Pick<
```
khelif96, line 5: Added a bunch of typescript utility methods for casing on types in an object.

src/context/toast.test.tsx
```diff
@@ -0,0 +1,228 @@
...
+import { useToastContext, ToastProvider } from "./toast";
+
+type useToastContextProps = ReturnType<typeof useToastContext>;
+
+// This type requires that the component only accepts one of the props
+type UseToastComponentProps = RequireAtMostOne<
+  {
+    [K in keyof useToastContextProps]: Parameters<useToastContextProps[K]>;
+  }
+>;
```
khelif96, line 14: This takes the return type of `useToastContext` and maps the returned `DispatchToast` object type. Into a Object with 
`Key: Function Parameter tuple` pairings.

This was necessary in order to expose these as component props below while preserving type safety.
Each callable function will be mapped to the correct function parameters as props. 

`RequireAtMostOne` also ensures that we can't pass more than one parameter into this test component. This ensures we have static type safety checks in place for this component instead of relying on implementation level validation.

src/context/toast.test.tsx
```diff
@@ -0,0 +1,228 @@
...
+
+describe("Real Toast", () => {
+  // Since useToastContext relies on the toastProvider which in turn relies on the react.createPortal api we cannot test it directly
+  // because the react.createPortal api is not available in the testing-library/react-hooks test environment. So we need to create a wrapper
+  // component that will render the toastProvider and useToastContext and test the useToastContext hook internally.
+  const UseToastComponent: React.FC<UseToastComponentProps> = (props) => {
```
khelif96, line 20: This is a test fixture around the useToastContext hook that mimics the implementation of a component using this hook.

src/types/utils.ts
```diff
@@ -1,2 +1,26 @@
...
+  T,
+  Exclude<keyof T, Keys>
+> &
+  {
+    [K in Keys]-?: Required<Pick<T, K>> &
+      Partial<Record<Exclude<Keys, K>, undefined>>;
+  }[Keys];
+
+/**  Helper that takes in an object of key/value pairs and returns an object of key/value pairs where at least one value is required */
+export type RequireAtLeastOne<T, Keys extends keyof T = keyof T> = Pick<
```
SupaJoon, line 15: I don't see this type imported anywhere but we could probably use this in a future test

khelif96: yup i added these types since they seemed to be something we could reuse in the future.

src/context/toast.test.tsx
SupaJoon: nit: this is kind of a confusing component name since we usually start our hooks with `use`

SupaJoon (approved):

SupaJoon:
It would be cool to use this method to test that our analytics events are dispatched. It could also be interesting if we can somehow chain the fake contexts together so we can supply multiple fake contexts to the component we are testing

khelif96:
> It would be cool to use this method to test that our analytics events are dispatched. It could also be interesting if we can somehow chain the fake contexts together so we can supply multiple fake contexts to the component we are testing

That should be super easy to do. We could just wrap the fake contexts around each other. And they would stack.
We could definitely utilize this same method for analytics events as well.

sophstad (approved):

PR #987, opened 2021-11-18
EVG-15805 ad block detection

khelif96:
[EVG-15805](https://jira.mongodb.org/browse/EVG-15805)

### Description 
This builds on top of #986 so only the commits starting at 11186a94f486c979d6d5bd2827a6a07d7dd7f133 are new
This only shows up when on the prod environment. I added this flag so we can avoid running into it locally. But i can remove it if y'all don't think its necessary. 
### Screenshots
image

### Testing 
Unit tests and tested  locally


src/hooks/useAdBlockDetection.ts
```diff
@@ -0,0 +1,52 @@
...
+import { useToastContext } from "context/toast";
+import { environmentalVariables } from "utils";
+import { get } from "utils/request";
+
+const { isProduction } = environmentalVariables;
+// This url is what newrelic downloads to support analytics it gets blocked by adblockers
+const AD_URL = "https://js-agent.newrelic.com/nr-spa-1158.min.js";
+
+/** useAdBlockDetection is a hook that detects if a user has an adblocker enabled. It checks this by trying to download
+ * a file from a url that is blocked by adblockers. If the file is not downloaded, the user has an adblocker enabled.
```
sophstad, line 13: This is an interesting approach. I like it, although I believe we run the risk of mistakenly labeling everyone as having Adblock enabled if this URL ever starts returning an error for whatever reason. However, I think that's a reasonable tradeoff -- it would be easy enough to address if it comes up. 🎉

sophstad (approved):
This is great!

src/hooks/useAdBlockDetection.ts
khelif96: Yup this was one of the few files i could try to download that failed with adblock. That doesn't involve me actually downloading some tracking script. And/or trying to hit some tracking endpoint. It seemed like the most innocent approach.

PR #1006, opened 2021-12-02
EVG-13178: Implement Previous Commit Selector

SupaJoon:
[EVG-13178](https://jira.mongodb.org/browse/EVG-13178), [EVG-15954](https://jira.mongodb.org/browse/EVG-15954)

### Description 
This PR adds the Previous Commit Selector for patches and mainline commit versions.

### Screenshots
https://user-images.githubusercontent.com/47064971/153246981-2f059891-8152-4662-b9e0-862d61cdedff.mov

### Testing 
- Added unit tests in `PreviousCommits.test.tsx`

### Other notes 
- Merged main, so token changes conflicts have already been resolved
- got rid of the console errors that came from `Filters.tsx`


khelif96 (changes requested):
image
I see this warning.

src/pages/task/actionButtons/PreviousCommits.tsx
```diff
@@ -0,0 +1,198 @@
...
+    data?.mainlineCommits.versions.find(({ version }) => version)?.version
+      .buildVariants ?? [];
+  if (buildVariants.length > 1) {
+    reportError(
+      "Multiple build variants matched previous commit search."
+    ).warning();
+  }
+  if (buildVariants[0]?.tasks.length > 1) {
+    reportError("Multiple tasks matched previous commit search.").warning();
+  }
```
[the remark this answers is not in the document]
SupaJoon: I don't think this scenario would occur unless there was an issue with the data model, resolver, or fetch parameter. I thought the reportError is a good sanity check to prevent the component from failing silently since we depend on the list having length 1 which I didn't think was an explicit guarantee

src/components/Table/Filters.tsx
```diff
@@ -146,7 +148,7 @@ const FilterWrapper = styled.div`
 `;
 
 interface StyledOutlinedProps {
-  active?: boolean;
+  active?: number;
```
minnakt, line 151: These error messages were clogging up my console so I fixed this file, but can change it back if you want me to 

Screen Shot 2022-02-09 at 9 50 36 AM

src/pages/task/actionButtons/PreviousCommits.tsx
```diff
@@ -0,0 +1,212 @@
...
+      <StyledSelect
+        size="small"
+        data-cy="previous-commits"
+        label="Previous commits for this task"
+        allowDeselect={false}
+        onChange={(v: CommitType) => setSelectState(v)}
+        value={selectState}
+        disabled={!versionMetadata?.baseVersion}
+      >
+        <Option value={CommitType.Base} key={CommitType.Base}>
```
khelif96, line 166: nit. key isn't needed

src/pages/task/actionButtons/PreviousCommits.tsx
```diff
@@ -0,0 +1,212 @@
...
+          Go to {versionMetadata?.isPatch ? "base commit" : "parent commit"}
+        </Option>
+        <Option value={CommitType.LastPassing} key={CommitType.LastPassing}>
+          Go to last passing version
+        </Option>
+        <Option value={CommitType.LastExecuted} key={CommitType.LastExecuted}>
+          Go to last executed version
+        </Option>
```
khelif96, line 174: What happens if the task never passed or its the first time it ever ran?

src/pages/task/actionButtons/PreviousCommits.tsx
```diff
@@ -0,0 +1,212 @@
...
+            skipOrderNumber,
+            buildVariantOptions: {
+              ...bvOptionsBase,
+              statuses: finishedTaskStatuses,
+            },
+          },
+        });
+      }
+    }
+  }, [parentTask]); // eslint-disable-line react-hooks/exhaustive-deps
```
khelif96, line 134: Could we avoid making these queries if the user never clicks on these buttons? I think the queries are fast enough where we can avoid making them preemptively. 
 Base commit/ parent commit make sense to me since they are the default values/ default state of the button.
But the other ones may never be hit so maybe it would make sense to only call the query if the user selects one of them.

khelif96 (changes requested):

src/pages/Task.tsx
```diff
@@ -76,7 +76,11 @@ export const Task: React.FC = () => {
   const isPatchOnCommitQueue = requester === commitQueueRequester;
 
   // Set the execution if it isnt provided
-  if (Number.isNaN(selectedExecution) && latestExecution !== undefined) {
+  if (
+    Number.isNaN(selectedExecution) &&
+    task?.id === id &&
```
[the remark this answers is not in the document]
minnakt: Ah okay. So this was something I was encountering when a task had multiple executions, and this was the only way I could think to fix it (at least quickly, haha).

When a task has multiple executions, and you click to go to the base task, for some reason the previous task's data will initially persist. You can see in the video attached that the task data's `id` doesn't match with the URL `id`. Since we base the `latestExecution` field off of the task data, that means it tries to go to execution 2 for the base task, which doesn't exist.

https://user-images.githubusercontent.com/47064971/153973587-02f74d57-6262-4f7f-9af6-1f874134d1eb.mov

This actually happens for any task, but most tasks just have 1 execution, so most of the time it doesn't cause an error. I don't really know why this data persists and couldn't find a way to "clear" it :(

src/pages/task/actionButtons/PreviousCommits.tsx
minnakt: Hmm so I checked using this project `http://localhost:3000/commits/docs-ecosystem?skipOrderNumber=8`

I'm pretty sure what happens is that all the links redirect to the parent task. Which is slightly unexpected behavior, because if it never passed it'll link to the parent task (which didn't pass) and if it didn't finish executing it'll link to the parent task (which hasn't finished executing). Maybe I can disable the `Go` button in these cases?

src/pages/task/actionButtons/PreviousCommits.tsx
```diff
@@ -0,0 +1,212 @@
...
+          Go to {versionMetadata?.isPatch ? "base commit" : "parent commit"}
+        </Option>
+        <Option value={CommitType.LastPassing} key={CommitType.LastPassing}>
+          Go to last passing version
+        </Option>
+        <Option value={CommitType.LastExecuted} key={CommitType.LastExecuted}>
+          Go to last executed version
+        </Option>
+      </StyledSelect>
+      <Button as={Link} to={link || "/"} disabled={!link} size="small">
```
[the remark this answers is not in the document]
minnakt: It's to prevent this error in the console:

Screen Shot 2022-02-15 at 11 06 48 AM

src/pages/Task.tsx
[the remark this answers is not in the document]
minnakt: nvm it should be fixed now!!! just needed a `useEffect` :P

src/pages/Task.tsx
```diff
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useEffect } from "react";
```
sophstad, line 1: [nit] As of React 17 there's no need to `import React` so may as well remove it

src/pages/Task.tsx
```diff
@@ -76,11 +76,13 @@ export const Task: React.FC = () => {
...
-      execution: `${latestExecution}`,
-    });
-  }
+  useEffect(() => {
+    if (Number.isNaN(selectedExecution) && latestExecution !== undefined) {
+      updateQueryParams({
+        execution: `${latestExecution}`,
+      });
+    }
+  }, [task]); // eslint-disable-line react-hooks/exhaustive-deps
```
sophstad, line 85: Is it not sufficient to include `selectedExecution` and `latestExecution` here? This would cause the hook to fire again whenever any field within the `task` updates.

minnakt: ahh true!! lemme give it a try P:

src/pages/task/actionButtons/PreviousCommits.tsx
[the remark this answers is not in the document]
minnakt: Resolving because this code changed and went away

minnakt:
hopefully this became less and not more confusing 😵‍💫

src/pages/task/actionButtons/PreviousCommits.tsx
```diff
@@ -0,0 +1,229 @@
...
+        type: "setLastExecutedTask",
+        task: getTaskFromMainlineCommitsQuery(data),
+      });
+    },
+  });
+
+  const { baseTask, versionMetadata, buildVariant, displayName } =
+    taskData?.task ?? {};
+  const { projectIdentifier, order } = versionMetadata?.baseVersion ?? {};
+  const skipOrderNumber = versionMetadata?.isPatch ? order + 1 : order;
```
khelif96, line 91: This logic seems wrong to me a patches order number is not the same as a mainline commits order number. 
Patch order numbers are only relevant to the user who created the patch. (How many patches the user created)

While a mainline commits order number is specific to the project. (How many mainline commits were created.)

khelif96: You would need to look at the base commits order number.

khelif96 (changes requested):

src/types/task.ts
```diff
@@ -100,3 +101,11 @@ export enum TaskStatus {
...
   ScheduledUmbrella = "scheduled-umbrella",
 }
+
+export enum CommitType {
+  Base = "base",
+  LastPassing = "lastPassing",
+  LastExecuted = "lastExecuted",
+}
+
+export type CommitTask = GetBaseVersionAndTaskQuery["task"]["baseTask"];
```
khelif96, line 111: These might make sense in a `task/actionButtons/types.ts` (or something similar) file instead. These are very specific to that component and would likely be better off living colocated where they are used.

khelif96 (approved):
Looks Great! Thanks for all of your work getting this through!
Just had one last comment

PR #1007, opened 2021-12-02
EVG-15921: Improve transformers and typing

sophstad:
EVG-15921

### Description 
- Transform data from GQL results to form state in shared Tabs component (rather than within each tab). This prevents the data from having to be memoized so as not to be re-processed on each render. It also avoids having to do two separate transformations of the data.
- Improving typing and organization of transform functions


khelif96 (approved):

khelif96:
Please merge in main to run the updated linter if you haven't already.

PR #1008, opened 2021-12-03
EVG-15895 Actually enforce the jest rules

khelif96:


sophstad (approved):
Verified rules are being applied by the commit hook using lint-staged 👍 assuming tests pass LGTM!

PR #1009, opened 2021-12-07
EVG-15909 Add Task History And Variant History filtering

khelif96:
[EVG-15909](https://jira.mongodb.org/browse/EVG-15909)

### Description 
Adds test filtering to the task and variant history pages. Users can now filter on test results.

I was deciding between 2 approaches. 
One which fetches test results alongside the rest of the data. and one that fetches them individually as the row is being rendered. 
I went with the latter because backfilling would have been unnecessarily expensive when ever a filter is modified. Since we would have to go back and update previously fetched rows data. In the first approach. 

The latter does have a drawback in that it performs more network requests. (One for each row rendered) But this can be solved in a follow up ticket where we can introduce some batching of network requests.

### Screenshots

Task history filtering
https://user-images.githubusercontent.com/4605522/145098375-9c383f48-a28e-4a20-b55f-a57abda5376b.mov

variant history filtering
https://user-images.githubusercontent.com/4605522/145098447-6c12b390-1199-48a4-b2b9-a41b5fb5a128.mov

Test Filtering on mongodb master projects (Test results in evg)
image

### Testing 
Tested on staging since thats the only place with enough data.
Also wrote some unit tests for some of the functions.


src/components/HistoryTable/HistoryTableContext.tsx
```diff
@@ -18,11 +22,31 @@ interface HistoryTableState {
...
 const HistoryTableDispatchContext = createContext<any | null>(null);
 
-const HistoryTableProvider: React.FC = ({ children }) => {
+interface HistoryTableProviderProps {
+  children: React.ReactNode;
+  initialState?: HistoryTableReducerState;
+}
+const HistoryTableProvider: React.FC<HistoryTableProviderProps> = ({
+  children,
+  initialState = {
```
khelif96, line 37: Extracted this out into a optional prop to make it easier to test the components that depend on this functionality. So we can pre seed the state.

src/components/HistoryTable/HistoryTableIcon/HistoryTableIcon.stories.tsx
```diff
@@ -4,12 +4,12 @@ import { TaskStatus } from "types/task";
 import { HistoryTableIcon } from "./index";
 
 const failingTests = [
-  { testName: "test a", testId: "1" },
-  { testName: "test b", testId: "2" },
-  { testName: "test c", testId: "3" },
-  { testName: "test looooonnnnnnnng name", testId: "4" },
-  { testName: "some other test", testId: "5" },
-  { testName: "test name d", testId: "6" },
+  "test a",
```
khelif96, line 7: This reflects the data we get from the backend.

src/test_utils/test-utils.tsx
```diff
@@ -1,5 +1,10 @@
 import React from "react";
-import { render, queries, RenderResult } from "@testing-library/react";
```
khelif96, line 2: Did some typescript updates in this file so the exported functions are properly typed.

src/utils/array/index.ts
```diff
@@ -1,5 +1,10 @@
-// This takes in an array and an id and safely inserts or removes the value from
-// the array so that there is never a duplicate value
+/**
```
khelif96, line 1: did some refactoring of these array utils. Adding some jsdoc documentation so we get intellisense descriptions. And updated the typescript types so they can be a little bit smarter.

SupaJoon (changes requested):
I'm unable to type in the test name input when only one build variant or task is selected. This is also the case for the buildvariant/task dropdown input. It's possible both history pages are affected by the same bug. Here is an example when you are pointed to staging: http://localhost:3000/task-history/mci/test-thirdparty?buildVariants=race-detector

khelif96:
> I'm unable to type in the test name input when only one build variant or task is selected. This is also the case for the buildvariant/task dropdown input. It's possible both history pages are affected by the same bug. Here is an example when you are pointed to staging: http://localhost:3000/task-history/mci/test-thirdparty?buildVariants=race-detector

Nice catch that was a render cycle a useMemo solved it

src/pages/TaskHistory.tsx
```diff
@@ -66,67 +69,94 @@ export const TaskHistory = () => {
...
+        <PageHeader>
+          <H2>Task Name: {taskName}</H2>
+          <PageHeaderContent>
+            <HistoryTableTestSearch />
+            <BuildVariantSelector projectId={projectId} taskName={taskName} />
+          </PageHeaderContent>
+        </PageHeader>
+        <PaginationFilterWrapper>
+          <BadgeWrapper>
+            <FilterBadges queryParamsToDisplay={queryParamsToDisplay} />
```
sophstad, line 131: It's a slightly weird UX that the "Clear all filters" button only appears once a string search has been added, but clicking it clears the selected build variants too. Maybe out of scope for this ticket, but would be nice to show this button once a build variant or task has been selected.

src/components/HistoryTable/HistoryTableIcon/index.tsx
```diff
@@ -39,15 +36,16 @@ export const HistoryTableIcon: React.FC<HistoryTableIconProps> = ({
...
-            <Body key={testId}>{testName}</Body>
+          {failingTests.map((testName) => (
+            <TestName key={testName}>{testName}</TestName>
           ))}
+          {loadingTestResults && <Skeleton active data-cy="skeleton" />}
         </TestNameContainer>
       </Tooltip>
     )}
   >
     <Container onClick={() => onClick()} data-cy="history-table-icon">
```
sophstad, line 47: This line is throwing an error for me when run from http://localhost:3000/variant-history/evergreen/ubuntu1604?tasks=docker-cleanup using `yarn prod`. Since `onClick` is an optional prop perhaps we want to provide a default noop?

```
Uncaught TypeError: onClick is not a function
    onClick index.tsx:47
    React 14
    unstable_runWithPriority scheduler.development.js:646
    React 4
index.tsx:47
```

https://user-images.githubusercontent.com/9298431/146421282-fcc19013-a37c-4a79-b85e-03a372c6c8e8.mov

src/pages/TaskHistory.tsx
```diff
@@ -66,67 +69,94 @@ export const TaskHistory = () => {
...
+      selectedBuildVariants?.length
+        ? buildVariantsForTaskName?.filter((bv) =>
+            selectedBuildVariants.includes(bv.buildVariant)
+          )
+        : buildVariantsForTaskName,
+    [buildVariantsForTaskName, selectedBuildVariants]
+  );
+
+  useEffect(() => {
+    console.log("Ran effect");
```
sophstad, line 115: Delete log

sophstad (changes requested):

src/components/HistoryTable/HistoryTableIcon/index.tsx
khelif96: Nice catch

src/pages/TaskHistory.tsx
khelif96: Great catch. https://jira.mongodb.org/browse/EVG-15998

SupaJoon (approved):

khelif96:
I am going to merge this so the branch does not go stale while i am on vacation. @sophstad  i addressed all of your comments if you have any more i am happy to reopen this pr or address them in a follow up.

PR #1011, opened 2021-12-13
Deskunking cypress

khelif96:
### Description 
Refactors of a bunch of our tests. Removed allot of dead code from our cypress tests. Refactored all flaky tests to make them not flaky.


cypress/support/commands.ts
```diff
@@ -67,10 +56,23 @@ Cypress.Commands.add("getInputByLabel", (label) =>
...
+
+Cypress.Commands.add(
+  "paste",
+  { prevSubject: true },
+  (selector: Element, pastePayload: string) => {
+    // We can paste directly into a textarea but the onchange event is not fired because cypress directly manipulates the dom elements value.
+    // So we need to manually type in the last character of the paste payload to trigger an onchange event.
+    const subString = pastePayload.substr(0, pastePayload.length - 1);
+    const lastChar = pastePayload.slice(-1);
+    cy.wrap(selector).invoke("val", subString);
```
SupaJoon, line 75: I think we can get a better representation of copy and paste if we actually write to a clipboard and then paste. Cypress shows us an example of this here: https://github.com/cypress-io/cypress-example-recipes/blob/master/examples/testing-dom__clipboard/cypress/integration/spec.js#L79

SupaJoon (changes requested):

cypress/integration/spawn/volume.ts
```diff
@@ -154,12 +154,6 @@ describe("Navigating to Spawn Volume page", () => {
     cy.contains(errorBannerCopy).should("exist");
   });
 
-  it("Switching tabs should clear the error toast.", () => {
```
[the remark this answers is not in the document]
khelif96: This behavior existed when we we used banners. But since the toasts are persistent in a context they exist throughout the app and disappear on their own over time. This test always failed initially and would only pass when it was automatically restarted by cypress. (which refreshed the page and made the toast go away)

cypress/integration/version/action_buttons.ts
```diff
@@ -26,19 +26,25 @@ describe("Action Buttons", () => {
     });
 
     it("Clicking ellipses dropdown shows ellipses options", () => {
+      cy.dataCy("card-dropdown").should("not.exist");
+
+      cy.dataCy("ellipses-btn").should("not.be.visible");
       cy.dataCy("ellipsis-btn").click();
+      cy.dataCy("card-dropdown").should("be.visible");
       cy.dataCy("card-dropdown").should("exist");
```
sophstad, line 34: Same note as Arjun made below

cypress/integration/version/action_buttons.ts
```diff
@@ -52,7 +58,7 @@ describe("Action Buttons", () => {
         cy.dataCy("unschedule-patch").click();
         cy.dataCy("abort-checkbox").check({ force: true });
         cy.get(popconfirmYesClassName).contains("Yes").click({ force: true });
-        cy.dataCy("toast").should("exist");
+        cy.dataCy("toast").contains("Success!").should("exist");
```
sophstad, line 61: It may be preferable to include the variant type in the toast's data-cy prop since we're often checking for a string to verify the type. No worries if you'd rather not tackle that here though!

cypress/support/commands.ts
khelif96: I tried this it doesn't seem to work in chrome. 
That file states that this method only works in electron. 
> // access to the clipboard reliably works in Electron browser
// in other browsers, there are popups asking for permission
// thus we should only run these tests in Electron

SupaJoon (approved):

cypress/integration/version/action_buttons.ts
khelif96: done!

cypress/support/commands.ts
```diff
@@ -67,10 +56,31 @@ Cypress.Commands.add("getInputByLabel", (label) =>
...
+  }
+);
+
+Cypress.Commands.add("validateToast", (status: string, message?: string) => {
+  cy.dataCy(`toast`).should("be.visible");
+  cy.dataCy("toast").should("have.attr", "data-variant", status);
+  if (message) {
+    cy.dataCy(`toast`).contains(message);
+  }
+});
```
sophstad, line 86: Perfect

sophstad (approved):

PR #1013, opened 2021-12-15
EVG-15080: Add Variables page

sophstad:
[EVG-<number>](https://jira.mongodb.org/browse/EVG-<number>)

### Description 

### Screenshots

### Testing 

### Evergreen PR 

### DIFF
--- a/src/components/ProjectSettingsTabs/VariablesTab/VariableRow.tsx
+++ b/src/components/ProjectSettingsTabs/VariablesTab/VariableRow.tsx
@@ -4,34 +4,21 @@ import { ObjectFieldTemplateProps } from "@rjsf/core";
 
 const { yellow } = uiColors;
 
-export const VariableRowFieldTemplate: React.FC<ObjectFieldTemplateProps> = ({
-  formData,
-  properties,
-  uiSchema,
-}) => {
-  const repoData = uiSchema?.options?.repoData;
+export const VariableRow: React.FC<
+  Pick<ObjectFieldTemplateProps, "formData" | "properties" | "uiSchema">
+> = ({ formData, properties, uiSchema }) => {
   const [variableName, variableValue, isPrivate] = properties;
-  if (!repoData) {
-    return (
-      <RowContainer>
-        <LeftColumn>{variableName.content}</LeftColumn>
-        <div>
-          {variableValue.content}
-          {isPrivate.content}
-        </div>
-      </RowContainer>
-    );
-  }
-
-  const { vars } = repoData;
-  const inRepo = vars.some(({ name }) => name === formData.name);
+  const repoData = uiSchema?.options?.repoData;
+  const inRepo = repoData
+    ? repoData.vars.some(({ varName }) => varName === formData.varName)
+    : false;
 
   return (
     <RowContainer>
-      <LeftColumn>
+      <LeftColumn showWarning={inRepo}>
         {variableName.content}
         {inRepo && (
-          <span>
+          <span data-cy="override-warning">
             This will override the variable of the same name defined in the
             repo.
           </span>
@@ -48,6 +35,12 @@ export const VariableRowFieldTemplate: React.FC<ObjectFieldTemplateProps> = ({
 const LeftColumn = styled.div`
   color: ${yellow.dark2};
   padding-right: 16px;
+
+  ${(props: { showWarning?: boolean }): string =>
+    props.showWarning &&
+    `input {
+    border-color: ${yellow.dark2};
+  }`}
 `;
 
 const RowContainer = styled.div`
--- a/src/components/ProjectSettingsTabs/VariablesTab/getFormSchema.ts
+++ b/src/components/ProjectSettingsTabs/VariablesTab/getFormSchema.ts
@@ -3,9 +3,9 @@ import { SpruceFormProps } from "components/SpruceForm";
 import { CardFieldTemplate } from "components/SpruceForm/FieldTemplates";
 import widgets from "components/SpruceForm/Widgets";
 import { FormState } from "./types";
-import { VariableRowFieldTemplate } from "./VariableRowFieldTemplate";
+import { VariableRow } from "./VariableRow";
 
-export const getFormData = (
+export const getFormSchema = (
   useRepoSettings: boolean,
   repoData?: FormState
 ): {
@@ -21,15 +21,15 @@ export const getFormData = (
         items: {
           type: "object" as "object",
           properties: {
-            name: {
+            varName: {
               type: "string" as "string",
               title: "Variable Name",
             },
-            value: {
+            varValue: {
               type: "string" as "string",
               title: "Variable",
             },
-            private: {
+            isPrivate: {
               type: "boolean" as "boolean",
               title: "Private",
             },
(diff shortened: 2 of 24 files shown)


src/components/ProjectSettingsTabs/VariablesTab/getFormSchema.ts
```diff
@@ -0,0 +1,100 @@
...
+              type: "string" as "string",
+              title: "Variable Name",
+            },
+            varValue: {
+              type: "string" as "string",
+              title: "Variable",
+            },
+            isPrivate: {
+              type: "boolean" as "boolean",
+              title: "Private",
```
malikchaya2, line 34: Are we planning to add the "tooltip explaining what private means" in the tooltips ticket?

malikchaya2 (commented):
image
It doesn't look like this functionality is implemented. Will this be done separately?

src/components/ProjectSettingsTabs/utils.ts
```diff
@@ -40,7 +40,6 @@ export const radioBoxOptions = (
 ];
 
 export const placeholderIf = (element: string | number) =>
-  element !== null &&
```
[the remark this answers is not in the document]
sophstad: Yeah, there's a valid case where the variable in the repo _is_ null, and we still want to show it as a placeholder. Checking strictly for `undefined` means that we'll only display the placeholder if the repo value actually does not exist (i.e., the conditionally chained `repoData?.variableName` argument returns `undefined`).

src/components/ProjectSettingsTabs/VariablesTab/getFormSchema.ts
sophstad: I didn't see any language for this in the [language Google Doc](https://docs.google.com/document/d/16o6CeDF2Z_7S9SZNouN00Spuio-pF8628zlIPLz34oU/edit), is this correct @ablack12?

sophstad:
> image
> It doesn't look like this functionality is implemented. Will this be done separately?

@malikchaya2 I posted a screenshot producing this in the description, you should be able to emulate it by saving a variable for the repo [here](http://localhost:3000/project/602d70a2b2373672ee493184/settings/access) and then adding the same variable to the evergreen project [here](http://localhost:3000/project/evergreen/settings/access). Let me know if that doesn't work!

src/components/ProjectSettingsTabs/VariablesTab/getFormSchema.ts
ablack12: Good catch -- I think we can use the same description as the legacy ("Users must be logged in to see this project.")

ablack12: Good catch -- should be something like "Private variables have redacted values on the Project Page and the API and cannot be modified" (although does that make it sound like they also can't be deleted?)

sophstad: @ablack12 Hmmm perhaps slightly. Maybe "Private variables have redacted values on the Project Page and the API and cannot be _updated_"?

ablack12: Yes I think that's better!

src/components/ProjectSettingsTabs/VariablesTab/transformers.ts
```diff
@@ -0,0 +1,43 @@
...
+      vars.privateVarsList.push(varName);
+    }
+    vars.vars[varName] = varValue;
+  });
+
+  return {
+    projectRef: { id },
+    vars,
+  };
+};
```
khelif96, line 43:
```suggestion
export const formToGql: FormToGqlFunction = (
  { vars: varsData }: FormState,
  id
) => {
  const vars = varsData.reduce(
    (acc, { varName, varValue, isPrivate }) => {
      if (!varName) return acc;
      if (isPrivate) {
        acc.privateVarsList.push(varName);
      }
      acc.vars[varName] = varValue;
      return acc;
    },
    {
      vars: {},
      privateVarsList: [],
    }
  );

  return {
    projectRef: { id },
    vars,
  };
};
```
image

lets try to minimize our use of type assertions since we lose  the type safety features typescript provides.

src/components/ProjectSettingsTabs/VariablesTab/VariableRow.tsx
```diff
@@ -0,0 +1,68 @@
...
+const RowContainer = styled.div`
+  display: flex;
+  margin-bottom: 16px;
+
+  > div {
+    flex-grow: 1;
+    max-width: 50%;
+  }
+
+  > div > div > div {
```
khelif96, line 65: What are these nested selectors used for?
Could we instead just  apply the styles to the elements we are intending to target.  Any change to the underlying structure of any of these elements will cause this styling to break.

src/components/ProjectSettingsTabs/VariablesTab/VariableRow.tsx
```diff
@@ -0,0 +1,68 @@
...
+    </RowContainer>
+  );
+};
+
+const OverrideWarning = styled.span`
+  svg {
+    margin-right: 4px;
+    margin-top: 1px;
+    vertical-align: text-top;
+  }
```
khelif96, line 42: Could we apply this style to the leafygreen Icon instead?

khelif96 (changes requested):
Noticed 2 issues when it comes to saving variables.
The first trying to save a variable that does not have a value does not error out and instead succeeds but the variable wont be visible on refresh. 

Second saving a variable and then navigating to a different tab and coming back doesn't show it. The only way to see it is by refreshing the page.  

https://user-images.githubusercontent.com/4605522/146606368-5fd23ee0-4e33-40fe-bf00-2235fe85ac35.mov


Now that the functionality of these front end and back end functionality of these pages is becoming more defined we should write some cypress tests to capture it.

ablack12 (approved):
Didn't code read but works correctly to me from a usage standpoint 🚀

malikchaya2 (approved):
🎉

sophstad:
@khelif96 

> trying to save a variable that does not have a value does not error out and instead succeeds but the variable wont be visible on refresh

Thanks, the save button is now disabled until both fields are populated.

> saving a variable and then navigating to a different tab and coming back doesn't show it

Good catch, this appears related to Apollo's caching layer. I think a reasonable approach would be to use `refetchQueries` instead of updating the cache directly. Caching will get more complex as the data and mutations scale and I think we're already coming up against this. The docs [here](https://www.apollographql.com/docs/react/data/mutations/#supported-methods) make a good case for the simplicity of `refetchQueries`, and I doubt users will click save so often that the added network requests affect performance. Another complication is that we'd need to use a custom a keyField equal to projectRef.id in the cache, and nested key fields are currently not supported by Apollo.  Let me know what you think!

src/components/ProjectSettingsTabs/VariablesTab/VariableRow.tsx
```diff
@@ -2,13 +2,38 @@ import styled from "@emotion/styled";
...
+
+// Return child fields to be rendered
+// Conditionally disable based on whether it has been flagged as such (i.e. is a private variable that has already been saved).
+const getFields = (
+  properties: ObjectFieldTemplateProps["properties"],
+  isDisabled: boolean
+): Array<Unpacked<ObjectFieldTemplateProps["properties"]>["content"]> =>
+  isDisabled
+    ? properties.map(disableField)
+    : properties.map(({ content }) => content);
```
khelif96, line 28: These seem like useful utilities i wonder if we will/need use them again for some of the other tabs?  If so it might make sense to preemptively place them in a shared utils folder.

khelif96 (approved):
LGTM with one comment/suggestion 🚀

PR #1029, opened 2022-01-06
EVG-15532: Add Plugins Page 

malikchaya2:
[EVG-15532](https://jira.mongodb.org/browse/EVG-15532)

### Description 
Added the Plugins page. 

### Screenshots
image
image

### Testing 
Add unit test. To preview locally when running a dev server using the latest main branch of evergreen:

- http://localhost:3000/project/evergreen/settings/plugins to view a project that inherits from a repo
- http://localhost:3000/project/spruce/settings/plugins to view a project that does not inherit from a repo
- http://localhost:3000/project/602d70a2b2373672ee493184/settings/plugins to view a repository's settings

### Evergreen PR
https://github.com/evergreen-ci/evergreen/pull/5325


src/components/ProjectSettingsTabs/PluginsTab/getFormSchema.ts
```diff
@@ -0,0 +1,182 @@
...
+  schema: {
+    type: "object" as "object",
+    properties: {
+      performanceSettings: {
+        type: "object" as "object",
+        title: "Performance Plugins",
+        properties: {
+          perfEnabled: {
+            type: ["boolean", "null"],
+            title: "Build-break notifications",
```
sophstad, line 26: [nit] I've been capitalizing the headings for these forms
```suggestion
            title: "Build-break Notifications",
```

src/components/ProjectSettingsTabs/PluginsTab/getFormSchema.ts
```diff
@@ -0,0 +1,182 @@
...
+            ),
+          },
+        },
+      },
+      buildBaronSettings: {
+        type: "object" as "object",
+        title: "Build Baron",
+        properties: {
+          ticketSearchProjects: {
+            type: ["array", "null"],
```
sophstad, line 39: Since there's no option for this field to default to repo value, `null` is not supported
```suggestion
            type: "array" as "array",
```

src/components/ProjectSettingsTabs/PluginsTab/getFormSchema.ts
```diff
@@ -0,0 +1,182 @@
...
+              properties: {
+                searchProject: {
+                  type: "string" as "string",
+                  title: "Search Project",
+                },
+              },
+            },
+          },
+          customTicket: {
+            type: ["boolean", "null"],
```
sophstad, line 52: Same thought as above
```suggestion
            type: "boolean" as "boolean",
```

src/components/ProjectSettingsTabs/PluginsTab/getFormSchema.ts
```diff
@@ -0,0 +1,182 @@
...
+          taskAnnotationSettings: {
+            title: "Custom Ticket Creation",
+            type: "object" as "object",
+            properties: {
+              fileTicketWebhook: {
+                type: "object" as "object",
+                title: "Webhook",
+                properties: {
+                  endpoint: {
+                    type: ["string", "null"],
```
sophstad, line 71: Same here

src/components/ProjectSettingsTabs/PluginsTab/getFormSchema.ts
```diff
@@ -0,0 +1,182 @@
...
+          useRepoSettings,
+        },
+      },
+      customTicket: {
+        "ui:widget": widgets.RadioBoxWidget,
+        "ui:showLabel": false,
+        "ui:data-cy": "enabled-radio-box",
+      },
+      ticketCreateProject: {
+        "ui:disabled": formData?.buildBaronSettings.customTicket === true,
```
sophstad, line 132: What do you think about using LeafyGreen's [segmented control](https://www.mongodb.design/component/segmented-control/example/) component to display the correct fields based on whether a user has chosen "custom ticket creation" or "build baron ticket creation"? I'm using it right now for some fields on the GitHub settings page and while it requires more logic on our end, I think it's a pretty effective pattern for displaying form fields like this. Interested in your thoughts!

src/components/ProjectSettingsTabs/PluginsTab/transformers.ts
```diff
@@ -0,0 +1,79 @@
...
+    performanceSettings: {
+      perfEnabled: projectRef.perfEnabled,
+    },
+    buildBaronSettings: {
+      ticketCreateProject: projectRef.buildBaronSettings.ticketCreateProject,
+      ticketSearchProjects: projectRef.buildBaronSettings.ticketSearchProjects
+        ? projectRef.buildBaronSettings.ticketSearchProjects.map(
+            (searchProject) => ({ searchProject })
+          )
+        : [],
```
sophstad, line 27: [nit] Just a lil syntactic sugar
```suggestion
      ticketSearchProjects:
        projectRef?.buildBaronSettings?.ticketSearchProjects?.map(
            (searchProject) => ({ searchProject })
          ) ?? [],
```

src/components/ProjectSettingsTabs/PluginsTab/transformers.ts
```diff
@@ -0,0 +1,79 @@
...
+    taskAnnotationSettings: {
+      fileTicketWebhook: {
+        endpoint: projectRef.taskAnnotationSettings.fileTicketWebhook.endpoint,
+        secret: projectRef.taskAnnotationSettings.fileTicketWebhook.secret,
+      },
+      jiraCustomFields: projectRef.taskAnnotationSettings.jiraCustomFields
+        ? projectRef.taskAnnotationSettings.jiraCustomFields.map(
+            ({ field, displayText }) => ({ field, displayText })
+          )
+        : [],
```
sophstad, line 39: Same as above
```suggestion
      jiraCustomFields:
        projectRef?.taskAnnotationSettings?.jiraCustomFields?.map(
          ({ field, displayText }) => ({ field, displayText })
        ) ?? [],
```

src/components/ProjectSettingsTabs/PluginsTab/transformers.ts
```diff
@@ -0,0 +1,79 @@
...
+      perfEnabled: projectRef.perfEnabled,
+    },
+    buildBaronSettings: {
+      ticketCreateProject: projectRef.buildBaronSettings.ticketCreateProject,
+      ticketSearchProjects: projectRef.buildBaronSettings.ticketSearchProjects
+        ? projectRef.buildBaronSettings.ticketSearchProjects.map(
+            (searchProject) => ({ searchProject })
+          )
+        : [],
+      customTicket: projectRef.buildBaronSettings.ticketCreateProject == null,
```
sophstad, line 28:
```suggestion
      customTicket: projectRef.buildBaronSettings.ticketCreateProject === null,
```

src/pages/projectSettings/getTabTitle.ts
```diff
@@ -33,6 +33,9 @@ export const getTabTitle = (
       [ProjectSettingsTabRoutes.PeriodicBuilds]: {
         title: "Periodic Builds",
       },
+      [ProjectSettingsTabRoutes.Plugins]: {
+        title: "Plugins & Annotations",
+      },
```
sophstad, line 38: The designs have "Plugins" last in the sidebar, just wanted to ensure this ordering is intentional!

sophstad (changes requested):

ablack12:
I'll take a look at this in the UI when conflicts are resolved; in the meantime, should the functionality of "only Custom Ticket Creation or Build Baron" is still accurate after the recent changes. 

Also, I think a lot of this wording isn't super clear (it looks like the design doesn't contain everything though). Calling one of the options Build Baron Ticket Creation is confusing when the whole section is labeled Build Baron; I think the "Build Baron" wording itself is outdated since only Evergreen still knows why the labeling is split like that. It seems like each of these sections could use some help text in the new UI, as opposed to just Ticket Create Project.

src/pages/projectSettings/getTabTitle.ts
malikchaya2: It felt more natural to me to have Event Log last

malikchaya2:
> I'll take a look at this in the UI when conflicts are resolved; in the meantime, should the functionality of "only Custom Ticket Creation or Build Baron" is still accurate after the recent changes.
> 
> Also, I think a lot of this wording isn't super clear (it looks like the design doesn't contain everything though). Calling one of the options Build Baron Ticket Creation is confusing when the whole section is labeled Build Baron so maybe the overall section should have a different label. It seems like each of these sections could use some help text in the new UI, as opposed to just Ticket Create Project.

The  "only Custom Ticket Creation or Build Baron" setup is still accurate, because it refers to what happens when the file ticket button is clicked, and it either files the ticket the traditional way or with a webhook. The recent changes only affect the "created tickets" list. 

I'll think about how I can change the sections and help text to make things clearer.

malikchaya2:
@ablack12 conflicts are resolved now

src/components/ProjectSettingsTabs/PluginsTab/getFormSchema.ts
malikchaya2: In the linked example, nothing changes when I toggle to a different fruit... hmm...

malikchaya2: This doesn't look very different than the radio buttons/toggles we are currently using. How would using this component change things?

malikchaya2: custom
)

ablack12: I think design was pretty against having fields be hidden based on toggles so we may want to run this by them.

malikchaya2: @sophstad Since you plan to use this design pattern as well, should we bring it up at the next design sync?

sophstad: Sure, although the use case for alias definitions is a little different and use of a toggle is already established in their designs. We should probably ask about this page specifically!

malikchaya2: sounds good, I'll add it to the agenda

ablack12: @malikchaya2 should custom jira fields only be available for JIRA ticket creation?

ablack12: Gotcha; Can Custom Jira Fields and Ticket Search Project be together then, since they're both configurable regardless of the toggle? (either at the top or bottom)

ablack12 (commented):
Not sure where to actually add my comments in the code but:
Why doesn't this section have any options for defaulting to repo? The repo has the option to set things so.
image

Also, how do we specify "I don't want to have any kind of ticket creation?" (should just leaving anything unpopulated work?) It seems like JIRA Ticket Creation is allowed to be always empty, but Custom Ticket Creation returns a vague error, even if I populate the webhook.
image

ablack12:
Also can these trash cans be in line?
image

src/components/ProjectSettingsTabs/PluginsTab/getFormSchema.ts
ablack12: > Gotcha; Can Custom Jira Fields and Ticket Search Project be together then, since they're both configurable regardless of the toggle? (either at the top or bottom)

@malikchaya2 can this be addressed?

malikchaya2: It is set up for assigned teams which is what they originally wanted the custom fields for. In order to populate assigned teams, it would need to be added as a custom field. 
https://github.com/evergreen-ci/spruce/blob/main/src/pages/task/taskTabs/buildBaronAndAnnotations/BBComponents.tsx#L88

malikchaya2: I thought about that, but decided against it because it will probably only be used by one team. I didn't want it to be too visible. Thoughts?

ablack12: I think it's okay to move even if most teams won't use it; if anything it's more visible where it is because users need to consider whether or not it's required for Custom Ticket / JIRA ticket creation.

malikchaya2: that's a good point, I'll move it

malikchaya2: I realized that you can't actually use `ticket search project` with custom ticket creation, so I will move that over.

ablack12:
I still don't seem able to save a project with build baron options (but the reorganization looks good to me!)
image

src/components/ProjectSettingsTabs/PluginsTab/getFormSchema.ts
```diff
@@ -0,0 +1,189 @@
...
+              [
+                "Build Baron Ticket Search and Create",
+                "Custom Ticket Creation",
+              ],
+              undefined
+            ),
+          },
+          ticketSearchProjects: {
+            type: "array" as "array",
+            title: "Ticket Search Projects",
```
ablack12, line 74: In the old UI it's required that Ticket Search Projects and Ticket Create Project both be populated if one is populated. Should that be maintained?

src/components/ProjectSettingsTabs/PluginsTab/getFormSchema.ts
```diff
@@ -0,0 +1,189 @@
...
+              createProject: {
+                type: "string" as "string",
+                title: "",
+              },
+            },
+          },
+
+          fileTicketWebhook: {
+            type: "object" as "object",
+            title: "Custom Ticket Creation",
```
ablack12, line 98: Likewise, if one of Endpoint/Secret is populated, should the other be required?

src/components/ProjectSettingsTabs/PluginsTab/getFormSchema.ts
```diff
@@ -0,0 +1,189 @@
...
+                items: {
+                  type: "object" as "object",
+                  properties: {
+                    field: {
+                      type: "string" as "string",
+                      title: "Field",
+                    },
+                    displayText: {
+                      type: "string" as "string",
+                      title: "Display Text",
```
ablack12, line 55: Should display text be required if Field is populated?

ablack12 (changes requested):

src/components/ProjectSettingsTabs/PluginsTab/getFormSchema.ts
malikchaya2: I was undecided about that. I don't see why one should depend on the other since one is used for the file ticket webhook, while the other is used to populate the "related tickets" list, they don't seem to be dependent on each other. However, since nobody has asked to be able to do one without the other, we have no reason to break that pattern now.

src/components/ProjectSettingsTabs/PluginsTab/getFormSchema.ts
malikchaya2: I don't personally think so. Especially since nobody is using it yet and we don't know what their setup is (and if they'll even use a secret). A secret without an endpoint would be nonsensical, but since we are assuming that it'll be used by intelligent users, I don't know if we need to block saving due to it, especially since it won't harm anything. I'm open to suggestions though.

src/components/ProjectSettingsTabs/PluginsTab/getFormSchema.ts
malikchaya2: I think it would be more of a nice to have. I don't think it needs to be required. Especially since it's very possible that some fields will have human-readable names (like "epic" instead of custom_1234).

src/components/ProjectSettingsTabs/PluginsTab/getFormSchema.ts
malikchaya2: If we do indeed decide to make it not required, we would need a one-line change to the backend function that checks is the build baron is configured.

src/components/ProjectSettingsTabs/PluginsTab/getFormSchema.ts
ablack12: I'd personally prefer to block it; I think it's out pattern to block users from doing nonsensical things -- a user can't save a variable without a key for example, I believe (cc @sophstad to verify).

src/components/ProjectSettingsTabs/PluginsTab/getFormSchema.ts
ablack12: I'm concerned that making it not required might also confuse other places that assume one exists without the other (the task page for example?) so that may be worth making as a different ticket but I defer to your decision here.

malikchaya2: right, that's why we would need to change the function that checks if the build baron is configured, which is what the other pages use. 
We have no reason to change it though, so I'll re-implement making them both required.

src/components/ProjectSettingsTabs/PluginsTab/getFormSchema.ts
sophstad: Correct, blocking save is the tack we've generally been taking so would make sense to continue here. This could be done using a custom validation function like we did for the variables page [here](https://github.com/evergreen-ci/spruce/blob/main/src/components/ProjectSettingsTabs/VariablesTab/VariablesTab.tsx#L53)

ablack12 (approved):

src/components/ProjectSettingsTabs/PluginsTab/transformers.ts
```diff
@@ -0,0 +1,86 @@
...
+      jiraCustomFields: buildBaronSettings.taskAnnotationSettings?.jiraCustomFields
+        .map(({ field, displayText }) => ({ field, displayText }))
+        .filter((str) => !!str),
+    },
+  };
+
+  return { projectRef };
+};
+
+export const buildBaronIf = (useBuildBaron: boolean, buildBaronSettings: any) =>
```
sophstad, line 64: A quick comment explaining this function would be appreciated!

src/components/ProjectSettingsTabs/PluginsTab/transformers.ts
```diff
@@ -0,0 +1,86 @@
...
+    buildBaronSettings: {
+      ticketCreateProject:
+        buildBaronSettings.ticketCreateProject?.createProject,
+      ticketSearchProjects: buildBaronSettings.ticketSearchProjects
+        .map(({ searchProject }) => searchProject)
+        .filter((str) => !!str),
+    },
+  };
+
+export const fileTicketWebhookIf = (
```
sophstad, line 76: Same here

src/components/SpruceForm/FieldTemplates.tsx
```diff
@@ -31,6 +31,7 @@ export const ArrayFieldTemplate: React.FC<ArrayFieldTemplateProps> = ({
   const id = idSchema.$id;
   const description = uiSchema["ui:description"] || schema.description;
   const buttonText = uiSchema["ui:buttonText"] || "Add";
+  const buttonDisabled = uiSchema["ui:buttonDisabled"] || false;
```
sophstad, line 34: This approach can be removed from this PR; I don't see it used and I added logic in #1050 so that the "Add" button is disabled if the `"ui:disabled"` option is set.

src/components/ProjectSettingsTabs/PluginsTab/getFormSchema.ts
```diff
@@ -0,0 +1,189 @@
...
+        type: "object" as "object",
+        title: "Build Baron and Task Annotations",
+        properties: {
+          taskAnnotationSettings: {
+            title: "",
+            type: "object" as "object",
+            properties: {
+              jiraCustomFields: {
+                type: "array" as "array",
+                title: "Custom Jira Fields",
```
sophstad, line 45: [nit] Standardize since we're using "JIRA" elsewhere on the page
```suggestion
                title: "Custom JIRA Fields",
```

src/components/ProjectSettingsTabs/PluginsTab/getFormSchema.ts
```diff
@@ -0,0 +1,189 @@
...
+      },
+    },
+    buildBaronSettings: {
+      "ui:rootFieldId": "buildBaron",
+      "ui:ObjectFieldTemplate": CardFieldTemplate,
+      taskAnnotationSettings: {
+        "ui:rootFieldId": "taskAnnotation",
+        jiraCustomFields: {
+          "ui:description":
+            "Add any custom Jira fields that you want displayed on any listed JIRA tickets, for example: assigned teams.",
```
sophstad, line 129:
```suggestion
            "Add any custom JIRA fields that you want displayed on any listed JIRA tickets, for example: assigned teams.",
```

src/components/ProjectSettingsTabs/PluginsTab/getFormSchema.ts
```diff
@@ -0,0 +1,189 @@
...
+          ...placeholderIf(
+            repoData?.projectPluginsSettings?.taskAnnotationSettings
+              ?.jiraCustomFields
+          ),
+          "ui:buttonText": "Add custom Jira field",
```
sophstad, line 134:
```suggestion
          "ui:buttonText": "Add custom JIRA field",
```

sophstad (approved):
🥳

src/components/SpruceForm/FieldTemplates.tsx
malikchaya2: Ah yeah it's left over from when custom jira fields was grouped with ticket creation. removed.

PR #1031, opened 2022-01-10
EVG-16063 Remove unused deps 

khelif96:
[EVG-16063](https://jira.mongodb.org/browse/EVG-16063)

### Description 
Removes some unused deps. 
as well as adds some deps to the package.json. Some of these did not explicitly exist in the package.json but only worked since they were peer dependencies of some other deps.


malikchaya2 (approved):

malikchaya2 (commented):
It looks like this is not on top of the latest main branch. I'd suggest merging main before merging.

PR #1034, opened 2022-01-11
EVG-13867 Wire up requester selector and Dropdown updates

khelif96:
[EVG-13867](https://jira.mongodb.org/browse/EVG-<number>)

### Description 
Created a reusable DropdownButton component which looks similar to the leafygreen select. Which serves as a basis for other Dropdowns we have.  I've also updated everywhere i can find where we use the old dropdown design 

Deprecated the dropdown component that comes bundled with TreeSelect. TreeSelect will now only be a TreeSelect
Added some tests for Treeselect 

Can select a specific requester on the waterfall page.

### Screenshots
image
image
image


src/utils/queryString/parseQueryString.ts
```diff
@@ -9,8 +9,5 @@ export const parseQueryString = (search: string): ParseQueryString =>
 export const getString = (param: string | string[] = ""): string =>
   Array.isArray(param) ? param[0] : param;
 
-export const getArray = (param: string | string[] = []): string[] =>
```
khelif96, line 12: `toArray` does the same thing and is more generic instead of focusing on just queryStrings

khelif96 (commented):
Sorry about the large PR

src/pages/commits/commitTypeSelect/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+    </Container>
+  );
+};
+
+const Container = styled.div`
+  display: flex;
+  flex-direction: column;
+`;
+
+const noFilterMessage = "No filters selected";
```
malikchaya2, line 75: It would probably be a good idea to defined this on the src/pages/commits level so that it can be shared between this file and StatusSelect

malikchaya2 (approved):
Approved with one suggestion

src/components/Dropdown/index.tsx
```diff
@@ -0,0 +1,143 @@
...
+  extends Omit<DropdownProps, "isOpen" | "setIsOpen"> {
+  ref?: React.Ref<DropdownWithRef>;
+}
+
+interface DropdownWithRefState {
+  isOpen: boolean;
+}
+/** DropdownWithRef is a class component that allows the implementer
+ *  to control its internal state methods with a ref in order to trigger state updates */
+class DropdownWithRef extends React.Component<
```
sophstad, line 76: Just to clarify, why exactly must this component be a class?

sophstad (approved):

src/components/Dropdown/index.tsx
khelif96: I wanted a way for a  component implementing this to be able to close the dropdown. Without giving up the state control this component has. In order to do that i wanted to expose a method an implementer component could use to trigger a state change. 
To do this i needed to use a ref to expose the components internal methods. Functional components don't support this functionality.

src/pages/commits/commitTypeSelect/index.tsx
khelif96: i moved this to a `constants/strings` directory. Since we also have filters outside of the commits page

PR #1040, opened 2022-01-12
EVG-15280 Consolidate task colors

khelif96:
[EVG-15280](https://jira.mongodb.org/browse/EVG-15280)

### Screenshots
image
image


src/components/TaskStatusBadge/index.tsx
```diff
@@ -70,19 +53,54 @@ const TaskStatusBadge: React.FC<TaskStatusBadgeProps> = ({ status }) => {
...
+  [TaskStatus.Started]: Variant.Yellow,
+  [TaskStatus.Dispatched]: Variant.Yellow,
+  [TaskStatus.Failed]: Variant.Red,
+  [TaskStatus.TestTimedOut]: Variant.Red,
+  [TaskStatus.TaskTimedOut]: Variant.Red,
+  [TaskStatus.Succeeded]: Variant.Green,
+  [TaskStatus.KnownIssue]: Variant.Red,
+  [TaskStatus.WillRun]: Variant.DarkGray,
+};
+const customBadgeColors = (status: string) => {
```
sophstad, line 84: This would require updating a few components, but I think it would be ideal to have `status` typed as a `TaskStatus`. I believe if you change [this line](https://github.com/evergreen-ci/spruce/blob/main/src/pages/Task.tsx#L104) to read

```js
<TaskStatusBadge status={TaskStatus[status]} />
```

and update its child components, we should be able to use `TaskStatus` safely.

src/utils/statuses/groupStatusesByUmbrellaStatus.ts
```diff
@@ -26,14 +27,15 @@ export const groupStatusesByUmbrellaStatus = (
     const umbrellaStatus = mapTaskStatusToUmbrellaStatus[stat.status];
     if (counts[umbrellaStatus]) {
       counts[umbrellaStatus].count += stat.count;
-      if (!counts[umbrellaStatus].statuses.includes(stat.status)) {
-        counts[umbrellaStatus].statuses.push(taskStatusToCopy[stat.status]);
-      }
+      counts[umbrellaStatus].statuses = deduplicatedAppend(
```
sophstad, line 30: Would making `ColorCount.statuses` a Set of strings eliminate the need for this method?

src/components/TaskStatusBadge/index.tsx
khelif96: `TaskStatus[status]` doesn't actually work. 
`status` is the value returned from the server which represents a value in the `TaskStatus` enum so we can't actually use at as a key for it. 

```
TaskStatus["success"] === undefined 
TaskStatus.Success === "success"
```

src/utils/statuses/groupStatusesByUmbrellaStatus.ts
khelif96: Discussed offline but commenting for visibility. The implementations of this method require an array so the implementers would need to cast the set into an array in order to use this field. Which makes implementing this method more complex. We opted instead to use a set within `deduplicatedAppend`  and have it return a deduped array instead.

src/utils/array/index.ts
```diff
@@ -18,6 +18,18 @@ export const toggleArray = <T>(value: T, array: T[]) => {
...
 };
 
+/**
+ * deduplicatedAppend takes in an array of values regardless of type and a new value and safely inserts the value if it doesn't exist in the array.
+ * if it does exist it will do nothing
+ * @param value The value to insert into the array.
+ * @param array The array to insert the value into.
+ */
+export const deduplicatedAppend = <T>(value: T, array: T[]) => {
+  let tempArray = new Set([...array]);
```
sophstad, line 28: `[...array]` = `array`
```suggestion
  let tempArray = new Set(array);
```

sophstad (approved):

PR #1045, opened 2022-01-18
EVG-16130 Update yarn lock

khelif96:
[EVG-16130](https://jira.mongodb.org/browse/EVG-16130)

### Description 
This would get regenerated every time i ran yarn which means it was probably not committed the first time.


minnakt (approved):
Just ran it myself and saw the same changes 🙂👊🔥

PR #1048, opened 2022-01-18
EVG-14429: Navigate to task tab when clicking on build variant badges

minnakt:
[EVG-14429](https://jira.mongodb.org/browse/EVG-14429)

### Description 
This PR makes it so that if a user is on a different tab like `Changes` or `Downstream Tasks` on the Patch page, clicking on a task status badge will bring them back to the `Tasks` tab (with the correct filters applied).

### Screenshots
https://user-images.githubusercontent.com/47064971/150021885-6b75986a-0d33-43e3-830b-cf9061bc7911.mov

### Testing 
- Edited e2e test in `version_route.ts`

### DIFF
--- a/cypress/integration/version/version_route.ts
+++ b/cypress/integration/version/version_route.ts
@@ -90,14 +90,14 @@ describe("Version route", () => {
       // click on a different tab first, so that we aren't on the task tab initially
       cy.dataCy("changes-tab").first().click();
       cy.dataCy("task-tab")
-        .invoke("attr", "aria-selected")
-        .should("equal", "false");
+        .should("have.attr", "aria-selected")
+        .and("equal", "false");
 
       // clicking on task status badge should move to the task tab
       cy.dataCy("grouped-task-status-badge").first().click();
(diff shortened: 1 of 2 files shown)


cypress/integration/version/version_route.ts
```diff
@@ -87,7 +87,17 @@ describe("Version route", () => {
     });
 
     it("Navigates to task tab and applies filters when clicking on grouped task status badge", () => {
+      // click on a different tab first, so that we aren't on the task tab initially
+      cy.dataCy("changes-tab").first().click();
+      cy.dataCy("task-tab")
+        .invoke("attr", "aria-selected")
```
sophstad, line 93: I'm not familiar with `invoke()`, but the docs suggest that it should be used to run a function. Perhaps the "should have" syntax used [here](https://github.com/evergreen-ci/spruce/blob/main/cypress/integration/version/unscheduled_patch/configure_patch.ts#L40-L44) could get the job done instead?

sophstad (approved):
💯

PR #1049, opened 2022-01-19
EVG-16119: Linkify & trim column headers on Variant/Task History pages

minnakt:
[EVG-16119](https://jira.mongodb.org/browse/EVG-16119)

### Description 
This PR linkifies the column headers on the variant and task history pages. It also trims column headers by inserting an ellipsis (`...`) if a variant or task name exceeds a given length. If a variant or task name has been trimmed, a tooltip displaying the full name will appear when hovering over it; otherwise, no tooltip appears.

Also adds a small CSS fix to solve a visual bug that was occurring on Safari.

### Screenshots
https://user-images.githubusercontent.com/47064971/150229673-5e561fbf-2cc0-425a-a580-592f27eef34c.mov

### Testing 
- Added unit tests for `ColumnHeaders.tsx` in `pages/variantHistory` and `pages/taskHistory` folders
- Added a unit test for new function `trimMiddleText` in `utils/string/string.test.ts`

### DIFF
--- a/src/components/HistoryTable/Cell.tsx
+++ b/src/components/HistoryTable/Cell.tsx
@@ -3,6 +3,7 @@ import { uiColors } from "@leafygreen-ui/palette";
 import Tooltip from "@leafygreen-ui/tooltip";
 import { Skeleton } from "antd";
 import { Link } from "react-router-dom";
+import { ConditionalWrapper } from "components/ConditionalWrapper";
 import { inactiveElementStyle, StyledRouterLink } from "components/styles";
 import { getTaskRoute } from "constants/routes";
 import { TaskStatus } from "types/task";
@@ -77,20 +78,21 @@ export const ColumnHeaderCell: React.FC<ColumnHeaderCellProps> = ({
   fullDisplayName,
 }) => (
   <HeaderCell data-cy="header-cell">
-    {trimmedDisplayName === fullDisplayName ? (
-      <StyledRouterLink to={link}>{fullDisplayName}</StyledRouterLink>
-    ) : (
-      <Tooltip
-        align="top"
-        justify="middle"
-        trigger={
-          <StyledRouterLink to={link}>{trimmedDisplayName}</StyledRouterLink>
-        }
-        triggerEvent="hover"
-      >
-        {fullDisplayName}
-      </Tooltip>
-    )}
+    <ConditionalWrapper
+      condition={trimmedDisplayName !== fullDisplayName}
+      wrapper={(children) => (
+        <Tooltip
+          align="top"
+          justify="middle"
+          trigger={children}
+          triggerEvent="hover"
+        >
+          {fullDisplayName}
+        </Tooltip>
+      )}
+    >
+      <StyledRouterLink to={link}>{trimmedDisplayName}</StyledRouterLink>
+    </ConditionalWrapper>
   </HeaderCell>
 );
 
--- a/src/components/HistoryTable/Cell.tsx
+++ b/src/components/HistoryTable/Cell.tsx
@@ -113,7 +113,7 @@ const Cell = styled.div<{ inactive?: boolean }>`
 `;
 
 export const HeaderCell = styled(Cell)`
+  word-break: break-all; // Safari
   word-wrap: anywhere;
-  word-break: break-word; // Safari
   text-align: center;
 `;
--- a/src/components/HistoryTable/Cell.tsx
+++ b/src/components/HistoryTable/Cell.tsx
@@ -1,8 +1,9 @@
 import styled from "@emotion/styled";
 import { uiColors } from "@leafygreen-ui/palette";
+import Tooltip from "@leafygreen-ui/tooltip";
 import { Skeleton } from "antd";
 import { Link } from "react-router-dom";
-import { inactiveElementStyle } from "components/styles";
+import { inactiveElementStyle, StyledRouterLink } from "components/styles";
 import { getTaskRoute } from "constants/routes";
 import { TaskStatus } from "types/task";
 import { HistoryTableIcon } from "./HistoryTableIcon";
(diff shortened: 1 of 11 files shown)


src/components/HistoryTable/Cell.tsx
```diff
@@ -71,6 +113,7 @@ const Cell = styled.div<{ inactive?: boolean }>`
 `;
 
 export const HeaderCell = styled(Cell)`
+  word-break: break-all; // Safari
   word-wrap: anywhere;
   text-align: center;
```
minnakt, line 118: On Safari before:
Screen Shot 2022-01-19 at 5 30 21 PM

On Safari now:
Screen Shot 2022-01-19 at 5 30 08 PM

src/pages/variantHistory/ColumnHeaders.tsx
```diff
@@ -1,17 +1,27 @@
...
+import { array, string } from "utils";
 
 const { mapStringArrayToObject } = array;
+const { trimMiddleText } = string;
 const { useHistoryTable } = context;
-const { HeaderCell } = Cell;
+const { LoadingCell, ColumnHeaderCell } = Cell;
+
+const maxLength = 50;
+const trailingLength = 10;
```
minnakt, line 12: I decided these values arbitrarily for both pages ... 🙂 Open to hearing other ways to trim the name

src/utils/string/index.ts
```diff
@@ -167,3 +167,24 @@ export const applyStrictRegex = (str: string) => `^${str}$`;
...
+ * @param {number} maxLength - Max length before trimming text
+ * @param {number} trailingLength - Desired length of trailing text
+ * @return {string} The original or trimmed text.
+ */
+export const trimMiddleText = (
+  text: string,
+  maxLength: number,
+  trailingLength: number
+): string => {
+  if (text.length > maxLength) {
```
khelif96, line 183: This should account for the additional ellipsis otherwise we could end up with a string as long or longer as the original string.

src/pages/taskHistory/ColumnHeaders.test.tsx
```diff
@@ -0,0 +1,176 @@
...
+};
+
+interface wrapperProps {
+  children: React.ReactNode;
+  mocks?: MockedResponse[];
+  state?: Partial<HistoryTableReducerState>;
+}
+
+const wrapper: React.FC<wrapperProps> = ({ children, mocks = [], state }) => (
+  <MockedProvider mocks={mocks}>
```
khelif96, line 38: Is the MockedProvider actually needed? it doesn't seem like we are actually mocking any queries.

src/components/HistoryTable/Cell.tsx
```diff
@@ -46,10 +47,51 @@ export const EmptyCell = () => (
...
+        align="top"
+        justify="middle"
+        trigger={
+          <StyledRouterLink to={link}>{trimmedDisplayName}</StyledRouterLink>
+        }
+        triggerEvent="hover"
+      >
+        {fullDisplayName}
+      </Tooltip>
+    )}
```
khelif96, line 93: We have a ConditionalWrapper component that you can probably use for something like this. https://github.com/evergreen-ci/spruce/blob/main/src/components/ConditionalWrapper/index.tsx

src/pages/variantHistory/ColumnHeaders.test.tsx
```diff
@@ -0,0 +1,156 @@
...
+};
+
+interface wrapperProps {
+  children: React.ReactNode;
+  mocks?: MockedResponse[];
+  state?: Partial<HistoryTableReducerState>;
+}
+
+const wrapper: React.FC<wrapperProps> = ({ children, mocks = [], state }) => (
+  <MockedProvider mocks={mocks}>
```
khelif96, line 38: Same comment as the other Test file.

khelif96 (changes requested):
Looks good. Just had a few comments

src/utils/string/index.ts
```diff
@@ -167,3 +167,24 @@ export const applyStrictRegex = (str: string) => `^${str}$`;
...
+    // eslint-disable-next-line
```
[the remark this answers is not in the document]
minnakt: Sorry it's something I forgot to remove 🙏 It doesn't affect the code underneath it in any way

src/pages/taskHistory/ColumnHeaders.test.tsx
minnakt: Ah no it's not needed 🙂 removing

src/pages/variantHistory/ColumnHeaders.tsx
minnakt, line 12: I think I should move these out considering I also use them in the test files... Is `src/constants/history.ts` (new file) a good place to do this?

src/utils/string/index.ts
```diff
@@ -167,3 +167,27 @@ export const applyStrictRegex = (str: string) => `^${str}$`;
...
+  text: string,
+  maxLength: number,
+  trailingLength: number
+): string => {
+  const ellipsis = "...";
+  if (text.length > maxLength) {
+    return (
+      text.substring(0, maxLength - trailingLength - ellipsis.length) +
+      ellipsis +
+      text.substring(text.length - trailingLength)
```
khelif96, line 189: This still doesn't technically trim the middle of the string. It seems biased towards the head of the string.  (As seen in the test case below) 
I think ideally it would try to trim from the middle
`task_name -> ta...me` instead of `t...ame` 

It might also make sense to simplify the function a little bit so its easier to implement. 
One way we could do this 

```
const trimStringFromMiddle = (str, maxLength) => {
  if (str.length <= maxLength) {
    return str;
  }
  const midpoint = Math.floor(maxLength / 2);
  return `${str.substr(0, midpoint)}...${str.substr(str.length - midpoint)}`;
}
```

minnakt: Hmm yeah I guess I didn't name this function really accurately, it's more like `trimTextToFixedTrailingLength`... 🙂 I'll try again and write the true middle cropping function

src/utils/string/index.ts
```diff
@@ -167,3 +167,29 @@ export const applyStrictRegex = (str: string) => `^${str}$`;
...
+
+  const midpoint = Math.floor(text.length / 2);
+  const frontOffset = Math.floor(numCharsToRemove / 2);
+  const backOffset = Math.ceil(numCharsToRemove / 2);
+  return (
+    text.substring(0, midpoint - frontOffset) +
+    ellipsis +
+    text.substring(midpoint + backOffset)
+  );
+};
```
minnakt, line 195: Okay, how about this 🙂

src/utils/string/index.ts
```diff
@@ -167,3 +167,29 @@ export const applyStrictRegex = (str: string) => `^${str}$`;
...
+
+/**
+ * Function that trims the middle portion of a string. ex: "EvergreenUI" -> "Ev...UI"
+ * The resulting length, if trimmed, is maxLength + 3 (due to ellipsis length).
+ * @param {string} text - Text to trim
+ * @param {number} maxLength - Max length before trimming text
+ * @return {string} The original or trimmed text.
+ */
+export const trimStringFromMiddle = (text: string, maxLength: number) => {
+  const ellipsis = "...";
```
sophstad, line 179: This is a good use case for the [ellipsis character](https://unicode-table.com/en/2026/): `…`. And since you cleverly used `ellipsis.length` below, it should be an easy change!

src/pages/taskHistory/ColumnHeaders.test.tsx
```diff
@@ -0,0 +1,184 @@
...
+  currentPage: 0,
+  pageCount: 0,
+  columns: [],
+  historyTableFilters: [],
+  commitCount: 10,
+  visibleColumns: [],
+  columnLimit: 7,
+};
+
+interface wrapperProps {
```
sophstad, line 32: [nit] Types and interfaces are capitalized
```suggestion
interface WrapperProps {
```

khelif96 (approved):

sophstad (approved):

PR #1050, opened 2022-01-19
EVG-16153: Update SpruceForm ui elements

sophstad:
EVG-16153

### Description 
Break up some of the UI work from EVG-15081. This PR includes general changes that are not specific to the GitHub/Commit Queue page, but are used by this page
- Improve support for `null` type fields
- Better array customization
- Add new `sectionTitle` flag to expand form's text hierarchy
- Support hiding input labels by using LeafyGreen's text input's `ariaLabelledBy` prop

### Evergreen PR 
- Update unit tests & snapshot

### DIFF
--- a/src/components/ProjectSettingsTabs/Context.tsx
+++ b/src/components/ProjectSettingsTabs/Context.tsx
@@ -7,11 +7,7 @@ import { ProjectSettingsTabRoutes } from "constants/routes";
 /* Given a diff between a form's previous and current state, check if it represents an array field added with no content. */
 const isArrayPushUpdate = (diff: object): boolean =>
   Object.values(diff).every((val) => {
-    if (val === null) {
-      return false;
-    }
-
-    if (typeof val !== "object") {
+    if (val === null || typeof val !== "object") {
       return false;
     }
     // An empty object represents a new array element (i.e. blank input field)
--- a/src/components/ProjectSettingsTabs/Context.tsx
+++ b/src/components/ProjectSettingsTabs/Context.tsx
@@ -7,6 +7,10 @@ import { ProjectSettingsTabRoutes } from "constants/routes";
 /* Given a diff between a form's previous and current state, check if it represents an array field added with no content. */
 const isArrayPushUpdate = (diff: object): boolean =>
   Object.values(diff).every((val) => {
+    if (val === null) {
+      return false;
+    }
+
     if (typeof val !== "object") {
       return false;
     }
@@ -73,9 +77,7 @@ const reducer = (state: TabState, action: Action): TabState => {
 interface ProjectSettingsState {
   tabs: TabState;
   saveTab: (tab: ProjectSettingsTabRoutes) => void;
-  getTab: (
-    tab: ProjectSettingsTabRoutes
-  ) => TabState[ProjectSettingsTabRoutes.General]; // TODO: update type as all tabs are implemented
+  getTab: (tab: ProjectSettingsTabRoutes) => FormDataProps;
   updateForm: (
     tab: ProjectSettingsTabRoutes,
     save?: boolean
--- a/src/components/SpruceForm/CustomFields.tsx
+++ b/src/components/SpruceForm/CustomFields.tsx
@@ -1,13 +1,22 @@
 import styled from "@emotion/styled";
-import { Description, Subtitle } from "@leafygreen-ui/typography";
-import { FieldProps } from "@rjsf/core";
+import { Description, H3, Subtitle } from "@leafygreen-ui/typography";
+import { Field } from "@rjsf/core";
 
-export const TitleField: React.FC<FieldProps> = ({ id, title }) => (
-  <>
-    {/* @ts-expect-error  */}
-    <StyledSubtitle id={id}>{title}</StyledSubtitle>
-  </>
-);
+export const TitleField: Field = ({ id, isSectionTitle, title }) => {
+  const Component = isSectionTitle ? StyledH3 : StyledSubtitle;
+  return (
+    <>
+      {/* @ts-expect-error  */}
+      <Component id={id}>{title}</Component>
+    </>
+  );
+};
+
+/* @ts-expect-error  */
+const StyledH3 = styled(H3)`
+  margin-top: 24px;
+  margin-bottom: 12px;
+`;
 
 /* @ts-expect-error  */
 const StyledSubtitle = styled(Subtitle)`
@@ -15,8 +24,12 @@ const StyledSubtitle = styled(Subtitle)`
   margin-bottom: 12px;
 `;
 
-export const DescriptionField: React.FC<FieldProps> = ({ id, description }) => (
-  <StyledDescription id={id}>{description}</StyledDescription>
+export const DescriptionField: Field = ({ id, description }) => (
+  <>
+    {description && (
+      <StyledDescription id={id}>{description}</StyledDescription>
+    )}
+  </>
 );
 
 const StyledDescription = styled(Description)`
--- a/src/components/SpruceForm/CustomFields.tsx
+++ b/src/components/SpruceForm/CustomFields.tsx
@@ -1,8 +1,15 @@
 import styled from "@emotion/styled";
 import { Description, H3, Subtitle } from "@leafygreen-ui/typography";
-import { Field } from "@rjsf/core";
+import { Field, FieldProps } from "@rjsf/core";
 
-export const TitleField: Field = ({ id, isSectionTitle, title }) => {
+type TitleFieldProps = Pick<FieldProps, "id" | "title" | "uiSchema">;
+
+export const TitleField: React.FC<TitleFieldProps> = ({
+  id,
+  title,
+  uiSchema,
+}) => {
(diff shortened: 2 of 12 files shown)


src/components/SpruceForm/FieldTemplates.tsx
```diff
@@ -30,9 +91,12 @@ export const ArrayFieldTemplate: React.FC<ArrayFieldTemplateProps> = ({
...
-  const buttonText = uiSchema["ui:buttonText"] || "Add";
+  const addButtonSize = uiSchema["ui:addButtonSize"] || "small";
+  const addButtonText = uiSchema["ui:addButtonText"] || "Add";
   const fullWidth = !!uiSchema["ui:fullWidth"];
   const showLabel = uiSchema["ui:showLabel"] !== false;
+  const topAlignDelete = uiSchema["ui:topAlignDelete"] ?? false;
+  const isDisabled = disabled || readonly;
   return (
     <>
       {showLabel && (
```
minnakt, line 102: (This is for line `103` not `102` but I can't highlight that line 😭)

I think `TitleField` here is missing the new props `isSectionTitle`? I see it `undefined` when I try to print it. Alternatively could make `isSectionTitle` false by default in `CustomFields.tsx `😊

Screen Shot 2022-01-20 at 10 21 45 AM

src/components/SpruceForm/CustomFields.tsx
```diff
@@ -1,22 +1,35 @@
...
-    <StyledSubtitle id={id}>{title}</StyledSubtitle>
-  </>
-);
+export const TitleField: Field = ({ id, isSectionTitle, title }) => {
+  const Component = isSectionTitle ? StyledH3 : StyledSubtitle;
+  return (
+    <>
+      {/* @ts-expect-error  */}
+      <Component id={id}>{title}</Component>
+    </>
```
minnakt, line 11: nit: it might be more readable if we named this `Component` variable something like `TitleFieldTag` or `TitleTag`

src/components/ProjectSettingsTabs/Context.tsx
```diff
@@ -7,6 +7,10 @@ import { ProjectSettingsTabRoutes } from "constants/routes";
 /* Given a diff between a form's previous and current state, check if it represents an array field added with no content. */
 const isArrayPushUpdate = (diff: object): boolean =>
   Object.values(diff).every((val) => {
+    if (val === null) {
+      return false;
+    }
+
     if (typeof val !== "object") {
       return false;
     }
```
minnakt, line 16: Maybe we can condense these two `if` statements into one since they both return `false`?  But I don't know if that would hurt readability 🤔 Your call!

src/components/SpruceForm/FieldTemplates.tsx
```diff
@@ -41,67 +105,42 @@ export const ArrayFieldTemplate: React.FC<ArrayFieldTemplateProps> = ({
...
-                  />
-                </DeleteButtonWrapper>
-              )}
-            </ArrayItemRow>
-          )
-        )}
+      <ArrayContainer fullWidth={fullWidth} hasChildren={!!items?.length}>
+        {items.map((p) => (
+          <ArrayItem key={p.key} topAlignDelete={topAlignDelete} {...p} />
+        ))}
```
minnakt, line 124: Sweet refactor 😲🔥

src/components/SpruceForm/CustomFields.tsx
```diff
@@ -1,22 +1,35 @@
...
+  );
+};
+
+/* @ts-expect-error  */
+const StyledH3 = styled(H3)`
+  margin-top: 24px;
+  margin-bottom: 12px;
```
khelif96, line 18: Can we intervals of 8px for margins so we can replace them with the leafygreen tokens.

src/components/SpruceForm/CustomFields.tsx
```diff
@@ -1,22 +1,35 @@
...
+`;
 
 /* @ts-expect-error  */
 const StyledSubtitle = styled(Subtitle)`
   margin-top: 16px;
   margin-bottom: 12px;
```
khelif96, line 24: same comment

src/components/SpruceForm/FieldTemplates.tsx
```diff
@@ -6,22 +6,83 @@ import {
...
   hidden,
-}) => !hidden && <div className={classNames}>{children}</div>;
+  id,
+  label,
+  schema,
+  uiSchema,
+}) => {
+  const isNullType = schema.type === "null";
+  // fields is incorrectly typed in FieldTemplateProps, so cast to avoid type error
+  const { TitleField } = fields as any;
```
khelif96, line 27: Is there something else we can cast this too so we don't completely lose typesafety?

src/components/SpruceForm/FieldTemplates.tsx
```diff
@@ -41,67 +105,42 @@ export const ArrayFieldTemplate: React.FC<ArrayFieldTemplateProps> = ({
...
 };
 
+type ArrayContainerProps = {
+  hasChildren: boolean;
+  fullWidth?: boolean;
+};
+
 const ArrayContainer = styled.div`
-  margin-bottom: 24px;
+  ${({ hasChildren }) => hasChildren && "margin-buttom: 24px;"}
```
khelif96, line 136: typo.
```suggestion
  ${({ hasChildren }) => hasChildren && "margin-bottom: 24px;"}
```

src/components/SpruceForm/FieldTemplates.tsx
```diff
@@ -2,26 +2,90 @@ import styled from "@emotion/styled";
...
-}) => !hidden && <div className={classNames}>{children}</div>;
+  id,
+  label,
+  schema,
+  uiSchema,
+}) => {
+  const isNullType = schema.type === "null";
+  // fields is incorrectly typed in FieldTemplateProps, so cast to avoid type error
+  const { TitleField } = (fields as unknown) as {
+    TitleField: React.StatelessComponent<Partial<FieldProps>>;
```
khelif96, line 29: This method has been deprecated. 
image

Also im a little confused on why the library is incorrectly typing this value.  The library thinks this should be an array so why are we treating it as an object?
Are we sure this works? Minnas comment below about the field being undefined makes me question if we are accessing the value properly.

src/components/SpruceForm/LeafyGreenWidgets.tsx
```diff
@@ -38,7 +43,9 @@ export const LeafyGreenTextInput: React.FC<WidgetProps> = ({
         <TextInput
           data-cy={dataCy}
           value={value === null || value === undefined ? null : `${value}`}
-          label={label}
+          // @ts-ignore
```
khelif96, line 46: nit. Can we use `@ts-expect-error` instead. If leafygreen every fixes this we will get an alert and we can remove the  statement.

src/components/SpruceForm/SpruceForm.test.tsx
```diff
@@ -35,12 +48,21 @@ describe("spruceForm", () => {
     fireEvent.change(queryByDataCy("valid-projects-input"), {
       target: { value: "new value" },
     });
+    await fireEvent.click(queryByDataCy("add-button"));
```
khelif96, line 51: await is unnecessary with fireEvent

src/components/SpruceForm/FieldTemplates.tsx
sophstad: The library types it as an array but it really is an object! Here's the result of console logging `fields`:
image
 [This](https://github.com/rjsf-team/react-jsonschema-form/blob/master/packages/core/index.d.ts#L26) instance of `fields` is correctly typed but for some reason the `FieldTemplateProps` property isn't. I also opted to use a StatelessComponent because that is how `Field` is typed by the library ([here](https://github.com/rjsf-team/react-jsonschema-form/blob/master/packages/core/index.d.ts#L152)). I could type it as a ClassComponent if you prefer.

The `undefined` Minna referenced above just had to do with the prop not being passed in every usage of the component, it's not related to this issue.

khelif96: The [Docs indicate](https://react-jsonschema-form.readthedocs.io/en/latest/advanced-customization/custom-templates/#fieldtemplate) that fields  is indeed an array in a FieldTemplate 

The linked instance is a [prop on the Form type](https://react-jsonschema-form.readthedocs.io/en/latest/api-reference/form-props/#fields)  which is a different type of attribute. 

Is it possible we are incorrectly using this component and it is actually not a FieldTemplate but instead another type of field?

sophstad: No, it's passed in via the `FieldTemplate` prop [here](https://github.com/evergreen-ci/spruce/blob/main/src/components/SpruceForm/index.tsx#L35) so the usage is as expected for a FieldTemplate. I'm pretty convinced that this is an issue on the library side, but implemented a different approach to avoid it altogether. Let me know what you think!

khelif96 (approved):

minnakt (approved):

PR #1052, opened 2022-01-20
EVG-16173 Escape regex characters on history table links

khelif96:
[EVG-16173](https://jira.mongodb.org/browse/EVG-16173)

### Description 
Since the history table applies a regex search and the tests table often has test names with characters that can be mistaken for regex we need to escape them before using them in the history table otherwise we won't be able to match on them.

### DIFF
--- a/src/utils/string/index.ts
+++ b/src/utils/string/index.ts
@@ -163,6 +163,14 @@ export const sortFunctionDate = (a, b, key) => {
 export const applyStrictRegex = (str: string) => `^${str}$`;
 
 /**
+ *
+ * @param str - A string that may contain regex operators.
+ * @return {string} A regex that matches on the input.
+ */
+export const escapeRegex = (str: string) =>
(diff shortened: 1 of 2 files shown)


src/utils/string/index.ts
```diff
@@ -162,6 +162,14 @@ export const sortFunctionDate = (a, b, key) => {
  */
 export const applyStrictRegex = (str: string) => `^${str}$`;
 
+/**
+ *
+ * @param str - A string that may contain regex operators.
+ * @return {string} A regex that matches on the input.
+ */
+export const escapeRegex = (str: string) =>
```
sophstad, line 170: [nit] Could include the return type
```suggestion
export const escapeRegex = (str: string): string =>
```

sophstad (approved):

src/utils/string/index.ts
khelif96: image
It gets inferred by typescript :)

PR #1053, opened 2022-01-20
EVG-15081: Add GitHub Page

sophstad:
EVG-15081

### Description
- Add GitHub page
-  Still to do:
    - "Git Tag Version Definitions" section will be completed in EVG-16117 as there was design required
    - "Patch Trigger Aliases" section will be completed in EVG-15084, as the "Aliases" page is where the data is actually managed

### Screenshots
image
image
image

### Testing 
Add integration tests. To preview locally when running a dev server using the latest main branch of evergreen:

- http://localhost:3000/project/evergreen/settings/access to view a project that inherits from a repo
- http://localhost:3000/project/spruce/settings/access to view a project that does not inherit from a repo
- http://localhost:3000/project/602d70a2b2373672ee493184/settings/access to view a repository's settings


src/components/ProjectSettingsTabs/GithubCommitQueue/getFormSchema.ts
```diff
@@ -0,0 +1,350 @@
...
+            oneOf: radioBoxOptions(
+              ["Enabled", "Disabled"],
+              repoData?.github?.prTestingEnabled
+            ),
+          },
+          prTesting: {
+            type: "object" as "object",
+            title: "GitHub Patch Definitions",
+            description:
+              "For patches created from GitHub pull requests, Evergreen will find variants that match each variant regex, and schedule tasks that match the corresponding task regex. All regular expressions must be valid Golang regular expressions.",
```
ablack12, line 94: I realize I wrote this text 🤦‍♀️  but it should probably say "regex/tag list". We could also tweak this to be more like the later definitions:
For patches created from GitHub pull requests, Evergreen will schedule only the tasks and variants matching the tags/regexes definitions. All regular expressions must be valid Golang regular expressions.

src/components/ProjectSettingsTabs/GithubCommitQueue/getFormSchema.ts
```diff
@@ -0,0 +1,350 @@
...
+            title: "GitHub Pull Request Testing",
+            ...(useRepoSettings === undefined && {
+              description:
+                "If enabled, then untracked branches will also use the file patterns defined here for PR testing.",
+            }),
+          },
+          prTestingEnabled: {
+            type: ["boolean", "null"],
+            oneOf: radioBoxOptions(
+              ["Enabled", "Disabled"],
```
ablack12, line 86: Wanted to call out that this behaves slightly different than the below (disabling Pull Request Testing doesn't prevent you from modifying github aliases).
image

src/components/ProjectSettingsTabs/GithubCommitQueue/getFormSchema.ts
```diff
@@ -0,0 +1,350 @@
...
+                "Default to Repo Patch Definition",
+              ],
+              {
+                $ref: "#/definitions/aliasArray",
+              }
+            ),
+          },
+          githubChecksEnabledTitle: {
+            type: "null",
+            title: "GitHub Commit Checks",
```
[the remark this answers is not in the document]
sophstad: Makes sense! To clarify for a project that is attached to a repo: should we only provide this warning if neither the project nor the repo has aliases defined?

src/components/ProjectSettingsTabs/GithubCommitQueue/getFormSchema.ts
```diff
@@ -0,0 +1,350 @@
...
+    type: "object" as "object",
+    properties: {
+      github: {
+        type: "object" as "object",
+        title: "GitHub",
+        properties: {
+          gitHubWebhooksEnabled: {
+            type: "null",
+            title: "GitHub Webhooks",
+            description: `GitHub webhooks ${
```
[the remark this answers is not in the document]
ablack12: I think that message should be tweaked slightly since Git Tag Versions and Github Checks also aren't available. 
Maybe the messaging could either give the full list or say "Github Features"? (The latter would probably only be reasonable if the content was disabled and not hidden, since the user may not know what it is they're missing out on.)

Possibly we could also add a disclaimer "Please reach out to #evergreen-users if you are still experiencing issues.

src/components/ProjectSettingsTabs/GithubCommitQueue/getFormSchema.ts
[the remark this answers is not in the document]
ablack12: If the project chooses "Override Repo Configuration" then it doesn't matter if the repo has aliases defined so we can still show the message.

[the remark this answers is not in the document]
ablack12: Actually also: There is now a [VersionConfigEnabled boolean](https://github.com/evergreen-ci/evergreen/blob/8a425038834326c212d65289e0c9e80e48d07e7e/rest/model/project.go#L466) (Malik created https://jira.mongodb.org/browse/EVG-16185 to add this). It would be awesome if this showed as an _error_ if version control is not enabled, and a warning if it is. In which case maybe this validation should be tacked onto that ticket.

src/components/ProjectSettingsTabs/GithubCommitQueue/getFormSchema.ts
```diff
@@ -0,0 +1,350 @@
...
+          },
+          prTestingEnabledTitle: {
+            type: "null",
+            title: "GitHub Pull Request Testing",
+            ...(useRepoSettings === undefined && {
+              description:
+                "If enabled, then untracked branches will also use the file patterns defined here for PR testing.",
+            }),
+          },
+          prTestingEnabled: {
```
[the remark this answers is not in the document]
sophstad: Discussed offline, EVG-16194 and EVG-16195 have been created to address this issue

src/components/ProjectSettingsTabs/AliasRow.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+export const AliasRow: SpruceFormProps["ObjectFieldTemplate"] = ({
+  disabled,
+  formData,
+  idSchema,
+  properties,
+  readonly,
+  uiSchema,
+}) => {
+  const [, alias, variant, variantTags, task, taskTags] = properties;
```
malikchaya2, line 38: Is the lone comma at the beginning intentional?

sophstad: Yes! It's [array destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#array_destructuring) in which the zeroth element is ignored.

src/components/SpruceForm/FieldTemplates.tsx
```diff
@@ -130,7 +130,7 @@ type ArrayContainerProps = {
 };
 
 const ArrayContainer = styled.div`
-  ${({ hasChildren }) => hasChildren && "margin-bottom: 24px;"}
+  ${({ hasChildren }) => hasChildren && "margin-buttom: 24px;"}
```
minnakt, line 133: seems like this typo came back?!  🤔

src/components/ProjectSettingsTabs/AliasRow.tsx
```diff
@@ -0,0 +1,159 @@
...
+                disabled={isDisabled}
+                aria-controls={`${VariantInput.Regex}-field`}
+              >
+                Regex
+              </SegmentedControlOption>
+            </StyledSegmentedControl>
+            {variantInput === VariantInput.Tags ? (
+              <div id={`${VariantInput.Tags}-field`}>{variantTags.content}</div>
+            ) : (
+              <div id={`${VariantInput.Regex}-field`}>{variant.content}</div>
```
minnakt, line 106: Is there any reason these `id`s can't be `classNames`? Or else there's a lot of elements with the same non-unique `id` (also applies to lines `135-137`)

src/components/ProjectSettingsTabs/GithubCommitQueue/GithubCommitQueueTab.tsx
```diff
@@ -0,0 +1,96 @@
...
+const pairHasError = (regex: string, tags: string[]) => {
+  const hasInvalidTags =
+    !tags || !tags?.length || tags.every((tag) => tag === "");
+  return regex === "" && hasInvalidTags;
+};
+
+const aliasHasError = ({ task, taskTags, variant, variantTags }: AliasType) =>
+  pairHasError(task, taskTags) || pairHasError(variant, variantTags);
+
+const validate = (formData: FormState, errors) => {
```
minnakt, line 73: Can add return types here -- `boolean` for the first two functions... honestly not sure for the last one, I guess just leave it if it's not easy to figure out 😅

src/components/ProjectSettingsTabs/AliasRow.tsx
```diff
@@ -0,0 +1,159 @@
...
+        taskTags.content.props.onChange([]);
+        break;
+      case TaskInput.Tags:
+        task.content.props.onChange("");
+        break;
+      default:
+    }
+  };
+
+  return (
```
minnakt, line 70: Would it make sense to default to the `Tags` tab if the regex field is empty? (At least for this page, all of the `SegmentedControl`s default to the `Regex` tab and I was kind of confused why it was empty until I clicked on it)

https://user-images.githubusercontent.com/47064971/151233514-db5add5a-d4c0-4516-8db5-4bf312480162.mov

src/components/ProjectSettingsTabs/AliasRow.tsx
malikchaya2: nice!

src/components/ProjectSettingsTabs/AliasRow.tsx
```diff
@@ -0,0 +1,159 @@
...
+          {alias.content}
+          <TaskRegexContainer>
+            <StyledSegmentedControl
+              label="Variant"
+              id="variant-input-control"
+              value={variantInput}
+              onChange={(value) => {
+                setVariantInput(value);
+                clearForm(value);
+              }}
```
malikchaya2, line 86: how would you feel about not clearing it between toggles? On the plugins page where there's a toggle between custom ticket creation versus default ticket creation, instead of clearing the form, I leave it there in case the user changes their mind and wants to come back to it, and instead make sure to only save the appropriate one when the form is saved.

sophstad: Definitely would prefer not clearing it—I went with this approach because it was significantly easier to implement 😅 I could try looking into it a bit more, or we could create a ticket as an improvement for later. Open to either!

malikchaya2: ah I definitely hear that! I would definitely suggest a separate ticket as opposed to looking into to it now :)

malikchaya2 (approved):

src/components/ProjectSettingsTabs/AliasRow.tsx
sophstad: EVG-16214

src/components/ProjectSettingsTabs/GithubCommitQueue/getFormSchema.ts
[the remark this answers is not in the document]
sophstad: EVG-16215 can handle validation for both cases.

cypress/integration/projectSettings/project_settings.ts
```diff
@@ -152,38 +201,60 @@ describe("Project Settings when defaulting to repo", () => {
...
-    cy.dataCy("save-settings-button").click();
-    cy.contains("Successfully updated repo");
+    it("Should enable save when the task and variant fields are filled in", () => {
+      cy.dataCy("variant-input").type(".*");
+      cy.get("#task-input-control").find("button").eq(1).click();
+      cy.dataCy("task-input").type(".*");
+      cy.dataCy("save-settings-button").should("not.be.disabled");
+    });
+
+    it("Disables Authorized Users section based on repo settings", () => {
```
ablack12, line 249: I would suggest  that these also hide when Trigger Versions With Git Tags is disabled (to be consistent with CQ and PR).

src/components/ProjectSettingsTabs/GithubCommitQueue/GithubCommitQueueTab.tsx
```diff
@@ -0,0 +1,117 @@
...
+      ),
+    [formData, gitHubWebhooksEnabled, repoData, useRepoSettings]
+  );
+
+  if (!formData) return null;
+
+  return (
+    <>
+      {!gitHubWebhooksEnabled && (
+        <Banner data-cy="disabled-webhook-banner" variant="warning">
```
ablack12, line 59: image

FYI this doesn't prevent the user from saving the page with Enabled set for these things, and it should be, but I'm not sure if that's what line 71 is about.

ablack12 (approved):
Left a couple of comments but no need to block merge on another review from me!

src/components/ProjectSettingsTabs/GithubCommitQueue/GithubCommitQueueTab.tsx
sophstad: Yeah, the entire page will be disabled if GitHub webhooks are not enabled once we fix the EVG-16208 bug. For the meantime I'm preventing disabling so that the page can be tested.

ablack12: Oh right as we literally discussed 🤦‍♀️

PR #1054, opened 2022-01-21
EVG-16120 Sticky column headers and refactor commits page

khelif96:
[EVG-16120](https://jira.mongodb.org/browse/EVG-16120)

### Description 
Refactored the commits page to have a more semantically correct html structure. Which now allows us to use the `position: sticky` css attribute.  

This also updates some sizes in preparation for EVG-16152
There is also a refactor of some storybook stories and some components to make it easier to render the project health view in a storybook story.
### Screenshots

https://user-images.githubusercontent.com/4605522/150557960-a1a713b0-19f3-41fe-9747-0d88048b3338.mov

### DIFF
--- a/src/pages/commits/ActiveCommits/ProjectHealth.stories.tsx
+++ b/src/pages/commits/ActiveCommits/ProjectHealth.stories.tsx
@@ -1,8 +1,7 @@
 import { useState } from "react";
 import { withKnobs, boolean, number } from "@storybook/addon-knobs";
 import StoryRouter from "storybook-react-router";
-import { MainlineCommitsQuery } from "gql/generated/types";
-import { ChartTypes } from "types/commits";
+import { ChartTypes, CommitVersion, Commit } from "types/commits";
 import { TaskStatus } from "types/task";
 import { CommitsWrapper } from "../CommitsWrapper";
 
@@ -33,12 +32,8 @@ export const ActualWaterfallPage = () => {
   );
 };
 
-type Commits = MainlineCommitsQuery["mainlineCommits"]["versions"];
-type Commit = Commits[0];
-type Version = Commit["version"];
-
 const buildVariantUpdateLength = (
-  buildVariant: Version["buildVariants"],
+  buildVariant: CommitVersion["buildVariants"],
   buildVariantCount: number
 ) => {
   if (buildVariant.length === buildVariantCount) {
@@ -57,7 +52,7 @@ const buildVariantUpdateLength = (
 };
 
 const taskUpdateLength = (
-  tasks: Version["buildVariants"][0]["tasks"],
+  tasks: CommitVersion["buildVariants"][0]["tasks"],
   taskCount: number
 ) => {
   if (tasks.length === taskCount) {
--- a/src/pages/commits/ActiveCommits/ProjectHealth.stories.tsx
+++ b/src/pages/commits/ActiveCommits/ProjectHealth.stories.tsx
@@ -7,7 +7,7 @@ import {
   FlexRowContainer,
   ProjectHealthWrapper,
 } from "../CommitsWrapper";
-import { InactiveCommits, InactiveCommitLine } from "../InactiveCommits/index";
+import InactiveCommits from "../InactiveCommits/index";
 import { Grid } from "./Grid";
 import { ActiveCommit } from "./index";
 import {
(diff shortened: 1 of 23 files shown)


src/pages/commits/CommitsWrapper.tsx
```diff
@@ -28,91 +36,130 @@ export const CommitsWrapper: React.FC<Props> = ({
...
+const getCommitWidth = (commit: Commit) => {
+  const { version, rolledUpVersions } = commit;
+  if (version) {
+    return 200;
+  }
+  if (rolledUpVersions) {
+    return 64;
+  }
+  throw new Error("Commit type not found");
+};
```
minnakt, line 127: It seems like `ChartContainer` in `CommitChart.tsx` is inheriting this `200px` width and it's causing a bug where the`Tooltip` is pretty far away from the actual bars in the graph (differs from current Spruce behavior)


Screen Shot 2022-01-21 at 4 51 47 PM

I don't know if this is the best solution but adding `width: fit-content` to the `ChartContainer` does fix this issue!

Screen Shot 2022-01-21 at 5 00 32 PM

src/pages/commits/CommitsWrapper.tsx
```diff
@@ -28,91 +36,130 @@ export const CommitsWrapper: React.FC<Props> = ({
...
+    }
+    return undefined;
+  }, [versions]);
+
+  const maxGroupedTaskStats = useMemo(() => {
+    if (versionToGroupedTaskStatsMap) {
+      return findMaxGroupedTaskStats(versionToGroupedTaskStatsMap);
+    }
+    return undefined;
+  }, [versionToGroupedTaskStatsMap]);
```
minnakt, line 53: are the `return undefined`s needed here? I removed them and observed the same behavior with or without, but I might have overlooked something

minnakt (commented):
I don’t have much to say but it looks great! I already thought it looked good before, but now it looks even better because the scrolling became really seamless. 🚀 The refactor makes a lot of sense considering the sticky functionality. 🙂

src/pages/commits/CommitsWrapper.tsx
```diff
@@ -28,91 +36,130 @@ export const CommitsWrapper: React.FC<Props> = ({
...
 
-export const ProjectHealthWrapper = styled.div`
-  display: flex;
-  flex-direction: column;
-  justify-content: flex-start;
-  align-items: flex-start;
-  height: 100%;
-  width: 100%;
-  position: relative;
+const CommitWrapper = styled.div<{ width: number }>`
```
sophstad, line 155: I like this syntax, much cleaner than including typing in the arrow function 👌

src/pages/commits/ActiveCommits/ProjectHealth.stories.tsx
```diff
@@ -1,77 +1,102 @@
...
+      hasTaskFilter={boolean("hasTaskFilter", false)}
+      hasFilters={boolean("hasFilters", false)}
+      onChangeChartType={setChartType}
+    />
+  );
+};
+
+type Commits = MainlineCommitsQuery["mainlineCommits"]["versions"];
+type Commit = Commits[0];
+type Version = Commit["version"];
```
sophstad, line 38: Should these be imported from `types/commits.ts`?

src/types/commits.ts
```diff
@@ -24,3 +26,8 @@ export enum CommitRequesterTypes {
   TriggerRequester = "trigger_request",
   GitTagRequester = "git_tag_request",
 }
+
+export type Commits = MainlineCommitsQuery["mainlineCommits"]["versions"];
+export type Commit = Commits[0];
```
sophstad, line 31: [nit] You could probably use the [Unpacked](https://github.com/evergreen-ci/spruce/blob/main/src/types/utils.ts#L2) helper instead

sophstad (approved):
LGTM once Minna's note about the tooltip position is addressed! I also think it could look nice to apply a small shadow to the bottom of the commit row, kinda like GitHub does:
image
But up to you if you want to implement that!

khelif96:
>I also think it could look nice to apply a small shadow to the bottom of the commit row, kinda like GitHub does: image But up to you if you want to implement that!

I looked into this unfortunately there isn't a clean way to toggle the box shadow based off of if the element is scrolled or not. One of the solutions i found would be to attach an event handler to the element and measure if its scrolled but at that point it becomes the solution i have outlined in the other pr. If design asks for it i can take a look at it. I do think it would be really neat if it did have the box shadow tho but it might not be worth the added complexity.

minnakt (approved):

PR #1055, opened 2022-01-21
EVG-16178 Add project triggers and move trigger types

malikchaya2:
[EVG-16178](https://jira.mongodb.org/browse/EVG-16178)

### Description 
Add the types needed by the notifications page, and move trigger types so a more re-usable location. 

### Testing 
existing tests


src/components/ProjectSettingsTabs/NotificationsTab/types.ts
```diff
@@ -0,0 +1,162 @@
...
+  buildRegexSelectors,
+  taskRegexSelectors,
+  requesterSubscriberConfig,
+  ResourceType,
+  Trigger,
+  failureTypeSubscriberConfig,
+} from "constants/triggers";
+import { validateDuration, validatePercentage } from "utils/validators";
+
+export const triggers: Trigger[] = [
```
khelif96, line 11: This should be in a constants file since this isn't a typescript type.

src/constants/triggers.ts
```diff
@@ -0,0 +1,89 @@
...
+
+type PayloadResourceIdKey = "in-version" | "in-build" | "id";
+export interface Trigger {
+  trigger: string;
+  label: string;
+  extraFields?: ExtraField[];
+  resourceType: ResourceType;
+  payloadResourceIdKey?: PayloadResourceIdKey;
+  regexSelectors?: RegexSelector[];
+}
```
khelif96, line 15: These should probably be defined in a types file since these are typescript types

khelif96 (approved):
Great Refactor! This will make another refactor i have planned of the notification modal easier. 

Just had 2 comments i think some variables and types might need to be swapped or moved to a more accurate file.

src/components/ProjectSettingsTabs/NotificationsTab/types.ts
```diff
@@ -0,0 +1,162 @@
...
+    extraFields: [requesterSubscriberConfig],
+  },
+  {
+    trigger: "failure",
+    resourceType: ResourceType.VERSION,
+    label: "Any version fails",
+    extraFields: [requesterSubscriberConfig],
+  },
+  {
+    trigger: "outcome",
```
ablack12, line 25: Can the triggers also be defined as constants (as well as the extra field Keys)? I think this would help make it clearer that evergreen relies on these fields being precise to build the notification (as opposed to label, for example, which is strictly a UI element). I realize that this isn't done in the legacy UI but I don't think that defining these as constants there was the right thing to do either.

An alternative would be to get this list of triggers from Evergreen/globals.go or notifications.go with a query, but I recognize that might be an unnecessary slowdown.

src/constants/triggers.ts
```diff
@@ -0,0 +1,89 @@
...
+
+export type RegexSelectorType = "display-name" | "build-variant";
+export interface RegexSelector {
+  type: RegexSelectorType;
+  typeLabel: string;
+}
+
+export const failureTypeSubscriberConfig = {
+  text: "Failure type",
```
ablack12, line 56: In some places we use capitalization across (ex: Build Variant Name) but not consistently. I'm sure this was a failure of the legacy UI but I think it would be good to make this consistent now -- personally I prefer capitalizing (Failure Type) but whatever works

src/components/ProjectSettingsTabs/NotificationsTab/types.ts
malikchaya2: good point, done.

src/constants/triggers.ts
malikchaya2: done

src/constants/triggers.ts
malikchaya2: done

src/components/ProjectSettingsTabs/NotificationsTab/types.ts
malikchaya2: done

src/components/PatchActionButtons/addNotification/PatchNotificationModal.tsx
```diff
@@ -93,25 +90,25 @@ export const triggers: Trigger[] = [
   {
     trigger: "outcome",
```
ablack12, line 91: Can this file be using triggerType.Outcome / etc?

src/constants/triggers.ts
```diff
@@ -0,0 +1,218 @@
...
+    trigger: TriggerType.FAILURE,
+    resourceType: ResourceType.BUILD,
+    label: "Any Build Fails",
+    regexSelectors: buildRegexSelectors,
+    extraFields: [requesterSubscriberConfig],
+  },
+  {
+    trigger: TriggerType.OUTCOME,
+    resourceType: ResourceType.TASK,
+    label: "Any task finishes",
```
ablack12, line 97: Any Task Finishes

src/constants/triggers.ts
```diff
@@ -0,0 +1,218 @@
...
+    trigger: TriggerType.FIRST_FAILURE_BUILD,
+    resourceType: ResourceType.TASK,
+    label: "The First Failure In Each Build Occurs",
+    regexSelectors: taskRegexSelectors,
+    extraFields: [requesterSubscriberConfig],
+  },
+  {
+    trigger: TriggerType.FIRST_FAILURE_VERSION_NAME,
+    resourceType: ResourceType.TASK,
+    label: "The First Failure In Each Version For Each Task Name Occurs",
```
ablack12, line 125: tiny nit: shouldn't capitalize "in" or "for" (also for the other labels)
https://www.webucator.com/article/how-to-capitalize-headings-and-titles/

ablack12 (approved):
LGTM just some small comments!

PR #1060, opened 2022-01-26
EVG-14849: Add Project Patches link in Navbar dropdown

minnakt:
[EVG-14849](https://jira.mongodb.org/browse/EVG-14849)

### Description 
This PR adds a `Project Patches` link in the navbar dropdown based on a user's most recently selected project. 

Information about a user's most recently selected project is stored in the cookie `mci-project-cookie`. If a user does not have this cookie, then this link defaults to `mongodb-mongo-master` (alternatively `evergreen` if running entirely locally).

### Screenshots
https://user-images.githubusercontent.com/47064971/151207754-9a26309f-c385-4bc9-b0f2-f57b146e4947.mov

### Testing 
- Added tests in `nav_bar.ts` (e2e)

### DIFF
--- a/cypress/integration/nav_bar.ts
+++ b/cypress/integration/nav_bar.ts
@@ -50,14 +50,14 @@ describe("Nav Bar", () => {
       .should("have.attr", "href")
       .and("include", LEGACY_URLS.distros);
   });
-  it("Nav Dropdown should link to patches page of most recent project if cookie exists", () => {
+  it.skip("Nav Dropdown should link to patches page of most recent project if cookie exists", () => {
     cy.setCookie("mci-project-cookie", "spruce");
     cy.visit(SPRUCE_URLS.userPatches);
     cy.dataCy("auxiliary-dropdown-link").click();
     cy.dataCy("auxiliary-dropdown-project-patches").click();
     cy.location("pathname").should("eq", "/project/spruce/patches");
   });
-  it("Nav Dropdown should link to patches page of default project in SpruceConfig if cookie does not exist", () => {
+  it.skip("Nav Dropdown should link to patches page of default project in SpruceConfig if cookie does not exist", () => {
     cy.clearCookie("mci-project-cookie");
     cy.visit(SPRUCE_URLS.userPatches);
     cy.dataCy("auxiliary-dropdown-link").click();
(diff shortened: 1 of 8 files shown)


sophstad (approved):
LGTM! I did find a weird unrelated bug though :0

https://user-images.githubusercontent.com/9298431/151262300-cd9cdfd8-580c-4192-985b-4eb94f0d70da.mov

minnakt:
> LGTM! I did find a weird unrelated bug on mainline commits though :0
> 
>  dropdown-bug.mov

I took a look at the weird bug and think I fixed it 🤔 

thank you for pointing it out! sherlock holmes level detective work 🕵️‍♂️❤️

src/components/Header/AuxiliaryDropdown.tsx
```diff
@@ -23,6 +36,11 @@ export const AuxiliaryDropdown = () => {
       href: `${uiURL}${legacyRoutes.projects}`,
       text: "Projects",
     },
+    {
```
khelif96, line 39: Since this is tied to the mainline commits project it might make sense to toggle this with the spruce-beta flag. The cookie only gets set from the waterfall page so its possible the user won't find this link useful since it would always link them to the default project.

cypress/integration/nav_bar.ts
```diff
@@ -50,4 +50,18 @@ describe("Nav Bar", () => {
       .should("have.attr", "href")
       .and("include", LEGACY_URLS.distros);
   });
+  it("Nav Dropdown should link to patches page of most recent project if cookie exists", () => {
+    cy.setCookie("mci-project-cookie", "spruce");
+    cy.visit(SPRUCE_URLS.userPatches);
+    cy.dataCy("auxiliary-dropdown-link").first().click();
```
khelif96, line 56: nit. is `.first()` necessary are there multiple elements with this data-cy selector?

src/components/Dropdown/index.tsx
```diff
@@ -38,7 +38,6 @@ const Dropdown: React.FC<DropdownProps> = ({
         ref={menuButtonRef}
         onClick={() => setIsOpen(!isOpen)}
         data-cy={dataCy}
-        id="searchable-dropdown"
```
khelif96, line 41: Any idea why/if this was causing the flicker?

minnakt: Yeah haha it's because the `Dropdown` is used in multiple places. For example, the Status Selector (the dropdown that Sophie could keep opening and closing through the bug) also uses the `Dropdown` which meant it had `id = searchable-dropdown`, so it was connected to the wrong label. 😅 I added an `id` field onto the `Dropdown` just now, lmk what you think!

khelif96 (approved):

PR #1061, opened 2022-01-27
EVG-15961: Fix usePatchAnalytics error on mainline commits

minnakt:
[EVG-15961](https://jira.mongodb.org/browse/EVG-15961)

### Description 
This PR addresses the issue where viewing a version (for example, this [one](https://spruce.mongodb.com/version/mongodb_mongo_v4.4_0b01caf40a5a610685c0bf590ff2fa54aadd0443/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)) causes `usePatchAnalytics` to error. Visiting one of these pages causes an error where we aren't able to find a corresponding patch for the id in the URL (see below). The solution is to instead find a corresponding version for the id in the URL.

Screen Shot 2022-01-26 at 11 01 14 PM

I added another function `useVersionAnalytics` and replaced all instances of `usePatchAnalytics` with it. I kept the `usePatchAnalytics` file though because it might be useful to have in the future (for example on the Reconfigure page).

### Testing 
- Manually checking the console for errors (using `yarn run prod`)


bynn (commented):
very good!

khelif96 (approved):
Great Refactor. LGTM 🪨 🚀

PR #1064, opened 2022-01-28
EVG-15813 Update bugsnag source map plugin and support multiple release stages for bugsnag tracking

khelif96:
[EVG-15813](https://jira.mongodb.org/browse/EVG-15813)

### Description 
This will require an update .cmdrc.json once merged.

We should be able to get more detailed bugsnag tracking  as well as release specific bug reports and bug reporting on staging and beta.


.evergreen.yml
```diff
@@ -291,6 +291,9 @@ functions:
       working_dir: spruce
       script: |
         export PATH=/opt/nodejs/node-v16.13.0-linux-x64/bin:$PATH
+        chmod +x appVersion.sh
+        . ./appVersion.sh
+        echo "Building Production version: $REACT_APP_VERSION"
```
khelif96, line 296: This loads the output of this script into an env variable to be used during the build stage.

appVersion.sh
```diff
@@ -0,0 +1,3 @@
+# This script runs appVersion.js which fetches the version number from the
+# package.json file and writes it to a env variable.
+export REACT_APP_VERSION=`node appVersion.js`
```
khelif96, line 3: Yess a shell script that calls a node js script. Easiest way to extract a variable from the package.json and have it live as an env variable. Open to suggestions.

upload-bugsnag-sourcemaps.js
```diff
@@ -1,47 +1,9 @@
...
-    Promise.all(files.map(uploadSourceMap)).catch((e) => {
-      console.log(e);
-    })
-  );
-}
-
-processSourceMaps();
+const { browser } = require("@bugsnag/source-maps");
+
+browser.uploadMultiple({
```
sophstad, line 3: 🔥

src/components/ErrorBoundary/ErrorBoundary.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+    jest.restoreAllMocks();
+    resetBugsnag();
+    restoreCalls.forEach((restore) => restore());
+  });
+  it("should not initialize bugsnag if on development environment", () => {
+    const mockBugsnag = jest.fn();
+    jest.spyOn(Bugsnag, "start").mockImplementation(mockBugsnag);
+    jest.spyOn(environmentalVariables, "isDevelopment").mockReturnValue(true);
+    mockEnvironmentalVariable("REACT_APP_VERSION", "1.0.0");
+    // mock isDevelopment and isProduction
```
sophstad, line 29: Is this comment outdated? We're not mocking isProduction here it seems

src/components/ErrorBoundary/ErrorBoundary.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+  it("should initialize bugsnag if running in production", () => {
+    const mockBugsnag = jest.fn();
+    jest.spyOn(Bugsnag, "start").mockImplementation(mockBugsnag);
+
+    mockEnvironmentalVariable("NODE_ENV", "production");
+    mockEnvironmentalVariable("REACT_APP_VERSION", "1.0.0");
+    mockEnvironmentalVariable("REACT_APP_RELEASE_STAGE", "production");
+
+    // mock isDevelopment and isProduction
```
sophstad, line 42: Same as above

src/components/ErrorBoundary/index.tsx
```diff
@@ -4,11 +4,17 @@ import BugsnagPluginReact from "@bugsnag/plugin-react";
...
+  isProduction,
+  getAppVersion,
+  getReleaseStage,
+} = environmentalVariables;
+
 let bugsnagStarted = false;
 
 // This error boundary is used during local development
-class DefaultErrorBoundary extends React.Component {
+class DefaultErrorBoundary extends React.Component<{}, { hasError: boolean }> {
```
sophstad, line 17: What's the purpose of the initial empty object?

khelif96: The initial object is the typing for the component props. Since i did not want to do anything with them i had to pass an empty type.  The second value is state and you have to pass in the props type before passing in the state type. It does not let me use an undefined value.

sophstad: Thanks, I've never really used class components with TypeScript! Cool 😎

sophstad (approved):

appVersion.sh
minnakt: I wonder if we can get rid of `appVersion.js` and do everything in this script. Could something like this work?

Screen Shot 2022-01-28 at 5 21 10 PM

(credits to this [stackoverflow post](https://stackoverflow.com/questions/47234740/read-name-property-of-package-json-file-with-bash) lol)

minnakt (approved):

appVersion.sh
khelif96: Wow Great find! 🔥

PR #1065, opened 2022-01-28
EVG-14895 Leave breadcrumbs for graphql queries

khelif96:
[EVG-14895](https://jira.mongodb.org/browse/EVG-14895)

### Description 
We should now leave bugsnag breadcrumbs with Graphql queries and the variables used with in them. Which should help us track down what occurred just before a crash. 

### Testing

image


minnakt (approved):
LGTM (Let's Get This Merged) 😎🔥

PR #1067, opened 2022-01-31
EVG-16152 update sizes to use tokens

khelif96:
[EVG-16152](https://jira.mongodb.org/browse/EVG-16152)

### Description 
I took the sizes supplied in [@leafygreen-ui/tokens](https://www.mongodb.design/component/tokens/example/) and mapped them to variables. (See constants/tokens) 

I then did a find and replace updating hard coded pixel values with the token constants.
I manually updated some values and mapped them to the appropriate approximated size. 
With +- 1px tolerance so (5px -> 4px) (7px - 8px) etc

Each commit is labeled accordingly so it should be simple to verify swaps were made correctly. 

I am debating if it makes sense for the tokens to be pixel values directly instead of numbers. Feel free to leave your thoughts.
EDIT: Commit 20d2827 updates the values to have `px` included. 
I also included a proposal for managing our `z-index` values in `constants/tokens.ts` as well.


src/components/CommitChartLabel/index.tsx
```diff
@@ -69,7 +70,7 @@ const LabelContainer = styled.div`
   width: 100%;
   display: flex;
   margin-top: 10px;
```
minnakt, line 72: Change to `8px`?

src/components/Spawn/ExpirationField.tsx
```diff
@@ -140,8 +141,7 @@ const FormContainer = styled.div`
 `;
 
 const PaddedBody = styled.span`
-  padding-left: 15px;
-  padding-right: 15px;
+  padding: 0 ${size.s};
   margin-top: 22px;
```
minnakt, line 145: `24px` is probably okay for `margin-top` here

src/pages/host/HostCard.tsx
```diff
@@ -40,7 +41,7 @@ const SiderCard = styled(Card)<StylingProps>`
   padding-bottom: 25px;
```
minnakt, line 41: `24px` is probably okay here for `padding-bottom`

minnakt (commented):
Thanks for replacing in so many files! I looked at the `zIndex` stuff and it makes sense to me at least... 🙂

One thing that strikes me as kind of weird is that we often use `2px`, `12px`, `20px`, `28px`, `48px` etc but since those values aren’t really close to any MongoDB spacing tokens it’s hard to gauge if we can replace them. Do we want to get rid of these values by all means or should we consider having our own custom spacing in addition to the tokens MongoDB has (which would affect how we map the sizes in `constants/tokens.ts`)?

khelif96:
> Thanks for replacing in so many files! I looked at the `zIndex` stuff and it makes sense to me at least... 🙂
> 
> One thing that strikes me as kind of weird is that we often use `2px`, `12px`, `20px`, `28px`, `48px` etc but since those values aren’t really close to any MongoDB spacing tokens it’s hard to gauge if we can replace them. Do we want to get rid of these values by all means or should we consider having our own custom spacing in addition to the tokens MongoDB has (which would affect how we map the sizes in `constants/tokens.ts`)?

Thats a great point and something i'm wondering as well. I think it makes sense to keep them as is for now and try to match the LG tokens as closely as we can. And making incremental updates to the parts of the code base that use those custom values to match the tokens. How do y'all feel about that?

sophstad (approved):
> I think it makes sense to keep them as is for now and try to match the LG tokens as closely as we can. And making incremental updates to the parts of the code base that use those custom values to match the tokens. How do y'all feel about that?

This sounds good to me! I don't think it's a huge problem if we have a few non-tokenized values across the codebase. Another option in those instances could be to use CSS `calc()` to add/subtract a small value from one of our base sizes, though I don't know if this would really be beneficial...

Overall LGTM!

khelif96:
> > I think it makes sense to keep them as is for now and try to match the LG tokens as closely as we can. And making incremental updates to the parts of the code base that use those custom values to match the tokens. How do y'all feel about that?
> 
> This sounds good to me! I don't think it's a huge problem if we have a few non-tokenized values across the codebase. Another option in those instances could be to use CSS `calc()` to add/subtract a small value from one of our base sizes, though I don't know if this would really be beneficial...
> 
> Overall LGTM!

I ended up updating a bunch of the sizes we had to reflect the token sizes instead. and i added some font size constants. cc @sophstad @minnakt

src/constants/tokens.ts
```diff
@@ -20,4 +20,12 @@ const zIndex = {
...
 } as const;
 
-export { size, zIndex };
+const fontSize = {
+  xs: "8px",
+  s: "12px",
+  m: "14px",
+  l: "16px",
+  xl: "18px",
+} as const;
```
sophstad, line 29: Interesting, this almost feels like _too_ many font sizes to me. Is this perhaps an opportunity to standardize some uses? E.g. move all 16px usages to either 14px or 18px

src/components/PageTitle.tsx
```diff
@@ -1,6 +1,7 @@
 import styled from "@emotion/styled";
 import { H2, Subtitle } from "@leafygreen-ui/typography";
 import { Skeleton } from "antd";
```
minnakt, line 3: Is there a bug with the badge placement? It's normally right after & inline with the title

**in this PR:**
Screen Shot 2022-02-08 at 4 05 39 PM

Screen Shot 2022-02-08 at 4 23 00 PM



**normally:**
Screen Shot 2022-02-08 at 4 06 13 PM

src/pages/commits/ActiveCommits/ChartToggle.tsx
```diff
@@ -3,6 +3,7 @@ import styled from "@emotion/styled";
 import { uiColors } from "@leafygreen-ui/palette";
```
minnakt, line 3: Padding looks a little off for the top-portion of the toggle (was this intentional)? 

Screen Shot 2022-02-08 at 4 14 12 PM

src/components/CommitChartLabel/index.tsx
```diff
@@ -78,7 +79,7 @@ const LabelContainer = styled.div`
 const LabelText = styled(Body)`
   color: ${gray.dark2};
   width: 100%;
-  font-size: 12px;
```
minnakt, line 81: Sorry for being picky but can we change this back to `12px`? It's okay if it doesn't have a token, I just think it looks nicer 😔

sophstad (approved):

src/pages/task/taskTabs/logs/logTypes/TaskEventLogLine.tsx
```diff
@@ -105,12 +106,12 @@ export const TaskEventLogLine: React.FC<TaskEventLogEntry> = ({
 };
 
 const Row = styled.div`
-  font-size: 1.2em;
-  padding-top: 8px;
-  margin-bottom: 8px;
+  font-size: ${fontSize.l};
```
minnakt, line 109: The large font size here looks a bit comically large while medium looks too small... can we make it `16px` 😭

src/pages/commits/ActiveCommits/ChartToggle.tsx
khelif96: Great catch!

src/components/PageTitle.tsx
khelif96: Great catch!

src/components/PageTitle.tsx
```diff
@@ -47,34 +47,26 @@ export const PageTitle: React.FC<Props> = ({
...
-              </span>
-            </TitleTypography>
-            <BadgeWrapper size={size}>{badge}</BadgeWrapper>
-          </FlexRow>
+          <TitleTypography size={size}>
+            <span data-cy="page-title">
+              {title}
+              {"  "}
+              <BadgeWrapper size={size}>{badge}</BadgeWrapper>
+            </span>
```
minnakt, line 55: You can remove the `BadgeWrapper` outside of the span to prevent the `{""}` weirdness (it will still work as intended

src/pages/commits/ActiveCommits/ChartToggle.tsx
```diff
@@ -31,14 +31,14 @@ export const ChartToggle: React.FC<{
...
+            <b>Absolute Number</b>
           </Radio>
           <Radio
             data-cy="cy-chart-percent-radio"
             id="chart-radio-percent"
             value={ChartTypes.Percentage}
           >
-            <Label htmlFor="chart-radio-percent">Percentage</Label>
+            <b>Percentage</b>
           </Radio>
```
minnakt, line 42: Padding is still off; I think for some reason leafygreen adds an `8px margin-top` to the radio buttons??

this `padding` will fix it:
Screen Shot 2022-02-08 at 5 32 31 PM

minnakt (approved):
Thank you for all your work on this!

PR #1069, opened 2022-02-01
EVG-15082: Add commit queue section

sophstad:
EVG-15082

### Description 
- Add Commit Queue card to GitHub/Commit Queue page
- Rename directory: `GithubCommitQueue` ➡️ `GithubCommitQueueTab` for consistency
- Stricter typing: `FormState` types should not extend `FormDataProps` and fix subsequent typecheck failures

### Screenshots
image

### Testing 
Update unit and integration tests. To preview locally when running a dev server using the latest main branch of evergreen:

- http://localhost:3000/project/evergreen/settings/github-commitqueue to view a project that inherits from a repo
- http://localhost:3000/project/spruce/settings/github-commitqueue to view a project that does not inherit from a repo
- http://localhost:3000/project/602d70a2b2373672ee493184/settings/github-commitqueue to view a repository's settings


src/components/ProjectSettingsTabs/GithubCommitQueueTab/getFormSchema.ts
```diff
@@ -3,7 +3,12 @@ import { SpruceFormProps } from "components/SpruceForm";
 import { CardFieldTemplate } from "components/SpruceForm/FieldTemplates";
```
malikchaya2, line 3: I noticed that the "Save changes on Page" button is enabled even after I refresh the page. I feel like it should be disabled at that point. Am I doing something wrong or is there a bug?

src/components/ProjectSettingsTabs/PluginsTab/getFormSchema.ts
```diff
@@ -171,18 +159,12 @@ export const getFormSchema = (
...
-              ?.fileTicketWebhook?.endpoint
-          ),
-        },
-        secret: {
-          ...placeholderIf(
-            repoData?.projectPluginsSettings?.buildBaronSettings
-              ?.fileTicketWebhook?.secret
-          ),
-        },
+        endpoint: placeholderIf(
```
malikchaya2, line 162: After this change, it shows (Default from Repo) without any other content. But that is probably it's own bug with the placeholderIf function. 
image

sophstad: Actually if I had to guess, the problem is likely that there's no data for this field in the local document for the project's parent repo. Is this something that is guaranteed to be set for a repo? If not, we should only show a placeholder if the repo value is set.

src/components/ProjectSettingsTabs/GithubCommitQueueTab/getFormSchema.ts
sophstad: It is a bug :( good catch!

src/components/ProjectSettingsTabs/PluginsTab/getFormSchema.ts
malikchaya2: yes, that's probably the case

src/components/ProjectSettingsTabs/GithubCommitQueueTab/getFormSchema.ts
```diff
@@ -178,6 +183,65 @@ export const getFormSchema = (
...
+        type: "object" as "object",
+        title: "Commit Queue",
+        properties: {
+          enabled: {
+            type: ["boolean", "null"],
+            oneOf: radioBoxOptions(
+              ["Enabled", "Disabled"],
+              repoData?.commitQueue?.enabled
+            ),
+          },
```
ablack12, line 196: Sorry this additional checkbox was added more recently than the figma design:
image
You can totally create a new PR to handle it, but it should work like a simple checkbox if you do want to add it here (whether this is enabled/disabled doesn't impact anything else in the section).

src/components/ProjectSettingsTabs/GithubCommitQueueTab/getFormSchema.ts
```diff
@@ -178,6 +183,65 @@ export const getFormSchema = (
...
+              ...insertIf(repoData, {
+                type: "string" as "string",
+                title: `Default to Repo (${repoData?.commitQueue?.mergeMethod})`,
+                enum: [""],
+              }),
+            ],
+          },
+          patchDefinitions: {
+            type: "object" as "object",
+            title: "Commit Queue Patch Definitions",
```
ablack12, line 229: I am able to press save with an empty tag or etc; although I do get this upset backend error but I wonder if it's possible have the UI prevent trying to do this in the first place.
image

ablack12 (approved):
Left a couple of thoughts but generally LGTM with Chaya's comments -- no need for me to review again 👍

cypress/integration/projectSettings/project_settings.ts
```diff
@@ -41,15 +41,44 @@ describe("Repo Settings", () => {
...
       cy.dataCy("add-button").contains("Add Patch Definition").parent().click();
 
       cy.dataCy("variant-tags-field").find("button").click();
       cy.dataCy("variant-tags-input").first().type("vtag");
 
       cy.dataCy("task-tags-field").find("button").click();
       cy.dataCy("task-tags-input").first().type("ttag");
+    });
+
+    it.only("Toggling disable commit queue hides inputs", () => {
```
minnakt, line 54: Is this `it.only` intentional? (i.e. won't be running any tests in this file except this one?)

minnakt (approved):
Looks good to me 🙂 thank you for fixing the broken `codegen` test! ❤️

cypress/integration/projectSettings/project_settings.ts
sophstad: Ah good catch 😰 thank you!!

src/components/ProjectSettingsTabs/GithubCommitQueueTab/getFormSchema.ts
sophstad: Cool, it looks like this field isn't returned by GraphQL yet so I've made some new tickets to complete this: EVG-16257, EVG-16258

src/components/ProjectSettingsTabs/GithubCommitQueueTab/getFormSchema.ts
sophstad: Thanks, added the same validation that's used for other aliases on this page 👍

src/components/ProjectSettingsTabs/GithubCommitQueueTab/getFormSchema.ts
sophstad: I did some digging and I believe this bug is a result of a bug in the form library: https://github.com/rjsf-team/react-jsonschema-form/issues/1201. The onChange handler fires many times on mount, causing the "hasChanges" flag to be set to `true` and the save button to be enabled.

I've come up with a better way to manage saveability but I think it's worth doing in another ticket: EVG-16268. Will merge this as-is!

PR #1079, opened 2022-02-02
EVG-16180: Limit Tooltip size by reducing number of commits shown

minnakt:
[EVG-16180](https://jira.mongodb.org/browse/EVG-16180)

### Description 
This PR addresses two issues on the `/commits` page related to the `Tooltip` that is attached to the `Inactive Commits` button:
1. If the `Tooltip` is larger than the scrollable space, it gets cut off and is not readable (see image in [JIRA](https://jira.mongodb.org/browse/EVG-16180) ticket).
2. If the `Tooltip` is particularly large, it can cause the commits page to expand weirdly. This is especially noticeable on pages where you can't scroll, because viewing a large `Tooltip` will elongate the page to make space for the Tooltip. ([here](https://spruce.mongodb.com/commits/sys-perf) is an example)

To fix these issues, I limited the contents inside the tooltip so that this overflow problem will never occur. This is likely a temporary solution, as from user testing sessions it seems like this component will have to change to accommodate user requests.

### Screenshots
tooltip-size-limit

### Testing 
- Changed a few lines in `InactiveCommits.test.tsx`


src/pages/commits/InactiveCommits/index.tsx
```diff
@@ -58,7 +61,8 @@ export const InactiveCommitButton: React.FC<InactiveCommitsProps> = ({
   }
 
   return (
-    <Tooltip
+    <StyledTooltip
+      scrolled={isScrolled ? 1 : 0}
```
minnakt, line 65: the only reason I did this is because of this

Screen Shot 2022-02-02 at 10 09 24 AM

src/pages/commits/CommitsWrapper.tsx
```diff
@@ -158,10 +158,10 @@ const CommitWrapper = styled.div<{ width: number }>`
   min-width: ${({ width }) => width * 0.75}px;
   margin: 0px ${size.xs};
 
-  &:first-child {
+  &:first-of-type {
     margin-left: 0;
   }
-  &:last-child {
+  &:last-of-type {
```
khelif96, line 164: thanks!

src/pages/commits/InactiveCommits/index.tsx
```diff
@@ -120,25 +122,38 @@ const ButtonContainer = styled.div`
...
+const TooltipTitleText = styled.div`
+  margin-bottom: ${size.xs};
+  font-weight: bold;
 `;
 
-const MAX_COMMIT_COUNT = 5;
+const CommitText = styled.div`
+  padding: ${size.xxs} 0;
+  word-break: break-all;
+  font-size: 13px;
```
khelif96, line 146: Does `fontSize.m` work here? 14px?

khelif96 (approved):
Thanks for all your work on this!

src/pages/commits/InactiveCommits/index.tsx
minnakt: I think I'll keep it because `14px` looks a little ugly 😔

PR #1082, opened 2022-02-04
EVG-16085: Prevent text from overlapping on Commits page

minnakt:
[EVG-16085](https://jira.mongodb.org/browse/EVG-16085)

### Description 
Currently, when you view the `/commits` page with a small screen you will see a few visual bugs (see image): 
1. "Unmatching" text will overlap with commits text
2. Text in the sticky headers will overflow and overlap the task results below
<p align="center">
Screen Shot 2022-02-04 at 11 54 29 AM
</p>

I modified some CSS to prevent this from happening. Most importantly, I gave the `CommitWrapper` a `min-width` because right now it shrinks proportionately to the screen size which causes the overflow to get worse.

### Screenshots
<p align="center">
Screen Shot 2022-02-04 at 12 17 34 PM
</p>

### Testing 
- Visually


sophstad (approved):
💯

PR #1085, opened 2022-02-07
EVG-13250: Disable all options except Reconfigure for unfinalized patch

minnakt:
[EVG-13250](https://jira.mongodb.org/browse/EVG-13250)

### Description 
This PR disables the options to `Schedule Tasks` / `Unschedule All Tasks` / `Restart` / `Add to Commit Queue` for an unfinalized patch.

#### Analytics Changes
This PR also addresses some issues with `usePatchAnalytics` that I missed earlier. I was wondering why I would get this error in Bugsnag whenever I interacted with an action button dropdown on the My Patches page:
<p align="center">
Screen Shot 2022-02-07 at 12 40 55 PM
</p>

It's because `usePatchAnalytics` and `useVersionAnalytics` always try to grab the `patchID`/`versionID` from the URL, which isn't always possible. For example, if you're on the My Patches page, it'll assume that your username is the `patchID`/`versionID`, causing the error above.

For this reason I slightly altered the analytics functions to take in an optional parameter for `patchID` or `versionID`.

### Screenshots
Screen Shot 2022-02-04 at 2 34 21 PM

### Testing 
- Manually checking the console for errors

### DIFF
--- a/src/analytics/patch/usePatchAnalytics.ts
+++ b/src/analytics/patch/usePatchAnalytics.ts
@@ -55,9 +55,12 @@ interface P extends Properties {
 }
 interface PatchAnalytics extends Analytics<Action> {}
 
-export const usePatchAnalytics = (): PatchAnalytics => {
+export const usePatchAnalytics = (patchId?: string): PatchAnalytics => {
   const userId = useGetUserQuery();
-  const { id } = useParams<{ id: string }>();
+  const { id: idFromURL } = useParams<{ id: string }>();
+
+  const id = patchId || idFromURL;
+
   const { data: eventData } = useQuery<PatchQuery, PatchQueryVariables>(
     GET_PATCH,
     {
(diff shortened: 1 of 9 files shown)


minnakt:
@khelif96 tagged you specifically since you know most about my last `useVersionAnalytics` PR  😅

src/analytics/patch/usePatchAnalytics.ts
```diff
@@ -55,9 +55,12 @@ interface P extends Properties {
 }
 interface PatchAnalytics extends Analytics<Action> {}
 
-export const usePatchAnalytics = (): PatchAnalytics => {
+export const usePatchAnalytics = (patchId?: string): PatchAnalytics => {
   const userId = useGetUserQuery();
-  const { id } = useParams<{ id: string }>();
+  const { id: idFromURL } = useParams<{ id: string }>();
+
+  const id = patchId || idFromURL;
```
khelif96, line 62: In this case i think it makes sense to not tie this hook to a specific implementation. It should not try to fetch a param from the url and it should instead always use a passed in param. So the implementer can pass in the appropriate id depending on the context it's implemented in.  (If its used on a patch page the implementer can pass in the appropriate id from the url. Or if its used in a card they can pass it in based off of that cards specific id.)

khelif96 (changes requested):

khelif96 (approved):

PR #1092, opened 2022-02-10
EVG-15084: Add Patch Aliases Page

sophstad:
EVG-15084

### Description
- Add first section of Patch Aliases page
- Add ability to render aliases with either accordions or expandable cards
- Refactor project settings utils: split into `form` and `alias` modules

### Screenshots
image

### Testing 
- Add unit test, update integration test
- To preview locally when running a dev server using the latest main branch of evergreen:
  - http://localhost:3000/project/evergreen/settings/patch-aliases to view a project that inherits from a repo
  - http://localhost:3000/project/spruce/settings/patch-aliases to view a project that does not inherit from a repo
  - http://localhost:3000/project/602d70a2b2373672ee493184/settings/patch-aliases to view a repository's settings


src/components/ProjectSettingsTabs/PatchAliasesTab/getFormSchema.ts
```diff
@@ -0,0 +1,68 @@
...
+    },
+    type: "object" as "object",
+    properties: {
+      patchAliases: {
+        title: "Patch Aliases",
+        description:
+          "Specify aliases to use with the CLI. Aliases may be specified multiple times. The result will be their union. All regular expressions must be valid Golang regular expressions. Use an alias with the --alias flag passed to the CLI patch command.",
+        ...overrideRadioBox(
+          "aliases",
+          ["Override Repo Patch Aliases", "Default to Repo Patch Aliases"],
```
malikchaya2, line 30: Is there an example of a project that default to a repo where the repo has patch aliases populated?

sophstad: No, but you could add one to the repo that I linked to in the description and then check out the evergreen project's settings (since evergreen defaults to that repo).

src/components/ProjectSettingsTabs/utils/alias.ts
```diff
@@ -0,0 +1,170 @@
...
+  task,
+  taskTags,
+  variant,
+  variantTags,
+}: AliasType): boolean =>
+  pairHasError(task, taskTags) || pairHasError(variant, variantTags);
+
+export const sortAliases = (aliases) =>
+  aliases.reduce(
+    (o, a) => {
```
minnakt, line 28: loving this new utils file 😎 can you add types to this function (if possible)?

src/components/ProjectSettingsTabs/AliasRow.tsx
```diff
@@ -71,90 +67,119 @@ export const AliasRow: SpruceFormProps["ObjectFieldTemplate"] = ({
...
+          <div data-cy="task-field" id={makeId(TaskInput.Regex)}>
+            {task.content}
+          </div>
+        )}
+      </TaskRegexContainer>
+    </>
+  );
+
+  return useExpandableCard ? (
+    <ExpandableCard
```
minnakt, line 147: do we want any spacing between the expandable cards? (the top and bottom of each card touch)

Screen Shot 2022-02-14 at 1 13 40 PM

src/components/ProjectSettingsTabs/AliasRow.tsx
```diff
@@ -71,90 +67,119 @@ export const AliasRow: SpruceFormProps["ObjectFieldTemplate"] = ({
...
+              size="small"
+            />
+          )}
+        </>
+      }
+    >
+      {contents}
+    </ExpandableCard>
+  ) : (
     <Accordion
```
minnakt, line 170: The alignment between the text and Delete button is a little off because of the margins on the `AccordionTitle`. 

Screen Shot 2022-02-14 at 1 23 12 PM

One way to fix it is by changing the `align-items` property on the `ArrayItemRow` to `baseline` but I'm not sure that's what you want in this situation... 

Screen Shot 2022-02-14 at 1 19 58 PM

sophstad: That had `align-items` rule had some unexpected ramifications, so I opted here to use a custom margin around the title to force alignment. Let me know what you think!

src/components/ProjectSettingsTabs/PatchAliasesTab/getFormSchema.ts
ablack12: I noticed that if I had a project that had overridden patch aliases, but then I clicked on "Default to Repo Patch Aliases" and then saved it, that the override wasn't deleted. (idk how to make those videos you do but note that this Save toast is from when I clicked "Default to Repo Patch Aliases" and then saved the page; i.e. the overrides are still there, even after refereshing)
image

ablack12 (changes requested):

src/components/ProjectSettingsTabs/AliasRow.tsx
minnakt: Oh yeah hmm I see the problem. Is there any reason this Delete button can't be with the `Accordion` title like how it's with the `ExpandableCard` title in `AliasRow.tsx`? From a glance the only difference I can see is the usage of `topAlignDelete`, which I wonder if that can be propagated down since `AliasRow` also has access to `uiSchema`.

src/components/ProjectSettingsTabs/PatchAliasesTab/getFormSchema.ts
sophstad: Thank you good catch! I've added some tests around this too.

src/components/ProjectSettingsTabs/PatchAliasesTab/getFormSchema.ts
```diff
@@ -0,0 +1,68 @@
...
+          {
+            $ref: "#/definitions/aliasArray",
+          }
+        ),
+      },
+    },
+  },
+  uiSchema: {
+    patchAliases: {
+      aliasesOverride: {
```
ablack12, line 40: One other thing: we allow users to save the project with Override checked but for aliases we can't actually override unless there are aliases defined so refreshing the page just shows this as the repo values again. Can we disable save unless aliases are actually added?
image

sophstad: This relates to the larger issue of "save button is enabled too much," which is tracked by EVG-16268. I'll prioritize that ticket for work soon since this seems to keep coming up.

ablack12: Oh nice. I'll add a comment to that ticket.

ablack12 (approved):

minnakt (approved):

PR #1093, opened 2022-02-10
EVG-15559 Use task stats query for build variants panel

khelif96:
[EVG-15559](https://jira.mongodb.org/browse/EVG-15559)

### Description 
This uses the query described in https://github.com/evergreen-ci/evergreen/pull/5385
### Screenshots
No visible changes

### Evergreen PR 
https://github.com/evergreen-ci/evergreen/pull/5385


src/pages/version/buildVariants/buildVariants.test.ts
```diff
@@ -1,128 +0,0 @@
-import { TaskStatus } from "types/task";
-import { groupTasksByUmbrellaStatus } from "./utils";
-
-describe("groupTasksByUmbrellaStatus", () => {
```
minnakt, line 4: Should we consider keeping this test file but modifying it to work with the new function? I'm curious as to why it was deleted

ablack12:
Since this has no visible changes I'm not sure I'm the best reviewer so removing myself.

minnakt (approved):
Nice, thank you for getting rid of the `GroupedTaskSquare` file because those extra layers were confusing to read. 😃 And yay for reusing the util function!

PR #1094, opened 2022-02-10
EVG-16300 Skip query unless modal is opened

khelif96:
[EVG-16300](https://jira.mongodb.org/browse/EVG-16300)

### Description 
Noticed the restart modal was performing a query on initial page load this is an expensive query. (Uses GetTasksByVersion under the hood) We shouldn't be making it for a part of the page that a user may never open.


minnakt (approved):
Nice 🧠

PR #1095, opened 2022-02-14
EVG-15070: Attach/detach from repo

sophstad:
[EVG-<number>](https://jira.mongodb.org/browse/EVG-<number>)

### Description 

### Screenshots

### Testing 

### Evergreen PR 

### DIFF
--- a/src/components/ProjectSettingsTabs/GeneralTab/Fields/MoveRepoField.tsx
+++ b/src/components/ProjectSettingsTabs/GeneralTab/Fields/MoveRepoField.tsx
@@ -1,10 +1,22 @@
 import { useState } from "react";
+import { useMutation } from "@apollo/client";
 import styled from "@emotion/styled";
 import { Field } from "@rjsf/core";
 import { Button } from "components/Button";
 import { ConfirmationModal } from "components/ConfirmationModal";
 import { SpruceForm } from "components/SpruceForm";
 import { size } from "constants/tokens";
+import { useToastContext } from "context/toast";
+import {
+  AttachProjectToRepoMutation,
+  AttachProjectToRepoMutationVariables,
+  DetachProjectFromRepoMutation,
+  DetachProjectFromRepoMutationVariables,
+} from "gql/generated/types";
+import {
+  ATTACH_PROJECT_TO_REPO,
+  DETACH_PROJECT_FROM_REPO,
+} from "gql/mutations";
 
 interface ModalProps {
   onCancel: () => void;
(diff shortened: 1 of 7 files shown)


src/gql/mutations/index.ts
```diff
@@ -1,9 +1,11 @@
 import ABORT_TASK from "./abort-task.graphql";
 import ADD_ANNOTATION from "./add-annotation.graphql";
 import ADD_FAVORITE_PROJECT from "./add-favorite-project.graphql";
+import ATTACH_PROJECT_TO_REPO from "./attach-project-to-repo.graphql";
 import ATTACH_VOLUME from "./attach-volume.graphql";
 import CLEAR_MY_SUBSCRIPTIONS from "./clear-my-subscriptions.graphql";
 import CREATE_PUBLIC_KEY from "./create-public-key.graphql";
+import DETACH_PROJECT_FROM_REPO from "./detach-project-from-repo.graphql";
```
malikchaya2, line 8: I noticed that after a project is detached from the repo, the "Save" button on the page becomes enabled. I'm guessing that that is not desired behavior?

sophstad: Right, I think all of these save issues can be saved for EVG-16268 because they stem from the same problem (the exception being if save is enabled when there is an error on the page—that kind of validation should be addressed now).

src/components/ProjectSettingsTabs/GeneralTab/Fields/MoveRepoField.tsx
```diff
@@ -1,10 +1,22 @@
 import { useState } from "react";
```
malikchaya2, line 1: Would it make sense to rename the file to something more general now that it also has attach and detach?

src/gql/mutations/index.ts
malikchaya2: makes sense, thanks!

src/components/ProjectSettingsTabs/GeneralTab/Fields/MoveRepoField.tsx
sophstad: Sure! Hm maybe... `RepoConfigField`?

malikchaya2: Nice, I couldn't think of anything. That works!

src/components/ProjectSettingsTabs/GeneralTab/Fields/RepoConfigField.tsx
```diff
@@ -0,0 +1,251 @@
...
+      onConfirm={() => {
+        if (shouldAttach) {
+          attachProjectToRepo();
+        } else {
+          detachProjectFromRepo();
+        }
+        handleClose();
+      }}
+      open={open}
+      title={`Are you sure you want to ${
```
ablack12, line 121: image
When I clicked "Save Changes on Page" for the General Section it now detaches me from repo, regardless of whether or not I had clicked into the modal.

ablack12 (changes requested):

src/components/ProjectSettingsTabs/GeneralTab/Fields/RepoConfigField.tsx
sophstad: That's funny, this is actually a consequence of EVG-16081, which began as "SaveProjectSettingsForSection incorrectly sets useRepoSettings." The behavior we're seeing there involves `useRepoSettings` being returned as `false` whenever the save resolver is called. You can verify that we're not _actually_ detaching from the repo by the fact that there's no network request being made to the `DetachProjectFromRepo` resolver.

src/components/ProjectSettingsTabs/GeneralTab/Fields/RepoConfigField.test.tsx
```diff
@@ -0,0 +1,255 @@
...
+    });
+
+    it("disables the confirm button on initial render", () => {
+      const mockOnConfirm = jest.fn();
+      const mockOnCancel = jest.fn();
+      const { getAllByText } = render(
+        <MoveRepoModal onCancel={mockOnCancel} onConfirm={mockOnConfirm} open />
+      );
+
+      const moveRepoButton = getAllByText("Move Repo")[1].closest("button");
```
khelif96, line 116: does `getByRole("button", { name: "Move Repo" })` work instead? https://testing-library.com/docs/queries/byrole/

khelif96: It helps target the button specifically without needed to rely on indexes since there are multiple elements named "Move Repo"

src/components/ProjectSettingsTabs/GeneralTab/Fields/RepoConfigField.tsx
```diff
@@ -0,0 +1,251 @@
...
+  onCancel,
+  onConfirm,
+  open,
+}) => {
+  const [formState, setFormState] = useState({});
+  const [hasError, setHasError] = useState(false);
+  const isDisabled =
+    Object.keys(formState).length === 0 ||
+    Object.values(formState).some((input) => input === "") ||
+    hasError;
```
khelif96, line 37: I wonder can we not rely on rjsf form validation instead of validating it outside?

khelif96 (approved):

PR #1102, opened 2022-02-17
EVG-15822: Add View Hosts link on Task Queue page

minnakt:
[EVG-15822](https://jira.mongodb.org/browse/EVG-15822), [EVG-16359](https://jira.mongodb.org/browse/EVG-16359)

### Description 
This PR adds a `View hosts` link on the Task Queue page. This link will redirect to the Hosts page, where the Hosts table will be filtered to only include that particular distro. Additionally, tasks from mainline commits now link to the Spruce UI instead of the legacy UI.

I also added a loading skeleton because this is what currently happens when the page is loading, which is not very pretty.

Screen Shot 2022-02-17 at 9 57 46 AM

### Screenshots
https://user-images.githubusercontent.com/47064971/154508868-4ed4c561-e43f-451c-8e66-4a77aaa79a2c.mov

### Testing 
- Modified Cypress tests in `taskQueue/route.ts`


src/pages/TaskQueue.tsx
```diff
@@ -59,41 +61,62 @@ export const TaskQueue = () => {
...
+            </SearchableDropdownWrapper>
+          </TableControlOuterRow>
+
+          {
+            /* Only show name & link if distro exists. */
+            selectedDistro && (
+              <TableHeader>
+                {/* @ts-expect-error  */}
+                <StyledH3> {selectedDistro.id} </StyledH3>
+                <StyledRouterLink to={`/hosts?distroId=${selectedDistro.id}`}>
```
khelif96, line 108: Instead of hard coding this link can you export it as a util function from here. https://github.com/evergreen-ci/spruce/blob/main/src/constants/routes.ts

Also do not hard code query params use a util function to generate them.

PR #1105, opened 2022-02-17
EVG-16268: Improve save button disabled state

sophstad:
EVG-16268

### Description 
- Improve the logic by which we determine whether a page has changes: compare the initial and current states of the page after they have been transformed into their GQL input shape (i.e. run `formToGql()`).
- Using these transform functions should ensure that save isn't enabled when a non-impactful change is made, for example adding a new empty field to a list of users. It also prevents saving when a change is made and then undone.
- Debounce this operation to improve performance

### Testing 
- Updated unit and integration tests


cypress/integration/projectSettings/project_settings.ts
```diff
@@ -253,13 +271,14 @@ describe("Project Settings when defaulting to repo", () => {
...
 
-  it("Loads with the save button disabled initially", () => {
+  it("Should not have the save button enabled on load", () => {
     cy.dataCy("save-settings-button").should("be.disabled");
   });
 
   it("Preserves edits to the form when navigating between settings tabs and does not show a warning modal", () => {
     cy.dataCy("spawn-host-input").should("have.value", "/path");
     cy.dataCy("spawn-host-input").type("/test");
+    cy.wait(HAS_CHANGES_TIMEOUT_MS); // eslint-disable-line cypress/no-unnecessary-waiting
```
sophstad, line 281: I experimented with using `cy.clock()` and `cy.tick()` but that had unexpected consequences on debounce in later tests, even after restoring the clock. Since we only need to wait in this one place I think it's no big deal.

malikchaya2 (approved):
🎉

ablack12 (approved):
We can probably create a new ticket for this bc I imagine it's not related to these changes, but Move to New Repo only takes effect when "Save" is pressed, not when Move Repo is pressed, which isn't right.

khelif96 (changes requested):

cypress/integration/projectSettings/project_settings.ts
sophstad: Since `setHasChanges` is debounced, we need to make sure it finishes running and sets `hasChanges` to true so that the modal appears on line 283.

sophstad: Oh sorry I got confused above—without waiting for debounce to complete, the "Shows a navigation warning modal when clicking on a header link" test will fail because the test will navigate to a new page via header link

khelif96: Would an assertion on the `Save Changes on Page` button state work instead?     

`cy.dataCy("save-settings-button").should("not.be.disabled");`

I'm hesitant on arbitrary waits since they tend to be flaky and unnecessarily slow down the test suite. We removed several of them in the past.

src/components/SpruceForm/FieldTemplates.tsx
```diff
@@ -154,7 +154,7 @@ const ArrayContainer = styled.div`
 const DeleteButtonWrapper = styled(ElementWrapper)`
   margin-left: ${size.s};
   // Align button with top of input unless it should specifically align to the top of the ArrayItemRow
-  align-items: ${({ topAlignDelete }: { topAlignDelete: boolean }) =>
+  margin-top: ${({ topAlignDelete }: { topAlignDelete: boolean }) =>
     topAlignDelete ? "0px" : "20px"};
```
khelif96, line 158: This is a common pattern we have been doing with lg components  i wonder if we could make it more universal. (Not actionable just a thought i had when i saw this)

khelif96 (commented):
Looks Really Good! Just had that last comment about removing the `cy.wait`

cypress/integration/projectSettings/project_settings.ts
sophstad: Yeah I understand what you're saying. But I think this is a unique case because we are testing that state is preserved when navigating to a new page (via `cy.dataCy("navitem-access").click()`). If there is no wait, then `setHasChanges` will not fire before loading the Access page and then we aren't able to effectively test that state is preserved between pages. Def agree about avoiding waits in general but I think there are a few uncommon things happening here that necessitate it.

khelif96: Sorry i should have clarified what i meant by the `Save Changes on Page` assertion. If you instead asserted that the `Save Changes on Page` button became un-disabled  it would accomplish the same affect as waiting for the debounce since its state only changes once the validation runs. It would have offered a similar outcome to waiting to the debounce without relying on a set interval.

PR #1107, opened 2022-02-18
EVG-16081: Remove useRepoSettings usage

sophstad:
EVG-16081

### Description 
- Remove usage of `useRepoSettings` throughout by assigning the current project a `ProjectVariant`: either `Project`, `AttachedProject`, or `Repo`


malikchaya2 (commented):
`projectVariant === ProjectVariant.AttachedProject` feels a bit less intuitively understandable than `useRepoSettings`. I wonder if there is a way to add some clarity around it, maybe add some comments in some places linking it to a project being defaulted to the repo or using repo settings??

src/components/ProjectSettingsTabs/AccessTab/AccessTab.tsx
```diff
@@ -21,8 +22,12 @@ export const AccessTab: React.FC<TabProps> = ({
   const onChange = updateForm(tab);
 
   const { fields, schema, uiSchema } = useMemo(
-    () => getFormSchema(useRepoSettings, useRepoSettings ? repoData : null),
-    [repoData, useRepoSettings]
+    () =>
+      getFormSchema(
+        projectVariant,
+        projectVariant === ProjectVariant.AttachedProject ? repoData : null
```
ablack12, line 28: I'm hesitant about the wording ProjectVariant since it reminds me of build variants (which was initially what I thought this was about). If we want to go ahead with something like this, could we try a different naming? (ProjectType or etc)

khelif96 (commented):
I'm not exactly sure on the context of this change why do we want to stop using useRepoSettings?

malikchaya2:
> @malikchaya2 I think there's a good case to be made for moving to an enum in general, since `useRepoSettings` is a boolean that was attempting to represent three distinct states. I added some details to the enum definition, let me know if there's anything else in particular that would be helpful.
Discussed over slack, I think the enum is great!

src/components/ProjectSettingsTabs/AccessTab/getFormSchema.ts
```diff
@@ -78,23 +78,19 @@ export const getFormSchema = (
...
       },
     },
   },
 });
 
-const getAdminsDescription = (useRepoSettings: boolean): string => {
-  // Repo page, where useRepoSettings field does not exist
-  if (useRepoSettings === undefined) {
+const getAdminsDescription = (projectType: ProjectType): string => {
+  if (projectType === ProjectType.Repo) {
```
malikchaya2, line 89: This is a great example of why the enum is superior to having one boolean.

src/components/ProjectSettingsTabs/GeneralTab/Fields/index.tsx
```diff
@@ -1,5 +1,4 @@
-import { FilesIgnoredFromCacheField } from "./FilesIgnoredFromCacheField";
 import { RepoConfigField } from "./RepoConfigField";
 import { RepotrackerField } from "./RepotrackerField";
 
-export { FilesIgnoredFromCacheField, RepoConfigField, RepotrackerField };
```
malikchaya2, line 5: Why do we want to remove this?

src/components/ProjectSettingsTabs/NotificationsTab/getFormSchema.ts
```diff
@@ -3,22 +3,18 @@ import { SpruceFormProps } from "components/SpruceForm";
...
 export const getFormSchema = (
-  useRepoSettings: boolean,
   repoData?: FormState
 ): {
   fields: Record<string, Field>;
   schema: SpruceFormProps["schema"];
   uiSchema: SpruceFormProps["uiSchema"];
 } => ({
-  fields: {
-    defaultSubscriptionsField: DefaultSubscriptionsField,
```
malikchaya2, line 20: can we maybe keep this here for now since I'll be working on that part of the page soon?

src/components/ProjectSettingsTabs/NotificationsTab/Fields/DefaultSubscriptionsField.tsx
```diff
@@ -1,50 +0,0 @@
-import { useState } from "react";
```
malikchaya2, line 1: Can we keep this for now as well?

malikchaya2 (approved):
approved with small comments

src/components/ProjectSettingsTabs/NotificationsTab/getFormSchema.ts
sophstad: Sorry -- removed because I thought we opted not to allow defaulting to repo for this section. Will re-add!

src/components/ProjectSettingsTabs/GeneralTab/Fields/index.tsx
sophstad: I updated the field to use an approach managed entirely in the schema so no custom field is necessary! You can see the changes in `GeneralTab/getFormSchema.ts`

src/components/ProjectSettingsTabs/NotificationsTab/getFormSchema.ts
malikchaya2: Oh actually you're right, we did. But now worries, I'll do that cleanup soon.

PR #1110, opened 2022-02-22
EVG-16278 Fix accordion title misalignment

khelif96:
[EVG-16278](https://jira.mongodb.org/browse/EVG-16278)

### Screenshots
before:
image
after:
image


src/components/VersionRestartModal/BuildVariantAccordian.tsx
```diff
@@ -81,3 +82,8 @@ const countMatchingTasks = (
 const BadgeWrapper = styled("div")`
   padding-left: 10px;
 `;
+
+const FlexContainer = styled("div")`
```
minnakt, line 86: Should we use `styled.div` notation since that's what we use everywhere else?

PR #1113, opened 2022-02-23
EVG-16301 Use Neutral language for project error

malikchaya2:
[EVG-16301](https://jira.mongodb.org/browse/EVG-16301)

### Description 
This error didn't work as is because it can be either a project or a repo. 

### Screenshots
image

### Testing 
none


src/pages/ProjectSettings.tsx
```diff
@@ -68,7 +68,7 @@ export const ProjectSettings: React.FC = () => {
     variables: { repoId: repoRefId },
     onError: (e) => {
       dispatchToast.error(
-        `There was an error loading the repo ${repoRefId}: ${e.message}`
```
ablack12, line 71: Will the message always contain a helpful enough error? It seems like the "Error!" already indicates there was an error and just printing e.message would be sufficient

ablack12 (changes requested):

src/pages/ProjectSettings.tsx
malikchaya2: I think by convention we wouldn't want to rely on just e.message since it's out of Spruce's control and may change. We'd want to wrap it in our own error explanation.

ablack12: I'd prefer to shorten it since I do think the title covers it but I'll defer to UI conventions; cc @khelif96 if you can weigh in

ablack12 (approved):

PR #1115, opened 2022-02-23
EVG-16110 Script to send deploy emails.

khelif96:
[EVG-16110](https://jira.mongodb.org/browse/EVG-16110)

### Description 
Script to send properly formatted deploy emails. Based off of https://gist.github.com/bsamek/02436a45ee1cc151fe075d75871d0aad

This should work both locally and from within evergreen.


email.sh
```diff
@@ -0,0 +1,83 @@
...
+fi
+
+# Fetch previous release tag and get the commits since that tag
+PREVIOUS_VERSION=$(git describe --tags  --abbrev=0  `git rev-list --tags --max-count=1 --skip=1`)
+git log --no-merges $PREVIOUS_VERSION..HEAD --pretty="format:%s (%h)" > body.txt
+
+# Determine which verson of evergreen is available and use that
+if ! [ -x "$(command -v evergreen)" ]
+then
+  echo 'Error: evergreen is not on $PATH.'
```
bsamek, line 39: Nit: I think this means there will be a line that says "Error" in every CI deploy, which is maybe misleading.

email.sh
```diff
@@ -0,0 +1,83 @@
...
+  echo 'Trying to use the local binary'
+  if ! [ -x "$(command -v ~/evergreen)" ]
+  then
+    echo 'Error: Could not find evergreen binary on home directory'
+    exit 1
+  else
+    echo 'Found evergreen binary on home directory'
+    EVERGREEN=~/evergreen
+    CREDENTIALS='-c .evergreen.yml'
```
bsamek, line 48: Where did this file come from? I would be surprised if it's already on the host with creds, but maybe I'm forgetting something.

email.sh
```diff
@@ -0,0 +1,83 @@
...
+COMMAND+="$DATE"
+COMMAND+=" Spruce Deploy $TASK_NAME"
+COMMAND+="'"
+COMMAND+=" -b"
+COMMAND+=" '"
+COMMAND+="$BODY_HTML"
+COMMAND+="'"
+echo $COMMAND
+eval $COMMAND
+rm body.txt
```
bsamek, line 83: This is probably unnecessary, since you're already in the task directory, which gets cleaned up automatically at the end of the task.

package.json
```diff
@@ -27,7 +27,7 @@
     "eslint-staged:src": "STRICT=1 eslint -c src/.eslintrc.js",
     "eslint-staged:cypress": "STRICT=1 eslint -c cypress/.eslintrc.js",
     "eslint:strict": "STRICT=1 eslint -c src/.eslintrc.js 'src/**/*.ts?(x)' && eslint -c cypress/.eslintrc.js 'cypress/**/*.ts'",
-    "notify-email": "env-cmd -e prod -r env/.cmdrc.json node cli-scripts/notify-email.js",
+    "notify-email": "env-cmd -e prod -r env/.cmdrc.json ./email.sh",
```
bsamek, line 30: Is this for local deploys? If so, should it invoke email.sh with bash? Or should email.sh start with a shebang specifying bash?

bsamek (changes requested):

email.sh
khelif96: This gets generated here. https://github.com/evergreen-ci/spruce/blob/73ece6563534d3fcb10f9193f307ac060fb360a5/.evergreen.yml#L257-L273

bsamek (approved):

PR #1119, opened 2022-02-28
EVG-15075 Create the Project dropdown

malikchaya2:
[EVG-15075](https://jira.mongodb.org/browse/EVG-15075)

### Description 
This moves the project dropdown to a more generic location and modifies it to accommodate the project settings page. 

### Screenshots
image

### Testing 
manual testing. 
To test: http://localhost:3000/project/sandbox/settings/general


src/pages/ProjectSettings.tsx
```diff
@@ -113,7 +114,12 @@ export const ProjectSettings: React.FC = () => {
   return (
     <ProjectSettingsProvider>
       <SideNav aria-label="Project Settings">
-        <SideNavGroup header="Project" />
+        <DropdownContainer>
```
ablack12, line 117: Can we not show "Favorites" if no favorites are selected?
image

Since unattached projects don't have a header, it makes them all look like favorites.

ablack12 (commented):
Looking at this page on localhost, I'm not sure where in the dropdown to click to see the repo (not sure if I have permission to do this though). http://localhost:3000/project/evergreen/settings/general shows that it _does_ default to a repo but I don't see the repo itself in the dropdown (the figma design shows the repo as a clickable header:)
image

src/components/projectSelect/index.tsx
```diff
@@ -83,3 +94,24 @@ export const ProjectSelect: React.FC<ProjectSelectProps> = ({
...
+  }
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  const { data, loading } = useQuery<
+    GetViewableProjectRefsQuery,
+    GetViewableProjectRefsQueryVariables
+  >(GET_VIEWABLE_PROJECTS);
+  const { viewableProjectRefs } = data || { viewableProjectRefs: [] };
+  // const projects = ;
+  return { projects: viewableProjectRefs.map((p) => ({ ...p })), loading };
+};
```
khelif96, line 117: You can't actually do this. Conditional hooks are not allowed because they break reacts ability to track the hooks. 
https://reactjs.org/docs/hooks-rules.html
The rules-of-hooks rule is there to prevent this. `useQuery` has a [param called skip](https://www.apollographql.com/docs/react/data/queries/#skip) you can use instead

src/components/projectSelect/index.tsx
```diff
@@ -83,3 +94,24 @@ export const ProjectSelect: React.FC<ProjectSelectProps> = ({
     />
   );
 };
+
+export const getProjects = (isProjectSettingsPage) => {
```
khelif96, line 98: Since this is a hook it should have the following naming convention `useHookName`  https://reactjs.org/docs/hooks-custom.html

src/pages/ProjectSettings.tsx
```diff
@@ -197,3 +203,7 @@ const tabRouteValues = Object.values(ProjectSettingsTabRoutes);
 const PageContainer = styled.div`
   display: flex;
 `;
+
+const DropdownContainer = styled.div`
+  padding: 0.5rem;
```
khelif96, line 208: can you use one of the [tokens instead](https://github.com/evergreen-ci/spruce/blob/main/src/constants/tokens.ts). `size.xs`  is the equivalent to 0.5rem

src/pages/Commits.tsx
```diff
@@ -28,10 +28,10 @@ import {
 } from "types/commits";
 import { TaskStatus } from "types/task";
 import { array, queryString } from "utils";
+import { ProjectSelect } from "../components/projectSelect";
```
khelif96, line 31: you should use an absolute import `components/projectSelect`

src/components/projectSelect/index.tsx
```diff
@@ -66,7 +70,14 @@ export const ProjectSelect: React.FC<ProjectSelectProps> = ({
...
         Cookies.set(CURRENT_PROJECT, projectIdentifier, { expires: 365 });
-        history.push(getCommitsRoute(projectIdentifier));
+        history.push(
+          isProjectSettingsPage
+            ? getProjectSettingsRoute(
+                projectIdentifier,
+                ProjectSettingsTabRoutes.General
+              )
+            : getCommitsRoute(projectIdentifier)
+        );
```
khelif96, line 80: Instead of casing on it we should pass this callback in as a parameter. So the component could be used more widely.

khelif96 (changes requested):

src/components/projectSelect/index.tsx
malikchaya2: it's pretty heavily cased otherwise, so I don't know if this will do much. If this is needed elsewhere, it'll require more work anyhow.

src/components/projectSelect/index.tsx
malikchaya2: fixed, thanks!

src/pages/Commits.tsx
malikchaya2: done

src/components/projectSelect/index.tsx
khelif96: I think as a general rule of reusable component design it would make sense to make this more generic. This component will also be used on the project patches page eventually.

malikchaya2:
> Looking at this page on localhost, I'm not sure where in the dropdown to click to see the repo (not sure if I have permission to do this though). http://localhost:3000/project/evergreen/settings/general shows that it _does_ default to a repo but I don't see the repo itself in the dropdown (the figma design shows the repo as a clickable header:) image

done

src/components/projectSelect/index.tsx
malikchaya2: done

src/components/projectSelect/ProjectOptionGroup.tsx
```diff
@@ -37,9 +39,18 @@ export const ProjectOptionGroup: React.FC<OptionGroupProps> = ({
...
+  repoIdentifier,
+  isProjectSettingsPage = false,
 }) => (
   <OptionGroupContainer>
-    <Overline>{name}</Overline>
+    {repoIdentifier === "" || !isProjectSettingsPage ? (
+      <Overline>{name} </Overline>
+    ) : (
+      <OverlineHover>
+        <Overline onClick={() => onClick(repoIdentifier)}>{name} </Overline>
```
ablack12, line 50: Sorry I'm not sure if this is the right place to actually comment this but I try 😅  I love the look/feel of the clickable header! Testing locally though, it doesn't look like it's directing to the right place. Here if I click "evergreen-ci/evergreen" it redirects to the page for evergreen smoke test.
image

src/components/projectSelect/ProjectOptionGroup.tsx
```diff
@@ -37,9 +39,18 @@ export const ProjectOptionGroup: React.FC<OptionGroupProps> = ({
   name,
   projects,
   onClick,
+  repoIdentifier,
```
sophstad, line 42: Since this component is otherwise totally generic, could we make this prop something more generic like `optionGroupLink` or something?

src/components/projectSelect/ProjectOptionGroup.tsx
```diff
@@ -37,9 +39,18 @@ export const ProjectOptionGroup: React.FC<OptionGroupProps> = ({
...
+  isProjectSettingsPage = false,
```
sophstad, line 43: What is the purpose of this prop? Would it be possible for the ternary below to only check if a repoIdentifier/optionGroupLink exists?

src/constants/routes.ts
```diff
@@ -221,8 +221,14 @@ export const getProjectPatchesRoute = (projectId: string) =>
 
 export const getProjectSettingsRoute = (
   projectId: string,
-  tab: ProjectSettingsTabRoutes
-) => `${paths.project}/${projectId}/${PageNames.Settings}/${tab}`;
+  tab?: ProjectSettingsTabRoutes
+) => {
+  if (!tab) {
+    return `${paths.project}/${projectId}/${PageNames.Settings}/${ProjectSettingsTabRoutes.General}`;
+  }
```
sophstad, line 228: I'd rather not hardcode directing to the general tab here since that is defined elsewhere as the default page and is theoretically subject to change. If no tab is defined, can we direct to `${paths.project}/${projectId}/${PageNames.Settings}`? I think then the user will be redirected to the General page by our router.

sophstad (changes requested):

src/components/projectSelect/ProjectOptionGroup.tsx
sophstad: Eh tbh maybe not necessary since this is a project selector component 😄 your call!

src/components/projectSelect/ProjectOptionGroup.tsx
malikchaya2: The headers should be clickable on the project settings page, but shouldn't be clickable on the waterfall. I'll add a comment to make it more readable.

src/constants/routes.ts
malikchaya2: good point, fixed

src/components/projectSelect/ProjectOptionGroup.tsx
sophstad: Since the waterfall doesn't provide a repoIdentifier, would it not be sufficient to check if one exists?

src/pages/ProjectSettings.tsx
```diff
@@ -110,10 +112,18 @@ export const ProjectSettings: React.FC = () => {
     ? projectType === ProjectType.Project || repoData
     : repoData;
```
khelif96, line 113: I know this isn't related but this ternary is confusing to me. Couldn't you just check? 

```suggestion
    const hasData = projectData || repoData
```

src/pages/ProjectSettings.tsx
```diff
@@ -110,10 +112,18 @@ export const ProjectSettings: React.FC = () => {
...
 
+  const repoDisplay = `${repoData?.repoSettings?.projectRef?.owner}/${repoData?.repoSettings?.projectRef?.repo}`;
```
khelif96, line 115: Can we do this instead. It makes the final statement more readable.  
 ```suggestion
 const { repoSettings } = repoData || {};
 const { projectRef } = repoSettings || {};
 const { owner, repo } = projectRef || {}
  const repoDisplay = `${owner}/${repo}`;
```

File: src/pages/ProjectSettings.tsx [hunk 13]
```diff
@@ -110,10 +112,18 @@ export const ProjectSettings: React.FC = () => {
...
+
   return (
     <ProjectSettingsProvider>
       <SideNav aria-label="Project Settings">
-        <SideNavGroup header="Project" />
+        <DropdownContainer>
+          <ProjectSelect
+            selectedProjectIdentifier={repoDisplay || identifier}
```
khelif96, line 122: `repoDisplay` will always be defined.  since be default it is "/". identifier will never be used.

src/components/projectSelect/index.tsx
```diff
@@ -4,35 +4,57 @@ import Cookies from "js-cookie";
...
+  GetViewableProjectRefsQuery,
+  GetViewableProjectRefsQueryVariables,
 } from "gql/generated/types";
-import { GET_PROJECTS } from "gql/queries";
+import { GET_PROJECTS, GET_VIEWABLE_PROJECTS } from "gql/queries";
 import { ProjectOptionGroup } from "./ProjectOptionGroup";
 
 interface ProjectSelectProps {
   selectedProjectIdentifier: string;
+  isProjectSettingsPage?: boolean;
```
khelif96, line 18: This components implementation is still tied to closely to the projectSettings page.

src/components/projectSelect/index.tsx
```diff
@@ -62,18 +84,24 @@ export const ProjectSelect: React.FC<ProjectSelectProps> = ({
...
       options={allProjects}
       onChange={(projectIdentifier: any) => {
         Cookies.set(CURRENT_PROJECT, projectIdentifier, { expires: 365 });
-        history.push(getCommitsRoute(projectIdentifier));
+        history.push(getRoute(projectIdentifier));
       }}
       optionRenderer={(projectGroup, onClick) => (
         <ProjectOptionGroup
-          key={projectGroup.name}
+          key={projectGroup.name + projectGroup.projects.length}
```
khelif96, line 99: This won't prevent duplicate keys if this was your intent with this change. Since you can have 2 project groups with the same length.

src/components/projectSelect/ProjectOptionGroup.tsx
```diff
@@ -37,9 +39,19 @@ export const ProjectOptionGroup: React.FC<OptionGroupProps> = ({
...
 }) => (
   <OptionGroupContainer>
-    <Overline>{name}</Overline>
+    {/* if it's the project settings page and it's not the "" group, make the header clickable */}
+    {repoIdentifier === "" || !isProjectSettingsPage ? (
+      <Overline>{name} </Overline>
+    ) : (
+      <OverlineHover>
+        <Overline onClick={() => onClick(repoIdentifier)}>{name} </Overline>
```
khelif96, line 51: Can you add a `role="button"` To help the browser know this is a clickable button. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role

khelif96: Usually the linter would warn you about this but since its a styled element it can't really tell its a div.

src/components/projectSelect/ProjectOptionGroup.tsx
```diff
@@ -71,3 +83,10 @@ const ProjectContainer = styled.div`
 const OptionGroupContainer = styled.div`
   padding: ${size.xs};
 `;
+
+const OverlineHover = styled.div`
```
khelif96, line 87: Could we apply these styled directly to `Overline` instead of a wrapping component. 
You can replace `styled.div` with `css` and pass the styled into `Overline`

src/components/projectSelect/ProjectOptionGroup.tsx
khelif96: I agree with @sophstad  I think trying to make this a little more generic would be nice. It would also alleviate some of the conditional logic that happens below.

src/components/projectSelect/ProjectOptionGroup.tsx
```diff
@@ -37,9 +39,19 @@ export const ProjectOptionGroup: React.FC<OptionGroupProps> = ({
   name,
   projects,
   onClick,
+  repoIdentifier,
+  isProjectSettingsPage = false,
 }) => (
   <OptionGroupContainer>
-    <Overline>{name}</Overline>
+    {/* if it's the project settings page and it's not the "" group, make the header clickable */}
+    {repoIdentifier === "" || !isProjectSettingsPage ? (
```
khelif96, line 47: This logic can be replaced by a simple conditional prop.

src/components/projectSelect/index.tsx
```diff
@@ -83,3 +111,23 @@ export const ProjectSelect: React.FC<ProjectSelectProps> = ({
     />
   );
 };
+
+export const getRepoIdentifier = (projects) => {
+  if (!projects || projects.length === 0) {
+    return "";
+  }
```
khelif96, line 118: I'm not sure this case needs to be handled. This shouldn't be called if projects doesn't exist.

src/components/projectSelect/index.tsx
```diff
@@ -83,3 +111,23 @@ export const ProjectSelect: React.FC<ProjectSelectProps> = ({
     />
   );
 };
+
+export const getRepoIdentifier = (projects) => {
```
khelif96, line 115: Can you add a a type for projects.

src/components/projectSelect/index.tsx
```diff
@@ -83,3 +111,23 @@ export const ProjectSelect: React.FC<ProjectSelectProps> = ({
...
+  if (!projects || projects.length === 0) {
+    return "";
+  }
+  return projects[0].repoRefId;
+};
+
+export const getProjects = (nonFilteredProjects, filteredProjects) => {
```
khelif96, line 122: Same comment about adding a type for these params.

khelif96 (changes requested):

src/pages/ProjectSettings.tsx
sophstad: Not exactly, if the project type is an attached project we need to ensure that there is both project and repo data, not one or the other. I agree it could be improved though now that we are have a `ProjectType` defined. Something like
```js
 const hasData = projectType === ProjectType.AttachedProject
     ? projectData && repoData
     : projectData || repoData;
    ? projectType === ProjectType.Project || repoData
    : repoData;
```
🤷‍♀️

src/components/projectSelect/ProjectOptionGroup.tsx
malikchaya2: we only want the hover for the project settings page, we don't want it for the waterfall.

src/components/projectSelect/index.tsx
khelif96: the fact that there is a special prop called `isProjectSettingsPage` makes this component less generic then it should be. I think just renaming this to the condition it toggles will make it more generic. 
`canClickOnRepoGroup` or something like that.

src/components/projectSelect/ProjectOptionGroup.tsx
malikchaya2: It feels too risky to rely on that because it's just accessing the repoIdentifier through the project ref. Right now GET_PROJECTS which is how the waterfall page fetches projects doesn't get the repoRefId, but that may change down the line and whoever makes the change will have no reason to suspect that adding it will completely change the project dropdown.

src/components/projectSelect/ProjectOptionGroup.tsx
malikchaya2: I'm assuming you are referring to isProjectSettingsPage and not repoIdentifier?

src/pages/ProjectSettings.tsx
malikchaya2: thank you for spotting that!

src/components/projectSelect/index.tsx
malikchaya2: This naming feels less intuitive to me. 
If I was a new engineer reading this code, it would make more sense to me that we are using different queries to fetch the projects based on `isProjectSettingsPage` or `!isProjectSettingsPage`. It would be less apparent why we are doing that based on wether or not we want the headers to be clickable. Clickable headers are only one part of what distinguishes the project page dropdown component. The main difference is which projects are being fetched. 

I'll rename it for the ProjectOptionGroup component though.

src/components/projectSelect/index.tsx
malikchaya2: true. removing.

src/components/projectSelect/ProjectOptionGroup.tsx
malikchaya2: done

src/components/projectSelect/index.tsx
malikchaya2: True, but I don't want to take the risk of accessing an array index without knowing with certainty that it exists.

src/components/projectSelect/ProjectOptionGroup.tsx
malikchaya2: how so?

malikchaya2: I think I know what you mean, I committed the change that I think you meant.

src/components/projectSelect/ProjectOptionGroup.tsx
khelif96: It can still be a conditional style.

src/components/projectSelect/index.tsx
```diff
@@ -4,35 +4,63 @@ import Cookies from "js-cookie";
...
-  const allProjects = [
-    { name: "Favorites", projects: favoriteProjects },
-    ...projects,
-  ];
+  const favorites =
+    favoriteProjects.length > 0
+      ? { name: "Favorites", projects: favoriteProjects }
+      : {};
+
+  const allProjects = [favorites, ...projects];
```
khelif96, line 63:
```suggestion
  const favorites = { name: "Favorites", projects: favoriteProjects }

  const allProjects = favoriteProjects.length > 0 ? [favorites, ...projects] : projects;
```

khelif96: This avoids adding an empty object to allProjects.

khelif96: I question if this is even necessary? I would imagine we would always want to show people their favorite projects even on the project settings page.

src/components/projectSelect/index.tsx
```diff
@@ -62,18 +90,27 @@ export const ProjectSelect: React.FC<ProjectSelectProps> = ({
   return (
     <SearchableDropdown
       label="Project"
-      value={selectedProject?.displayName || selectedProject?.identifier}
+      value={
+        selectedProject?.displayName ||
+        selectedProject?.identifier ||
+        selectedProjectIdentifier
```
khelif96, line 96: Why are these 3 necessary? i would imagine the original code is enough. 
if `selectedProjectIdentifier` does not exist we shouldn't be showing it at all.

src/components/projectSelect/index.tsx
```diff
@@ -62,18 +90,27 @@ export const ProjectSelect: React.FC<ProjectSelectProps> = ({
...
+      }
       options={allProjects}
       onChange={(projectIdentifier: any) => {
```
khelif96, line 99: I know this was already here but can we type `projectIdentifier` as a string instead of any

src/components/projectSelect/index.tsx
khelif96: fair enough but in that case maybe we should error?

src/components/projectSelect/ProjectOptionGroup.tsx
```diff
@@ -37,9 +39,21 @@ export const ProjectOptionGroup: React.FC<OptionGroupProps> = ({
...
 }) => (
   <OptionGroupContainer>
-    <Overline>{name}</Overline>
+    {/* if it's the project settings page and it's not the "" group, make the header clickable */}
+    {canClickOnRepoGroup ? (
+      <OverlineHover>
+        <Overline role="button" onClick={() => onClick(repoIdentifier)}>
+          {name}
+        </Overline>
+      </OverlineHover>
```
khelif96, line 52: On smaller project names the clickable area is difficult to select.
image

src/pages/ProjectSettings.tsx
khelif96: Thanks for clarifying i wasn't sure how important projectType was. 
Could we use @sophstad  suggestion with a slight change so that hasData is a boolean and not an object.

```js
 const hasData = projectType === ProjectType.AttachedProject
   ? !!projectData && !!repoData
   : !!projectData || !!repoData;
```
```js
let projectData = { "some": "object"}
undefined
let repoData = { "other": "object" }
undefined
let hasData = true ? projectData && repoData : projectData || repoData;
undefined
hasData
{other: 'object'}
let hasData = false ? projectData && repoData : projectData || repoData;
undefined
hasData
{some: 'object'}
hasData = false ? !!projectData && !!repoData : !!projectData || !!repoData;
true
hasData
true
hasData = true ? !!projectData && !!repoData : !!projectData || !!repoData;
true
projectData = undefined
undefined
hasData = true ? !!projectData && !!repoData : !!projectData || !!repoData;
false
hasData = false ? !!projectData && !!repoData : !!projectData || !!repoData;
true
repoData = undefined
undefined
hasData = false ? !!projectData && !!repoData : !!projectData || !!repoData;
false
```

khelif96 (changes requested):
Tested this on staging
This does not handle projects with large names. 
image

There are also some projects with no name clicking on them causes the page to error. 
image

src/components/projectSelect/ProjectOptionGroup.tsx
malikchaya2: I actually can't add it to Overline because we don't own it, it's part of leafygreen typography

src/components/projectSelect/index.tsx
malikchaya2: I'm not sure which part you are questioning is necessary. We do show people their favorite projects even on the project settings page.

src/components/projectSelect/index.tsx
malikchaya2: It's necessary for repos. We set selectedProjectIdentifier to be the repo display name.

src/components/projectSelect/index.tsx
malikchaya2: your suggestion results in type errors later on.

src/pages/ProjectSettings.tsx
malikchaya2: Can you please create a ticket for it? This is not related to this PR so I'd prefer to keep the work separate.

src/components/projectSelect/index.tsx
malikchaya2: nvm, it doesn't

src/components/projectSelect/index.tsx
malikchaya2: I can, but I don't know that it is necessary, because the user wouldn't know what to do with an error like that.

src/components/projectSelect/index.tsx
malikchaya2: image
string doesn't work as a type here

src/components/projectSelect/ProjectOptionGroup.tsx
malikchaya2: I added some padding to the hover.

src/components/projectSelect/index.tsx
```diff
@@ -13,6 +13,12 @@ import {
 import { GET_PROJECTS, GET_VIEWABLE_PROJECTS } from "gql/queries";
 import { ProjectOptionGroup } from "./ProjectOptionGroup";
 
+type project = {
```
sophstad, line 16: [nit] types are capitalized
```suggestion
type Project = {
```

src/pages/ProjectSettings.tsx
```diff
@@ -118,18 +118,27 @@ export const ProjectSettings: React.FC = () => {
     ? projectType === ProjectType.Project || repoData
     : repoData;
 
-  const repoDisplay = `${repoData?.repoSettings?.projectRef?.owner}/${repoData?.repoSettings?.projectRef?.repo}`;
+  const { repoSettings } = repoData || {};
+  const { projectRef } = repoSettings || {};
+  const { owner, repo } = projectRef || {};
+
+  const repoDisplay = owner && repo ? `${owner}/${repo}` : identifier;
```
sophstad, line 125: Since this variable is passed to `selectedProjectIdentifier`, it is used to determine the display name for a repo or project, not just a repo. And if we're viewing a project, I think it'd be preferable to show the `identifier` instead of owner/repo, no? Currently when looking at http://localhost:3000/project/evergreen/settings/general it shows owner/repo when "evergreen" is what I'd expect to see there.

src/components/projectSelect/index.tsx
```diff
@@ -112,14 +120,17 @@ export const ProjectSelect: React.FC<ProjectSelectProps> = ({
   );
 };
 
-export const getRepoIdentifier = (projects) => {
+export const getRepoIdentifier = (projects: project[]) => {
```
sophstad, line 123: [nit] `getRepoId` may make more sense since repos don't have an identifier

sophstad (changes requested):

src/pages/ProjectSettings.tsx
```diff
@@ -116,9 +118,22 @@ export const ProjectSettings: React.FC = () => {
...
     <ProjectSettingsProvider>
       <SideNav aria-label="Project Settings">
+        <DropdownContainer>
+          <ProjectSelect
+            selectedProjectIdentifier={repoDisplay}
+            getRoute={getProjectSettingsRoute}
+            isProjectSettingsPage
+          />
+        </DropdownContainer>
         <SideNavGroup header="Project" />
```
sophstad, line 137: These sidebar elements can also go in the same side nav group:

```js
<SideNavGroup>
  <DropdownContainer>
    <ProjectSelect ... />
  </DropdownContainer
  <CreateProjectModal ... />
</SideNavGroup>

File: src/pages/ProjectSettings.tsx (hunk 11 above)
Comment (khelif96) [reply src/pages/ProjectSettings.tsx:119]: Done :) https://jira.mongodb.org/browse/EVG-16515

File: src/components/projectSelect/index.tsx (hunk 22 above)
Comment (khelif96) [reply src/components/projectSelect/index.tsx]: I'm questioning the code changes around the favoriteProjects logic in that case.

File: src/components/projectSelect/index.tsx (hunk 24 above)
Comment (khelif96) [reply src/components/projectSelect/index.tsx:98]: Looks like we need some better typescript generic handling :) 
https://jira.mongodb.org/browse/EVG-16516

File: src/components/projectSelect/index.tsx (hunk 19 above)
Comment (malikchaya2) [reply src/components/projectSelect/index.tsx:126]: perfect, I like that better than an error toast to the user. I'll do that.

File: src/pages/ProjectSettings.tsx (hunk 29 above)
Comment (malikchaya2) [reply src/pages/ProjectSettings.tsx:137]: It matches the design better when it's outside of it
image

File: src/components/projectSelect/index.tsx (hunk 24 above)
Comment (malikchaya2) [reply src/components/projectSelect/index.tsx:98]: :)

File: src/components/projectSelect/index.tsx (hunk 19 above)
Comment (malikchaya2) [reply src/components/projectSelect/index.tsx]: just added this

File: src/components/projectSelect/index.tsx [hunk 30]
```diff
@@ -83,3 +119,30 @@ export const ProjectSelect: React.FC<ProjectSelectProps> = ({
...
   );
 };
+
+const { reportError } = errorReporting;
+
+export const getRepoId = (projects: Project[]) => {
+  if (!projects || projects.length === 0) {
+    reportError(
+      new Error("No projects provided when fetching repo id")
+    ).warning();
```
khelif96, line 129: Also needs a return to exit.

src/components/projectSelect/ProjectOptionGroup.tsx
```diff
@@ -71,3 +85,11 @@ const ProjectContainer = styled.div`
 const OptionGroupContainer = styled.div`
   padding: ${size.xs};
 `;
+
+const OverlineHover = styled.div`
+  :hover {
+    cursor: pointer;
+    background-color: ${gray.light1};
+    padding: ${size.xs};
```
khelif96, line 93: The padding on hover is a little jarring maybe it should always be there.

src/components/projectSelect/ProjectOptionGroup.tsx
khelif96: If you replace styled.div with `css` from `@emotion/core` and pass it into Overline as part of the `css` prop

khelif96 (changes requested):
image
Adding a HR like the designs have would def make this look better.

image

This def should not have any horizontal scroll. Can we add a word-break to the Project Options to avoid this.

src/components/projectSelect/index.tsx
malikchaya2: done

src/components/projectSelect/index.tsx
```diff
@@ -83,3 +119,31 @@ export const ProjectSelect: React.FC<ProjectSelectProps> = ({
     />
   );
 };
+
+const { reportError } = errorReporting;
+
+export const getRepoId = (projects: Project[]) => {
+  if (!projects || projects.length === 0) {
+    reportError(
+      new Error("No projects provided when fetching repo id")
```
malikchaya2, line 128: @khelif96 this error shows up in the test. I don't think we should error here because there is probably just a point where it's undefined and it'll probably keep logging.

khelif96: If thats the case then it would also show up in production. So maybe we need to guard against this case better?

malikchaya2: I don't think it'll ever be a problem, it'll be there by the time the page is loaded.

sophstad (approved):

PR #1133, opened 2022-03-03
EVG-16356 Fix task grouping with build variant filter applied

khelif96:
[EVG-16356](https://jira.mongodb.org/browse/EVG-16356)

### Description 
Noticed a edge case where a lone variant filter would cause the task status icons to appear. 
### Screenshots
Before:
image

After:
image

### Testing 
Wrote tests for this edge case and tested

### DIFF
--- a/src/pages/commits/utils.ts
+++ b/src/pages/commits/utils.ts
@@ -89,6 +89,8 @@ const generateBuildVariantOptionsForTaskIconsFromState = (
           FAILED_STATUSES
         );
       }
+    } else {
+      statusesToShow = FAILED_STATUSES;
     }
   } else {
     statusesToShow = FAILED_STATUSES;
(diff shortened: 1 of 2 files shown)


src/pages/commits/utils.ts
```diff
@@ -89,6 +89,8 @@ const generateBuildVariantOptionsForTaskIconsFromState = (
           FAILED_STATUSES
         );
       }
+    } else {
+      statusesToShow = FAILED_STATUSES;
     }
   } else {
     statusesToShow = FAILED_STATUSES;
```
minnakt, line 96: Do you think it would be possible to change this block of `if-else`s to this?

```
  if (hasTasks) {
    statusesToShow = filterState.statuses;
  } else if (hasStatuses) {
    const onlyHasNonFailingStatuses =
      arrayIntersection(filterState.statuses, FAILED_STATUSES).length === 0;
    if (onlyHasNonFailingStatuses) {
      shouldShowTaskIcons = false;
    } else {
      statusesToShow = arrayIntersection(filterState.statuses, FAILED_STATUSES);
    }
  } else {
    statusesToShow = FAILED_STATUSES;
  }
```

This would still pass the tests and take away one of the `else` blocks, but all the scenarios can get confusing so I might have missed something 😵‍💫

minnakt (approved):
Nice! 🐛

PR #1134, opened 2022-03-07
EVG-16183 Add mainline commits e2e tests 

khelif96:
[EVG-16183](https://jira.mongodb.org/browse/EVG-16183)

### Description 
Add e2e tests for much of the mainline commits functionality. Wrote task history tests as well but have them skipped since the links are not accessible.
They do work however 
image


src/pages/commits/CommitsWrapper.tsx
```diff
@@ -19,27 +26,40 @@ import {
...
+  const onChangeChartType = (chartType: ChartTypes): void => {
+    updateQueryParams({
+      [ChartToggleQueryParams.chartType]: chartType,
+    });
+  };
+
+  const chartType =
+    (getString(parsed[ChartToggleQueryParams.chartType]) as ChartTypes) ||
+    DEFAULT_CHART_TYPE;
+
```
minnakt, line 62: I think it would be possible to have this entire logic (toggling the chart, updating query params) in `ChartToggle.tsx`. This might be a slight improvement because then all of the logic would be localized to where the actual code is for the chart toggle

cypress/integration/projectHealth/commits.ts
```diff
@@ -0,0 +1,257 @@
...
+    });
+    it("applying a build variant filter should show all task statuses by default", () => {
+      cy.getInputByLabel("Add New Build Variant Filter").type("Ubuntu");
+      cy.getInputByLabel("Add New Build Variant Filter").type("{enter}");
+      cy.dataCy("filter-badge").should("exist");
+      cy.dataCy("filter-badge").should("have.length", 1);
+      cy.dataCy("filter-badge").should("have.text", "buildVariants : Ubuntu");
+      cy.location("search").should("contain", "?buildVariants=Ubuntu");
+      cy.dataCy("grouped-task-status-badge").should("exist");
+      cy.dataCy("grouped-task-status-badge").should("have.length", 5);
```
minnakt, line 120: I kind of wonder if we need all of these `exist` assertions before these `have.length` assertions. I would think if the assertion `have.length` passes, then that would implicitly mean the element exists (and therefore there would be no need to test for it)

minnakt (commented):
These tests are looking great 💥, tysm!!!

src/pages/commits/CommitsWrapper.tsx
khelif96: yeah i'm not a big fan of this pattern either. I think extracting this type of logic into a hook reminiscent of `useState` would be the best approach. 
`useQueryState()` or something like that 
```
const [chartType, setChartType] = useQueryState(ChartToggleQueryParams.chartType, DEFAULT_CHART_TYPE);
```

How do you feel about some thing like that it could reduce allot of this boiler plate logic we use?

khelif96: it would operate in the same way as `useState` and trigger rerenders in the same way.

minnakt: I think it's a good idea and worth investigating in a separate ticket, it would definitely help if it can centralize most or all of the URL logic we copy/paste everywhere 😲

cypress/integration/projectHealth/commits.ts
khelif96: When i wrote them i intended them to be a safe guard for trying to click on elements that may not exist yet. Since cypress tends to wait for elements to render when it makes assertions. I may have gone a little overboard :)

minnakt (approved):
Sorry I missed this yesterday, looks great!!! 😼

PR #1135, opened 2022-03-07
EVG-16117: Add Git Tag field

sophstad:
EVG-16117

### Description 
- Add git tag field
- Edit general alias implementation to make it more modular
  - Remove AliasRow component
  - Allow any array form to use expandable cards (not tied to aliases)
- Improve UX so that toggling between regex/tag input does not clear the input

### Screenshots
image

### Testing 
- Added unit tests
- Updated Cypress tests


ablack12:
I tried saving this page with defined git tag aliases and received an error code.
image

ablack12 (approved):
I imagine this isn't related to your change, and I'm not sure it's something we can every fix (or at least not as part of the POC 😄 ) but I noticed that when I type in the text box it takes a full beat for what I type to become visible; not sure if this is a debounce thing or etc but wanted to call it out.

sophstad:
> I imagine this isn't related to your change, and I'm not sure it's something we can every fix (or at least not as part of the POC 😄 ) but I noticed that when I type in the text box it takes a full beat for what I type to become visible; not sure if this is a debounce thing or etc but wanted to call it out.

Yeah, I think this is less of a debounce issue and more of a "there's a lot of stuff on this page" problem 😢. I made EVG-16512 to look into it

cypress/integration/projectSettings/project_settings.ts
```diff
@@ -268,6 +264,32 @@ describe("Project Settings when not defaulting to repo", () => {
...
+
+    it("Allows adding a git tag alias", () => {
+      cy.dataCy("git-tag-enabled-radio-box").children().first().click();
+      cy.dataCy("add-button").contains("Add Git Tag").parent().click();
+      cy.dataCy("git-tag-input").type("myGitTag");
+      cy.dataCy("remote-path-input").type("./evergreen.yml");
+    });
+
+    it("Updates the Require Signed field", () => {
+      cy.dataCy("require-signed-radio-box").children().first().click();
```
khelif96, line 281: There is no assertion here.

src/components/SpruceForm/LeafyGreenWidgets.tsx
```diff
@@ -299,6 +303,51 @@ export const LeafyGreenTextArea: React.FC<WidgetProps> = ({
...
+  value,
+}) => {
+  const {
+    "aria-controls": ariaControls,
+    "data-cy": dataCy,
+    enumOptions,
+  } = options;
+
+  if (!Array.isArray(enumOptions)) {
+    console.error("Non-Array passed into Segmented Control");
```
khelif96, line 320: We should log this to bugsnag instead using `reportError`.

src/components/SpruceForm/LeafyGreenWidgets.tsx
```diff
@@ -299,6 +303,51 @@ export const LeafyGreenTextArea: React.FC<WidgetProps> = ({
...
+    return null;
+  }
+
+  const idPrefix = id.substring(0, id.lastIndexOf("_"));
```
khelif96, line 324: i'm curious about this transformation?  Why is this necessary?

sophstad: Good point, I was experimenting with programmatically computing the IDs that the segmented control controls, but it's way more straightforward to specify them manually. Removed!

khelif96 (approved):

PR #1145, opened 2022-03-10
EVG-16470 Fix crashes when we fail to load column headers

khelif96:
[EVG-16470](https://jira.mongodb.org/browse/EVG-16470)

### Description 
Fix crashes that occurred when trying to load the history pages for tasks that have no prior commits with the same name.
Add some error reporting when we encounter this scenario. We have removed all links to these pages in https://jira.mongodb.org/browse/EVG-16384


minnakt (approved):
🐛

PR #1147, opened 2022-03-10
EVG-15086: Add Virtual Workstation page

sophstad:
EVG-15086

### Description 
- Add Virtual Workstation page
- The [designs](https://www.figma.com/file/Z0rfsPjVKglRq9baDhsSux/Project-Settings-Sections?node-id=212%3A1328) have a slick draggable interface for this page to order shell commands, but for now I'm using RJSF's default button-based ordering interface. Implementing dragging felt like a separate chunk of work so I opened a new ticket for it EVG-16525

### Screenshots
image

### Testing 
- Update unit and integration tests


src/components/ProjectSettingsTabs/AccessTab/getFormSchema.ts
```diff
@@ -79,6 +79,7 @@ export const getFormSchema = (
       admins: {
         "ui:addButtonText": "Add Username",
         "ui:description": getAdminsDescription(projectType),
+        "ui:orderable": false,
```
sophstad, line 82: I just implemented a way of ordering items in a list via arrow buttons -- RJSF defaults this setting to true, but most of our pages don't use it, so I had to set them all to false.

src/components/ProjectSettingsTabs/Header.tsx
```diff
@@ -92,7 +94,6 @@ export const Header: React.FC<Props> = ({
           });
 
     save(newData, mapRouteToSection[tab]);
-    saveTab(tab);
```
sophstad, line 95: This saveTab function is used to disable the save button; we only want to call it if the save operation succeeded

src/components/ProjectSettingsTabs/AccessTab/getFormSchema.ts
minnakt: This might be a silly question but is there a reason we couldn't do something like `const isOrderable = uiSchema["ui:orderable"] ?? false` and rely on that for rendering?

src/components/SpruceForm/FieldTemplates.tsx
```diff
@@ -91,9 +100,27 @@ const ArrayItem: React.FC<
       {children}
     </ExpandableCard>
   ) : (
-    <ArrayItemRow key={index}>
+    <ArrayItemRow key={index} border={border} index={index}>
+      {(hasMoveUp || hasMoveDown) && !readonly && (
```
malikchaya2, line 104: Why are we using `readonly` here instead of `isDisabled`?

src/components/SpruceForm/FieldTemplates.tsx
```diff
@@ -106,8 +133,27 @@ const TitleWrapper = styled.span`
...
+
+  > :not(:last-of-type) {
+    margin-bottom: ${size.xs};
+  }
+`;
+
+const ArrayItemRow = styled.div<{ border: boolean; index: number }>`
   display: flex;
+  ${({ border, index }) =>
+    border && index === 0 && `border-top: 1px solid ${gray.light1}`};
```
malikchaya2, line 150: size.xxs is 1px

src/pages/projectSettings/Tabs.tsx
```diff
@@ -45,6 +45,10 @@ export const ProjectSettingsTabs: React.FC<Props> = ({
   const { setInitialData } = useProjectSettingsContext();
 
   const projectId = projectData?.projectRef?.id;
+  const repoId = repoData?.projectRef?.id;
+  const {
+    projectRef: { identifier },
+  } = projectData ?? { projectRef: {} };
```
minnakt, line 51: Is there a reason this can't be `const identifier = projectData?.projectRef?.identifier`?

src/components/SpruceForm/FieldTemplates.tsx
khelif96: size.xxs is 4px :)

cypress/integration/projectSettings/project_settings.ts
```diff
@@ -529,4 +559,35 @@ describe("Project Settings when defaulting to repo", () => {
...
+        cy.get("input").should("be.disabled");
+        cy.get("textarea").should("be.disabled");
+      });
+    });
+
+    it("Allows overriding without adding a command", () => {
+      cy.dataCy("commands-override-radio-box")
+        .find("input")
+        .first()
+        .parent()
```
khelif96, line 581: i wonder does `cy.getInputByLabel` work instead?. You should be able to target the exact input element you want as long as its label is defined correctly (Which i assume it would because accessibility)

src/components/ProjectSettingsTabs/VirtualWorkstationTab/CommandRow.tsx
```diff
@@ -0,0 +1,29 @@
+import styled from "@emotion/styled";
+import { SpruceFormProps } from "components/SpruceForm";
```
khelif96, line 2: Was there a ticket to move these files to a more page specific directory? I remember us discussing it a while ago.

src/components/ProjectSettingsTabs/VirtualWorkstationTab/CommandRow.tsx
```diff
@@ -0,0 +1,29 @@
...
+};
+
+const LeftColumn = styled.div`
+  padding-right: ${size.s};
+`;
+
+const RowContainer = styled.div`
+  display: flex;
+
+  > div {
```
khelif96, line 25: Wouldn't this style apply to both divs anyways im curious why we don't apply it to them directly instead?

src/components/SpruceForm/FieldTemplates.tsx
malikchaya2: ah my bad, I misread. thanks!

src/components/SpruceForm/LeafyGreenWidgets.tsx
```diff
@@ -19,9 +19,18 @@ import ElementWrapper from "./ElementWrapper";
...
   // Don't display empty input errors as these are too visually noisy
-  rawErrors?.filter((err) => !err.startsWith("should match format")) ?? [];
+  const errorMsgs = [
+    "should match format",
+    "should NOT be shorter than 1 characters",
+  ];
+  return (
+    rawErrors?.filter((err) => !errorMsgs.some((msg) => err.startsWith(msg))) ??
+    []
+  );
```
khelif96, line 31: I was wondering how you felt about RJSF's pattern for handling custom error messages. https://react-jsonschema-form.readthedocs.io/en/latest/usage/validation/#custom-error-messages. 
It seems to case on the validators as opposed to the error messages they output.

ablack12 (approved):

src/components/SpruceForm/FieldTemplates.tsx
sophstad: I figured there could be an instance where we'd want to show disabled buttons for an item that is disabled but not read-only. To me read-only definitely implies no interaction and I found the page was less cluttered hiding them, but I left them there when a form element is `disabled` for flexibility (these are two different settings in the library). They definitely could be hidden though.

src/components/ProjectSettingsTabs/VirtualWorkstationTab/CommandRow.tsx
sophstad: Yes! EVG-16542 I'll pull it into next sprint 👍

khelif96 (approved):
Great Work! 💯

PR #1148, opened 2022-03-11
EVG-16367 Add canCreateProject check to the createProject Modal

malikchaya2:
[EVG-16367](https://jira.mongodb.org/browse/EVG-16367)

### Description 
This restricts the createProject button and modal to people who have the right permissions.


src/components/ProjectSettingsTabs/CreateProjectModal.tsx
```diff
@@ -70,7 +72,16 @@ interface Props {
...
-  const isAdmin = true; // todo after EVG-16353
+
+  // USER PERMISSIONS QUERY
+  const { data: userPermissionsData } = useQuery<GetUserPermissionsQuery>(
+    GET_USER_PERMISSIONS
+  );
+  const { user } = userPermissionsData || {};
+  const { permissions } = user || {};
+  const canCreateProject = permissions?.canCreateProject;
+  console.log("userPermissionsData: ", userPermissionsData);
```
khelif96, line 83: console.log

src/components/ProjectSettingsTabs/CreateProjectModal.tsx
```diff
@@ -109,13 +120,13 @@ export const CreateProjectModal: React.FC<Props> = ({ owner, repo }) => {
     refetchQueries: ["ProjectSettings", "RepoSettings"],
   });
 
-  if (!isAdmin) {
+  if (!canCreateProject) {
```
khelif96, line 123: can we fallback to something?

src/gql/queries/get-user-permissions.graphql
```diff
@@ -0,0 +1,9 @@
+query GetUserPermissions {
+  user {
+    userId
+    permissions {
+      canCreateProject
+      userId
```
khelif96, line 6: userId isn't needed here.

src/components/ProjectSettingsTabs/CreateProjectModal.tsx
malikchaya2: Can you please elaborate a bit more? What kind of fallback? If someone doesn't have permissions, we don't want them to see the button. Do you mean like an empty div or something?

src/components/ProjectSettingsTabs/CreateProjectModal.tsx
malikchaya2: thanks!

src/components/ProjectSettingsTabs/CreateProjectModal.tsx
khelif96: Sorry i missed the context of this change i didn't realize this was for the button that toggles the modal.

khelif96 (approved):

ablack12:
Not reviewing since there isn't a visual change to check on but let me know if that changes.

src/gql/generated/types.ts
```diff
@@ -3909,6 +3909,15 @@ export type GetUserConfigQuery = {
   }>;
 };
 
+export type GetUserPermissionsQueryVariables = Exact<{ [key: string]: never }>;
+
+export type GetUserPermissionsQuery = {
+  user: {
+    userId: string;
+    permissions: { canCreateProject: boolean; userId: string };
```
sophstad, line 3917: Looks like `yarn codegen` needs to be run again; this `userId` field was removed from the query.

sophstad (changes requested):

src/gql/generated/types.ts
```diff
@@ -3912,10 +3912,7 @@ export type GetUserConfigQuery = {
 export type GetUserPermissionsQueryVariables = Exact<{ [key: string]: never }>;
 
 export type GetUserPermissionsQuery = {
-  user: {
-    userId: string;
```
sophstad: I think this `userId` should stay to help with Apollo caching (maybe @khelif96 can confirm)

khelif96: Yes userId should stay for the user object. My earlier comment was only in reference to the permissions object.

malikchaya2: restored

sophstad (approved):

PR #1149, opened 2022-03-11
EVG-16316: Prevent unnecessary PatchTasks query

minnakt:
[EVG-16316](https://jira.mongodb.org/browse/EVG-16316)

### Description 
When a user loads the version page, there is an initial request to fetch `PatchTasks` with no sorts or filters applied. The page then updates through `useEffect` to add the default sorts, which cancels the initial request and dispatches a separate request. 

This PR prevents that initial request by adding a `skip` param to the Apollo query which skips the query if there are no query params in the URL. There should be query params in the URL at any given time, because applying/removing any filter will always add a `page` query param and `Clear all filters` always applies the default sort.

### Note
I rewrote some of the tab logic for because it was causing problems due to race conditions (parent & child components trying to update the URL at the same time). The new code should make it so that first the URL is updated in the parent component, then the state is updated to trigger URL updates in the child components.

### Screenshots
No visible changes

### Testing 
- Manually checked network request on page load with and without `skip` applied, saw two requests for `PatchTasks` without `skip` applied, and only one request for `PatchTasks`  with `skip` applied
- Add tests for tabs in `page_tabs.ts` (Cypress)


minnakt:
nvm don't review yet; there's one scenario where this is an issue 😔

khelif96 (approved):

PR #1155, opened 2022-03-16
EVG-16475: Disable Previous Commits Selector button by default

minnakt:
[EVG-16475](https://jira.mongodb.org/browse/EVG-16475)

### Description 
The Go button for the Previous Commits Selector is currently enabled by default, which is problematic because the user can click on it while a task is still being fetched. The task link is only correct after the task has been fetched, so the button should be disabled by default instead.

This PR also entirely removes the Previous Commits Selector on execution task pages.

### Screenshots
(First tab is task that doesn't exist on parent, second tab is task that exists on parent, and third tab is an execution task.)

https://user-images.githubusercontent.com/47064971/158634949-408c84b8-318f-4891-8205-35177d4fd18c.mov

### Testing 
- Added more tests in `PreviousCommits.test.tsx` for patch specific and mainline commits specific scenarios

### DIFF
--- a/src/pages/task/actionButtons/reducer.ts
+++ b/src/pages/task/actionButtons/reducer.ts
@@ -18,6 +18,9 @@ interface State {
   hasFetchedLastExecuted: boolean;
 }
 
+// a link cannot be null, so it's common to use # as a substitute.
+const nullLink = "#";
+
 // Note that the state doesn't include a field for shouldFetchParent / hasFetchedParent as the parent task
 // is always fetched on render.
 export const initialState: State = {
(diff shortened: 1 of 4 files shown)


src/pages/task/actionButtons/reducer.ts
```diff
@@ -18,15 +18,18 @@ interface State {
   hasFetchedLastExecuted: boolean;
```
khelif96, line 18: Sorry i hadn't noticed this before this should be in a folder on its own. 
The PreviousCommits logic and tests should be in the same folder. They can live as a  sub directory in actionButtons.

khelif96 (approved):
LGTM

PR #1156, opened 2022-03-16
EVG-16512: Improve form performance

sophstad:
EVG-16512

### Description 
- Update RJSF to improve project settings form performance, particularly on GitHub/Commit Queue page. This made such a difference that I didn't look into any other approaches, but we can revisit if things start to get sluggish again.
- Remove `showErrors` option for two reasons:
  1. The [`transformErrors` function](https://github.com/evergreen-ci/spruce/blob/000302e68d2106b7c6ebdae83789a5d2816a9e20/src/components/SpruceForm/errors.ts#L3) supresses all of the errors that needed this option  
  2. RJSF added a `hideError` option, so we should use the supported way if we need this in the future

### Testing 
- Manual testing

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
     "@leafygreen-ui/toggle": "7.0.4",
     "@leafygreen-ui/tooltip": "6.3.1",
     "@leafygreen-ui/typography": "8.1.0",
-    "@rjsf/core": "3.2.1",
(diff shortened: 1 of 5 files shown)


khelif96 (approved):
Glad you were able to find a fix for the performance problems!

minnakt (approved):
Yessss upgrading stuff 😎

PR #1159, opened 2022-03-18
EVG-13488: Surface recently terminated hosts in SpawnHostTable

minnakt:
[EVG-13488](https://jira.mongodb.org/browse/EVG-13488)

### Description 
This PR makes it so that users can also see their hosts that have been terminated in the past 5 minutes on the My Hosts page.

### Screenshots
https://user-images.githubusercontent.com/47064971/159028556-f7680001-d839-4e87-8822-87b690331357.mov

### Testing 
- Asked Bynn if he would hover over the info icon at the top and he said "No" so I put it in the table instead
- Manually 

### Evergreen PR 
- evergreen-ci/evergreen/pull/5499


src/components/Spawn/Layout.tsx
```diff
@@ -101,3 +101,5 @@ export const PaddedButton = styled(Button)`
   margin-right: ${size.xxs};
   flex-grow: 0;
 `;
+
+export const tooltipWidth = "250px";
```
minnakt, line 105: I gave a width to all of the tooltips because [PD-1542](https://jira.mongodb.org/browse/PD-1542) is still a problem, leading to weird behaviors like:

Screen Shot 2022-03-15 at 3 57 27 PM

Screen Shot 2022-03-15 at 3 57 36 PM

src/components/Spawn/IconTooltip.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { size } from "constants/tokens";
+
+interface IconTooltipProps {
+  tooltipText: string;
+  iconType: string;
+  color: string;
+  shade: string;
+}
+
+export const IconTooltip: React.FC<IconTooltipProps> = ({
```
khelif96, line 14: I like this pattern and it will also be used for Projects Track Multiple Branches. We should move this to a  more generic location.
image

src/components/Spawn/IconTooltip.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { uiColors } from "@leafygreen-ui/palette";
+import Tooltip from "@leafygreen-ui/tooltip";
+import Icon from "components/Icon";
+import { size } from "constants/tokens";
+
+interface IconTooltipProps {
+  tooltipText: string;
+  iconType: string;
+  color: string;
+  shade: string;
```
khelif96, line 11: I think this would make more sense for it to accept any color as a prop as opposed to relying on keys of uiColors.

khelif96 (changes requested):
This looks great just had a few comments

src/components/PageTitle.tsx
```diff
@@ -61,7 +61,8 @@ export const PageTitle: React.FC<Props> = ({
 const BadgeWrapper = styled.div<TitleTypographyProps>`
   display: inline-flex;
   margin-left: ${({ size }) => (size === "large" ? tokenSize.m : tokenSize.s)};
-  vertical-align: text-bottom;
+  vertical-align: ${({ size }) =>
+    size === "large" ? "middle" : "text-bottom"};
```
khelif96, line 65: Since this affects a bunch of pages can you spot check this to make sure nothing looks weird on the other pages its used.

src/components/IconTooltip.tsx
```diff
@@ -0,0 +1,44 @@
+import styled from "@emotion/styled";
+import Tooltip from "@leafygreen-ui/tooltip";
+import Icon from "components/Icon";
+import { size } from "constants/tokens";
+
+interface IconTooltipProps {
+  tooltipText: string;
+  iconType: string;
+  color: string;
```
khelif96, line 9: how do you feel about making this optional and have it fallback to black.

src/components/IconTooltip.tsx
```diff
@@ -0,0 +1,44 @@
+import styled from "@emotion/styled";
+import Tooltip from "@leafygreen-ui/tooltip";
+import Icon from "components/Icon";
+import { size } from "constants/tokens";
+
+interface IconTooltipProps {
+  tooltipText: string;
+  iconType: string;
```
khelif96, line 8: this can be typed as `keyof glyphs` 
glyphs is exported by `components/Icon`

khelif96 (approved):

src/components/PageTitle.tsx
minnakt: The only page that uses `size = large` is a Host page. So every other page that uses `PageTitle` should look exactly the same as deployed Spruce.

PR #1160, opened 2022-03-18
EVG-16380 Jump to commit on history tables

khelif96:
[EVG-16380](https://jira.mongodb.org/browse/EVG-16380)

### Description 
Support the ability to jump to a commit. If a skip order number is passed in  the history table will continuously load in commits until the desired commit is loaded. We do this to support scrolling back to previous commits. I will likely split up some of the unrelated history page refactor work into a separate pr to make reviewing easier. 

- [X] Can specify a commit number and it will jump to said commit
- [X] Can Scroll back up to view previous commits in accordance to https://jira.mongodb.org/browse/EVG-5536 
- [x] Support for inactive commits (This is debatable since they are collapsed by default.)
- [x] Wrote Unit Tests
- [x] Wrote Cypress Tests
- [X] Refactored pages to make them easier to read. 
- [x] Fixed Folded Commits state management to make them more resilient to virtualization.

### Screenshots

https://user-images.githubusercontent.com/4605522/159084886-68a3fbef-9365-4112-8764-23eb020ea322.mov

### Testing 
Not yet can't get local-evergreen working. But tested on staging and prod

### Evergreen PR 


src/components/HistoryTable/HistoryTable.tsx
```diff
@@ -38,7 +40,39 @@ const HistoryTable: React.FC<HistoryTableProps> = ({
...
-  }, [recentlyFetchedCommits]);
+  }, [processedCommitCount]);
+
+  useEffect(() => {
+    if (processedCommitCount > 0) {
+      if (listRef.current) {
+        listRef.current.resetAfterIndex(0);
+      }
+    }
+  }, [processedCommitCount]);
```
minnakt, line 51: can this `useEffect` be combined with the one above it?

src/components/HistoryTable/types.ts
```diff
@@ -1,12 +1,30 @@
...
+      rowHeight: number;
+      date: Date;
+      selected: boolean;
+    }
+  | {
+      type: rowType.DATE_SEPARATOR;
+      date: Date;
+      rowHeight: number;
+      selected: boolean;
+    }
```
minnakt, line 19: Would it make sense/is it possible to exclude the `selected` field in the date separator type? we always designate this as `selected: false` which means it will never be selectable, so maybe we can omit this field?

src/components/HistoryTable/utils.ts
```diff
@@ -31,24 +40,38 @@ export const processCommits = (
...
         processedCommits.push({
           type: rowType.COMMIT,
           commit: version,
           date: version.createTime,
+          selected: version.order === selectedCommitOrder,
           rowHeight: COMMIT_HEIGHT,
         });
+        if (version.order === selectedCommitOrder) {
+          selectedCommitRowIndex = processedCommits.length - 1;
+        }
```
minnakt, line 66: If I had to guess I think the `selectedCommitRowIndex` is off by one? This only happens from the `/commits` page, not on any Task History buttons

https://user-images.githubusercontent.com/47064971/160643333-fdc26631-4d94-492b-b802-7a8499a9a2a4.mov

src/components/HistoryTable/utils.ts
```diff
@@ -31,24 +40,38 @@ export const processCommits = (
           type: rowType.COMMIT,
           commit: version,
           date: version.createTime,
+          selected: version.order === selectedCommitOrder,
           rowHeight: COMMIT_HEIGHT,
         });
+
+        if (version.order === selectedCommitOrder) {
+          selectedCommitRowIndex = processedCommits.length - 1;
+        }
```
minnakt, line 49: Since this code is repeated on line 64 I believe you can pull it out to the top level if statement `if (commitType === rowType.COMMIT)`. 

Tbh you might even be able to move the whole thing up with something like this:

Screen Shot 2022-03-29 at 12 19 50 PM

src/components/HistoryTable/historyTableContextReducer.ts
```diff
@@ -9,7 +9,9 @@ type Action =
   | { type: "prevPageColumns" }
```
minnakt, line 9: I don’t completely understand what triggers this bug/if it’s even worth dealing with but when you click on on the same build variant from two different versions, it won't jump to commit the second time

https://user-images.githubusercontent.com/47064971/160662041-204e4c3d-eec8-4fd6-9d27-cf73ed77f783.mov

minnakt (commented):
looking great! 😎

src/components/HistoryTable/HistoryTableContext.tsx
```diff
@@ -72,21 +87,23 @@ const HistoryTableProvider: React.FC<HistoryTableProviderProps> = ({
   const getItem = (index: number) => processedCommits[index];
 
   const getItemHeight = (index: number) =>
-    processedCommits[index]?.rowHeight || COMMIT_HEIGHT;
+    processedCommits[index]?.rowHeight || DEFAULT_HEIGHT;
```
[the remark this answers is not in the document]
khelif96: Great idea!

src/components/HistoryTable/HistoryTable.tsx
```diff
@@ -38,7 +40,39 @@ const HistoryTable: React.FC<HistoryTableProps> = ({
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [selectedCommit?.loaded]);
+
+  useEffect(() => {
+    if (selectedCommit) {
+      if (!selectedCommit.loaded && !selectedCommit.visited) {
+        loadMoreItems();
+      }
+    }
+  }, [selectedCommit, processedCommitCount, loadMoreItems]);
```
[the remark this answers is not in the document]
khelif96: yes we want to run this check when ever we fetch new commits. To see if selected has changed.

src/components/HistoryTable/historyTableContextReducer.ts
khelif96: Figured it out it was our favorite frenemy. The Apollo cache. Since the version already lived in the cache it wouldn't be processed the same way for some reason. If i HAD to guess there was likely some weird JS comparison behavior going on here. Possibly trying to match on a reference to the value in the apollo cache? https://github.com/khelif96/spruce/blob/1c15e03e19a667669d9a5d11a2c4d279423fa781/src/components/HistoryTable/utils.ts#L64 
Setting the `fetchPolicy` to `network-only` so it would always compare to a fresh value fixed it. There isn't any negative impact for using `network-only` here since we don't ever try to fetch older values on this page. They are cached by the context since we need to process them before using them.

https://user-images.githubusercontent.com/4605522/160713001-ee3a7670-ff90-475d-a8ae-96c55e510ca3.mov

src/components/HistoryTable/utils.ts
khelif96: Swapped it out with switch case statement which looks more readable imo lmk what you think.

src/components/HistoryTable/types.ts
khelif96: image
it breaks the typings :( 
VariantHistoryRow and taskHistoryRow have no way of knowing that a CommitRowType is a dateSeperator.

src/components/HistoryTable/historyTableContextReducer.ts
```diff
@@ -127,6 +147,24 @@ export const reducer = (state: HistoryTableReducerState, action: Action) => {
...
+      return {
+        ...state,
+        selectedCommit: {
+          ...state.selectedCommit,
+          visited: true,
+        },
+      };
     default:
       throw new Error(`Unknown reducer action${action}`);
   }
```
minnakt, line 170: unrelated nit but can you add a space before `${action}` 😅

minnakt (commented):
thanks for answering all my questions! I have a few more... 🐛

src/components/HistoryTable/HistoryTable.tsx
```diff
@@ -35,10 +37,36 @@ const HistoryTable: React.FC<HistoryTableProps> = ({
...
+          markSelectedVisited();
+        }
+      }
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [selectedCommit?.loaded]);
+
+  useEffect(() => {
+    if (selectedCommit) {
+      if (!selectedCommit.loaded && !selectedCommit.visited) {
```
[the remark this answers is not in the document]
khelif96: Good point.

src/components/HistoryTable/HistoryTableContext.tsx
```diff
@@ -26,10 +29,20 @@ interface HistoryTableState {
   columnLimit: number;
   historyTableFilters: TestFilter[];
   setHistoryTableFilters: (filters: TestFilter[]) => void;
+  setSelectedCommit: (order: number) => void;
+  markSelectedVisited: () => void;
+  selectedCommit: {
+    order: number;
+    loaded: boolean;
+    visited: boolean;
+    rowIndex: number;
```
SupaJoon, line 38: rowIndex and loaded could could be combined into the same variable if we calculate the rowIndex based on existing commits and not just new ones.

src/components/HistoryTable/HistoryTable.tsx
```diff
@@ -35,10 +37,37 @@ const HistoryTable: React.FC<HistoryTableProps> = ({
...
+        selectedCommit.rowIndex
+      ) {
+        if (listRef.current) {
+          listRef.current.scrollToItem(selectedCommit.rowIndex, "center");
+          markSelectedVisited();
+        }
+      }
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [selectedCommit?.loaded]);
```
SupaJoon, line 61: if the rowIndex changes, we should scroll to it. We don't actually need visited or loaded if we look at all commits to see if we can find the selected commits rowIndex.

khelif96: Great point!

src/components/HistoryTable/utils.ts
```diff
@@ -12,10 +13,18 @@ import { mainlineCommits, CommitRowType, rowType } from "./types";
...
+  selectedCommitRow,
+}: {
+  newCommits: mainlineCommits["versions"];
+  existingCommits: CommitRowType[];
+  selectedCommitOrder: number | null;
+  selectedCommitRow: number | null;
+}) => {
+  let selectedCommitRowIndex: number | null = selectedCommitRow;
   const processedCommits: CommitRowType[] = [...existingCommits];
   for (let i = 0; i < newCommits.length; i++) {
```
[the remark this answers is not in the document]
khelif96: We would have had to run this function in order to generate the rows in the first place. So this code had to have been ran on the initial rows.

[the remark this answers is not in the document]
SupaJoon: That makes sense that his function is ran after new commits are fetched. The rowIndex won't be calculated if new commits don't need to be loaded. This is why the row isn't highlighted when selecting 2 buildvariant labels that belong to the same response page. We should separate the rowIndex calculation from the commit processing.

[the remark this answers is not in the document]
khelif96: How would the initial rows be rendered if it does not run this?  The  `processedCommits` are stored in a context which wraps this page. If you navigate away and return the context will be a fresh instance. It will have to rerun this function in order to process the data into rows regardless of how they were fetched. 
You can see this behavior by adding a `console.log` within this function. It will step through each version no matter which way it was fetched. (A cached copy from the apollo cache from a previous page or a new copy fetched over the network). For what ever reason the strict `===` comparison is not matching the `selectedCommitOrder` with the `version.order` when we compare it against the cached copy. Yet everything else is working.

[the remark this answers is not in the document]
khelif96: Discussed offline opened https://jira.mongodb.org/browse/EVG-16642

src/components/HistoryTable/HistoryTableRow/FoldedCommit.tsx
```diff
@@ -16,56 +24,68 @@ interface FoldedCommitProps {
...
+  }: FoldedCommitProps) => {
+    // The commits are expanded if the height of the element is not equal to FOLDED_COMMITS_HEIGHT.
+    const { height } = style;
 
-  const columns = Array.from(Array(numVisibleCols)).map((_, idx) => (
-    <EmptyCell key={`loading_row_${idx}`} /> // eslint-disable-line react/no-array-index-key
-  ));
+    // The virtualized table will unmount the row when it is scrolled out of view but it will cache its height in memory.
+    // This means we can't rely on the state of the row to determine if it is expanded or not.
+    // So we instead look at its height which is cached by the table and determine if it is expanded or not.
```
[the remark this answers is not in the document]
khelif96: afaict resetAfterIndex tells react-window to recalculate all of the rows including the heights. Since we rely on the row height to determine the state of this component we store its value in the context so we have a guarantee that its state will be preserved despite how react-window decides to process it.

src/components/HistoryTable/historyTableContextReducer.ts
```diff
@@ -127,8 +141,26 @@ export const reducer = (state: HistoryTableReducerState, action: Action) => {
...
+        },
+      };
+    case "markSelectedVisited":
+      return {
+        ...state,
+        selectedCommit: {
+          ...state.selectedCommit,
+          visited: true,
+        },
+      };
```
minnakt, line 161: I think we can remove this action since there is no longer a `visited` property

src/components/HistoryTable/historyTableContextReducer.ts
```diff
@@ -127,8 +141,26 @@ export const reducer = (state: HistoryTableReducerState, action: Action) => {
...
       };
+    case "setSelectedCommit":
+      return {
+        ...state,
+        selectedCommit: {
+          order: action.order,
+          loaded: state.commitCache[action.order] != null,
+          visited: false,
+          rowIndex: null,
+        },
```
minnakt, line 152: remove `loaded` and `visited` properties!

src/components/HistoryTable/HistoryTable.tsx
```diff
@@ -24,21 +24,42 @@ const HistoryTable: React.FC<HistoryTableProps> = ({
...
+      if (listRef.current) {
+        listRef.current.resetAfterIndex(0);
+      }
+    }
+  }, [processedCommitCount]);
+
+  useEffect(() => {
+    if (selectedCommit && selectedCommit.rowIndex && listRef.current) {
+      listRef.current.scrollToItem(selectedCommit.rowIndex, "center");
+      markSelectedVisited();
```
minnakt, line 51: remove `markSelectedVisited()` function call

src/components/HistoryTable/utils.ts
```diff
@@ -24,57 +33,83 @@ export const processCommits = (
...
+              commit: version,
+              date: version.createTime,
+              selected: version.order === selectedCommitOrder,
+              rowHeight: COMMIT_HEIGHT,
+            });
+          }
+          if (version.order === selectedCommitOrder) {
+            selectedCommitRowIndex = processedCommits.length - 1;
+          }
+        }
```
minnakt, line 66: Bump on this: is there any reason we shouldn't consolidate `version.order === selectedCommitOrder` into a single variable, since it's repeated three times in this block?

src/components/HistoryTable/HistoryTable.tsx
```diff
@@ -24,21 +24,42 @@ const HistoryTable: React.FC<HistoryTableProps> = ({
...
+  }, [selectedCommit?.rowIndex]);
+
+  useEffect(() => {
+    if (selectedCommit) {
+      if (!selectedCommit.rowIndex) {
+        loadMoreItems();
+      }
     }
-  }, [recentlyFetchedCommits]);
+  }, [selectedCommit, loadMoreItems]);
```
minnakt, line 62: It almost seems like this hook could be combined with the one above it since the one above is basically `(selectedCommit && selectedCommit.rowIndex)` and the one here is `(selectedCommit && !selectedCommit.rowIndex)`. 🤔

Though maybe it would be more confusing. Just food for thought!!

src/components/HistoryTable/utils.ts
khelif96: oops sorry missed this.

src/components/HistoryTable/HistoryTable.tsx
khelif96: sgtm

src/pages/task/ActionButtons.tsx
```diff
@@ -271,7 +274,9 @@ export const ActionButtons: React.FC<Props> = ({
               onClick={() => {
                 taskAnalytics.sendEvent({ name: "Click See History Button" });
               }}
-              to={getTaskHistoryRoute(projectIdentifier, taskName)}
+              to={getTaskHistoryRoute(project.identifier, displayName, {
+                selectedCommit: order,
+              })}
```
minnakt, line 279: Would you mind double checking this? It's not working for me, for some reason the `order` is `undefined`. I might be doing something wrong though 🥲

minnakt (approved):
sheeesh!!!! 🥶🥶 looks awesome! thank you for all your work on this!

khelif96:
Since PR tests are broken heres a patch build. https://spruce.mongodb.com/version/624618c01e2d17279da12118/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

PR #1163, opened 2022-03-22
EVG-16547: Add y-axis labels on Mainline Commits chart 

minnakt:
[EVG-16547](https://jira.mongodb.org/browse/EVG-16547)

### Description 
This PR adds labels for both Absolute/Percentage options on the mainline commits chart. 

For the Absolute option, I slightly changed the way we scale bar heights just because it would be kind of odd if there were 73 total tasks and the y-axis values were `[0, 14.6, 29.2, 43.8, 58.4, 73]`. I also had to squeeze the chart with a little more padding to make room for the labels.

### Screenshots
https://user-images.githubusercontent.com/47064971/159536168-88f4fcd5-7931-43f7-93ad-2d52093b344f.mov

### Testing 
- Added tests for new `range` function
- Added tests for new `roundMax` function


src/pages/commits/ActiveCommits/utils.ts
```diff
@@ -59,9 +59,31 @@ export function calculateBarHeight(
   if (chartType === ChartTypes.Percentage) {
     return `${(value / total) * 100}%`;
   }
-  return `${(value / max) * 100}%`;
+  const maxRounded = roundMax(max);
+  return `${(value / maxRounded) * 100}%`;
 }
 
+export const roundMax = (max: number) => {
```
minnakt, line 66: Cut offs in this function are entirely arbitrary and made by my own judgment

sophstad (approved):
Just a general note but have we considered that the "View Options" panel perhaps covers the content a bit too aggressively? It's impossible to know if this graph reads 8, 9, or 10 without hovering, which IMO defeats the point of an easily readable chart. Maybe we should run this by design?
image

This work looks great! 🤠

minnakt:
> Just a general note but have we considered that the "View Options" panel perhaps covers the content a bit too aggressively? It's impossible to know if this graph reads 8, 9, or 10 without hovering, which IMO defeats the point of an easily readable chart. Maybe we should run this by design? image
> 
> This work looks great! 🤠

Yeah, I also noticed it gets in the way a lot 😓 I can play around with it a little. If I don't see a way to fix it, I can bring up during the design sync this week

khelif96:
> > Just a general note but have we considered that the "View Options" panel perhaps covers the content a bit too aggressively? It's impossible to know if this graph reads 8, 9, or 10 without hovering, which IMO defeats the point of an easily readable chart. Maybe we should run this by design? image
> > This work looks great! 🤠
> 
> Yeah, I also noticed it gets in the way a lot 😓 I can play around with it a little. If I don't see a way to fix it, I can bring up during the design sync this week
There are actually updated designs for this. 
https://www.figma.com/file/TxlYD1vlJPgZ8S2Z8sZfe9/Mainline-Commit-UI?node-id=2019%3A7093

https://jira.mongodb.org/browse/EVG-16600

image

khelif96 (approved):
Great Work! 🥳

PR #1164, opened 2022-03-22
EVG-16361 Update test config to test snapshots and unit tests individually

khelif96:
[EVG-16361](https://jira.mongodb.org/browse/EVG-16361)

### Description 
The coverage task has been a source of confusion since it would often include test failures related to to snapshot tests but it was not clear that they were related. I removed that task and created a new task called snapshot which only includes snapshot tests and displays the results on the tests tab. 
The tests tab will no longer test snapshots and will only test unit tests.

### Testing 
Passing patch
https://spruce.mongodb.com/version/6239fb95e3c3310e237bc2f3/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

Patch with failing unit tests and failing snapshot tests
https://spruce.mongodb.com/version/6239fbd461837d27ce86d6ff/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC


.evergreen.yml
```diff
@@ -135,14 +135,14 @@ functions:
...
-  yarn-coverage:
+    
+  yarn-snapshot:
     command: shell.exec
     params:
       working_dir: spruce
       script: |
         export PATH=/opt/nodejs/node-v16.13.0-linux-x64/bin:$PATH
-        yarn coverage
+        yarn test --reporters="jest-junit" --reporters="default" --testPathPattern="storybook.test.ts"
```
minnakt, line 145: Does this need a `-u`? (honest question, I'm not sure what the answer is 😵‍💫)

minnakt (approved):
I think it might be worthwhile to make a patch where one test fails and the other passes before merging, just to make sure they test separate things. But it looks great! Thank you! 💚🐛

.evergreen.yml
khelif96: no it does not `-u` is a flag to regenerate snapshots. Which was causing the test to never fail since the failing snapshots would always be updated with updated ones.

khelif96:
> I think it might be worthwhile to make a patch where one test fails and the other passes before merging, just to make sure they test separate things. But it looks great! Thank you! 💚🐛

You can look at the test results tab for the 2 tasks and see that there are 2 failing tests. One is failing because of failing snapshots and the other is failing for failing unit tests.

PR #1166, opened 2022-03-22
EVG-16542: Reorganize project settings

sophstad:
EVG-16542

### Description 
Lots of files but I promise it's just moving things around!
- Move page elements like `Header`, `CreateProjectModal`, etc. into `src/pages/projectSettings`
- Move tab directories into `src/pages/projectSettings/tabs`


khelif96 (approved):
Looks great to me! 
Only had one comment about the location of the context.
Thanks for doing this!

PR #1167, opened 2022-03-23
EVG-16091: Linkify inactive commits tooltip

minnakt:
[EVG-16091](https://jira.mongodb.org/browse/EVG-16091)

### Description 
This PR makes it possible to click on the inactive commits on the Project Health page. If the inactive commits do not fit within the tooltip, the user can open a modal where they can see all of the inactive commits. Also works for unmatching commits.

### Screenshots
https://user-images.githubusercontent.com/47064971/159737509-f0932c46-0489-4bbc-bf78-1a502abd2d23.mov

### Testing 
- Modified test in `projectHealth/commits.ts (Cypress)`
- Added test in `InactiveCommits.test.tsx`
- For manual testing I recommend `sys-perf` on `prod`, they have a lot of inactive commits


src/components/DisplayModal.tsx
```diff
@@ -21,18 +22,25 @@ export const DisplayModal: React.FC<DisplayModalProps> = ({
   setOpen,
   size,
   title,
+  popoverZIndex = zIndex.modal,
```
sophstad, line 25: > Note: I made the modal use a different z-index than zIndex.modal because the ChartToggle has a higher z-index than that, causing it to be visually on top of the modal (which is wrong).

I was playing around with the `ChartToggle` component and noticed that the `Container` doesn't really seem to do anything — when the page is loaded it has a height of 0, and removing it doesn't change anything (unless I am missing something...). Removing the container also means we can use the regular `zIndex.modal` here. Let me know if you find a bug with that though 😳

minnakt: you are completely correct thank you!!! ❤️

khelif96 (approved):
Looks Great to me!

PR #1168, opened 2022-03-23
EVG-16543: Task Status Icon Legend

SupaJoon:
[EVG-16543](https://jira.mongodb.org/browse/EVG-16543)

### Description 
These code changes add the task status icon legend and button to access on the bottom right of the page. 

### Screenshots
Screen Shot 2022-03-23 at 1 19 16 PM


src/components/TaskStatusIconLegend/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+  );
+};
+
+const StyledIcon = styled(Icon)`
+  cursor: pointer;
+`;
+
+const Row = styled.div`
+  display: flex;
+  align-items: top;
```
minnakt, line 65: This should probably be `flex-start`, it says `top` is an invalid property value here

Screen Shot 2022-03-23 at 3 18 05 PM

src/components/TaskStatusIconLegend/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+  width: 125px;
```
minnakt, line 66: Do you think we could use `display: grid` instead, so we don't have to specify a particular width on these items? Something like this?

Screen Shot 2022-03-23 at 3 29 25 PM

Result:

Screen Shot 2022-03-23 at 3 27 36 PM

src/components/TaskStatusIconLegend/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+        justify="end"
+        active={isActive}
+        usePortal
+        spacing={spacing[2]}
+      >
+        <PopoverContainer>
+          <LegendContent />
+        </PopoverContainer>
+      </Popover>
+    </IconButton>
```
minnakt, line 55: I think it might be beneficial to provide some other way of closing this `Popover` besides clicking again on the `IconButton`. Since there's no `X` button, I kind of expected it to close when I clicked outside, and it took me a moment to figure out how to close it

minnakt (commented):
Looks great 🐛 Btw, the JIRA ticket number in the PR title and description link seems to be incorrect. I think you meant this one: [EVG-16543](https://jira.mongodb.org/browse/EVG-16543)

src/components/TaskStatusIconLegend/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+      }}
+      aria-label="Task Status Icon Legend"
+    >
+      <StyledIcon glyph="QuestionMarkWithCircle" />
+      <Popover
+        align="top"
+        justify="end"
+        active={isActive}
+        usePortal
+        spacing={spacing[2]}
```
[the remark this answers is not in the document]
SupaJoon: The spacing prop only accepts a number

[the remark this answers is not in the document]
khelif96: she's referring to the size constant in the tokens file.

```js
import { size } from "constants/tokens"
...
spacing={size.xs}
/>
```
is the equivalent to what you are doing.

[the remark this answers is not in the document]
SupaJoon: size.xs from constants/tokens is type string but the spacing prop in the Popover component only accepts a number
Screen Shot 2022-03-24 at 10 25 53 AM

src/components/TaskStatusIconLegend/index.tsx
SupaJoon: I added an X button on the popover. I think some users may benefit from having the popover visible while interacting with the page so I didn't include the functionality to close on outside click

src/components/TaskStatusIconLegend/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+);
+
+export const TaskStatusIconLegend: React.FC = () => {
+  const [isActive, setIsActive] = useState(false);
+
+  return (
+    <IconButton
+      onClick={() => {
+        setIsActive(!isActive);
+      }}
```
khelif96, line 44: Looks like the click listener is acting on the tooltip. When i click anywhere on the tooltip it closes and when i click outside it does not. I would expect the opposite behavior it should only close when we click outside the tooltip.

src/components/TaskStatusIconLegend/index.tsx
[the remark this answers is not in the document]
khelif96: Oops thanks for clarifying.

src/components/TaskStatusIconLegend/index.tsx
SupaJoon: I think some users may want to reference the legend as they interact between the waterfall and history pages so I decided to exclude the outside click functionality. I think clicking anywhere on the popover to close makes sense because there are no other click events in the tooltip and it becomes easier to close. Although the design guidelines section isn't filled out, the popover on [mongodb.design](https://www.mongodb.design/component/popover/example/) behaves this way. Let me know what you think

src/pages/Version.tsx
```diff
@@ -135,7 +135,8 @@ export const VersionPage: React.FC = () => {
   }, [data]);
 
   const { version } = data || {};
-  const { status, patch, isPatch, revision, message, order } = version || {};
+  const { status, patch, isPatch, revision, message, order, activated } =
```
minnakt, line 138: I think this file was committed accidentally

src/pages/version/ActionButtons.tsx
```diff
@@ -23,6 +23,7 @@ interface ActionButtonProps {
   versionId: string;
   childPatches: Partial<Patch>[];
   isPatch: boolean;
+  activated: boolean;
 }
 
 export const ActionButtons: React.FC<ActionButtonProps> = ({
```
minnakt, line 29: I think this file was also committed accidentally

src/components/TaskStatusIconLegend/index.tsx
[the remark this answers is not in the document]
minnakt: Do you think we need the `spacing` here? The default spacing is `10` and this changes it to `8`, which I feel might not make a huge difference

src/components/TaskStatusIconLegend/index.tsx
minnakt: Personally, I'm not a fan of the default leafygreen `Popover` behavior... both `mui` and `antd` popovers don't close when clicking on the popover content, which is what I prefer. This doesn't have to be done in this ticket but maybe we should consider creating a modified leafygreen `Popover` component in the future, since this issue has come up a few times

minnakt (approved):
looks awesome! 🔥😼

src/components/TaskStatusIconLegend/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+const TitleContainer = styled.div`
+  margin-bottom: ${size.m};
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+`;
+
+const StyledPopoverContainer = styled(PopoverContainer)`
+  border-radius: ${size.xs};
```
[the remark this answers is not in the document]
SupaJoon: @minnakt I was unable to see the popover flush with the browser window in chrome and firefox in staging unless I zoom into the view with the trackpad. 
Screen Shot 2022-03-28 at 9 31 37 AM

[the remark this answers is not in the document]
minnakt: oh!!! that's weird, I wonder why I was seeing that. it looks great!!! 😇

PR #1171, opened 2022-03-24
EVG-16520: Allow scheduling inactive mainline commit versions

SupaJoon:
[EVG-16520](https://jira.mongodb.org/browse/EVG-16520)

### Description 
The schedule button/modal on the version page is now accessible for inactive versions. 

### Testing 
You can test this on staging or prod by clicking on the Schedule button in the top right corner of the version page for an inactive version.


src/pages/version/ActionButtons.tsx
```diff
@@ -75,7 +77,7 @@ export const ActionButtons: React.FC<ActionButtonProps> = ({
         <ScheduleTasks
           versionId={versionId}
           isButton
-          disabled={isPatchOnCommitQueue}
+          disabled={isPatchOnCommitQueue && activated}
```
khelif96, line 80: should this be isPatchOnCommitQueue || activated?

SupaJoon: I don't think so. `isPatchOnCommitQueue` is true for unactivated mainline commits so `isPatchOnCommitQueue || activated` will disable the button when it shouldn't.

khelif96 (approved):

PR #1176, opened 2022-03-25
EVG-16627 Fix flaky my patches test

khelif96:
[EVG-16627](https://jira.mongodb.org/browse/EVG-16627)

### Description 
Noticed there was a flaky cypress test caused by cypress trying to assert on a value that wasn't on the screen yet. 
Since cypress would rerun flaky tests it was never reported but it was slowing down the test suite. This fixes it by using the correct selector for the assertion. 

I also replaced a superfluous test with a simpler assertion.


khelif96:
Failing test seems unrelated

SupaJoon (approved):

PR #1177, opened 2022-03-28
EVG-16616 Add more context to the Patch Trigger Aliases modals

malikchaya2:
[EVG-16616](https://jira.mongodb.org/browse/EVG-16616)

### Description 
When setting up patch aliases for Spruce, we thought there was a bug because we didn't realize that the parent changes were not automatically applied. I added this help text to help avoid confusion in the future. 

### Screenshots
image

### Testing 
visually on dev


ablack12 (approved):
Oops haha! Thanks!

khelif96 (approved):

PR #1178, opened 2022-03-28
EVG-16580: Show more headers on history pages if the screen is wide enough

SupaJoon:
[EVG-16580](https://jira.mongodb.org/browse/EVG-16580)

### Description 
The number of columns rendered on the history page will change depending on the table width. The current page will be reset to 0 when the number of visible columns changes. 

### Screenshots

https://user-images.githubusercontent.com/10734386/160463677-48594a0e-8fc9-4f21-9bba-5ac30c641d8e.mov

### Testing 
  Jest


yarn.lock
```diff
@@ -12439,6 +12439,11 @@ lodash.templatesettings@^4.0.0:
   dependencies:
     lodash._reinterpolate "^3.0.0"
 
+lodash.throttle@^4.1.1:
```
sophstad, line 12442: I think you should run `yarn install` again to update this file since it still includes the caret

src/components/HistoryTable/HistoryTable.tsx
```diff
@@ -12,21 +13,36 @@ interface HistoryTableProps {
...
+  width: number;
 }
 const HistoryTable: React.FC<HistoryTableProps> = ({
   loadMoreItems,
   recentlyFetchedCommits,
   children,
+  height,
+  width,
 }) => {
+  useEffect(() => {}, [width, height]);
```
sophstad, line 26: Should this line be deleted? Everything seems to work fine without it.

src/components/HistoryTable/historyTableContextReducer.ts
```diff
@@ -122,6 +127,22 @@ export const reducer = (state: HistoryTableReducerState, action: Action) => {
...
+      return {
+        ...state,
+        visibleColumns: state.columns.slice(
+          0,
+          calcColumnLimitFromWidth(action.width)
+        ),
+        columnLimit: calcColumnLimitFromWidth(action.width),
+        currentPage: 0,
+        pageCount: Math.ceil(
+          state.columns.length / calcColumnLimitFromWidth(action.width)
```
sophstad, line 143: [nit] You could save `calcColumnLimitFromWidth(action.width)` as a variable since it's used a few times.

sophstad (approved):

src/components/HistoryTable/historyTableContextReducer.ts
SupaJoon: I would have to move the variable outside of the switch statement because of the rule [no-case-declarations](https://eslint.org/docs/rules/no-case-declarations). I think a neater way to do this would be to create a separate function that performs the logic in the case by accepting a state and action and returning the state but we haven't really established that pattern for the rest of our reducer cases.

src/components/HistoryTable/utils.ts
```diff
@@ -114,3 +117,10 @@ export const toggleRowSizeAtIndex = (
   }
   return newCommits;
 };
+
+export const calcColumnLimitFromWidth = (tableWidth: number) => {
+  const colLimit = Math.floor(
+    (tableWidth - ROW_LABEL_WIDTH) / COLUMN_LABEL_WIDTH
+  );
+  return colLimit > DEFAULT_COLUMN_LIMIT ? colLimit : DEFAULT_COLUMN_LIMIT;
```
khelif96, line 125: This might not be necessary. If a user shrinks their browser it might make more sense to only show enough columns that actually fit instead of trying to squeeze in more columns. The labels end up getting squished.
image

src/components/HistoryTable/historyTableContextReducer.ts
khelif96: You would actually just need to add a `{` to the case declaration. We do something similar on line 111

src/components/HistoryTable/utils.ts
SupaJoon: Good idea!
Screen Shot 2022-03-29 at 1 52 02 PM

src/components/HistoryTable/HistoryTable.tsx
```diff
@@ -12,21 +13,35 @@ interface HistoryTableProps {
...
   const listRef = useRef<List>(null);
 
+  const throttleOnChangeTableWidth = useMemo(
+    () => throttle(onChangeTableWidth, 400),
+    [onChangeTableWidth]
+  );
+
+  useEffect(() => {
+    throttleOnChangeTableWidth(width);
+  }, [width, throttleOnChangeTableWidth]);
```
khelif96, line 43: https://github.com/bvaughn/react-virtualized/blob/master/docs/AutoSizer.md Looks like autosizer supports an onResize callback. We don't need this useEffect.

khelif96 (changes requested):

src/components/HistoryTable/HistoryTable.tsx
```diff
@@ -13,20 +14,27 @@ interface HistoryTableProps {
   recentlyFetchedCommits: MainlineCommitsForHistoryQuery["mainlineCommits"];
   children: ComponentType<ListChildComponentProps<any>>;
 }
-const HistoryTable: React.FC<HistoryTableProps> = ({
+const HistoryTable = ({
   loadMoreItems,
   recentlyFetchedCommits,
   children,
-}) => {
+}: HistoryTableProps): JSX.Element => {
```
khelif96, line 21: nit: you don't need to type the return type it is implicit.

src/components/HistoryTable/constants.ts
```diff
@@ -9,12 +9,20 @@ const queryParamsToDisplay = new Set([
 const FOLDED_COMMITS_HEIGHT = 40;
 const COMMIT_HEIGHT = 120;
 const DATE_SEPARATOR_HEIGHT = 40;
+const DEFAULT_HEIGHT = 100;
```
khelif96, line 12: its not immediately clear to me what DEFAULT_HEIGHT is or if its being used.

khelif96 (approved):
🔥 🔥 🔥

PR #1181, opened 2022-03-30
EVG-15088: Add Periodic Builds page

sophstad:
EVG-15088

### Description 

### Screenshots
image

### Testing 

### Evergreen PR 


src/components/DatePicker.tsx
```diff
@@ -1,8 +1,25 @@
...
+  return (
+    <InteractionRing disabled={disabled}>
+      <DatePicker {...props} style={leafygreenInputStyle} />
+    </InteractionRing>
+  );
+};
+
+const leafygreenInputStyle = {
+  border: `1px solid ${gray.base}`,
+  borderRadius: "4px",
```
SupaJoon, line 23:
```suggestion
  borderRadius: ${size.xxs},
```

src/pages/projectSettings/tabs/PeriodicBuildsTab/PeriodicBuildsTab.tsx
```diff
@@ -1 +1,54 @@
...
+import {
+  usePopulateForm,
+  useProjectSettingsContext,
+} from "pages/projectSettings/Context";
+import { getFormSchema } from "./getFormSchema";
+import { TabProps } from "./types";
+
+const tab = ProjectSettingsTabRoutes.PeriodicBuilds;
+
+const getInitialFormState = (projectData, repoData) => {
```
SupaJoon, line 14: could we add types for this function

SupaJoon (approved):

khelif96 (approved):

src/pages/projectSettings/tabs/PeriodicBuildsTab/getFormSchema.ts
```diff
@@ -0,0 +1,126 @@
...
+      "periodicBuilds",
+      ["Override Repo Commands", "Default to Repo Commands"],
+      {
+        type: "array" as "array",
+        default: [],
+        items: {
+          type: "object" as "object",
+          properties: {
+            intervalHours: {
+              type: "number" as "number",
```
ablack12, line 31: I don't know if this is relevant to your PR but if I type a non-integer into the text box it shows the error, and then when I press the back button it just takes away the error but doesn't actually delete the character.
image

src/pages/projectSettings/tabs/PeriodicBuildsTab/getFormSchema.ts
```diff
@@ -0,0 +1,126 @@
...
+              type: "string" as "string",
+              title: "Alias",
+              default: "",
+            },
+            message: {
+              type: "string" as "string",
+              title: "Message",
+              default: "",
+            },
+            timezone: {
```
ablack12, line 50: Does everybody have a default timezone? Or else should we populating this with our default timezone (which I assume is us-east)
image

src/components/DatePicker.tsx
```diff
@@ -1,8 +1,26 @@
...
+      <DatePicker {...props} style={leafygreenInputStyle} />
+    </InteractionRing>
+  );
+};
+
+const leafygreenInputStyle = {
+  border: `1px solid ${gray.base}`,
+  borderRadius: size.xxs,
+  transition: "border-color 150ms ease-in-out",
+};
```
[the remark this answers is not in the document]
sophstad: I decided to use [antd's `style` prop](https://ant.design/components/date-picker/#Common-API) directly since it specifically targets the `input` element. This seems more future-proof than targeting a node that could be subject to future changes to the component's internals.

src/pages/projectSettings/tabs/PeriodicBuildsTab/getFormSchema.ts
```diff
@@ -0,0 +1,126 @@
...
+              default: timezone,
+              oneOf: timeZones.map(({ str, value }) => ({
+                type: "string" as "string",
+                title: str,
+                enum: [value],
+              })),
+            },
+            nextRunTime: {
+              type: "string" as "string",
+              title: "Next Run Time",
```
ablack12, line 62: This time picker isn't working as I'd expect. Every time I click "OK", it translates my time. For example, if I try to click OK on 23:00 for 4/8
image
it gives me 3 on 4/9: 
image

src/pages/projectSettings/tabs/PeriodicBuildsTab/getFormSchema.ts
```diff
@@ -0,0 +1,126 @@
+import { Field } from "@rjsf/core";
+import { SpruceFormProps } from "components/SpruceForm";
+import widgets from "components/SpruceForm/Widgets";
+import { timeZones } from "constants/fieldMaps";
+import { form, ProjectType } from "../utils";
+
+const { overrideRadioBox } = form;
+
+export const getFormSchema = (
```
ablack12, line 9: I'm pretty sure I've configured this correctly so I'm not sure why I'm not able to save the page.
image

ablack12 (changes requested):

src/pages/projectSettings/tabs/PeriodicBuildsTab/getFormSchema.ts
sophstad: Not related to these changes but it is being tracked! EVG-16641

src/pages/projectSettings/tabs/PeriodicBuildsTab/getFormSchema.ts
sophstad: Users are not required to have a timezone set. In places like the hosts table, if a timezone is not set, we use the user's local time. I can go with that approach here unless you think defaulting to Eastern Time makes more sense.

ablack12: Using the user's local time makes sense, I just think this should show something.

ablack12 (approved):
LGTM with nit

PR #1184, opened 2022-03-31
EVG-15817: Project health analytics

SupaJoon:
[EVG-15817](https://jira.mongodb.org/browse/EVG-15817)

### Description 
Analytic events for history tables and commits charts. `useProjectHealthAnalytics` describes the elements being tracked. 

### Testing 
Click on elements from the commits chart and history tables pages and make sure a analytics event is logged in the developer tools console.


src/components/FilterBadges/index.tsx
```diff
@@ -11,9 +11,13 @@ const { parseQueryString } = queryString;
 
 interface FilterBadgesProps {
   queryParamsToDisplay: Set<string>;
+  onRemoveAnalytics?: () => void;
+  onClearAllAnalytics?: () => void;
```
khelif96, line 15: nit. `onRemove` and `onClearAll`. I think theres value in exposing these callbacks and theres no need to tie them to analytics

src/components/HistoryTable/HistoryTableRow/FoldedCommit.tsx
```diff
@@ -34,6 +38,7 @@ export const FoldedCommit = memo(
     numVisibleCols,
     style,
     selected,
+    sendAnalytics,
```
khelif96, line 41: any reason we don't wrap all of the analytics logic inside this component i don't think theres any reason to pass it in a conditional prop. I think it makes sense to just define the hook inline in this component.

src/components/HistoryTable/HistoryTableTestSearch/HistoryTableTestSearch.tsx
```diff
@@ -8,8 +8,16 @@ import { queryString, url } from "utils";
...
+  failedTests: string[];
+}
+interface HistoryTableTestSearchProps {
+  sendAnalytics?: (v: HistoryTableTestSearchAnalytics) => void;
+}
 
-export const HistoryTableTestSearch = () => {
+export const HistoryTableTestSearch = ({
+  sendAnalytics,
+}: HistoryTableTestSearchProps): JSX.Element => {
```
khelif96, line 20: This is implicit.

src/components/HistoryTable/HistoryTableTestSearch/HistoryTableTestSearch.tsx
```diff
@@ -8,8 +8,16 @@ import { queryString, url } from "utils";
 
 const { upsertQueryParam } = url;
 const { parseQueryString } = queryString;
+export interface HistoryTableTestSearchAnalytics {
+  failedTests: string[];
+}
+interface HistoryTableTestSearchProps {
+  sendAnalytics?: (v: HistoryTableTestSearchAnalytics) => void;
```
khelif96, line 15: Same comment about defining this inline instead of passing it in.

src/components/TupleSelect/index.tsx
```diff
@@ -16,10 +16,18 @@ type option = {
...
+  value: string;
+}
 interface TupleSelectProps {
   options: option[];
+  sendAnalytics?: (t: TupleAnalytics) => void;
 }
-export const TupleSelect: React.FC<TupleSelectProps> = ({ options }) => {
+export const TupleSelect: React.FC<TupleSelectProps> = ({
+  options,
+  sendAnalytics,
```
khelif96, line 29: Same comment about inlining this.

src/components/projectSelect/index.tsx
```diff
@@ -24,11 +24,17 @@ interface ProjectSelectProps {
...
 }
+export interface ProjectSelectAnalytics {
+  projectIdentifier: string;
+}
+
 export const ProjectSelect: React.FC<ProjectSelectProps> = ({
   selectedProjectIdentifier,
   isProjectSettingsPage = false,
   getRoute,
+  sendAnalytics,
```
khelif96, line 37: Same comment about inlining.

src/pages/commits/InactiveCommits/index.tsx
```diff
@@ -71,7 +73,16 @@ export const InactiveCommitButton: React.FC<InactiveCommitsProps> = ({
         align="bottom"
         justify="middle"
         trigger={
-          <ButtonContainer role="button">
+          <ButtonContainer
+            onClick={() => {
+              sendEvent({
+                name: "Toggle commit chart label tooltip",
+                type: tooltipType,
+                toggle: tooltipRef?.current ? "close" : "open",
```
khelif96, line 81: This logic is confusing why is this ref necessary?

src/pages/taskHistory/TaskHistoryRow.tsx
```diff
@@ -38,6 +42,9 @@ const TaskHistoryRow = memo((props: ListChildComponentProps) => {
               failingTests={failingTests}
               label={label}
               loading={loading}
+              sendAnalytics={(v: TaskCellAnalytics) => {
+                sendEvent({ name: "Click task history task cell", ...v });
```
khelif96, line 46: Maybe we call this event `Click Task Cell` and have the specific page be an option. That way we can query for all task cell clicks.

src/pages/variantHistory/VariantHistoryRow.tsx
```diff
@@ -33,6 +37,9 @@ const VariantHistoryRow = memo((props: ListChildComponentProps) => {
               task={t}
               failingTests={failingTests}
               label={label}
+              sendAnalytics={(v: TaskCellAnalytics) => {
+                sendEvent({ name: "Click variant history task cell", ...v });
```
khelif96, line 41: same comment as above.

src/utils/string/jiraLinkify.tsx
```diff
@@ -4,10 +4,15 @@ import { getJiraTicketUrl } from "constants/externalResources";
...
 export const jiraLinkify = (
   unlinkified: string | React.ReactNodeArray,
-  jiraHost: string
+  jiraHost: string,
+  onClick?: (match: string) => () => void
 ) =>
   reactStringReplace(unlinkified, /([A-Z]{1,10}-\d{1,6})/g, (match, i) => (
-    <StyledLink key={`${match}${i}`} href={getJiraTicketUrl(jiraHost, match)}>
+    <StyledLink
+      onClick={onClick ? onClick(match) : undefined}
```
khelif96, line 12: we can assign a default value to the onClick callback above and avoid this ternary.

src/pages/commits/ActiveCommits/ChartToggle.tsx
```diff
@@ -12,9 +13,14 @@ export const ChartToggle: React.FC<{
   currentChartType: ChartTypes;
   onChangeChartType: (chartType: ChartTypes) => void;
 }> = ({ currentChartType, onChangeChartType }) => {
+  const { sendEvent } = useProjectHealthAnalytics();
   const onChange = (e: React.ChangeEvent<HTMLInputElement>) => {
     const chartType = e.target.value as ChartTypes;
     onChangeChartType(chartType);
+    sendEvent({
```
khelif96, line 20: Noticed this error when toggling the chart toggle. I'm not sure if it was introduced in this pr or not. 
image

khelif96 (changes requested):

src/components/HistoryTable/HistoryTableRow/FoldedCommit.tsx
SupaJoon: The only reason to pass in an analytics prop is so VariantHistoryRow and TaskHistoryRow can differentiate the analytic event name. If we were to instantiate the sendEvent hook in this file, the only way to differentiate between "task" and "variant" would be to examine the URL which would require switch logic and seems less precise.

src/pages/commits/InactiveCommits/index.tsx
SupaJoon: The ref is placed on the tooltip body which exists when it's open and is null when it's closed. The alternative to this solution would have been converting the tooltip from controlled to uncontrolled which would actually add more complexity and state change directions since we would have to control the outside click with another event handler. Ideally the tooltip component exposes it's state in the onClick callback.

src/components/HistoryTable/HistoryTableTestSearch/HistoryTableTestSearch.tsx
SupaJoon: discussed offline

khelif96: can we use React.VFC now that we figured it out?

src/analytics/projectHealth/useProjectHealthAnalytics.ts
```diff
@@ -0,0 +1,73 @@
...
+      name: "Submit variant history failed test filter";
+    } & HistoryTableTestSearchAnalytics)
+  | { name: "Clear all commit chart badges" }
+  | { name: "Clear all task history badges" }
+  | { name: "Clear all variant history badges" }
+  | { name: "Click commit chart grouped status badge"; statuses: string[] }
+  | { name: "Click commit chart task status icon"; status: string }
+  | { name: "Click commit chart variant label"; variant: string }
+  | { name: "Click commit chart view option"; viewOption: string }
+  | { name: "Click commit label jira link"; jiraTicket: string }
```
khelif96, line 37: jiraTicket too. it is to identifying and will not provide much value.  Since we only care if they did click on a Jira link not what the ticket was.

src/analytics/projectHealth/useProjectHealthAnalytics.ts
```diff
@@ -0,0 +1,73 @@
...
+  | { name: "Click commit label version link"; versionId: string }
+  | { name: "Click task history column header"; variant: string }
+  | { name: "Click variant history column header"; task: string }
+  | { name: "Click variant history task selector"; tasks: string[] }
```
khelif96, line 41: Same arguments. about task ids and variant names being too specific.

src/analytics/projectHealth/useProjectHealthAnalytics.ts
```diff
@@ -0,0 +1,73 @@
...
+  | { name: "Filter commit chart by requester"; requesters: string[] }
+  | { name: "Filter commit chart by task status"; statuses: string[] }
```
khelif96, line 43: requesters and statuses make sense since they are not identifying and are more limited and generic.

src/analytics/projectHealth/useProjectHealthAnalytics.ts
```diff
@@ -0,0 +1,73 @@
...
+      name: "Toggle commit chart label tooltip";
+      type: "Unmatching" | "Inactive";
+      toggle: "open" | "close";
+    };
+
+interface P extends Properties {}
+interface Analytics extends A<Action> {}
+
+export const useProjectHealthAnalytics = (): Analytics => {
+  const userId = useGetUserQuery();
```
khelif96, line 64: `userId` is set in `useAnalyticsAttributes`. It is defined globally for the new relic instance. 
https://github.com/evergreen-ci/spruce/blob/7fb209195d4e6cc4eb3f9ade81732548b8d999bf/src/analytics/useAnalyticsAttributes.ts

src/analytics/projectHealth/useProjectHealthAnalytics.ts
```diff
@@ -0,0 +1,73 @@
...
+  | { name: "Click commit label version link"; versionId: string }
+  | { name: "Click task history column header"; variant: string }
+  | { name: "Click variant history column header"; task: string }
+  | { name: "Click variant history task selector"; tasks: string[] }
+  | { name: "Filter commit chart by requester"; requesters: string[] }
+  | { name: "Filter commit chart by task status"; statuses: string[] }
+  | { name: "Open task icons legend" }
+  | { name: "Remove commit chart badge" }
+  | { name: "Remove task history badge" }
+  | { name: "Remove variant history badge" }
```
khelif96, line 47: We don't need page specific analytics. 
New relic automatically logs where the events originate.
image

src/components/HistoryTable/HistoryTableRow/FoldedCommit.tsx
```diff
@@ -14,6 +14,14 @@ import {
 } from "../constants";
 import { RowContainer } from "./styles";
 
+export interface FoldedCommitAnalytics {
+  toggle: "open" | "close";
+}
+
+export const foldedCommitAnalytics = (
+  isVisible: boolean
+): FoldedCommitAnalytics => ({ toggle: isVisible ? "open" : "close" });
```
khelif96, line 23: This pattern is a bit confusing. I think it can be avoided by just containing the analytics logic in this file. The url helps differentiate where the event is originating.  (See above comments)

src/components/HistoryTable/HistoryTableTestSearch/HistoryTableTestSearch.tsx
```diff
@@ -8,8 +8,16 @@ import { queryString, url } from "utils";
 
 const { upsertQueryParam } = url;
 const { parseQueryString } = queryString;
+export interface HistoryTableTestSearchAnalytics {
+  failedTests: string[];
+}
+interface HistoryTableTestSearchProps {
+  onSubmit?: (failedTests: string[]) => void;
+}
```
khelif96, line 16: Same comment about the folded commits analytics

src/components/TaskStatusIconLegend/index.tsx
```diff
@@ -26,13 +27,17 @@ export const LegendContent = () => (
...
   const [isActive, setIsActive] = useState(false);
 
   return (
     <div>
       <IconButton
         onClick={() => {
           setIsActive(!isActive);
+          if (!isActive) {
+            sendEvent({ name: "Open task icons legend" });
+          }
```
khelif96, line 40: Might be nice to capture when they open and close it see above comment.

src/pages/Commits.tsx
```diff
@@ -128,11 +139,27 @@ export const Commits = () => {
             <ProjectSelect
               selectedProjectIdentifier={projectId}
               getRoute={getCommitsRoute}
+              sendAnalytics={({
```
khelif96, line 142: `onChange`

src/pages/commits/ActiveCommits/ChartToggle.tsx
khelif96: https://jira.mongodb.org/browse/EVG-16689
Opened a ticket.

src/pages/commits/InactiveCommits/index.tsx
khelif96: in this case i think it would be better to not log the toggle. Imo the additional complexity isn't worth the extra detail.

src/pages/commits/StatusSelect.tsx
```diff
@@ -8,8 +9,15 @@ import { useStatusesFilter } from "hooks";
...
 export const StatusSelect = () => {
+  const { sendEvent } = useProjectHealthAnalytics();
   const { inputValue, setAndSubmitInputValue } = useStatusesFilter({
     urlParam: PatchTasksQueryParams.Statuses,
+    sendAnalyticsEvent: (_, filterValue: string[]) => {
+      sendEvent({
+        name: "Filter commit chart by task status",
+        statuses: filterValue,
+      });
+    },
```
khelif96, line 20: This event belongs in the onChange callback of TreeSelect not in the hook.

src/pages/taskHistory/TaskHistoryRow.tsx
```diff
@@ -38,6 +42,13 @@ const TaskHistoryRow = memo((props: ListChildComponentProps) => {
               failingTests={failingTests}
               label={label}
               loading={loading}
+              onClick={(v: TaskCellAnalytics) => {
+                sendEvent({
+                  name: "Click task cell",
+                  page: "Task history",
+                  ...v,
+                });
+              }}
```
khelif96, line 51: We don't need to log page or any additional information. Since the page is logged automatically. Maybe the task status.

src/utils/string/jiraLinkify.tsx
khelif96: I think you missed this comment.

khelif96 (changes requested):

src/pages/commits/InactiveCommits/index.tsx
```diff
@@ -110,23 +125,54 @@ export const InactiveCommitButton: React.VFC<InactiveCommitsProps> = ({
...
+      </CommitTitleText>
+      <CommitBodyText>
+        {v.author} -{" "}
+        {jiraLinkify(message, jiraHost, () => {
+          sendEvent({
+            name: "Click commit label",
+            commitType: "inactive",
+            link: "jira",
+          });
+        })}{" "}
```
khelif96, line 169: Nice catch linkifying the jira ticket.

src/pages/commits/InactiveCommits/index.tsx
```diff
@@ -110,23 +125,54 @@ export const InactiveCommitButton: React.VFC<InactiveCommitsProps> = ({
...
+        (#
+        {v.order})
```
khelif96, line 171: nit. unnecessary new line?

src/utils/string/jiraLinkify.tsx
```diff
@@ -5,11 +5,11 @@ import { getJiraTicketUrl } from "constants/externalResources";
 export const jiraLinkify = (
   unlinkified: string | React.ReactNodeArray,
   jiraHost: string,
-  onClick?: (match: string) => () => void
+  onClick: () => void = () => {}
```
khelif96, line 8: if the prop is required it does not need a default value. we should make this optional again.

khelif96 (approved):
Great work on this! Thank you! 🔥

PR #1187, opened 2022-04-01
EVG-16548: Bold date in commit chart label

SupaJoon:
[EVG-16548](https://jira.mongodb.org/browse/EVG-16548)

### Description 
I used the b tag instead of a leafygreen typography component to avoid adding styles for positioning. The only LG component that accepts `weight` is body and places the text on a new line. 

### Screenshots
Screen Shot 2022-04-01 at 2 13 00 PM


src/components/CommitChartLabel/index.tsx
```diff
@@ -48,7 +48,7 @@ const CommitChartLabel: React.FC<Props> = ({
         <StyledRouterLink to={getVersionRoute(versionId)}>
           {shortenGithash(githash)}
         </StyledRouterLink>{" "}
-        {shortDate(createDate)}
+        <b>{shortDate(createDate)}</b>
```
minnakt, line 51: I'm not sure how but I think the bold might have caused a small bug where all of the commit labels became scrollable 🤔

https://user-images.githubusercontent.com/47064971/161333443-6c7e564c-aecb-4eac-b2f2-40ba41a7a526.mov

(I double checked on `main` as well and didn't see this behavior)

SupaJoon: The bold lettering added .5px to the commit chart label height which enabled the scroll bar. I removed the height since we already limit the height by truncating the commit description.

minnakt (approved):
💚🐛

PR #1198, opened 2022-04-04
EVG-16567 Collapse unmatching commits

khelif96:
[EVG-16567](https://jira.mongodb.org/browse/EVG-16567)

### Description 
Commits that do not contain matching tasks will be collapsed by default.
### Screenshots
before: 
image

after:
image

### Testing 

### Evergreen PR 


src/gql/queries/get-mainline-commits-for-history.graphql
```diff
@@ -2,7 +2,10 @@ query MainlineCommitsForHistory(
   $mainlineCommitsOptions: MainlineCommitsOptions!
```
minnakt, line 2: Random sidenote: I encountered this bug where it keeps leaping to the `selectedCommit` when I scroll to some loading rows. 

https://user-images.githubusercontent.com/47064971/162007284-ffbe5287-0927-4744-ae8b-045d2ea9fe7b.mov

I _think (???)_ it might be fixed by changing the dependency hook to this in `HistoryTable.tsx`:

Screen Shot 2022-04-06 at 11 02 53 AM

This PR might be a good place to fix it!

src/gql/queries/get-mainline-commits-for-history.graphql
```diff
@@ -2,7 +2,10 @@ query MainlineCommitsForHistory(
...
   $buildVariantOptions: BuildVariantOptions!
 ) {
-  mainlineCommits(options: $mainlineCommitsOptions) {
+  mainlineCommits(
```
minnakt, line 5: Just to confirm based on the behavior in this PR, the collapsing should only happen when visiting a page and not when applying any task/variant filters? If so then maybe the description in the ticket [EVG-16567](https://jira.mongodb.org/browse/EVG-16567) should be updated

minnakt (commented):
I think this looks good, just had 1 comment and 1 question!

SupaJoon (commented):
Should the commit that scrolled to in this video be inactive? 
http://localhost:3000/variant-history/spruce/ubuntu1604?selectedCommit=1399&visibleColumns=compile

https://user-images.githubusercontent.com/10734386/162017646-1cefa184-c122-488d-87c2-ee33b6c7de6d.mov

khelif96:
> Should the commit that scrolled to in this video be inactive? http://localhost:3000/variant-history/spruce/ubuntu1604?selectedCommit=1399&visibleColumns=compile
> 
>  Screen.Recording.2022-04-06.at.11.56.52.AM.mov

Nope you have a column filter applied. Column filters are only toggled on the UI and are not included in the folding. (There are other tasks in that variant that are active). Otherwise we would have to recalculate the entire table every time someone changes a column filter.

src/gql/queries/get-mainline-commits-for-history.graphql
khelif96: Wow Great Catch!!!!

minnakt: Hmm this bug is still occurring for me... would you mind taking a second look? (But if it seems not easily fixable, then we can just look at it another time 😅)

src/components/HistoryTable/HistoryTable.tsx
```diff
@@ -52,13 +53,21 @@ const HistoryTable = ({
...
   }, [processedCommitCount]);
 
+  console.log({ processedCommitCount });
   useEffect(() => {
-    if (selectedCommit && selectedCommit.rowIndex && listRef.current) {
+    if (
+      selectedCommit &&
+      selectedCommit.rowIndex &&
+      listRef.current &&
+      !selectedCommit.visited
```
minnakt, line 62: Maybe I am wrong but I still don't understand why this `visited` property is needed.

For example, I changed the hook to this, and the bug doesn't seem to occur anymore:

Screen Shot 2022-04-08 at 10 03 18 AM

Which sort of makes me feel like this can be done without the `visited` property, but let me know if there's something I am overlooking

PR #1202, opened 2022-04-05
EVG-16544: Order common build variants on Project Health page

minnakt:
[EVG-16544](https://jira.mongodb.org/browse/EVG-16544)

### Description 
This PR orders the build variants on the Project Health page so that it is easier to compare the build results between two versions.

To do this I constructed a build variant dictionary (`Object`) which maps each build variant to a `priority`. The higher the `priority`, the higher the build variant will appear on the page. The dictionary also keeps track of `iconHeight` and `cardHeight` to prevent ugly misalignment issues. For example, even though one build variant has more task icons in the image below, they are rendered such that they take up the same amount of space.

Screen Shot 2022-04-04 at 11 45 23 PM

### Note
It doesn't work perfectly every time. For example in the image below you can see that`~ Shared Library Enterprise RHEL 8.0 (BSONColumn validate)` is not aligned across every column, because one column ran out of other build variants. I did explore rendering an empty cell in these cases to force alignment, but it would introduce a lot of weird looking white space.

Screen Shot 2022-04-05 at 9 56 29 AM

### Screenshots

https://user-images.githubusercontent.com/47064971/161790152-e75278e2-201c-40ca-8b3a-a36e8e870d17.mov

### Testing 
- Added test in `ActiveCommits/utils.ts` for new function `constructBuildVariantDict`
- Edited storybook `ProjectHealth.stories.ts` to reflect the changes


src/pages/commits/constants.ts
```diff
@@ -1,4 +1,21 @@
...
 const gridHeight = 226;
 
-export { commitChartHeight, gridHeight };
+const TASK_ICONS_PER_ROW = 6;
+const TASK_ICON_HEIGHT = 24;
+const TASK_ICON_PADDING = 16;
+
+const GROUPED_BADGE_PER_ROW = 2;
+const GROUPED_BADGE_HEIGHT = 40;
+const GROUPED_BADGE_PADDING = 8;
```
khelif96, line 10: Where do these values come from?

khelif96 (approved):
This is awesome! We should make sure this is documented in the onboarding docs as well. It could just use some extra tests but apart from that LGTM!

src/pages/commits/constants.ts
minnakt: The `TASK_ICONS_PER_ROW` and `GROUPED_BADGE_PER_ROW` are more observational values. From the image below you can see that at most there will be 6 task icons per row, and 2 grouped badges per row:

Screen Shot 2022-04-08 at 11 22 44 AM

The padding and height correspond to the actual padding/heights of the task icons and grouped badges. To make it clearer, I altered the styles of these components so that these constants are used when appropriate 😇

src/pages/commits/ActiveCommits/BuildVariantCard.tsx
```diff
@@ -55,7 +58,7 @@ export const BuildVariantCard: React.VFC<Props> = ({
       >
         {buildVariantDisplayName}
       </Label>
-      {render}
+      <div style={{ height }}>{render}</div>
```
SupaJoon, line 61: can we use a styled component here?

src/pages/commits/ActiveCommits/utils.ts
```diff
@@ -49,6 +47,68 @@ export const getAllTaskStatsGroupedByColor = (
...
+    // skip if inactive/unmatching
+    if (version) {
+      // Deduplicate build variants and build variant stats by consolidating into a
+      // single object.
+      const allBuildVariants = [
+        ...version.buildVariants,
+        ...version.buildVariantStats,
+      ].reduce((acc, curr) => {
+        const { variant } = curr;
+        acc[variant] = { ...(acc[variant] || {}), ...curr };
```
SupaJoon, line 67:
```suggestion
        acc[variant] = { ...acc[variant], ...curr };
```

src/pages/commits/ActiveCommits/utils.ts
```diff
@@ -49,6 +47,68 @@ export const getAllTaskStatsGroupedByColor = (
...
+        return acc;
+      }, {});
+
+      // Construct build variant dict which will contain information needed for
+      // rendering.
+      Object.keys(allBuildVariants).reduce((acc, curr) => {
+        const { tasks, statusCounts, variant } = allBuildVariants[curr];
```
SupaJoon, line 74:
```suggestion
      Object.values(allBuildVariants).reduce((acc,{ tasks, statusCounts, variant }) => {
```

src/types/commits.ts
```diff
@@ -32,3 +32,10 @@ export type Commits = MainlineCommitsQuery["mainlineCommits"]["versions"];
 export type Commit = Unpacked<Commits>;
 export type CommitVersion = Commit["version"];
 export type CommitRolledUpVersions = Commit["rolledUpVersions"];
+export type BuildVariantDict = {
+  [key: string]: {
```
SupaJoon, line 36: can we change key to a more descriptive name

src/pages/commits/CommitsWrapper.tsx
```diff
@@ -80,6 +81,12 @@ export const CommitsWrapper: React.VFC<Props> = ({
     }
   }, [versionToGroupedTaskStatsMap]);
 
+  const buildVariantDict = useMemo(() => {
+    if (versions) {
+      return constructBuildVariantDict(versions);
+    }
+  }, [versions]);
```
SupaJoon, line 88:
```suggestion
  const buildVariantDict = useMemo(() =>  
      versions ? constructBuildVariantDict(versions) : undefined 
  ,[versions]);
```

SupaJoon (approved):

PR #1205, opened 2022-04-06
EVG-16681 Update React.FC to be React.VFC

khelif96:
[EVG-16681](https://jira.mongodb.org/browse/EVG-16681)

### Description 
This PR updates our usage of the `React.FC` prop to `React.VFC` which is functionally the same as `React.FC` without the explicit `children` prop.
 @SupaJoon  and I discovered this new type after discussing typing react components without `React.FC` like so
```ts
const SomeComponent = (props: SomePropType) => (...someReturn)
```
While the above example is explicit about the props the component accepts, avoiding a situation in which there is an incorrectly added `children` prop as is the case with `React.FC` it does not have an explicit return type defined.
This can potentially result in a user error in which we accidentally return the wrong thing from a component. 
`React.VFC` solves this issue by not including the unexpected `children` prop while preserving the the return type safety.

I think in general if you are creating a new component that does **require** the `children` prop it is safe to use either `React.FC` or `React.VFC` and explicitly type `children`. 

This is all subject to change, there is supposedly the plan to update `React.FC` to have the same behavior as `React.VFC` by default in the next major release. (React 18) 
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/46643#issuecomment-801487166

But since we will likely not be upgrading to React 18 in the near future i think it's ok to make this change for now and update it when we do.

Read more here. (Yes the naming sucks)
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/46643

### DIFF
--- a/src/components/DisplayModal.tsx
+++ b/src/components/DisplayModal.tsx
@@ -15,7 +15,7 @@ interface DisplayModalProps {
   children: React.ReactNode;
 }
 
-export const DisplayModal: React.FC<DisplayModalProps> = ({
+export const DisplayModal: React.VFC<DisplayModalProps> = ({
   children,
   "data-cy": dataCy,
   open,
--- a/src/components/DisplayModal.tsx
+++ b/src/components/DisplayModal.tsx
@@ -12,6 +12,7 @@ interface DisplayModalProps {
   ) => void | React.Dispatch<React.SetStateAction<boolean>>;
   size?: ModalSize;
   title?: string;
+  children: React.ReactNode;
 }
 
 export const DisplayModal: React.FC<DisplayModalProps> = ({
--- a/src/components/Spawn/Layout.tsx
+++ b/src/components/Spawn/Layout.tsx
@@ -1,6 +1,7 @@
 import React from "react";
 import styled from "@emotion/styled";
-import Button from "@leafygreen-ui/button";
+import { ExtendableBox } from "@leafygreen-ui/box";
+import Button, { ButtonProps } from "@leafygreen-ui/button";
 import { Body, H2 } from "@leafygreen-ui/typography";
 import { Table } from "antd";
 import Badge from "components/Badge";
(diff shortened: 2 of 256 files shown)


src/components/DisplayModal.tsx
```diff
@@ -12,6 +12,7 @@ interface DisplayModalProps {
   ) => void | React.Dispatch<React.SetStateAction<boolean>>;
   size?: ModalSize;
   title?: string;
+  children: React.ReactNode;
 }
 
 export const DisplayModal: React.FC<DisplayModalProps> = ({
```
sophstad, line 18:
```suggestion
export const DisplayModal: React.VFC<DisplayModalProps> = ({
```

src/context/Providers.tsx
```diff
@@ -3,7 +3,9 @@ import LeafyGreenProvider from "@leafygreen-ui/leafygreen-provider";
 import { AuthProvider } from "context/auth";
 import { ToastProvider } from "context/toast";
 
-export const ContextProviders: React.FC = ({ children }) => (
+export const ContextProviders: React.FC<{ children: React.ReactNode }> = ({
```
sophstad, line 6:
```suggestion
export const ContextProviders: React.VFC<{ children: React.ReactNode }> = ({
```

sophstad (approved):
LGTM once all remaining instances of `React.FC` are converted into `React.VFC` 👍 thanks!!

src/components/Spawn/Layout.tsx
```diff
@@ -22,18 +23,25 @@ export const StyledBadge = styled(Badge)`
   margin: 0 ${size.xs};
 `;
 
-export const PlusButton = ({ children, ...props }) => (
-  <Button {...{ ...props, glyph: <Icon glyph="Plus" /> }} as="button">
-    {children}
-  </Button>
-);
+export const PlusButton: ExtendableBox<
```
khelif96, line 26: This button was never actually typed correctly so it broke when we updated the Leafygreen Button and the props changed.

SupaJoon (approved):

PR #1207, opened 2022-04-07
EVG-16691 Add Queries for ProjectEvents and RepoEvents

malikchaya2:
[EVG-16691](https://jira.mongodb.org/browse/EVG-16691)

### Description 
This adds the queries that will be needed for the event log page. This includes a lot of fragments and files so I wanted to separate it out into its own PR so that it doesn't become overwhelming. 

### Evergreen PR
https://github.com/evergreen-ci/evergreen/pull/5552


src/gql/fragments/projectSettings/githubCommitQueue.graphql
```diff
@@ -35,3 +35,41 @@ fragment repoGithubCommitQueue on RepoSettings {
...
+    gitTagVersionsEnabled
+    gitTagAuthorizedUsers
+    gitTagAuthorizedTeams
+    commitQueue {
+      enabled
+      requireSigned
+      mergeMethod
+      message
+    }
+  }
```
sophstad, line 55: [nit] It may be preferable to make two fragments on `Project` and `RepoRef` to use within this file... then we can repeat the same thing only twice instead of 4x 😅

src/gql/generated/types.ts
```diff
@@ -1502,7 +1502,7 @@ export type ProjectEventSettings = {
 
 export type RepoEventSettings = {
   githubWebhooksEnabled: Scalars["Boolean"];
-  projectRef?: Maybe<Project>;
+  projectRef?: Maybe<RepoRef>;
```
sophstad, line 1505: This app change is a breaking change, but since the event page isn't used anywhere or tested yet, I think there should be no impact 👍

src/gql/fragments/projectSettings/projectEventSettings.graphql
```diff
@@ -0,0 +1,107 @@
...
+      alias      
+    }     
+    periodicBuilds{
+      id
+      configFile
+      intervalHours
+      alias
+      message
+      nextRunTime
+    }
```
sophstad, line 94: You probably worked on this before #1181 was merged in, but now you can use a periodic builds fragment!

sophstad (approved):
Just some minor notes!

src/gql/fragments/projectSettings/githubCommitQueue.graphql
malikchaya2: good point, thanks!

PR #1208, opened 2022-04-08
EVG-16700 Fix infinite loading bug on history tables and bug with cached commits not being scrolled to.

khelif96:
[EVG-16700](https://jira.mongodb.org/browse/EVG-16700)

### Description 

This fixes a few bugs  in the history table.

The first was the history table was endlessly loading data. This was due to a faulty exit condition in the logic to load more commits for the tables jump to commit functionality it would always default to loading new rows. I added a check to explicitly check if we already loaded the row and if we did it wouldn't attempt to load anymore. I think it makes sense to be explicit about this value since it was difficult to trace down. Being more explicit about these values and what they do will make it easier to understand whats going on in the future.

The second issue was the table would continuously jump back to the selected commit as new data was being loaded in by the user scrolling. This was caused because the conditional that guards the `jumpToRow` function had no method to determine if it has already scrolled to that row. This was an issue because the `useEffect` would get triggered when ever the table data was changed and then it would run that function. 

This also fixes a bug where the history table was unable to jump to the selected commit if the commits originated from the apollo cache. That was due to a race condition in which the `selectedCommit` was being set after the table processed the first batch of commits that lived in the cache. This was solved by having the table parse the `processedCommits` array if it knows the table for sure has the value. 

 
### Screenshots

https://user-images.githubusercontent.com/4605522/162510430-b0fd1071-f77d-4072-b1fd-dc83e45adf2c.mov


src/components/HistoryTable/historyTableContextReducer.ts
```diff
@@ -75,36 +84,29 @@ export const reducer = (state: HistoryTableReducerState, action: Action) => {
...
-        action.numCommits,
-        action.index
-      );
-      return {
-        ...state,
-        processedCommits: newProcessedCommits,
-      };
-    }
-    case "addColumns":
+
```
minnakt, line 102: nit. unnecessary new line

src/components/HistoryTable/historyTableContextReducer.ts
```diff
@@ -141,32 +156,50 @@ export const reducer = (state: HistoryTableReducerState, action: Action) => {
...
+      let loaded = false;
+      const updatedProcessedCommits = state.processedCommits;
+      // If we already loaded the commit we need to search for it in the processed commits
+      // to set its selected state to true. This is in the case where we have the commit in the cache.
+      const hasMatchingCommit = state.commitCache.has(action.order);
+      if (hasMatchingCommit) {
+        rowIndex = state.processedCommits.findIndex((commit) =>
+          commitOrderToRowIndex(action.order, commit)
+        );
+        updatedProcessedCommits[rowIndex].selected = true;
```
minnakt, line 175: Do we need to do this on `L175`? I was fiddling around and printing the before and after values for `state.processedCommits` when this condition ran, and it seemed like the `selected` field was already marked `true` for the commit by default (which might make sense based on how `processCommits` is written)

src/components/HistoryTable/HistoryTable.tsx
```diff
@@ -52,13 +53,28 @@ const HistoryTable: React.VFC<HistoryTableProps> = ({
...
+      listRef.current
+    ) {
       listRef.current.scrollToItem(selectedCommit.rowIndex, "center");
-    } else {
+      // We mark the selected commit as visited so that we don't keep trying to scroll to it.
+      markSelectedRowVisited();
+      // If we have not loaded the commit yet we need to keep fetching rows until we do.
+    } else if (selectedCommit && !selectedCommit.loaded) {
       loadMoreItems();
     }
```
minnakt, line 71: It might be more concise if we used a guard here

```
if (selectedCommit) {
    if (selectedCommit.rowIndex &&...) { // stuff...}
    else if (!selectedCommit.loaded) { //stuff...}
}
```

In this case I think you can also extract the fields you want like `const { rowIndex, visited, loaded } = selectedCommit;` and use those in the conditionals, instead of repeating `selectedCommit` in multiple places

minnakt (approved):
🪲🤺 defeated that bug

src/components/HistoryTable/historyTableContextReducer.ts
khelif96: The bug was caused by `ingestCommits` and `setSelectedCommit` racing. Since `ingestCommits` would run immediately before `setSelectedCommit`  had a chance to run for commits that already existed in the cache. So if the commits are processed before we call set `setSelectedCommit` `processCommits` would have never set the selected field. Its not necessary for new commits but is for cached commits.

PR #1211, opened 2022-04-11
EVG-12900: Add Task Duration tab on patches and versions

minnakt:
[EVG-12900](https://jira.mongodb.org/browse/EVG-12900)

### Description 
This PR adds a Task Duration tab for patches and versions which can be used to compare task run times. The task duration is shown in a table which is paginated and sorted by duration descending by default.

### Note
It's only possible to sort by task duration. I feel like this makes sense as Leafygreen doesn't support multisort and it would be somewhat meaningless to sort by anything else on the Task Duration tab. There is also the option to filter by task name or build variant.

### Screenshots
https://user-images.githubusercontent.com/47064971/162787972-f263b110-26b4-4e5a-84ca-223d9b208b85.mov

### Testing 
- Added test in `page_tabs.ts` (Cypress) but it is skipped due to the `isBeta` flag
- Added tests in `version/task_duration.ts` (Cypress) but it is skipped due to the `isBeta` flag

### Evergreen PR
- evergreen-ci/evergreen/pull/5563


src/pages/version/taskDuration/TaskDurationTable.tsx
```diff
@@ -0,0 +1,246 @@
...
+    </TableWrapper>
+  );
+};
+
+export const findMax = (patchTasks: PatchTasksQuery["patchTasks"]["tasks"]) =>
+  patchTasks.reduce((prev, curr) => {
+    const prevTimeTaken = prev?.taskDuration;
+    const currTimeTaken = curr?.taskDuration;
+    return prevTimeTaken > currTimeTaken ? prev : curr;
+  }).taskDuration;
```
khelif96, line 170: `Math.max(patchTasks.map(t => t.taskDuration))`

src/pages/version/taskDuration/TaskDurationTable.tsx
```diff
@@ -0,0 +1,246 @@
...
+      padding-right: ${size.l};
+    }
+    :nth-of-type(2) {
+      width: 20%;
+      padding-right: ${size.l};
+    }
+    :nth-of-type(3) {
+      span {
+        width: 100%;
+      }
```
khelif96, line 213: Does styling the `TableHeaders` directly work?

src/utils/queryString/sortString.ts
```diff
@@ -36,6 +36,23 @@ export const toSortString = (
...
+    if (sort.Direction !== undefined) {
+      const singleSortString = `${sort.Key}:${sort.Direction}`;
+      sortStrings = sortStrings.concat(singleSortString);
+    }
+  });
+
+  return sortStrings.some((s) => s)
+    ? sortStrings.filter(Boolean).join(";")
+    : undefined;
+};
```
khelif96, line 54: I discovered recently that queryString can do this natively. It  can generate a query param like this natively. https://www.npmjs.com/package/query-string

khelif96 (changes requested):

src/gql/queries/get-patch-tasks.graphql
```diff
@@ -42,6 +43,7 @@ query PatchTasks(
         baseStatus
         buildVariantDisplayName
         projectIdentifier
+        taskDuration
```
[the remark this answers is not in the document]
minnakt: I do show them in the table! (They are an expandable row, one is actually viewable on the big patch on `dev`)

I thought it might be helpful for a user to see which one of the execution tasks is taking really long, if the display task is taking very long. I can remove it though 😲

src/pages/version/TaskDuration.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+  const { data, startPolling, stopPolling } = useQuery<
+    PatchTasksQuery,
+    PatchTasksQueryVariables
+  >(GET_PATCH_TASKS, {
+    variables: queryVariables,
+    skip: noQueryVariables,
+    pollInterval,
+    fetchPolicy: "network-only",
+    nextFetchPolicy: "cache-and-network",
```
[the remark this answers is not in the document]
minnakt: It is because when I used `cache-first`, the tasks weren't being updated after reconfiguring a patch

https://user-images.githubusercontent.com/47064971/163219854-fa932cd7-c2fd-4b32-9b87-1905ccc54d3a.mov

src/utils/queryString/sortString.ts
minnakt: I gave this a try but I'm not sure if it'll work as intended...

I read the documentation and I think what needs to happen is that the `SortOrder[]` needs to be combined into an object that looks like `{ sorts: ["STATUS:ASC", "DURATION:ASC", "NAME:DESC"] }` and then use the `stringify() `function with  a separator of `;`.

This function will return `"sorts=STATUS:ASC;DURATION:ASC;NAME:DESC"`, but because `updateQueryParams` updates a specific field the param will end up looking like `"sorts=sorts=STATUS:ASC;DURATION:ASC;NAME:DESC"` which is not correct. I might just be not understanding something though 🥲

src/pages/version/TaskDuration.tsx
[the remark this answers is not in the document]
SupaJoon: I'm not completely sure even after reading the docs but I think nextFetchPolicy will actually read updates from cache and make a network request for the next poll interval. If this is the case, we may be okay omitting the cache read if another component isn't updating the cache while TaskDuration is being rendered.

cypress/integration/page_tabs.ts
```diff
@@ -28,29 +41,29 @@ describe("Tabs", () => {
...
-        "sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC"
+        "sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC%3BDURATION%3ADESC"
       );
 
       cy.dataCy("task-tab").first().click();
       locationPathEquals(patches.tasks.route);
       cy.location("search").should(
         "contain",
-        "sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC"
+        "sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC%3BDURATION%3ADESC"
```
khelif96, line 51: Should duration be a query param when you aren't looking at the task tab?

src/gql/queries/get-patch-task-durations.graphql
```diff
@@ -0,0 +1,35 @@
...
+      displayName
+      buildVariantDisplayName
+      timeTaken
+      executionTasksFull {
+        id
+        execution
+        displayName
+        status
+        buildVariantDisplayName
+        timeTaken
```
khelif96, line 31: It doesn't look like this value is being resolved for execution tasks.
image

src/pages/version/TaskDuration.tsx
```diff
@@ -0,0 +1,152 @@
...
+import { TaskDurationTable } from "./taskDuration/TaskDurationTable";
+
+const { parseQueryString, parseSortString, getString } = queryString;
+
+const { getPageFromSearch, getLimitFromSearch } = url;
+interface Props {
+  taskCount: number;
+}
+
+export const TaskDuration: React.VFC<Props> = ({ taskCount }) => {
```
khelif96, line 32: Why does taskCount get passed in?

src/pages/version/TaskDuration.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+  const updateQueryParams = useUpdateURLQueryParams();
+  const queryVariables = getQueryVariables(search, versionId);
+  const noQueryVariables = !search.length;
+
+  const { limit, page } = queryVariables;
+
+  const { sorts } = parseQueryString(search);
+  const allSorts = parseSortString(sorts);
+  const defaultSortMethod = "STATUS:ASC;BASE_STATUS:DESC;DURATION:DESC";
```
khelif96, line 47: You can't sort on status on the taskDuration table.

src/pages/version/TaskDuration.tsx
```diff
@@ -0,0 +1,152 @@
...
+  }, [search]); // eslint-disable-line react-hooks/exhaustive-deps
+
+  const { data, startPolling, stopPolling } = useQuery<
+    PatchTaskDurationsQuery,
+    PatchTaskDurationsQueryVariables
+  >(GET_PATCH_TASK_DURATIONS, {
+    variables: queryVariables,
+    skip: noQueryVariables,
+    pollInterval,
+    fetchPolicy: "network-only",
```
khelif96, line 64: Why the network-only fetch policy?

src/pages/version/TaskDuration.tsx
```diff
@@ -0,0 +1,152 @@
...
+            data-cy="tasks-table-pagination"
+            pageSize={limit}
+            value={page}
+            totalResults={patchTasks?.count}
+          />
+          <PageSizeSelector
+            data-cy="tasks-table-page-size-selector"
+            value={limit}
+            sendAnalyticsEvent={() => sendEvent({ name: "Change Page Size" })}
+          />
```
khelif96, line 111: Can we share these between the tasks and task duration table?

src/pages/version/Tasks.tsx
```diff
@@ -131,9 +131,14 @@ const getQueryVariables = (
     [PatchTasksQueryParams.Sorts]: sorts,
   } = parseQueryString(search);
 
+  // Don't include duration when sorting on PatchTasks table.
+  const filteredSorts = parseSortString(sorts).filter(
+    (sort) => sort.Key !== "DURATION"
+  );
```
khelif96, line 137: This doesn't seem necessary? maybe the two tabs should handle their own sorting and query param logic?

src/pages/version/taskDuration/TaskDurationTable.tsx
```diff
@@ -0,0 +1,259 @@
...
+                  status,
+                  buildVariantDisplayName,
+                  timeTaken,
+                } = task;
+                const barWidth = calculateBarWidth(timeTaken, max);
+                const barColor =
+                  mapTaskToBarchartColor[mapTaskStatusToUmbrellaStatus[status]];
+                return (
+                  <Row key={id}>
+                    <Cell>
```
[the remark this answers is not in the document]
khelif96: I think you can just put a displayTaskRow here? Or does LG not catch it?

src/pages/version/taskDuration/TaskDurationTable.tsx
```diff
@@ -0,0 +1,259 @@
...
+  border-top: 3px solid ${gray.light2};
+  // Styling the Cell directly doesn't work, so styling is done through the table.
+  td {
+    word-break: break-all;
+    padding-right: ${size.m};
+    :nth-of-type(3) {
+      span {
+        width: 100%;
+    }
+  }
```
khelif96, line 219: Why are these necessary?

src/pages/version/taskDuration/TaskDurationTable.tsx
```diff
@@ -0,0 +1,259 @@
...
+  const durations = tasks.map((t) => t.timeTaken);
+  return Math.max(...durations);
+};
+
+const calculateBarWidth = (value: number, max: number) =>
+  max ? `${(value / max) * 100}%` : "0%";
+
+const TableWrapper = styled.div`
+  border-top: 3px solid ${gray.light2};
+  // Styling the Cell directly doesn't work, so styling is done through the table.
```
khelif96, line 211: I was able to apply styles directly to the cell.
image

You need to use classNames https://emotion.sh/docs/class-names to style the cells.
I think LG does [some matching under the hood](https://github.com/mongodb/leafygreen-ui/blob/89390382406cf0a45896714f9c8aff17ad35042f/packages/lib/src/index.ts#L61) to render its elements. I think emotion breaks this behavior since it creates a new component with a new display name.
https://github.com/mongodb/leafygreen-ui/blob/main/packages/table/src/Row.tsx

Alternatively you can set the displayName of your styled component to `Cell`  to trick LG into thinking its the same thing.

khelif96: might be worth bring up with them if theres a preferred way.

khelif96 (changes requested):
image
Wow these server builds do go on for a REALLLLLLY long time but i don't think for this long.

src/pages/version/taskDuration/TaskDurationTable.tsx
minnakt: I went with the second option (setting the displayName), but only because using the `ClassNames` notation breaks the execution tasks 🥲

I think Leafygreen's nested row policy must be strict, it seems like `Row` needs to be the direct descendant in order for it to work...

src/pages/version/taskDuration/TaskDurationTable.tsx
minnakt: removed

src/pages/version/TaskDuration.tsx
minnakt: Comes from this discussion above https://github.com/evergreen-ci/spruce/pull/1211#discussion_r849642703

src/pages/version/taskDuration/TaskDurationTable.tsx
```diff
@@ -0,0 +1,279 @@
+import styled from "@emotion/styled";
```
minnakt, line 1: Sorry this file is still not great, I don't know if it's possible to simplify any further given the restrictions with the Table.

src/pages/version/taskDuration/TaskDurationTable.tsx
```diff
@@ -0,0 +1,279 @@
...
+  align-items: center;
+`;
+
+const TaskNameCell = styled(Cell)`
+  word-break: break-all;
+  padding-right: ${size.m};
+  svg {
+    margin-left: ${size.xxs};
+    vertical-align: text-bottom;
+  }
```
khelif96, line 246: You can apply a wrapper to the contents in the task Cell and do an `align-items: center` to align the task icons. (Which is what i assume this is doing)

Regardless i am not really a big fan of how this looks for larger task names. Can we just add another column for task status?

khelif96 (changes requested):

src/pages/version/taskDuration/TaskDurationTable.tsx
```diff
@@ -0,0 +1,279 @@
...
+                    <BuildVariantCell>
+                      {task.buildVariantDisplayName}
+                    </BuildVariantCell>
+                    <DurationCell>
+                      <DurationBar width={barWidth} color={barColor} />
+                      <DurationLabel>
+                        {msToDuration(task.timeTaken)}
+                      </DurationLabel>
+                    </DurationCell>
+                  </Row>
```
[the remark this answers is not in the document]
minnakt: I wish 😞 I've tried extracting the whole thing into a component, the row contents into a component, the individual cells into a component... it doesn't accept anything

minnakt (commented):
Commenting for update

src/pages/version/TaskDuration.tsx
```diff
@@ -0,0 +1,78 @@
...
+  };
+
+  const { data, startPolling, stopPolling } = useQuery<
+    PatchTaskDurationsQuery,
+    PatchTaskDurationsQueryVariables
+  >(GET_PATCH_TASK_DURATIONS, {
+    variables: queryVariables,
+    skip: noQueryVariables,
+    pollInterval,
+    fetchPolicy: "network-only",
```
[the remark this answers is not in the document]
khelif96: This is because of the ternary on Line 66 https://github.com/evergreen-ci/spruce/pull/1211/files#diff-79c22cee051ee418ff8d94e75958e9b84cf35cee255bf7a5db8fb636846dd552R66-R68 
The table gets unmounted every time we load something. Since tasks becomes undefined.

src/pages/version/taskDuration/TaskDurationTable.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+  const { tasks } = patchTasks ?? {};
+  const maxTimeTaken = findMaxTimeTaken(tasks);
+
+  return !tasks ? (
+    <Skeleton active title={false} paragraph={{ rows: 8 }} />
+  ) : (
+    <TableWrapper>
+      <Table
+        data={tasks}
```
khelif96, line 71:
```suggestion
  const { tasks = [] } = patchTasks ?? {};
  const maxTimeTaken = findMaxTimeTaken(tasks);

  return (
    <TableWrapper>
      <Table
        data={tasks}
```

src/pages/version/TaskDuration.tsx
```diff
@@ -0,0 +1,78 @@
...
+}
+
+export const TaskDuration: React.VFC<Props> = ({
+  taskCount,
+  queryVariables,
+}) => {
+  const dispatchToast = useToastContext();
+  const updateQueryParams = useUpdateURLQueryParams();
+
+  const noQueryVariables = !Object.keys(queryVariables).length;
```
khelif96, line 28: This will never be met. VersionID is always defined.

src/pages/version/TaskDuration.tsx
```diff
@@ -0,0 +1,78 @@
...
+} from "gql/generated/types";
+import { GET_PATCH_TASK_DURATIONS } from "gql/queries";
+import { usePolling } from "hooks";
+import { useUpdateURLQueryParams } from "hooks/useUpdateURLQueryParams";
+import { TableControl } from "./TableControl";
+import { TaskDurationTable } from "./taskDuration/TaskDurationTable";
+
+interface Props {
+  taskCount: number;
+  queryVariables: PatchTasksQueryVariables;
```
khelif96, line 18: Why do we pass this in instead of calling the hook?

src/pages/version/taskDuration/TaskDurationTable.tsx
minnakt: discussed offline!

src/pages/version/TaskDuration.tsx
minnakt: Woops, possibly misunderstood your comment [here](https://github.com/evergreen-ci/spruce/pull/1211#discussion_r861210816). I put it back to the way it was

minnakt:
If you see a bug where there is abnormally large display task rows, it's because nested rows take up 1px. 

Screen Shot 2022-05-03 at 2 04 36 PM

This bug has since been resolved in [PD-1655](https://jira.mongodb.org/browse/PD-1655).

---

I tried to upgrade the Table by one version to get this fix but it wouldn't compile🧍most likely dependency issues

src/components/TablePopover/TableFilterPopover.tsx
```diff
@@ -0,0 +1,72 @@
...
+  return (
+    <div>
+      <IconWrapper
+        active={active}
+        onClick={() => setActive(!active)}
+        data-cy={dataCy}
+        ref={buttonRef}
+      >
+        <Icon glyph="Filter" small="xsmall" color={iconColor} />
+      </IconWrapper>
```
khelif96, line 48: Does an IconButton work here?

src/pages/version/TaskDuration.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { TaskDurationTable } from "./taskDuration/TaskDurationTable";
+import { useQueryVariables } from "./useQueryVariables";
+
+const { parseQueryString } = queryString;
+
+interface Props {
+  taskCount: number;
+}
+
+export const TaskDuration: React.VFC<Props> = ({ taskCount }) => {
```
khelif96, line 24: nit. default export?

src/pages/version/Tasks.tsx
```diff
@@ -32,23 +20,19 @@ export const Tasks: React.VFC<Props> = ({ taskCount }) => {
...
-  const { sendEvent } = useVersionAnalytics(versionId);
   const dispatchToast = useToastContext();
 
   const updateQueryParams = useUpdateURLQueryParams();
-  const queryVariables = getQueryVariables(search, versionId);
   const noQueryVariables = !search.length;
-
+  const queryVariables = useQueryVariables(search, versionId);
   const { sorts, limit, page } = queryVariables;
   const defaultSortMethod = "STATUS:ASC;BASE_STATUS:DESC";
```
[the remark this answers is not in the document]
khelif96: Reminder.

src/pages/version/useQueryVariables.ts
```diff
@@ -0,0 +1,42 @@
...
+  } = queryParams;
+
+  // This should be reworked once the antd tables are removed.
+  // At the current state, sorts & duration will never both be defined.
+  let sortsToApply: SortOrder[];
+  if (sorts) {
+    sortsToApply = parseSortString(sorts);
+  }
+  if (duration) {
+    sortsToApply = parseSortString(`DURATION:${duration}`);
```
khelif96, line 29: TaskSortCategory.DURATION

src/pages/version/taskDuration/TaskDurationTable.tsx
```diff
@@ -0,0 +1,169 @@
...
+  });
+
+  const variantFilter = useFilterInputChangeHandler({
+    urlParam: PatchTasksQueryParams.Variant,
+    ...filterProps,
+  });
+
+  const handleDurationSort = (direction: string) => {
+    updateQueryParams({
+      duration: direction.toUpperCase(),
```
khelif96, line 58: TaskSortCategory.DURATION

khelif96 (approved):
Approved Pending the comments being addressed and our offline discussion.

PR #1212, opened 2022-04-11
EVG-16600: Collapse chart View options control

SupaJoon:
[Evg-16600](https://jira.mongodb.org/browse/Evg-16600)

### Description 
renders the commit chart type toggle within an accordion toggle

### Screenshots
Screen Shot 2022-04-11 at 5 44 16 PM


src/pages/commits/ActiveCommits/ChartToggle.tsx
```diff
@@ -17,40 +17,43 @@ export const ChartToggle: React.VFC<{
...
-              `}
-            >
-              Percentage
-            </Radio>
-          </StyledRadioGroup>
-        </ToggleWrapper>
-      )}
-    </ClassNames>
+    <AccordionContainer>
+      <Accordion title="View options">
```
khelif96, line 21: I think this should be expanded by default. So it's visible to the user.

minnakt: I kind of appreciate that it's closed by default, since the user probably won't change this very often

If it was always open it could be annoying when it blocks the graph bars (the component gets completely re-rendered after fetching data, so the open/close state isn't preserved)

https://user-images.githubusercontent.com/47064971/162991718-ff596504-d38f-46ea-b347-72b853b826c7.mov

khelif96: Could we save its state as a cookie? I think it makes sense to at least show it by default so new users are aware of it and they can dismiss it on their own. We can discuss it in the design sync as well.

src/components/Accordion/index.tsx
```diff
@@ -26,40 +27,40 @@ export const Accordion: React.VFC<AccordionProps> = ({
   titleTag,
   onToggle = () => {},
   useIndent = true,
+  isOpen,
```
khelif96, line 30: These changes weren't necessary `defaultOpen` would have worked.

src/pages/commits/ActiveCommits/ChartToggle.tsx
```diff
@@ -23,40 +25,47 @@ export const ChartToggle: React.VFC<{
...
-            </Radio>
-          </StyledRadioGroup>
-        </ToggleWrapper>
-      )}
-    </ClassNames>
+    <AccordionContainer>
+      <Accordion
+        title="View options"
+        isOpen={isOpen}
+        onToggle={onToggleAccordion}
```
khelif96, line 32:
```suggestion
      <Accordion
        title="View options"
        defaultOpen={isOpen}
        onToggle={onToggleAccordion}
```

src/pages/Commits.tsx
```diff
@@ -40,6 +43,15 @@ export const Commits = () => {
...
   const { search } = useLocation();
+  const hasSeenChartToggle = Boolean(Cookies.get(HAS_SEEN_COMMIT_CHART_TOGGLE));
+  const [isOpenChartToggle, setIsOpenChartToggle] = useState(
+    !hasSeenChartToggle
+  );
+  useEffect(() => {
+    if (!hasSeenChartToggle && !isOpenChartToggle) {
+      Cookies.set(HAS_SEEN_COMMIT_CHART_TOGGLE, "true");
+    }
+  }, [hasSeenChartToggle, isOpenChartToggle]);
```
khelif96, line 54:
```suggestion
  const shouldChartOpen = Boolean(Cookies.get(COMMIT_CHART_TOGGLE));
const onToggleChart = ({isVisible) => {
Cookies.set(COMMIT_CHART_TOGGLE, isVisible)
```

src/pages/Commits.tsx
```diff
@@ -170,6 +182,10 @@ export const Commits = () => {
           />
         </PaginationWrapper>
         <CommitsWrapper
+          isOpenChartToggle={isOpenChartToggle}
+          onToggleChartViewOptionsAccordion={({ isVisible }) =>
+            setIsOpenChartToggle(isVisible)
+          }
```
khelif96, line 188:
```suggestion
          isOpenChartToggle={shouldChartOpen}
          onToggleChartViewOptionsAccordion={onToggleChart}
```

src/pages/commits/CommitsChart.tsx
```diff
@@ -29,12 +29,20 @@ interface Props {
   versions?: Commits;
   hasTaskFilter?: boolean;
   hasError?: boolean;
+  isOpenChartToggle: boolean;
```
khelif96, line 32: nit. `isChartOpen`

src/pages/commits/CommitsChart.tsx
```diff
@@ -29,12 +29,20 @@ interface Props {
...
+  onToggleChartViewOptionsAccordion: ({
+    isVisible,
+  }: {
+    isVisible: boolean;
+  }) => void;
```
khelif96, line 37: `onToggleChart: (isVisible: boolean) => void`

src/constants/cookies.ts
```diff
@@ -5,3 +5,4 @@ export const getNotificationTriggerCookie = (type: string) =>
   `${type}-notification-trigger`;
 export const SLACK_NOTIFICATION_BANNER = "has-closed-slack-banner";
 export const CURRENT_PROJECT = "mci-project-cookie";
+export const HAS_SEEN_COMMIT_CHART_TOGGLE = "has-seen-commit-chart-toggle";
```
khelif96, line 8: `COMMIT_CHART_TOGGLE`

src/pages/Commits.tsx
khelif96: It makes more sense to store the last state the toggle was on.

src/pages/commits/CommitsWrapper.tsx
```diff
@@ -26,6 +26,10 @@ interface Props {
   isLoading: boolean;
   hasTaskFilter: boolean;
   hasFilters: boolean;
+  isOpenChartToggle: boolean;
```
khelif96, line 29: `isChartToggleOpen`

src/pages/Commits.tsx
khelif96, line 54: Since the commit chart toggle component is only used by the commit chart its state logic should in the same file. It doesn't need to be drilled through from the Commits page.

khelif96 (changes requested):

src/components/Accordion/index.tsx
SupaJoon: if we set defaultOpen, the chart will rerender every time new chart data is loaded. You can see this from Minna's video: https://github.com/evergreen-ci/spruce/pull/1212#discussion_r848544519

src/pages/Commits.tsx
SupaJoon: I remember in the design meeting we discussed that the chart should only be open to the user once and then closed after. I can run this by Lara again

SupaJoon: The state is controlled in the parent component because the CommitChart will rerender everytime new data is loaded which would reset the state.

src/pages/commits/CommitsChart.tsx
SupaJoon: I actually prefer to read `isOpenChart` as opposed to `isChartOpen`. Since it describes what the data represents and then what it belongs to.

src/pages/commits/CommitsChart.tsx
SupaJoon: Accordion actually accepts an interface for the callback 
https://github.com/evergreen-ci/spruce/pull/1212/files#diff-65dca1bb07024427b8595b463e4e0c3e469f2c0f106e6c928a30fd41fb96acfdL14

SupaJoon:
I'll use the cookie to control the chart toggle state. I'll omit a value for defaultOpen and instead supply isOpen to the Accordion so the component is completely controlled by the cookie value instead both the cookie and Accordion internal state.

src/pages/commits/ActiveCommits/ChartToggle.tsx
```diff
@@ -2,17 +2,24 @@ import { ClassNames } from "@emotion/react";
...
 import { useProjectHealthAnalytics } from "analytics/projectHealth/useProjectHealthAnalytics";
+import { Accordion } from "components/Accordion";
 import { size } from "constants/tokens";
 import { ChartTypes } from "types/commits";
 
 const { gray } = uiColors;
 
 export const ChartToggle: React.VFC<{
+  onToggleAccordion: (nextState: { isVisible: boolean }) => void;
+  defaultOpenAccordion: boolean;
```
khelif96, line 14: Do we need to pass these props in? i think the state logic for this component would make sense living in this file. Since it is only used for the chart.

khelif96 (approved):
Just  add one comment

PR #1215, opened 2022-04-12
EVG-15896: Prompt user for GitHub username on My Patches page

minnakt:
[EVG-15896](https://jira.mongodb.org/browse/EVG-15896)

### Description 
This PR adds a banner on the My Patches page if a user has not set their Github username. Note that the banner is not dismissible but is only shown on the My Patches page.

### Screenshots
(Is green the right color? 🤔)

Screen Shot 2022-04-12 at 2 23 18 PM

### Testing 
- Added test in `banners/github_username_banner.ts` (Cypress)


src/components/Banners/GithubUsernameBanner.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { useLocation } from "react-router-dom";
+import { paths } from "constants/routes";
+import { size } from "constants/tokens";
+import { GetUserSettingsQuery } from "gql/generated/types";
+import { GET_USER_SETTINGS } from "gql/queries";
+
+const { green } = uiColors;
+export const GithubUsernameBanner = () => {
+  const { pathname } = useLocation();
+  const isPatchesPage = pathname.startsWith(paths.user);
```
khelif96, line 14: Does this make sense to live in the `PatchesPage` component. That way it will always be rendered on a Patches Page
https://github.com/evergreen-ci/spruce/blob/f7a503ba9a4f553655dbfca46fbba6cf5e500cfc/src/components/PatchesPage/index.tsx

khelif96: Otherwise react-router exports a matchPath function. https://github.com/evergreen-ci/spruce/blob/e036af8357365d7be7cc6513b4a6e49af53b626b/src/hooks/useLegacyUIURL.ts

src/components/Banners/GithubUsernameBanner.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+  // USER SETTINGS QUERY
+  const { data: userSettingsData } = useQuery<GetUserSettingsQuery>(
+    GET_USER_SETTINGS
```
khelif96, line 18: We should avoid doing a query if we don't need to and are not on the correct page.

src/components/Banners/GithubUsernameBanner.tsx
```diff
@@ -0,0 +1,47 @@
...
+  );
+};
+
+const Banner = styled.div`
+  transition: max-height 0.3s ease-in-out;
+  align-items: center;
+  background-color: ${green.light2};
+  display: flex;
+  padding: ${size.xxs} ${size.s};
+`;
```
khelif96, line 43: I wonder if we can share some styles from some of the other banners?

src/components/Banners/GithubUsernameBanner.tsx
```diff
@@ -0,0 +1,47 @@
...
+        this to map GitHub pull requests to your Evergreen user account.
+      </Banner>
+    )
+  );
+};
+
+const Banner = styled.div`
+  transition: max-height 0.3s ease-in-out;
+  align-items: center;
+  background-color: ${green.light2};
```
khelif96, line 40: Do you think yellow would look better here? Since we want the user to take action.

src/components/Banners/GithubUsernameBanner.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const { githubUser } = userSettings || {};
+  const { lastKnownAs } = githubUser || {};
+  const hasNoGithubUser = lastKnownAs === "";
+
+  return (
+    isPatchesPage &&
+    hasNoGithubUser && (
+      <Banner data-cy="github-username-banner">
+        <IconWithMargin glyph="InfoWithCircle" />
+        Please set your GitHub username on the settings page. Evergreen uses
```
khelif96, line 30: can we linkify settings page.

khelif96 (changes requested):

khelif96 (approved):

PR #1219, opened 2022-04-14
EVG-16716 Compute eventLog diff

malikchaya2:
[EVG-16716](https://jira.mongodb.org/browse/EVG-16716)

### Description 
Add functions needed to calculate and reduce event log diffs so that they can later be put into columns on the event log page. 

### Testing 
Testing on staging

### DIFF
--- a/src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.test.ts
+++ b/src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.test.ts
@@ -1,0 +1,98 @@
+import { getEventDiffLines } from "./EventLogDiffs";
+
+const beforeAddition = {
+  __typename: "ProjectEventSettings",
+  projectRef: {
+    __typename: "Project",
+    identifier: "viewTest",
+    patchTriggerAliases: [],
+  },
+  vars: {
+    __typename: "ProjectVars",
+    vars: { newVariable: "so new" },
+  },
+};
+
+const afterAddition = {
+  __typename: "ProjectEventSettings",
+  projectRef: {
+    __typename: "Project",
+    identifier: "viewTest",
+    patchTriggerAliases: [
+      {
+        __typename: "PatchTriggerAlias",
+        alias: "newAlias",
+        childProjectIdentifier: "evg",
+      },
+    ],
+  },
+  vars: {
+    __typename: "ProjectVars",
+    vars: { newVariable: "so new" },
+  },
+};
+
+const beforeUpdate = {
+  __typename: "ProjectEventSettings",
+  projectRef: {
+    __typename: "Project",
+    identifier: "viewTest",
+    patchTriggerAliases: [
+      {
+        __typename: "PatchTriggerAlias",
+        alias: "newAlias",
+        childProjectIdentifier: "evg",
+      },
+    ],
+  },
+  vars: {
+    __typename: "ProjectVars",
+    vars: { newVariable: "so new" },
+  },
+};
+const afterUpdate = {
+  __typename: "ProjectEventSettings",
+  projectRef: {
+    __typename: "Project",
+    identifier: "viewTest",
+    patchTriggerAliases: [
+      {
+        __typename: "PatchTriggerAlias",
+        alias: "noLongerNewAlias",
+        childProjectIdentifier: "evg",
+      },
+    ],
+  },
+  vars: {
+    __typename: "ProjectVars",
+    vars: { newVariable: "so new" },
+  },
+};
+
+describe("should transform event diffs to key, before and after", () => {
+  it("should transform updates", () => {
+    const diffLines = getEventDiffLines(beforeUpdate, afterUpdate);
+    expect(diffLines).toStrictEqual([
+      {
+        key: "projectRef.patchTriggerAliases[0].alias",
+        before: "newAlias",
+        after: "noLongerNewAlias",
+      },
+    ]);
+  });
+  it("should transform additions", () => {
+    const diffLines = getEventDiffLines(beforeAddition, afterAddition);
+    expect(diffLines).toStrictEqual([
+      {
+        key: "projectRef.patchTriggerAliases[0].alias",
+        before: undefined,
+        after: "newAlias",
+      },
+      {
+        key: "projectRef.patchTriggerAliases[0].childProjectIdentifier",
+        before: undefined,
+        after: "evg",
+      },
+    ]);
+  });
+});
--- a/src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.ts
+++ b/src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.ts
@@ -1,0 +1,64 @@
+import { diff } from "deep-object-diff";
+import { ProjectEventSettings } from "gql/generated/types";
+import { Subset } from "types/utils";
+import { string } from "utils";
+
+const { omitTypename } = string;
+
+const isObject = (val) => val && typeof val === "object" && !Array.isArray(val);
+
+const addDelimiter = (a: string, b: string) => (a ? `${a}.${b}` : b);
+
+const getDiffProperties = (eventObj: object) => {
+  const paths = (obj = {}, head = "") =>
+    Object.entries(obj).reduce((event, [key, value]) => {
+      const fullPath = addDelimiter(head, key);
+
+      return isObject(value)
+        ? event.concat(paths(value, fullPath))
+        : event.concat(fullPath);
+    }, []);
+
+  return paths(eventObj);
+};
+
+const formatArrayElements = (eventKey: string) =>
+  eventKey.replace(/.[0-9]./g, (x) => `[${x[1]}].`);
+
+const getNestedObject = (nestedObj: object, pathArr: string[]) =>
+  pathArr.reduce(
+    (obj, key) => (obj && obj[key] !== "undefined" ? obj[key] : undefined),
+    nestedObj
+  );
+
+type EventDiffLine = {
+  key: string;
+  before: string;
+  after: string;
+};
+
+export const getEventDiffLines = (
+  before: Subset<ProjectEventSettings>,
+  after: Subset<ProjectEventSettings>
+): EventDiffLine[] => {
+  const eventDiff = omitTypename(diff(before, after));
+  const pathKeys = getDiffProperties(eventDiff);
+
+  const eventDiffLines = pathKeys.map((key) => {
+    const pathsArray = key.split(".");
+    const previousValue = getNestedObject(before, pathsArray);
+    const changedValue = getNestedObject(eventDiff, pathsArray);
+
+    const formattedKey = formatArrayElements(key);
+
+    const line = {
+      key: formattedKey,
+      before: previousValue,
+      after: changedValue,
+    };
+
+    return line;
+  });
+
+  return eventDiffLines.filter((el) => el !== null);
+};
--- a/src/types/utils.ts
+++ b/src/types/utils.ts
@@ -39,3 +39,8 @@ export type InvertedObject<T extends Record<PropertyKey, PropertyKey>> = {
 export type OneOf<T1, T2> =
   | (T1 & Partial<Record<Exclude<keyof T2, keyof T1>, never>>)
   | (T2 & Partial<Record<Exclude<keyof T1, keyof T2>, never>>);
+
+/** Helper to require only a partial set of fields on a nested object */
+export type Subset<K> = {
+  [S in keyof K]?: K[S] extends object ? Subset<K[S]> : K[S];
+};


src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.tsx
```diff
@@ -0,0 +1,52 @@
+import { diff } from "deep-object-diff";
```
khelif96, line 1: There is no `jsx` in this file so the file type should be `.ts`.

src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+const getDiffProperties = (eventObj) => {
```
khelif96: Can you type this function.

src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.tsx
```diff
@@ -0,0 +1,55 @@
...
+    nestedObj
+  );
+
+type EventDiffLine = {
+  key: string;
+  before: any;
+  after: any;
+};
+
+export const getEventDiffLines = (
```
khelif96, line 33: This file needs some tests

khelif96 (changes requested):
Can you explain why this code is needed? `deep-object-diff` recursively navigates objects and returns the diffs.

src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.tsx
malikchaya2: done

src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.tsx
```diff
@@ -0,0 +1,55 @@
...
+  before: object,
+  after: object
+): EventDiffLine[] => {
+  const eventDiff = diff(before, after);
+  const pathKeys = getDiffProperties(diff(before, after));
```
sophstad: No need to call `diff` again
```suggestion
  const pathKeys = getDiffProperties(eventDiff);
```

src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.tsx
malikchaya2: done

src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.ts
```diff
@@ -0,0 +1,55 @@
+import { diff } from "deep-object-diff";
+
+const getDiffProperties = (eventObj: object) => {
+  const isObject = (val) =>
+    val && typeof val === "object" && !Array.isArray(val);
```
sophstad, line 5: This function should exist outside of `getDiffProperties()`

src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  const addDelimiter = (a, b) => (a ? `${a}.${b}` : b);
```
sophstad, line 7: This function can be moved outside of `getDiffProperties()`, too.

src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.ts
```diff
@@ -0,0 +1,55 @@
...
+  before: any;
+  after: any;
+};
+
+export const getEventDiffLines = (
+  before: object,
+  after: object
+): EventDiffLine[] => {
+  const eventDiff = diff(before, after);
+  const pathKeys = getDiffProperties(diff(before, after));
```
sophstad, line 38: No need to call `diff()` again
```suggestion
  const pathKeys = getDiffProperties(eventDiff);
```

src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.ts
```diff
@@ -0,0 +1,55 @@
...
+  const eventDiffLines: EventDiffLine[] = [];
+
+  pathKeys.forEach((key) => {
```
sophstad, line 41: Since the goal of this loop is to generate a 1:1 array with pathKeys, you can use `.map()` instead

src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.ts
```diff
@@ -0,0 +1,55 @@
...
+    const pathsArray = key.split(".");
+    const previousValue = getNestedObject(before, pathsArray);
+    const changedValue = getNestedObject(eventDiff, pathsArray);
+    const line = {
+      key,
```
sophstad, line 46: IMO it would be more clear (and correct) if array access in the final output was indicated with brackets, i.e. `projectRef.patchTriggerAliases[0].alias` instead of `projectRef.patchTriggerAliases.0.alias`.

sophstad (changes requested):
Agree that some tests would be good, particularly around ensuring that adding/removing objects in an array is properly handled

src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.ts
malikchaya2: oops, I refactored and forgot to update. thanks!

src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.tsx
malikchaya2: done

src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.ts
malikchaya2: done

src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.ts
malikchaya2: done

src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.test.ts
```diff
@@ -0,0 +1,203 @@
+import { getEventDiffLines } from "./EventLogDiffs";
+
+const before = {
```
sophstad, line 3: Can we trim these objects down to the relevant fields for the test? That will help clarify what we're testing I think

src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.test.ts
```diff
@@ -0,0 +1,203 @@
...
+  aliases: [],
+  githubWebhooksEnabled: false,
+};
+
+describe("event diff lines", () => {
+  it("should transform event diffs to key, before and after", () => {
+    const diffLines = getEventDiffLines(before, after);
+    expect(diffLines).toStrictEqual([
+      {
+        key: "projectRef.patchTriggerAliases[0]alias",
```
sophstad, line 197: There should still be a period after the `]` bracket since that's how you would access a property of an object in an array

src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.test.ts
```diff
@@ -0,0 +1,203 @@
...
+        before: "newAlias",
+        after: "noLongerNewAlias",
+      },
+    ]);
+  });
```
sophstad, line 202: Cases where an entire element is being added and removed from an array would be useful to test, I think

sophstad (changes requested):

src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.test.ts
malikchaya2: done

src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.test.ts
malikchaya2: done

src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.test.ts
malikchaya2: done

src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.test.ts
```diff
@@ -0,0 +1,98 @@
...
+      {
+        key: "projectRef.patchTriggerAliases[0].alias",
+        before: undefined,
+        after: "newAlias",
+      },
+      {
+        key: "projectRef.patchTriggerAliases[0].childProjectIdentifier",
+        before: undefined,
+        after: "evg",
+      },
```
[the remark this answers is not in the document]
malikchaya2: I was contemplating this as well, and I'm not sure that I agree. 
We see both of these examples on the legacy page. 

Here we have it with each on it's own line
image

here we have it all on one line: 
image

In my opinion, listing it all separately makes it easier to visually discern in one quick glance. 

Each event will be separated in its own card. Therefore, it will already be one distinct unit representing the change.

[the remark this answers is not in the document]
ablack12: I agree -- you only really go into the event log to see how something specific has changed, so having it broken up for that is helpful.

src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.ts
```diff
@@ -0,0 +1,62 @@
...
+  after: object
+): EventDiffLine[] => {
+  const eventDiff = diff(before, after);
+  const pathKeys = getDiffProperties(eventDiff);
+
+  const eventDiffLines = pathKeys.map((key) => {
+    const pathsArray = key.split(".");
+    if (pathsArray[pathsArray.length - 1] === "__typename") {
+      return null;
+    }
```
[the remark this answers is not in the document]
malikchaya2: I tried that, it's harder to do it that way because of the nesting.

src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.test.ts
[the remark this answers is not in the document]
sophstad: True, but wouldn't the latter expand to look more like this than what that first chart looks like?
| Property | Before | After |
|----------|--------|-------|
| proj_ref.owner_name |  | a|
| ... | | |
| pop_ref.patch_trigger_aliases[0].alias |  | newAlias |
| pop_ref.patch_trigger_aliases[0].child_project_id |  | mci |
| pop_ref.patch_trigger_aliases[0].child_project_identifier |  | evg |
| pop_ref.patch_trigger_aliases[0].task_specifiers[0].patch_alias |  | __github |
| pop_ref.patch_trigger_aliases[0].task_specifiers[0].task_regex |  | "" |
| pop_ref.patch_trigger_aliases[0].task_specifiers[0].variant_regex |  | "" |
| pop_ref.patch_trigger_aliases[0].status |  | "" |
| pop_ref.patch_trigger_aliases[0].parent_as_module |  | "" |

To me that comes across as less clear that an entire new patch trigger alias was added. But if you prefer it then that's okay with me!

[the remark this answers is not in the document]
malikchaya2: Why don't we first set it up like this, and then maybe iterate on it later if we get user feedback?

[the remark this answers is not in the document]
malikchaya2: I tend to like this because for example, it being broken up into lines would make it easier to spot what is wrong with the alias -- for example, the module wasn't added.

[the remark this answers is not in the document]
malikchaya2: As for the chart you posted, I think I'll filter out any empty strings when I create the columns to make it clearer.

src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.ts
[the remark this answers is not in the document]
sophstad: Interesting, what were you seeing? The tests indicate that deep nesting and arrays should be handled https://github.com/evergreen-ci/spruce/blob/3e009e4af1a0ffd4477a9b8fdcf41a44947305c0/src/utils/string/string.test.ts#L160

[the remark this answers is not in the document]
malikchaya2: I was able to make it work. done.

src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.ts
```diff
@@ -0,0 +1,62 @@
...
+    const line = {
+      key: formattedKey,
+      before: previousValue,
+      after: changedValue,
+    };
+
+    return line;
+  });
+
+  return eventDiffLines.filter((el) => el != null);
```
sophstad, line 61:
```suggestion
  return eventDiffLines.filter((el) => el !== null);
```

sophstad (approved):
💯

src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.ts
```diff
@@ -0,0 +1,62 @@
...
+
+type EventDiffLine = {
+  key: string;
+  before: any;
+  after: any;
+};
+
+export const getEventDiffLines = (
+  before: object,
+  after: object
```
khelif96, line 40: We can more strongly type these values. So they accurately describe the objects within.

malikchaya2: done

src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.ts
```diff
@@ -0,0 +1,64 @@
...
+type EventDiffLine = {
+  key: string;
+  before: string;
+  after: string;
+};
+
+export const getEventDiffLines = (
+  before: Subset<ProjectEventSettings>,
+  after: Subset<ProjectEventSettings>
+): EventDiffLine[] => {
```
khelif96, line 43:
```ts
export const getEventDiffLines<T extends object> = (
  before: Subset<T>,
  after: Subset<T>
): EventDiffLine[] => {
```

khelif96 (approved):

PR #1221, opened 2022-04-15
EVG-16740 Control LG tooltip enabled behavior on waterfall icons

khelif96:
[EVG-16740](https://jira.mongodb.org/browse/EVG-16740)

### Description 
Currently LG tooltips try to render as soon as you hover over them. This is problematic if a user mouses over several icons because they will all try to render and then stop rendering which causes the page to jitter when rendering all of the tooltips that are never used.  

### Screenshots

https://user-images.githubusercontent.com/4605522/163607717-312bc841-1b13-4359-8486-baf61abfb5ca.mov

### DIFF
--- a/src/pages/commits/ActiveCommits/buildVariantCard/WaterfallTaskStatusIcon.tsx
+++ b/src/pages/commits/ActiveCommits/buildVariantCard/WaterfallTaskStatusIcon.tsx
@@ -43,7 +43,7 @@ export const WaterfallTaskStatusIcon: React.VFC<WaterfallTaskStatusIconProps> = 
   const failedTestDifference = filteredTestCount - testResults?.length;
 
   let timeout;
-  const onHover = () => {
+  const onMouseEnter = () => {
     timeout = setTimeout(() => {
       setEnabled(true);
       // Only query failing test names if the task has failed.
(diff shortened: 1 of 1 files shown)


src/pages/commits/ActiveCommits/buildVariantCard/WaterfallTaskStatusIcon.tsx
```diff
@@ -33,29 +34,45 @@ export const WaterfallTaskStatusIcon: React.VFC<WaterfallTaskStatusIconProps> =
...
+    if (timeout) {
+      clearTimeout(timeout);
     }
   };
-
+  useEffect(() => () => {
+    if (timeout) {
+      clearTimeout(timeout);
+    }
+  });
```
minnakt, line 65: I think this is running on every hover (at least that's what I'm seeing when I add a print statement here). I think we just want this to run when the component unmounts?

khelif96: Nice catch. I shouldn't have listened to eslint and removed the dep array.

minnakt (approved):
looks good to me :)

PR #1223, opened 2022-04-18
EVG-16594: Add items to front of array

sophstad:
EVG-16594

### Description 
- Add items to front of arrays since this better matches our designs (which have the "Add" button before all array elements)
  - Notable exception: "commands" array on the virtual workstations page. Since commands run in sequence from top to bottom, the "Add" button appears at the end of the array and items are appended to the end.
- Bonus fixes
  - Correctly save whether a project should override a repo's files ignored from cache
  - Fix layout of "Move to New Repo" and "Attach/Detach from Repo" buttons on General settings page

### Screenshots

https://user-images.githubusercontent.com/9298431/163862317-b90b0b91-4503-43aa-8ca2-eaaac5602b30.mov

https://user-images.githubusercontent.com/9298431/163862179-4bd8ef38-de70-4d93-97aa-3c8a52deff9c.mov

### Testing 
- Update unit and e2e tests


src/components/SpruceForm/FieldTemplates.tsx
```diff
@@ -257,19 +277,7 @@ export const ArrayFieldTemplate: React.VFC<ArrayFieldTemplateProps> = ({
...
-            disabled={isDisabled}
-            leftGlyph={<Icon glyph="Plus" />}
-            onClick={onAddClick}
-            size={addButtonSize}
-          >
-            {addButtonText}
-          </Button>
-        </ElementWrapper>
-      )}
+      {!readonly && canAdd && !addToEnd && addButton}
```
khelif96, line 280: Is there  a way to make this conditional more clear on what its doing? e.g. turn addButton into a component and use JSX syntax so its more readable.

khelif96: why doesn't this need an AddButtonContainer?

src/components/SpruceForm/FieldTemplates.tsx
```diff
@@ -249,6 +247,28 @@ export const ArrayFieldTemplate: React.VFC<ArrayFieldTemplateProps> = ({
...
   const isDisabled = disabled || readonly;
+
+  const addButtonSize = uiSchema["ui:addButtonSize"] || "small";
+  const addButtonText = uiSchema["ui:addButtonText"] || "Add";
+  // Override RJSF's default array behavior; add new elements to beginning of array unless otherwise specified.
+  const addToEnd = uiSchema["ui:addToEnd"] ?? false;
+  const handleAddClick =
+    items.length && !addToEnd ? items[0].onAddIndexClick(0) : onAddClick;
+
+  const addButton = (
```
khelif96, line 258: This should live as its own component  outside of this and ingest props otherwise it is redeclared on each render.

src/components/SpruceForm/FieldTemplates.tsx
```diff
@@ -291,11 +299,18 @@ export const ArrayFieldTemplate: React.VFC<ArrayFieldTemplateProps> = ({
             />
           ))
         )}
+        {!readonly && canAdd && addToEnd && (
```
khelif96, line 302: We can extract these repeated conditionals into a common variable to make this more readable.

```ts
L257: const buttonAtEnd = !readOnly && canAdd && addToEnd;
...
L280: {!buttonAtEnd && (<AddButt.... AddButtonContainer>)
...
L302: {buttonAtEnd && (<AddButt.... AddButtonContainer>)
```

khelif96 (changes requested):

src/components/SpruceForm/FieldTemplates.tsx
sophstad: Since the goal here is to have two (almost) identical buttons that render at different positions in the form, making a separate component and including it twice resulted in a lot of repetition. I thought a `useMemo` made more sense here so that any future changes to the button would be reflected in both instances.

ablack12 (commented):
Screen recordings LGTM 👍

khelif96 (approved):

PR #1227, opened 2022-04-19
EVG-15998 Refactor query param logic in filter badges and prevent clearing unrelated query params.

khelif96:
[EVG-15598](https://jira.mongodb.org/browse/EVG-15598)

### Description 
Filter badges now only consume  and render badges as props they no longer update query params I extracted that logic into a hook. Clearing all will not delete unrelated query params.
### Screenshots
No visible changes.

### DIFF
--- a/src/components/FilterBadges/FilterBadges.stories.tsx
+++ b/src/components/FilterBadges/FilterBadges.stories.tsx
@@ -1,49 +1,76 @@
-import { withKnobs, text, button } from "@storybook/addon-knobs";
-import { withQuery } from "@storybook/addon-queryparams";
-import { useLocation, BrowserRouter } from "react-router-dom";
-import { useUpdateURLQueryParams } from "hooks/useUpdateURLQueryParams";
-import { ProjectFilterOptions } from "types/commits";
-import { queryString, url } from "utils";
-import { FilterBadges } from ".";
-
-const { upsertQueryParam } = url;
-const { parseQueryString } = queryString;
+import { useState } from "react";
+import Button from "@leafygreen-ui/button";
+import TextInput from "@leafygreen-ui/text-input";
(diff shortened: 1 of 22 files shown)


src/components/FilterBadges/useFilterBadgeQueryParams.test.tsx
```diff
@@ -0,0 +1,133 @@
...
+    const { queryAllByDataCy } = render(Content, {
+      route: `/commits/evergreen?buildVariants=variant1`,
+      path: "/commits/:projectId",
+    });
+    expect(queryAllByDataCy("filter-badge")).toHaveLength(1);
+  });
+
+  it("should render multiple filter badges with the same key but different values", () => {
+    const { queryAllByDataCy } = render(Content, {
+      route: `/commits/evergreen?buildVariants=variant1,variant2`,
```
minnakt, line 41: nit. can we replace all of the uses of `` in this file with normal quotes?

src/components/FilterBadges/FilterBadges.stories.tsx
```diff
@@ -1,49 +1,76 @@
...
+        onChange={(e) => setBadgeKey(e.target.value)}
+      />
+      <TextInput
+        placeholder="value"
+        label="value"
+        value={badgeValue}
+        onChange={(e) => setBadgeValue(e.target.value)}
+        onSubmit={handleAdd}
+      />
+      <Button onClick={() => onAdd(badgeKey, badgeValue)}>Add</Button>
```
minnakt, line 73: This seems a little strange to me

Maybe `L71` should be deleted, and `L73` should say `onClick={() => handleAdd()}`?

minnakt (approved):
looks good to me :)

PR #1179, opened 2022-03-28
EVG-16545: Highlight matching task icons on Project Health page

minnakt:
[EVG-16545](https://jira.mongodb.org/browse/EVG-16545)

### Description 
This PR makes is so that hovering over task icons will highlight matching task icons on the Project Health page.

### Screenshots

https://user-images.githubusercontent.com/47064971/160472916-d4dc1f9d-f609-41be-a477-7579daa14f53.mov

### Testing 
- Added `CommitsContext.test.tsx`
- Modified `storybook` files and `test` files to pass with Context changes (storybook also reflects the highlighting icons  change)
- Wrote a very crude implementation of the build variants ordering (not included in this PR) but seemed to work with the Context stuff so I think it should be okay... 😶‍🌫️

PR #1229, opened 2022-04-20
EVG-16661: Inject global style on task icon hover

minnakt:
[EVG-16661](https://jira.mongodb.org/browse/EVG-16661)

### Description 
This PR removes the function `hoverTaskIcons` (which defines task icon hover behavior through plain JavaScript) in favor of the functions `injectGlobalStyle` and `removeGlobalStyle` (which define task icon hover behavior by injecting and removing styling from the document `head`).

### Screenshots
No visible changes (hovering should work the same way).

### Testing 
- Added test in [ActiveCommits/utils.test.ts](https://github.com/evergreen-ci/spruce/compare/main...minnakt:EVG-16661?expand=1#diff-08fc51da8c2c8361328053ce1c9b5a469774f69ab15747e6c2e48385ef99deda)


src/pages/commits/ActiveCommits/utils.ts
```diff
@@ -154,26 +154,22 @@ export const getStatusesWithZeroCount = (colors: ColorCount[]) => {
...
+const taskIconStyle = "task-icon-style";
+
+export const removeGlobalStyle = () => {
+  document.getElementById(taskIconStyle)?.remove();
+};
+
+export const injectGlobalStyle = (taskIdentifier: string) => {
+  removeGlobalStyle();
+
+  const hoverStyle = document.createElement("style");
```
sophstad, line 167: I wonder if we could incorporate [Emotion's Global styles](https://emotion.sh/docs/globals) instead?

khelif96 (changes requested):

src/pages/commits/ActiveCommits/buildVariantCard/WaterfallTaskStatusIcon.tsx
```diff
@@ -64,55 +65,65 @@ export const WaterfallTaskStatusIcon: React.VFC<WaterfallTaskStatusIconProps> =
...
-            ))}
-            {failedTestDifference > 0 && (
-              <div>and {failedTestDifference} more</div>
-            )}
-          </>
-        )}
-      </div>
-    </Tooltip>
+    <>
+      {enabled && (
```
khelif96, line 72: enabled is toggled by a timeout so any change will be delayed. 

I tried to replace this with a non timed out state value and that was also slower then what we had before. Reacts Virtual dom is working against us in this case since there are so many elements on the screen. The approach we had before technically by passed it the virtual dom by injecting styles directly into the dom.

src/pages/commits/ActiveCommits/utils.ts
```diff
@@ -154,26 +154,22 @@ export const getStatusesWithZeroCount = (colors: ColorCount[]) => {
...
+  hoverStyle.id = taskIconStyle;
+  hoverStyle.innerHTML = `
+    div[data-task-icon]:not([data-task-icon="${taskIdentifier}"]) {
```
khelif96, line 170: What do you think about this? Normally css selector performance is really good but it might make a difference on a page like this with thousands of dom nodes. https://csswizardry.com/2011/09/writing-efficient-css-selectors/

minnakt: Initially I was using the `classNames` instead of an attribute, but it gets attached to some other fluff due to the `styled` library, like this:

Screen Shot 2022-04-25 at 1 43 50 PM

Arjun also weighed in on it on this comment [here](https://github.com/evergreen-ci/spruce/pull/1179#discussion_r838557151).

---

I could add an `ID` to the surrounding container  and do something like `#commits-build-variants-container div[data-task-icon]:not([data-task-icon="${taskIdentifier}"])` but I'm not sure how much it would help considering that the attribute portion is supposed to be the slow part.

But also, since this article was written 11 years ago when browsers weren't as optimized as they were today, I wonder if it's something even worth worrying about?

src/pages/commits/ActiveCommits/buildVariantCard/WaterfallTaskStatusIcon.test.tsx
```diff
@@ -53,6 +56,29 @@ describe("waterfallTaskStatusIcon", () => {
...
+    userEvent.hover(queryByDataCy("waterfall-task-status-icon"));
+    await waitFor(() => {
+      expect(injectGlobalStyle).toHaveBeenCalledTimes(1);
+      expect(injectGlobalStyle).toHaveBeenCalledWith(props.identifier);
+    });
+
+    userEvent.unhover(queryByDataCy("waterfall-task-status-icon"));
+    await waitFor(() => {
+      expect(removeGlobalStyle).toHaveBeenCalledTimes(1);
+    });
```
khelif96, line 80: Should we assert that they had some styles applied?

khelif96 (approved):

src/pages/commits/ActiveCommits/buildVariantCard/WaterfallTaskStatusIcon.test.tsx
minnakt: I'm going to forgo this because as far as I can tell the styles on the element itself doesn't change (maybe some magic happening?). I could check that the document head is updated, but since I do that in the other test file anyways, it'd probably be unnecessary to do it twice

PR #1232, opened 2022-04-21
EVG-15089  Create the Event Log Page

malikchaya2:
[EVG-15089](https://jira.mongodb.org/browse/EVG-15089)

### Description 
This adds the event log page

### Screenshots
eventlog

### Testing 
[ Tested on staging](http://localhost:3000/project/viewTest/settings/event-log)


src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
```diff
@@ -38,11 +48,108 @@ export const EventLogTab: React.FC = () => {
...
+    </div>
   );
 };
+
+interface Props {
+  timestamp: Date;
+  user: string;
+}
+
+export const EventLogHeader: React.VFC<Props> = ({ user, timestamp }) => (
```
minnakt, line 111: nit. does this need to be exported? :0

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
```diff
@@ -38,11 +48,108 @@ export const EventLogTab: React.FC = () => {
...
+  <StyledHeader>
+    <div>
+      <H3>{getDateCopy(timestamp)}</H3>
+    </div>
```
minnakt, line 115: nit. this wrapping `div` might be unnecessary

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
```diff
@@ -38,11 +48,108 @@ export const EventLogTab: React.FC = () => {
...
+  padding: ${size.m};
+  width: 120%;
+`;
+
+const StyledRow = styled(Row)`
+  td:first-of-type {
+    width: 80%;
+    word-break: break-all;
+  }
+`;
```
minnakt, line 132: I ran into this issue too where I couldn't figure out how to directly style the `Cell`! Mohamed left a [comment](https://github.com/evergreen-ci/spruce/pull/1211#discussion_r855542333) on this other PR of mine, maybe it's worth trying this?

minnakt (approved):
Since the values in the table look kind of code-y, I was wondering if it might be worth showing the values in monospace font? Here's what it would look like: 

Screen Shot 2022-04-22 at 1 04 29 PM

Otherwise, it looks good to me :)

yarn.lock
```diff
@@ -7775,6 +7775,11 @@ dedent@^0.7.0:
   resolved "https://registry.yarnpkg.com/dedent/-/dedent-0.7.0.tgz#2495ddbaf6eb874abb0e1be9df22d2e5a544326c"
   integrity sha1-JJXduvbrh0q7Dhvp3yLS5aVEMmw=
 
+deep-diff@^1.0.2:
```
SupaJoon, line 7778: can we remove changes to this file since package.json didn't change

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
```diff
@@ -38,11 +48,108 @@ export const EventLogTab: React.FC = () => {
...
+                    sortBy={(datum: EventDiffLine) => datum?.before}
+                  />,
+                  <TableHeader
+                    key="after"
+                    label={<span data-cy="after">After</span>}
+                    sortBy={(datum: EventDiffLine) => datum?.after}
+                  />,
+                ]}
+              >
+                {({ datum }) => (
```
SupaJoon, line 90: could we add a type for datum? my vsc ts compiler is having a hard time understanding the type

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
```diff
@@ -38,11 +48,108 @@ export const EventLogTab: React.FC = () => {
...
+const StyledHeader = styled("div")`
+  padding-bottom: ${size.l};
+  padding-left: ${size.xxs};
+`;
+
+const getEventValue = (value: any) => {
+  if (value === true) {
+    return "true";
+  }
+  if (value === false) {
```
SupaJoon, line 143: before and after from EventLineDiff are both strings. when is value a boolean?

SupaJoon (changes requested):

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
khelif96: datum should infer the type passed into the data prop for the table so we should type that better instead.

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
```diff
@@ -38,11 +48,108 @@ export const EventLogTab: React.FC = () => {
...
+
   return (
-    <>
-      <div>This is the event log page for {identifier}</div>
-      <div> {JSON.stringify(projectEventData)}</div>
-      <div> {JSON.stringify(repoEventData)}</div>
-    </>
+    <div>
+      <div data-cy="event-log">
+        {((eventData as logEntry[]) || []).map(
```
khelif96, line 65: We shouldn't need this type aliasing.

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
```diff
@@ -38,11 +48,108 @@ export const EventLogTab: React.FC = () => {
...
+          ({ user, timestamp, before, after }) => (
+            /* @ts-expect-error */
+            <EventLogCard key={`event_log_${timestamp}`}>
+              <EventLogHeader user={user} timestamp={timestamp} />
+              <Table
+                data={getEventDiffLines(before, after)}
+                columns={[
+                  <TableHeader
+                    key="key"
+                    label={<span data-cy="key">Property</span>}
```
khelif96, line 75: Are these spans necessary? Does the default table styling work?

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
```diff
@@ -38,11 +48,108 @@ export const EventLogTab: React.FC = () => {
...
+                  />,
+                  <TableHeader
+                    key="after"
+                    label={<span data-cy="after">After</span>}
+                    sortBy={(datum: EventDiffLine) => datum?.after}
+                  />,
+                ]}
+              >
+                {({ datum }) => (
+                  <StyledRow key={datum?.key} data-cy="event-log-table-row">
```
khelif96, line 91: datum should always be defined so the optional chaining is not necessary.

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
```diff
@@ -38,11 +48,108 @@ export const EventLogTab: React.FC = () => {
...
+                    <Cell>{datum?.key ? datum?.key : ""}</Cell>
+                    <Cell>{getEventValue(datum?.before)}</Cell>
+                    <Cell>{getEventValue(datum?.after)}</Cell>
```
khelif96, line 94: Same comments as above.

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
```diff
@@ -38,11 +48,108 @@ export const EventLogTab: React.FC = () => {
...
+    ? repoEventData?.repoEvents?.eventLogEntries
+    : projectEventData?.projectEvents?.eventLogEntries;
+
   return (
-    <>
-      <div>This is the event log page for {identifier}</div>
-      <div> {JSON.stringify(projectEventData)}</div>
-      <div> {JSON.stringify(repoEventData)}</div>
-    </>
+    <div>
```
khelif96, line 63: Nit extra wrapping divs.

khelif96 (changes requested):

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
malikchaya2: we need it for the data-cy

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
malikchaya2: sometimes they are booleans not strings

yarn.lock
malikchaya2: Thanks!

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
malikchaya2: it's needed in order to be able to call .map on it

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
malikchaya2: It's of type EventDiffLine and my compiler picks it up. I'm not really sure where to define it better, I'm not able to add a type to that line. 

image

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
```diff
@@ -38,11 +47,121 @@ export const EventLogTab: React.FC = () => {
...
+                />,
+                <TableHeader
+                  key="before"
+                  label={<span data-cy="before">Before</span>}
+                  sortBy={(datum: EventDiffLine) => datum.before}
+                />,
+                <TableHeader
+                  key="after"
+                  label={<span data-cy="after">After</span>}
+                  sortBy={(datum: EventDiffLine) => datum.after}
```
sophstad, line 84: Do we really want to allow any sorting in this table? _Perhaps_ sorting alphabetically by key could be useful, but I don't think before and after need sortBy props.

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
```diff
@@ -38,11 +47,121 @@ export const EventLogTab: React.FC = () => {
...
+                />,
+              ]}
+            >
+              {({ datum }) => (
+                <Row key={datum.key} data-cy="event-log-table-row">
+                  <Cell>
+                    <StyledCell>{datum.key ?? ""}</StyledCell>
```
sophstad, line 91: Can we define `StyledCell` as `styled(Cell)` instead of using more nesting?

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
```diff
@@ -38,11 +47,121 @@ export const EventLogTab: React.FC = () => {
...
+                  sortBy={(datum: EventDiffLine) => datum.key}
+                />,
+                <TableHeader
+                  key="before"
+                  label={<span data-cy="before">Before</span>}
+                  sortBy={(datum: EventDiffLine) => datum.before}
+                />,
+                <TableHeader
+                  key="after"
+                  label={<span data-cy="after">After</span>}
```
sophstad, line 83: [nit] You can apply these `data-cy` props to the TableHeader components instead of adding spans

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
```diff
@@ -38,11 +47,121 @@ export const EventLogTab: React.FC = () => {
...
+  width: 150%;
+`;
+
+const StyledCell = styled("div")`
+  word-break: break-all;
+  font-size: 12px;
+  font-family: Menlo, Monaco, Consolas, "Courier New", monospace;
+`;
+
+const StyledHeader = styled("div")`
```
sophstad, line 135: [nit] Would something like `H2`, `H3` or `Subtitle` from the LG typography package, which may have padding built in, work here? If not, applying these styles to a header element like `h3` is more semantically correct than a `div`, which is a block element.

sophstad (changes requested):
For this change, I wanted to get a better sense of why this object is displayed as one change but other array updates have all their properties displayed individually!
image

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
```diff
@@ -38,11 +47,121 @@ export const EventLogTab: React.FC = () => {
...
+    ? repoEventData?.repoEvents?.eventLogEntries
+    : projectEventData?.projectEvents?.eventLogEntries;
+
+  return (
+    <div data-cy="event-log">
+      {((eventData as logEntry[]) || []).map(
+        ({ user, timestamp, before, after }) => (
+          /* @ts-expect-error */
+          <EventLogCard key={`event_log_${timestamp}`}>
+            <EventLogHeader user={user} timestamp={timestamp} />
```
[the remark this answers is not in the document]
khelif96: image

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
khelif96, line 91: Would key ever be undefined? I noticed we are using it as the key for the row above and if thats the case it can't be.

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
```diff
@@ -38,11 +47,121 @@ export const EventLogTab: React.FC = () => {
...
+    return "";
+  }
+
+  if (value === "") {
+    return '""';
+  }
+
+  const splitArray = JSON.stringify(value).split(",");
+
   return (
```
khelif96, line 157: We should be consistent about this functions return type. Returning either a string or jsx not both.

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
```diff
@@ -38,11 +47,121 @@ export const EventLogTab: React.FC = () => {
...
-    <>
-      <div>This is the event log page for {identifier}</div>
-      <div> {JSON.stringify(projectEventData)}</div>
-      <div> {JSON.stringify(repoEventData)}</div>
-    </>
+    <div>
```
khelif96, line 158:
```ts
const getEventValue = (value: boolean | string): string => {
  if (value === null || value === undefined) {
    return "";
  }
  if (typeof value === "boolean") {
    return value ? "true" : "false";
  }
  const transformedValue = JSON.stringify(value);
  if (Array.isArray(transformedValue)) {
    return transformedValue.join(", ");
  }

  return transformedValue ? String(transformedValue) : "";
};

```

This outputs the same results.

khelif96: And adds some type safety.

khelif96 (changes requested):

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
malikchaya2: this solution does not handle arrays nicely
image

malikchaya2: I'll take some parts of it and plug it into the existing function though

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
malikchaya2: I need both though, depending on what it is... Otherwise the array is hard to read.

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
```diff
@@ -38,11 +47,121 @@ export const EventLogTab: React.FC = () => {
...
+  font-size: 12px;
+  font-family: Menlo, Monaco, Consolas, "Courier New", monospace;
+`;
+
+const StyledHeader = styled("div")`
+  padding-bottom: ${size.l};
+  padding-left: ${size.xxs};
+`;
+
+const getEventValue = (value: any) => {
```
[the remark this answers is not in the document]
malikchaya2: but I also need to be able to handle an array

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
malikchaya2: I don't think it would ever be undefined

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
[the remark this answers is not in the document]
malikchaya2: that section doesn't exist :)

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
```diff
@@ -38,11 +47,121 @@ export const EventLogTab: React.FC = () => {
...
+    before?: RepoEventSettingsFragment | ProjectEventSettingsFragment;
+    after?: RepoEventSettingsFragment | ProjectEventSettingsFragment;
+  };
+
+  const eventData = isRepo
+    ? repoEventData?.repoEvents?.eventLogEntries
+    : projectEventData?.projectEvents?.eventLogEntries;
+
+  return (
+    <div data-cy="event-log">
```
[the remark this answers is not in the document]
malikchaya2: unfortunately yes

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
```diff
@@ -38,11 +47,121 @@ export const EventLogTab: React.FC = () => {
...
+
+    <div> {user} </div>
+  </StyledHeader>
+);
+
+/* @ts-expect-error */
+const EventLogCard = styled(Card)`
+  margin-bottom: 48px;
+  padding: ${size.m};
+  width: 150%;
```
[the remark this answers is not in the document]
malikchaya2: It's harder because it's used on all pages where we want it to be smaller.

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
```diff
@@ -38,11 +47,121 @@ export const EventLogTab: React.FC = () => {
...
+  <StyledHeader>
+    <H3>{getDateCopy(timestamp)}</H3>
+
+    <div> {user} </div>
+  </StyledHeader>
+);
+
+/* @ts-expect-error */
+const EventLogCard = styled(Card)`
+  margin-bottom: 48px;
```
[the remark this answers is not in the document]
malikchaya2: I don't love how it looks with size.l but it's not the end of the world. I'll use it.

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
```diff
@@ -38,11 +47,121 @@ export const EventLogTab: React.FC = () => {
...
+
+  const eventData = isRepo
+    ? repoEventData?.repoEvents?.eventLogEntries
+    : projectEventData?.projectEvents?.eventLogEntries;
+
+  return (
+    <div data-cy="event-log">
+      {((eventData as logEntry[]) || []).map(
+        ({ user, timestamp, before, after }) => (
+          /* @ts-expect-error */
```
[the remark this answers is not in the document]
malikchaya2: same

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
```diff
@@ -38,11 +47,121 @@ export const EventLogTab: React.FC = () => {
...
+    after?: RepoEventSettingsFragment | ProjectEventSettingsFragment;
+  };
+
+  const eventData = isRepo
+    ? repoEventData?.repoEvents?.eventLogEntries
+    : projectEventData?.projectEvents?.eventLogEntries;
+
+  return (
+    <div data-cy="event-log">
+      {((eventData as logEntry[]) || []).map(
```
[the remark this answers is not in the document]
malikchaya2: oh one second, lol

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
```diff
@@ -38,11 +47,121 @@ export const EventLogTab: React.FC = () => {
     },
   });
 
+  type logEntry = {
+    timestamp: Date;
+    user: string;
+    before?: RepoEventSettingsFragment | ProjectEventSettingsFragment;
+    after?: RepoEventSettingsFragment | ProjectEventSettingsFragment;
+  };
```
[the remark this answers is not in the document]
malikchaya2: sure

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
malikchaya2: LG doesn't let

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
malikchaya2: I'm already using H3, I need the padding on the entire header which includes the user field as well

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
malikchaya2: I'm following the design. I have no opinion about this but I don't know if design or product does.

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
```diff
@@ -38,11 +47,121 @@ export const EventLogTab: React.FC = () => {
...
+
+  const eventData = isRepo
```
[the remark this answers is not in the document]
malikchaya2: good catch, thank you!

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
```diff
@@ -38,11 +47,121 @@ export const EventLogTab: React.FC = () => {
...
+const EventLogCard = styled(Card)`
+  margin-bottom: 48px;
+  padding: ${size.m};
+  width: 150%;
+`;
+
+const StyledCell = styled("div")`
+  word-break: break-all;
+  font-size: 12px;
+  font-family: Menlo, Monaco, Consolas, "Courier New", monospace;
```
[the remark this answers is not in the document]
malikchaya2: switched to pre

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
```diff
@@ -38,11 +55,112 @@ export const EventLogTab: React.FC = () => {
...
+const EventLogCard = styled(Card)`
+  margin-bottom: ${size.l};
+  padding: ${size.m};
+  width: 150%;
+`;
+
+const StyledCell = styled("pre")`
+  word-break: break-all;
+  font-size: 12px;
+  font-family: Menlo, Monaco, Consolas, "Courier New", monospace;
```
sophstad, line 130: This line can be deleted now because `pre` will apply the monospace styling.

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
```diff
@@ -38,11 +55,112 @@ export const EventLogTab: React.FC = () => {
...
+);
+
+/* @ts-expect-error */
+const EventLogCard = styled(Card)`
+  margin-bottom: ${size.l};
+  padding: ${size.m};
+  width: 150%;
+`;
+
+const StyledCell = styled("pre")`
```
sophstad, line 127: [nit]
```suggestion
const StyledCell = styled.pre`
```

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
```diff
@@ -38,11 +55,112 @@ export const EventLogTab: React.FC = () => {
...
+  word-break: break-all;
+  font-size: 12px;
+  font-family: Menlo, Monaco, Consolas, "Courier New", monospace;
+`;
+
+const StyledHeader = styled("div")`
```
sophstad, line 133: [nit]
```suggestion
const StyledHeader = styled.div`
```

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
```diff
@@ -38,11 +55,112 @@ export const EventLogTab: React.FC = () => {
...
+
+interface Props {
+  timestamp: Date;
+  user: string;
+}
+
+const EventLogHeader: React.VFC<Props> = ({ user, timestamp }) => (
+  <StyledHeader>
+    <H3>{getDateCopy(timestamp)}</H3>
+    <div> {user} </div>
```
sophstad, line 116:
```suggestion
    <div>{user}</div>
```

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
```diff
@@ -38,11 +55,112 @@ export const EventLogTab: React.FC = () => {
...
+  if (value === null || value === undefined) {
+    return "";
+  }
+  if (typeof value === "boolean") {
+    return value ? "true" : "false";
+  }
+
+  if (value === "") {
+    return '""';
+  }
```
sophstad, line 148: [nit] This check isn't necessary because the next `if` statement will catch empty strings!

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
```diff
@@ -38,11 +55,112 @@ export const EventLogTab: React.FC = () => {
     },
   });
 
+  const eventData: LogEntry[] = isRepo
```
sophstad, line 58: When I load a valid project that I suppose lacks event data, I see this error:
image

Do you think it could be cleaned up?

src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.ts
```diff
@@ -31,15 +31,15 @@ const getNestedObject = (nestedObj: object, pathArr: string[]) =>
...
 };
 
 export const getEventDiffLines = (
-  before: Subset<ProjectEventSettings>,
-  after: Subset<ProjectEventSettings>
+  before: Subset<ProjectEventSettings> | Subset<RepoEventSettings>,
+  after: Subset<ProjectEventSettings> | Subset<RepoEventSettings>
 ): EventDiffLine[] => {
   const eventDiff = omitTypename(diff(before, after));
   const pathKeys = getDiffProperties(eventDiff);
```
SupaJoon, line 45: can we add a type for pathKeys? the type is currently `any`

src/pages/projectSettings/Tabs.tsx
```diff
@@ -209,25 +209,15 @@ export const ProjectSettingsTabs: React.VFC<Props> = ({
           />
         )}
       />
-      <TabRoute
-        Component={EventLogTab}
+
+      <Route
         path={routes.projectSettingsEventLog}
-        tab={ProjectSettingsTabRoutes.EventLog}
+        render={(props) => <EventLogTab {...props} projectType={projectType} />}
```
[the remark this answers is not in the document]
malikchaya2: We have the same setup for all tabs. @sophstad should we create a ticket to look into this?

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
malikchaya2: yes, but let's do it in a separate ticket.

malikchaya2: wait no, I misunderstood. I'll do that now.

malikchaya2: This should be solved on the backend, the resolver shouldn't be returning an error in this case. I'll create a ticket.

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
SupaJoon: JSON.stringify will return a string. I think you may have wanted JSON.parse

SupaJoon (approved):

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
malikchaya2: JSON.parse doesn't work well for this use case

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
```diff
@@ -38,11 +55,107 @@ export const EventLogTab: React.FC = () => {
...
+const getEventValue = (value: any) => {
+  if (value === null || value === undefined) {
+    return <></>;
+  }
+  if (typeof value === "boolean") {
+    return value ? <> true</> : <> false</>;
+  }
+
+  if (typeof value === "string") {
+    return <> &quot;{value}&quot;</>;
```
sophstad, line 146:
```suggestion
    return <>&quot;{value}&quot;</>;
```

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
```diff
@@ -38,11 +55,107 @@ export const EventLogTab: React.FC = () => {
...
+  padding-bottom: ${size.l};
+  padding-left: ${size.xxs};
+`;
+
+const getEventValue = (value: any) => {
+  if (value === null || value === undefined) {
+    return <></>;
+  }
+  if (typeof value === "boolean") {
+    return value ? <> true</> : <> false</>;
```
sophstad, line 142:
```suggestion
    return value ? <>true</> : <>false</>;
```

sophstad (approved):

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
khelif96:
```
const getEventValue = (
  value: null | undefined | boolean | string | Array<any>
) => {
  if (value === null || value === undefined) {
    return "";
  }
  if (typeof value === "boolean") {
    return String(value);
  }

  if (typeof value === "string") {
    return `"${value}"`;
  }
  if (Array.isArray(value)) {
    return JSON.stringify(value).replaceAll(",", ",\n");
  }
};
```

Produces the same output without the string parsing and extra divs
image

src/gql/fragments/projectSettings/projectEventSettings.graphql
```diff
@@ -12,7 +12,6 @@
 
 fragment projectEventSettings on ProjectEventSettings {
   projectRef {
-    id
```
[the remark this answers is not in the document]
khelif96: The cache is probably overwriting the values since it thinks they are the same object.

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
```diff
@@ -1,21 +1,38 @@
...
+  before?: RepoEventSettingsFragment | ProjectEventSettingsFragment;
+  after?: RepoEventSettingsFragment | ProjectEventSettingsFragment;
+};
+
+export const EventLogTab: React.VFC<{
+  projectType: ProjectType;
+}> = ({ projectType }) => {
   const { identifier } = useParams<{ identifier: string }>();
-  const isRepo = validateObjectId(identifier);
+  const isRepo = projectType === ProjectType.Repo;
```
khelif96, line 33: what should this page show if isRepo  is false?

src/gql/fragments/projectSettings/projectEventSettings.graphql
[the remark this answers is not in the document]
khelif96: In case you actually do end up needing the id field the best solution would be to add a custom merge policy for this case. https://www.apollographql.com/docs/react/caching/cache-field-behavior

khelif96 (approved):
Assuming the rest of my comments are addressed.

malikchaya2:
test failure is unrelated and failing on master

PR #1233, opened 2022-04-21
EVG-13847: Show ETA in commit chart tooltip

SupaJoon:
[EVG-13847](https://jira.mongodb.org/browse/EVG-13847)

### Description 
Show the eta for dispatched tasks in the commit chart tooltip

### Screenshots
Screen Shot 2022-04-21 at 5 02 22 PM

### Testing
Rendered in storybook: http://localhost:9009/?path=/story/commit-charts--absolute-chart

### Evergreen PR 
https://github.com/evergreen-ci/evergreen/pull/5605


src/pages/commits/ActiveCommits/CommitChartTooltip.tsx
```diff
@@ -10,18 +10,22 @@ import { size, zIndex } from "constants/tokens";
...
   trigger: React.ReactElement;
+  eta?: Date;
 }
 
 export const CommitChartTooltip: React.VFC<Props> = ({
   groupedTaskStats,
   trigger,
+  eta,
 }) => {
+  const etaString = msToDuration(new Date(eta).valueOf() - Date.now());
```
sophstad, line 28: More of a general observation but it looks like `eta` is really a string instead of a Date type (since we need to create a new instance of a `Date`). I wonder if there is something that could be improved with our gqlgen/codegen config so that `Time` scalars are accurately reflected as JS Date objects.

sophstad (approved):
This looks great! It would be nice to have the live timer like we do on the task page but I'm sure that would greatly increase the complexity here so LGTM!

SupaJoon:
> This looks great! It would be nice to have the live timer like we do on the task page but I'm sure that would greatly increase the complexity here so LGTM!

I could actually refactor some of the eta 

> This looks great! It would be nice to have the live timer like we do on the task page but I'm sure that would greatly increase the complexity here so LGTM!

I can probably hookify some of the logic in the ETATimer component to make this simpler

PR #1235, opened 2022-04-22
EVG-16286 Add regex validators to commits page filters

khelif96:
[EVG-16286](https://jira.mongodb.org/browse/EVG-16286)

### Description 
Adds a validator for bad regular expressions.
Create a hook that wraps `upsertQueryParam`. 
Update `IconTooltip` to make it more flexible 
Update `TextInputWithGlyph` to make it more flexible and reuse it in other components.
Stories for the new components.

https://user-images.githubusercontent.com/4605522/164772443-d464bd0a-4b42-4113-ae98-9c08383b62a3.mov


src/components/SearchableDropdown/index.tsx
```diff
@@ -128,14 +128,12 @@ const SearchableDropdown = <T extends {}>({
...
       buttonText = value.join(", ");
-    } else {
-      buttonText = value.toString();
     }
   }
 
   return (
     <Container>
-      <InputLabel htmlFor={`searchable-dropdown-${label}`}>{label}</InputLabel>
+      <Label htmlFor={`searchable-dropdown-${label}`}>{label}</Label>
```
minnakt, line 136: Nice, this was bothering me for a while

src/components/TupleSelect/index.tsx
```diff
@@ -20,37 +18,43 @@ type option = {
...
-      <Input.Group compact>
-        <Select
-          style={{ width: "30%" }}
+      <InputGroup>
+        <GroupedSelect
           value={selected}
           onChange={(v) => setSelected(v)}
           aria-label="Select Drop Down"
           data-cy="tuple-select-dropdown"
+          aria-labelledby="filter-input"
```
minnakt, line 57: What's the reason we need two `aria-label` properties?

src/components/TupleSelect/index.tsx
```diff
@@ -20,37 +18,43 @@ type option = {
...
   return (
     <Container>
       <Label htmlFor="filter-input">
         Add New {selectedOption.displayName} Filter
       </Label>
-      <Input.Group compact>
-        <Select
-          style={{ width: "30%" }}
+      <InputGroup>
+        <GroupedSelect
```
minnakt, line 52: It doesn't like when I click on the placeholder.

https://user-images.githubusercontent.com/47064971/165418048-1607b5f9-e87f-4dc9-95af-647b5a05b385.mov

You can get rid of the placeholder by adding the props `allowDeselect={false}` though, which will make this problem go away :)

src/components/TupleSelect/index.tsx
```diff
@@ -61,31 +65,80 @@ export const TupleSelect: React.VFC<TupleSelectProps> = ({
...
+
+const InputGroup = styled.div`
+  align-items: center;
+  display: flex;
+  flex-direction: row;
 `;
+
+const InactiveIconWrapper = styled.div`
+  margin: 6px;
+`;
```
minnakt, line 121: This seems kind of hacky to me... 🤔

A less hacky solution (but still not that pretty) would be to put a fixed `width` & `justify-content: center` property in the `IconWrapper` style in `TextInputWithGlyph`

Screen Shot 2022-04-26 at 9 23 29 PM

result:

https://user-images.githubusercontent.com/47064971/165419665-d696856c-4e4c-434e-9521-95001387f285.mov

src/components/HistoryTable/HistoryTableTestSearch/HistoryTableTestSearch.tsx
```diff
@@ -1,75 +1,83 @@
...
-  position: absolute;
-  height: 100%;
-  top: 10px; //temporary
-  align-self: flex-end;
-  margin-right: 10px;
-  &:hover {
-    cursor: pointer;
-  }
+const InactiveIconWrapper = styled.div`
+  margin: 6px;
```
minnakt, line 82: same comment

src/components/TupleSelect/index.tsx
```diff
@@ -61,31 +65,80 @@ export const TupleSelect: React.VFC<TupleSelectProps> = ({
...
-      </Input.Group>
+      </InputGroup>
     </Container>
   );
 };
 
 const Container = styled.div`
   display: flex;
   flex-direction: column;
+  position: relative;
```
minnakt, line 110: nit. I think you don't need the `position: relative` here

src/components/TupleSelect/index.tsx
khelif96: Nice catch!

src/components/HistoryTable/HistoryTableTestSearch/HistoryTableTestSearch.tsx
khelif96: Nice catch remnants of my n'thousandth iteration at making this reusable.

src/components/TextInputWithGlyph/index.tsx
```diff
@@ -1,33 +1,35 @@
...
+  height: 36px; /* height of LG text-input */
   position: absolute;
-  align-self: flex-end;
-  margin-right: 10px;
-  bottom: 10px;
-  &:hover {
-    cursor: pointer;
-  }
+  right: ${size.xs};
+  width: ${size.m};
```
minnakt, line 31: I'm a little unsure about `24px` because the icon button's default width is actually `28px`, so it's kind of strange (the content is bigger than the container)

It does work though so maybe it's fine 🙂

src/components/HistoryTable/HistoryTableTestSearch/HistoryTableTestSearch.tsx
minnakt: reminder to remove this style as it's no longer necessary

src/pages/Commits.tsx
```diff
@@ -118,7 +119,10 @@ export const Commits = () => {
   const { badges, handleOnRemove, handleClearAll } = useFilterBadgeQueryParams(
     queryParamsToDisplay
   );
-  const onSubmitTupleSelect = ({ category }: { category: string }) => {
+  const onSubmit = useUpsertQueryParams();
+
+  const onSubmitTupleSelect = ({ category, value }) => {
+    onSubmit({ category, value });
```
minnakt, line 125: I wonder if it would make more sense to call `useUpsertQueryParams` inside of the `TupleSelect` component since that's the only place where it's used on this page. 🤔 At least, that's the pattern that the `HistoryTableTestSearch` uses

minnakt (approved):
lgtm! 🤠🔥👊

src/components/HistoryTable/HistoryTableTestSearch/HistoryTableTestSearch.tsx
khelif96: oof.

PR #1240, opened 2022-04-28
EVG-16590: Use projectIdentifier in Project Patches route

minnakt:
[EVG-16590](https://jira.mongodb.org/browse/EVG-16590)

### Description 
This PR makes it so that the `projectIdentifier` is used instead of the `id` (`projectId`) in the project patches route.

The reason this still works despite there being no changes in the Evergreen repo is because the [ById](https://github.com/evergreen-ci/evergreen/blob/main/model/project_ref.go#L1210-L1217) function used to look up projects actually matches on either the project ID or identifier. In fact, if you go to the [sandbox](https://spruce-staging.corp.mongodb.com/project/5ffe393097b1d3759dd3c1a3/patches?commitQueue=false) project on spruce-staging and replace the ID in the url with `sandbox`, you'll see that this functionality already works.

### Screenshots
- No visible changes (only the links have changed)

### Testing 
- Manually on prod; an example of a project that has a different ID than its identifier is the [patch-analysis](https://spruce.mongodb.com/project/602e77ec5623435862026af3/patches) project

### DIFF
--- a/src/components/Breadcrumb/index.tsx
+++ b/src/components/Breadcrumb/index.tsx
@@ -142,7 +142,7 @@ const VersionBreadcrumb: React.VFC<VersionBreadcrumbProps> = ({
   analytics,
   isTask,
 }) => {
-  const { project, projectIdentifier, revision, id } = versionMetadata;
+  const { projectIdentifier, revision, id } = versionMetadata;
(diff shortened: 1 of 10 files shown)


SupaJoon:
we may be able to remove the projectId field from patchesPage.graphql and get-task.graphql

PR #1241, opened 2022-04-28
EVG-15068: Add Duplicate Project functionality

sophstad:
EVG-15968

### Description 
- For projects, add dropdown menu allowing users to either create a new project or duplicate the current one
- Redirect users to this new project on submission
- Fix bug where owner and repo were not being prepopulated for repos on the "Create New Project" modal

### Screenshots
image
image

### Testing 
While running a local dev server, duplicate the local `evergreen` project: http://localhost:3000/project/evergreen/settings


src/components/projectSelect/index.tsx
```diff
@@ -129,13 +128,8 @@ export const ProjectSelect: React.VFC<ProjectSelectProps> = ({
   );
 };
 
-const { reportError } = errorReporting;
-
 export const getRepoId = (projects: Project[]) => {
   if (!projects || projects.length === 0) {
-    reportError(
-      new Error("No projects provided when fetching repo id")
-    ).warning();
```
sophstad, line 138: This was throwing warnings on every render, but it's not actually an error if a repo exists with no projects attached to it.

src/pages/projectSettings/CreateDuplicateProjectButton.tsx
```diff
@@ -0,0 +1,103 @@
...
+          data-cy="new-project-menu"
+        >
+          <MenuItem
+            data-cy="create-project-button"
+            onClick={() => {
+              setMenuOpen(false);
+              setCreateModalOpen(true);
+            }}
+          >
+            Create New Project
```
ablack12, line 76: Would it be possible to refresh the dropdown after duplicating a project?

src/pages/projectSettings/CopyProjectModal.tsx
```diff
@@ -0,0 +1,103 @@
...
+    CopyProjectMutation,
+    CopyProjectMutationVariables
+  >(COPY_PROJECT, {
+    onCompleted({ copyProject: { identifier } }) {
+      dispatchToast.success(`Successfully created the project: ${identifier}`);
+      replace(getProjectSettingsRoute(identifier));
+    },
+    onError(err) {
+      dispatchToast.error(
+        `There was an error creating the project: ${err.message}`
```
ablack12, line 47: For many errors, we still create the project just some things may not be carried over. For example, 
image

When copying the spruce project I get errors about admins, however [at that point in the code|https://github.com/evergreen-ci/evergreen/blob/df23005997ff5ce934cc6114ae353a27ea42ab3e/rest/data/project_settings.go#L76] we've already copied everything else. Should we still redirect then and indicate t's a partial error? (We could modify the copy function to still return the item and/or try all of the pieces and return the entire error if that's better)

ablack12 (commented):
This probably isn't related to your changes but when I try to edit a local project that isn't evergreeen (spruce for example) the save changes button doesn't become available even when I click things. Is this purposeful or already tracked?

src/pages/projectSettings/CopyProjectModal.tsx
sophstad: Discussed offline, this will require additional back and front end work. EVG-16784 has been created to capture these changes.

ablack12 (approved):

src/pages/projectSettings/CreateDuplicateProjectButton.tsx
```diff
@@ -0,0 +1,103 @@
...
+export const CreateDuplicateProjectButton: React.VFC<Props> = ({
+  id,
+  label,
+  owner,
+  projectType,
+  repo,
+}) => {
+  const { data } = useQuery<GetUserPermissionsQuery>(GET_USER_PERMISSIONS);
+  const {
+    user: {
```
SupaJoon, line 41: what is represented by the object after the colon?

SupaJoon (approved):

src/pages/projectSettings/CreateDuplicateProjectButton.tsx
sophstad: It's a list of permissions that the current user possesses. We can add more fields in the future to conditionally render features in Spruce
https://github.com/evergreen-ci/evergreen/blob/bd8cf591918180b803cf0232cac9146b87599a31/graphql/schema.graphql#L1129-L1132

minnakt (approved):

PR #1010, opened 2021-12-09
Improve yarn dev speeds

sophstad:
### Description 
- Upgrade from Yarn Classic to Yarn Modern (EVG-15690)
  - Use [Yarn `link:` protocol](https://yarnpkg.com/features/protocols), their [recommended way](https://yarnpkg.com/features/protocols/#why-is-the-link-protocol-recommended-over-aliases-for-path-mapping) of handling path aliases (i.e. writing `components/...` instead of `src/components/...`)
  - Upgrade Storybook to be compatible
    - Note that we are blocked from updating to the latest Storybook versions due to API changes introduced in React Router
    - Use custom resolutions in package.json to prevent `storybook-addon-apollo-client` from introducing React Router v6 as a dependency 
- Remove references to [Plug'n'Play](https://yarnpkg.com/features/pnp) since using it's pretty nascent and switching over isn't on our roadmap

No upgrading is required to start using Yarn Modern; this is configured by the `.yarnrc` directory added to Spruce.

### Performance Improvements

I saw the following improvements on my local machine:

| | Before | After |
| --- | --- | --- |
| `yarn install` | 1:14 | 0:50 |
| `yarn dev` | 1:50 | 1:15 |

### Testing 
- Tested in staging and dev environments

PR #1252, opened 2022-05-03
EVG-15690: Upgrade to yarn 3

SupaJoon:
[EVG-15690(https://jira.mongodb.org/browse/EVG-15690)

### Description 
Followed the migration guide [here](https://yarnpkg.com/getting-started/migration) without installing Plug n Play
Upgrade your global yarn installation and trying running `yarn --version` when you check out this PR to confirm that version 3 is being used. 
The aliases changed the import sort order for files in the last commit.


src/App.test.tsx
```diff
@@ -1,6 +1,6 @@
 import React from "react";
-import App from "App";
 import { render, act } from "test_utils";
+import App from "App";
```
khelif96, line 3: Would pnp prevent this? 
The eslint rules that organize sorts explicitly sort internal deps after external ones and this breaks that. Is there a way we can do this without breaking that sort order?

khelif96 (changes requested):

khelif96 (commented):
It also seems like yarn-install got slower not faster. 
This pr.
https://evergreen.mongodb.com/task_log_raw/spruce_ubuntu1604_e2e_test_patch_6f535df6742c603d1c7c27e4b7754db374e05744_6271424d57e85a7a79ae9124_22_05_03_14_55_11/0?type=T#L610
Base commit.
https://evergreen.mongodb.com/task_log_raw/spruce_ubuntu1604_e2e_test_6f535df6742c603d1c7c27e4b7754db374e05744_22_05_02_15_34_19/0?type=T#L510

src/App.test.tsx
sophstad: I _think_ making this change to the ESLint config would prevent all the import sort changes

https://github.com/evergreen-ci/spruce/pull/1010/files#diff-b3d81df455893664810b193106f3543c920c80f9f4281b5541b949a8e1790769

SupaJoon:
> It also seems like yarn-install got slower not faster. This pr. https://evergreen.mongodb.com/task_log_raw/spruce_ubuntu1604_e2e_test_patch_6f535df6742c603d1c7c27e4b7754db374e05744_6271424d57e85a7a79ae9124_22_05_03_14_55_11/0?type=T#L610 Base commit. https://evergreen.mongodb.com/task_log_raw/spruce_ubuntu1604_e2e_test_6f535df6742c603d1c7c27e4b7754db374e05744_22_05_02_15_34_19/0?type=T#L510

We should expect some variability between installation times. I did 5 complete local installations with

Yarn v1.22.18:
1m 52s
1m 5s
1m 24s
40s 27ms
1m 34s

Yarn v3.2.0:
2m 20s
2m 20s
3m 5s
2min 16s
2min 8s

src/.eslintrc.js
```diff
@@ -63,7 +63,7 @@ module.exports = {
           },
           {
             pattern:
-              "(analytics|components|constants|context|gql|hoc|hooks|pages|types|utils)/**",
+              "{App,analytics,components,constants,context,gql,hoc,hooks,pages,test_utils,types,utils}{,/**}",
```
sophstad, line 66: It would probably make more sense to remove `App` here and include it as a relative import in src/index.tsx
```js
import App from "./App"
```
```suggestion
              "{analytics,components,constants,context,gql,hoc,hooks,pages,test_utils,types,utils}{,/**}",
```

package.json
```diff
@@ -95,14 +95,20 @@
     "@leafygreen-ui/tooltip": "6.3.1",
     "@leafygreen-ui/typography": "8.1.0",
     "@rjsf/core": "4.1.0",
+    "analytics": "link:./src/analytics",
```
sophstad, line 98: I would advocate removing `config/modules.js` since we're using this link format now. The webpack aliases introduced there have the same purpose as these.

sophstad (changes requested):

package.json
SupaJoon: I removed the baseUrl parameter from our tsconfig which is used in the `config/modules.js` to generate the webpack config since I'll like to avoid making changes to our build system until we redo it.

src/App.test.tsx
khelif96: I actually think this change breaks the minimatch expression matching used by `import/order`.
https://www.digitalocean.com/community/tools/glob?comments=true&glob=%7Banalytics%2Ccomponents%2Cconstants%2Ccontext%2Cgql%2Choc%2Chooks%2Cpages%2Ctest_utils%2Ctypes%2Cutils%7D%7B%2C%2F%2A%2A%7D%22&matches=false&tests=components%2Ftest&tests=analytics%2Ftest

https://www.digitalocean.com/community/tools/glob?comments=true&glob=%2B%28analytics%7Ccomponents%7Cconstants%7Ccontext%7Cgql%7Choc%7Chooks%7Cpages%7Ctypes%7Cutils%29%2F%2A%2A&matches=false&tests=components%2Ftest&tests=analytics%2Ftest

sophstad: Hm, isn't the problem with your first link that it contains an extraneous double quote?
image

Removing it passes the checks:
https://www.digitalocean.com/community/tools/glob?comments=true&glob=%7Banalytics%2Ccomponents%2Cconstants%2Ccontext%2Cgql%2Choc%2Chooks%2Cpages%2Ctest_utils%2Ctypes%2Cutils%7D%7B%2C%2F%2A%2A%7D&matches=false&tests=components%2Ftest&tests=analytics%2Ftest

src/.eslintrc.js
```diff
@@ -63,7 +63,7 @@ module.exports = {
           },
           {
             pattern:
-              "(analytics|components|constants|context|gql|hoc|hooks|pages|types|utils)/**",
+              "{analytics,components,constants,context,gql,hoc,hooks,pages,test_utils,types,utils}{,/**}",
```
khelif96, line 66: This change actually breaks the `import/order` plugin. See my comment here.
 https://github.com/evergreen-ci/spruce/pull/1252#discussion_r867010584

image

khelif96:
> https://evergreen.mongodb.com/task_log_raw/spruce_ubuntu1604_e2e_test_6f535df6742c603d1c7c27e4b7754db374e05744_22_05_02_15_34_19/0?type=T#L510

I think this only applies to local installs since yarn 3 is able to rely on a cache. CI builds it will always be slower since it can't rely on a cache and has to write to a cache. In this case should we tell yarn to skip the cache on CI?

khelif96:
> > It also seems like yarn-install got slower not faster. This pr. https://evergreen.mongodb.com/task_log_raw/spruce_ubuntu1604_e2e_test_patch_6f535df6742c603d1c7c27e4b7754db374e05744_6271424d57e85a7a79ae9124_22_05_03_14_55_11/0?type=T#L610 Base commit. https://evergreen.mongodb.com/task_log_raw/spruce_ubuntu1604_e2e_test_6f535df6742c603d1c7c27e4b7754db374e05744_22_05_02_15_34_19/0?type=T#L510
> 
> The runs you linked differed by about 14s. We should expect some variability between installation times. I did 5 complete local installations with **v3.2.0:** 1m 52s 1m 5s 1m 24s 40s 27ms 1m 34s
> 
> **v1.22.18:** 2m 20s 2m 53s 3m 5s 2min 16s 2min 8s

I think this only applies to local installs since yarn 3 is able to rely on a cache. CI builds will always be slower since it can't rely on a cache and has to generate a cache every time. In this case should we tell yarn to skip the cache if its on CI?

src/App.test.tsx
khelif96: Wow great catch! My copy and pasting skills aren't what they used to be.
It did not error in my editor earlier so i assumed the problem was with the statement. But i guess eslint needed to restart. 
image

src/.eslintrc.js
khelif96, line 66: I still think there is something up with this statement. 
image
```ts
import Tooltip from "@leafygreen-ui/tooltip";
import { useLocation } from "react-router";
import { useSpawnAnalytics } from "analytics";
import { ConditionalWrapper } from "components/ConditionalWrapper";
import { PlusButton } from "components/Spawn";
import {
  MyHostsQuery,
  MyHostsQueryVariables,
  GetSpruceConfigQuery,
  GetSpruceConfigQueryVariables,
} from "gql/generated/types";
import { GET_MY_HOSTS, GET_SPRUCE_CONFIG } from "gql/queries";
import { SpawnHostModal } from "pages/spawn/spawnHost/index";
import { HostStatus } from "types/host";
import { queryString } from "utils";
```

SupaJoon:
> > > It also seems like yarn-install got slower not faster. This pr. https://evergreen.mongodb.com/task_log_raw/spruce_ubuntu1604_e2e_test_patch_6f535df6742c603d1c7c27e4b7754db374e05744_6271424d57e85a7a79ae9124_22_05_03_14_55_11/0?type=T#L610 Base commit. https://evergreen.mongodb.com/task_log_raw/spruce_ubuntu1604_e2e_test_6f535df6742c603d1c7c27e4b7754db374e05744_22_05_02_15_34_19/0?type=T#L510
> > 
> > 
> > The runs you linked differed by about 14s. We should expect some variability between installation times. I did 5 complete local installations with **v3.2.0:** 1m 52s 1m 5s 1m 24s 40s 27ms 1m 34s
> > **v1.22.18:** 2m 20s 2m 53s 3m 5s 2min 16s 2min 8s
> 
> I think this only applies to local installs since yarn 3 is able to rely on a cache. CI builds will always be slower since it can't rely on a cache and has to generate a cache every time. In this case should we tell yarn to skip the cache if its on CI?

That's a good point about the yarn cache. Unfortunately, it's possible to disable it based on the [configuration options](https://yarnpkg.com/configuration/yarnrc).

khelif96:
Since caching is only really effective locally and has no benefit on CI i would push back on doing this.  We don't reinstall `node_modules` locally often enough during daily development where we would benefit as much. (Typically only when we upgrade dependencies but the cache wouldn't be valid anyways.) Install and task run times seem to be slower across the board with this change whereas the original goal was to speed them up. 
I think we would benefit more by getting a shared cache working for tasks and upgrading the build system for faster dev server startups and compilations.  What do you think @SupaJoon @sophstad 

Base Commit
image
This Commit
image

SupaJoon:
> Since caching is only really effective locally and has no benefit on CI i would push back on doing this. We don't reinstall `node_modules` locally often enough during daily development where we would benefit as much. (Typically only when we upgrade dependencies but the cache wouldn't be valid anyways.) Install and task run times seem to be slower across the board with this change whereas the original goal was to speed them up. I think we would benefit more by getting a shared cache working for tasks and upgrading the build system for faster dev server startups and compilations. What do you think @SupaJoon @sophstad
> 
> Base Commit image This Commit image

I also see that yarn 3 is slower across the board and the [docs](https://yarnpkg.com/getting-started/qa#is-yarn-faster-than-other-package-managers) mention speed isn't actually a main objective of the project. pnpm is a more promising alternative but we will need to have it installed on ubuntu1604-small before testing. I will close this ticket and experiment with pnpm as part of a separate ticket.

PR #1256, opened 2022-05-04
EVG-16814 Fix EventLog types 

malikchaya2:
[EVG-16814](https://jira.mongodb.org/browse/EVG-16814)

### Description 
There was some strange behavior around the before and after types because some functions up the chain were not typed correctly.

### DIFF
--- a/src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.ts
+++ b/src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.ts
@@ -27,7 +27,7 @@ const formatArrayElements = (eventKey: string): string =>
 const getNestedObject = (nestedObj: object, pathArr: string[]): EventValue =>
   pathArr.reduce((obj, key) => (obj ? obj[key] : undefined), nestedObj);
 
-export type EventValue = boolean | string | Array<any>;
+export type EventValue = boolean | string | Array<string | boolean | object>;
 export type EventDiffLine = {
   key: string;
   before: EventValue;
--- a/src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.ts
+++ b/src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.ts
@@ -25,10 +25,7 @@ const formatArrayElements = (eventKey: string): string =>
   eventKey.replace(/.[0-9]./g, (x) => `[${x[1]}].`);
 
 const getNestedObject = (nestedObj: object, pathArr: string[]): EventValue =>
-  pathArr.reduce(
-    (obj, key) => (obj && obj[key] !== "undefined" ? obj[key] : undefined),
-    nestedObj
-  );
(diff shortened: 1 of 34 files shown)


src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.ts
```diff
@@ -18,23 +18,23 @@ const getDiffProperties = (eventObj: object) => {
...
   eventKey.replace(/.[0-9]./g, (x) => `[${x[1]}].`);
 
-const getNestedObject = (nestedObj: object, pathArr: string[]) =>
+const getNestedObject = (nestedObj: object, pathArr: string[]): EventValue =>
   pathArr.reduce(
     (obj, key) => (obj && obj[key] !== "undefined" ? obj[key] : undefined),
     nestedObj
   );
 
+export type EventValue = boolean | string | Array<any>;
```
khelif96, line 33: I think this `Array<any>` type was what was causing the issue in the first place. It might be possible to more strongly type this value. If we know it can only ever be an array of strings or booleans we can define it as `Array<string | boolean>`

khelif96 (changes requested):

src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.ts
malikchaya2: I think it's more likely to be an object. This isn't what was causing the problem, there were functions that were returning any type.

src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.ts
```diff
@@ -18,23 +18,23 @@ const getDiffProperties = (eventObj: object) => {
...
 };
 
-const formatArrayElements = (eventKey: string) =>
+const formatArrayElements = (eventKey: string): string =>
   eventKey.replace(/.[0-9]./g, (x) => `[${x[1]}].`);
 
-const getNestedObject = (nestedObj: object, pathArr: string[]) =>
+const getNestedObject = (nestedObj: object, pathArr: string[]): EventValue =>
   pathArr.reduce(
     (obj, key) => (obj && obj[key] !== "undefined" ? obj[key] : undefined),
```
khelif96, line 29: This function confuses me. Like i get what it does but are there objects with the string "undefined" as the value instead of undefined?

malikchaya2: that check shouldn't be there, I removed it.

src/pages/projectSettings/tabs/EventLogTab/EventLogDiffs.ts
khelif96: I would advocate for another type other any if possibe

khelif96 (approved):

PR #1258, opened 2022-05-05
EVG-14731: Indicate when commits will be preserved in Enqueue Modal

minnakt:
[EVG-14731](https://jira.mongodb.org/browse/EVG-14731)

### Description 
This PR makes it so that if a user tries to enqueue a patch that was created with the `--preserve-commits` option, there will be 1) a message that says the commits will be preserved and 2) no text box as the commit message cannot be edited.

Note: The enqueue button is only enabled for patches. So this modal would never be reachable from a version.

### Screenshots
https://user-images.githubusercontent.com/47064971/166992260-66986e3a-5f8f-47b3-a1dd-f6639811dcc3.mov

### Testing 
- manually, example patch [here](https://spruce.mongodb.com/user/minna.kt/patches?commitQueue=true&patchName=preserve%20commits)

### DIFF
--- a/src/components/PatchActionButtons/EnqueuePatch.tsx
+++ b/src/components/PatchActionButtons/EnqueuePatch.tsx
@@ -1,15 +1,6 @@
 import { useState } from "react";
-import { useQuery } from "@apollo/client";
 import { DropdownItem } from "components/ButtonDropdown";
-import {
-  CodeChangesQuery,
-  CodeChangesQueryVariables,
-} from "gql/generated/types";
-import { GET_CODE_CHANGES } from "gql/queries";
 import { EnqueuePatchModal } from "pages/version/index";
-import { commits } from "utils";
-
-const { shouldPreserveCommits } = commits;
 
 interface EnqueuePatchProps {
   patchId: string;
@@ -32,19 +23,6 @@ export const EnqueuePatch: React.VFC<EnqueuePatchProps> = ({
       ? visibilityControl
       : fallbackVisibilityControl;
 
-  const { data, previousData } = useQuery<
-    CodeChangesQuery,
-    CodeChangesQueryVariables
-  >(GET_CODE_CHANGES, {
-    variables: { id: patchId },
-    skip: !isVisible,
-  });
-  const { patch } = data ?? previousData ?? {};
-  const { moduleCodeChanges } = patch ?? {};
-  const preserveCommits = moduleCodeChanges?.length
-    ? shouldPreserveCommits(moduleCodeChanges[0].fileDiffs)
-    : false;
-
   return (
     <>
       <DropdownItem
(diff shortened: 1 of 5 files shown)


src/components/PatchActionButtons/EnqueuePatch.tsx
```diff
@@ -23,6 +32,19 @@ export const EnqueuePatch: React.VFC<EnqueuePatchProps> = ({
...
       : fallbackVisibilityControl;
 
+  const { data, previousData } = useQuery<
+    CodeChangesQuery,
+    CodeChangesQueryVariables
+  >(GET_CODE_CHANGES, {
+    variables: { id: patchId },
+    skip: !isVisible,
+  });
+  const { patch } = data ?? previousData ?? {};
```
minnakt, line 42: I did this to prevent this glitchy weirdness 😞

https://user-images.githubusercontent.com/47064971/167157497-c1d21f1d-b9b9-4f93-94e5-808a26ac2d9b.mov

khelif96: How does this prevent the modal from doing that?

src/utils/commits/index.ts
```diff
@@ -1 +1,19 @@
...
+ * @return {boolean} true if commits are preserved, false if otherwise
+ */
+export const shouldPreserveCommits = (
+  fileDiffs: FileDiffsFragment[]
+): boolean => {
+  if (fileDiffs.length && fileDiffs[0].description !== "") {
+    return true;
+  }
+  return false;
+};
```
khelif96, line 19: This should probably live somewhere by enqueue modal i think its only used by that. Unless you think it might be used elsewhere.

minnakt: It's used by `CodeChanges.tsx`!

src/components/PatchActionButtons/EnqueuePatch.tsx
minnakt: When the state of the modal is no longer visible the `data` gets reset to `undefined` so `preserveCommits` becomes interpreted as `false`

Something interesting is that if I keep the `antd` Modal (as opposed to this LG Modal), the glitchiness doesn't happen

minnakt: I experimented by replacing our `VersionRestartModal` (which has a similar `skip: !visible` query) with a LG Modal as well and the same thing happens 😞

https://user-images.githubusercontent.com/47064971/167202303-61f62823-b5cf-4d6f-939a-0cb28cb20369.mov

khelif96: interesting... i would be curious to know why this happens. Does Apollo set the data to undefined when the component unmounts? Does LG unmount the modal in a different way than  antd?

khelif96 (approved):

src/components/PatchActionButtons/EnqueuePatch.tsx
minnakt: Yeah 🤔 Maybe `antd` preserves the content somehow before the modal fully closes? They do have a swanky animation on their modal which possibly presents the user from noticing it

PR #1277, opened 2022-05-17
EVG-16782: Remove repo event queries

sophstad:
EVG-16782

### Description 
- Remove calls to `RepoEventLogsQuery` and `RepoEventSettings` so that we can safely update types on the backend
- Event Log pages for repos safely load an empty page

### Evergreen PR
Depended on by https://github.com/evergreen-ci/evergreen/pull/5668


malikchaya2 (changes requested):
I just added a comment to the [evergreen PR](https://github.com/evergreen-ci/evergreen/pull/5668#pullrequestreview-976094010). I'm not sure I understand how this switch solves the problem we were seeing.

khelif96 (commented):
To make sure i am understanding this. Is the goal of this PR and the linked evergreen PR to make it safe to update Repo Event to have nullable values instead of non nullable?

sophstad:
> To make sure i am understanding this. Is the goal of this PR and the linked evergreen PR to make it safe to update Repo Event to have nullable values instead of non nullable?

@khelif96 Yes! And that will be done via changing the resolver return type from `RepoEventSettings` to `ProjectEventSettings`, which is nullable.

khelif96:
> 

So I see that `RepoEventSettings` and `ProjectEventSettings` use different types for `projectRef` one of which is a `Project` the other is a `RepoRef` afaict the fields with in `Project` and `RepoRef`  share some fields they are not a 1:1 match. Is there a plan to update all of the references to all of those mismatched fields?

sophstad:
> So I see that `RepoEventSettings` and `ProjectEventSettings` use different types for `projectRef` one of which is a `Project` the other is a `RepoRef`. afaict the fields within `Project` and `RepoRef` share some fields but they are not a 1:1 match. Is there a plan to update all of the references to all of those mismatched fields?

@khelif96 this discussion is probably better suited for the app PR since that's where the types are being changed :)

khelif96:
> > So I see that `RepoEventSettings` and `ProjectEventSettings` use different types for `projectRef` one of which is a `Project` the other is a `RepoRef`. afaict the fields within `Project` and `RepoRef` share some fields but they are not a 1:1 match. Is there a plan to update all of the references to all of those mismatched fields?
> 
> @khelif96 this discussion is probably better suited for the app PR since that's where the types are being changed :)

Sure just moved it over.

malikchaya2 (approved):

PR #1279, opened 2022-05-17
EVG-16188 Replace page dropdown with leafygreen one

khelif96:
[EVG-16188](https://jira.mongodb.org/browse/EVG-16188)

### Description 
Didn't need to update much thanks to @minnakt . I removed the antd version and added some tests.
Leafygreen doesn't support `dataCy` on `Select` yet so i needed to get creative. 
Refactored out the query param logic into a hook so the component can be used with out it.

### Screenshots
image

### Testing 
Wrote some unit tests and updated cypress


src/components/PageSizeSelector/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+interface Props {
+  value: number;
+  "data-cy"?: string;
+  onChange?: (i: number) => void;
+}
+
+/**
+ * `data-cy` prop is not currently supported by @leafygreen-ui/select
+ */
```
sophstad, line 14: I'm confused by this comment because the `data-cy` prop is still being used. Do you mean that we're including it anyway so that the prop is in place once LeafyGreen supports it?

src/components/PageSizeSelector/usePageSizeSelector.ts
```diff
@@ -0,0 +1,14 @@
+import { RECENT_PAGE_SIZE_KEY } from "constants/index";
+import { useUpdateURLQueryParams } from "hooks";
+
+/** usePageSizeSelector updates the page size query param and saves the page size to local storage */
+const usePageSizeSelector = () => {
+  const updateURLQueryParams = useUpdateURLQueryParams();
+  const setPageSize = (pageSize: number) => {
+    localStorage.setItem(RECENT_PAGE_SIZE_KEY, `${pageSize}`);
+    updateURLQueryParams({ limit: pageSize.toString(), page: "0" });
```
sophstad, line 9: [nit] Since `${pageSize}` and `pageSize.toString()` do the same thing, you could stringify pageSize once and use the variable for both

src/components/PageSizeSelector/usePageSizeSelector.ts
```diff
@@ -0,0 +1,14 @@
+import { RECENT_PAGE_SIZE_KEY } from "constants/index";
+import { useUpdateURLQueryParams } from "hooks";
+
+/** usePageSizeSelector updates the page size query param and saves the page size to local storage */
+const usePageSizeSelector = () => {
+  const updateURLQueryParams = useUpdateURLQueryParams();
+  const setPageSize = (pageSize: number) => {
```
sophstad, line 7: Since we usually make a call to analytics when setting the page size, would it make sense for `setPageSize` to take in an optional Analytics object and log the action if one exists? This could help avoid avoid repeating identical "Change Page Size" event names across the codebase.

cypress/integration/page_size_selector.ts
```diff
@@ -1,104 +0,0 @@
-// / <reference types="Cypress" />
```
khelif96, line 1: Removed this test since the PageSizeSelector is effectively being tested in several other tests this seems fairly redundant.

src/components/PageSizeSelector/index.tsx
khelif96: Yeah that was the intention but i just removed it all together and replaced it with a spread.

src/components/PageSizeSelector.tsx
```diff
@@ -1,80 +0,0 @@
...
-  onClick,
-  sendAnalyticsEvent = () => undefined,
-  useLeafygreen = false,
-}) => {
-  const { replace } = useHistory();
-  const { search, pathname } = useLocation();
-
-  const handleChange =
-    onClick ||
-    ((pageSize: number) => {
```
minnakt, line 32: I kind of liked this pattern, and was wondering if there's a reason we went away from it

I thought it made sense that `PageSizeSelector` could call an optional `onChange` function if it’s defined, otherwise it can just always default to using `setPageSize`. Similarly the analytics event could be another optional prop function. 

```
const handlePageSize = (pageSize: number) => {
	if (onChange) {
	    onChange(pageSize);
	} else {
	    setPageSize(pageSize);
        }
	sendEvent?.({name: "Change Page Size"});
}
```

That way there'd be no reason to import `usePageSizeSelector` across multiple places in the codebase, it could just be imported in `PageSizeSelector`

src/components/PageSizeSelector/index.tsx
```diff
@@ -0,0 +1,36 @@
+import styled from "@emotion/styled";
+import { Select, Option } from "@leafygreen-ui/select";
+import { PAGE_SIZES } from "constants/index";
+import usePageSizeSelector from "./usePageSizeSelector";
+
+interface Props {
+  value: number;
+  onChange?: (i: number) => void;
```
minnakt, line 8: nit. Since it seems like the `onChange` prop is always being called, I don't think it should be optional

cypress/integration/host/host_events.ts
```diff
@@ -11,11 +13,10 @@ describe("Host events", () => {
...
   it("host events display the correct text", () => {
     cy.visit(pathWithEvents);
-    cy.dataCy("host-event-table-page-size-selector").click();
-    cy.dataCy("host-event-table-page-size-selector-100").click();
-
-    const dataCyTableRows = "[data-cy=host-events-table] tr td:first-child";
-    cy.get(dataCyTableRows).should("have.length.of.at.most", 32);
+    clickOnPageSizeBtnAndAssertURLandTableSize(
+      100,
+      "[data-cy=host-events-table] tr td:first-child"
```
minnakt, line 18: nit. Similar to the other files, can you assign a variable `const dataCyTableRows = "[data-cy=host-events-table] tr td:first-child"` and use it here? I think it will be nice for when we replace the antd! :)

cypress/integration/host/host_events.ts
```diff
@@ -160,8 +161,11 @@ describe("Host events", () => {
       },
     ];
     cy.visit(pathWithEvents);
-    cy.dataCy("host-event-table-page-size-selector").click();
-    cy.dataCy("host-event-table-page-size-selector-100").click();
+    clickOnPageSizeBtnAndAssertURLandTableSize(
+      100,
+      "[data-cy=host-events-table] tr td:first-child"
```
minnakt, line 166: same comment

cypress/utils/index.ts
```diff
@@ -54,18 +54,14 @@ export const clickOnPageBtnAndAssertURLandTableResults = (
 /**
  * Assert limit query param and table length after clicking on a page size button
  * @param pageSize new page size
- * @param dataCyPageSizeSelectorToggle data-cy for page size dropdown toggle
- * @param dataCyPageSizeBtn data-cy for page size button to select page size
  * @param dataCyTableRows dat-cy for table rows
  */
 export const clickOnPageSizeBtnAndAssertURLandTableSize = (
```
minnakt, line 59: suggestion. You might also be able to use this in `job_logs.ts`. There is a test for changing the page size, but it's not as nice as this util function!

minnakt (approved):
LGTM!!! thanks for explain your reasoning to me 😇

src/components/PageSizeSelector/usePageSizeSelector.ts
sophstad: Got it, that makes sense. Looking ahead though, it does feel less than ideal that event names are magic strings used across the codebase. Do you think something like a string enum of all events we use could help with standardization going forward?

khelif96: All of our Analytics events are all [strongly typed](https://github.com/evergreen-ci/spruce/blob/e3b3a20574a54e6457bab69805c6eee9bf694e4b/src/analytics/version/useVersionAnalytics.ts#L12). So they aren't technically "magic strings" since they are all supported by typescripts type safety.
image

But standardizing some common ones for our often reused patterns like Table pagination could be useful. I'm totally up to ticket the work.

sophstad: Wow I totally didn't know about that typing 🤯 thanks for sharing!

PR #1281, opened 2022-05-19
EVG-16935 Allow viewing the project page on staging

malikchaya2:
[EVG-16935](https://jira.mongodb.org/browse/EVG-16935)

### Description 
This makes the project page viewable on staging instead of the coming soon page 

### Testing 
Tested manually on staging

### DIFF
--- a/src/pages/ProjectSettings.tsx
+++ b/src/pages/ProjectSettings.tsx
@@ -30,10 +30,10 @@ import { getTabTitle } from "./projectSettings/getTabTitle";
 import { ProjectSettingsTabs } from "./projectSettings/Tabs";
 import { ProjectType } from "./projectSettings/tabs/utils";
 
-const { isProduction } = environmentalVariables;
+const { isDevelopment } = environmentalVariables;
 const { validateObjectId } = validators;
 
-const disablePage = isProduction();
+const disablePage = !isDevelopment;
 
 export const ProjectSettings: React.VFC = () => {
   usePageTitle(`Project Settings`);
--- a/src/pages/ProjectSettings.tsx
+++ b/src/pages/ProjectSettings.tsx
@@ -33,7 +33,7 @@ import { ProjectType } from "./projectSettings/tabs/utils";
 const { isDevelopment } = environmentalVariables;
 const { validateObjectId } = validators;
 
-const disablePage = !isDevelopment;
+const disablePage = !isDevelopment();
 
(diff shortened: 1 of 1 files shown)


src/pages/ProjectSettings.tsx
```diff
@@ -30,10 +30,10 @@ import { getTabTitle } from "./projectSettings/getTabTitle";
 import { ProjectSettingsTabs } from "./projectSettings/Tabs";
 import { ProjectType } from "./projectSettings/tabs/utils";
 
-const { isProduction } = environmentalVariables;
+const { isDevelopment } = environmentalVariables;
 const { validateObjectId } = validators;
 
-const disablePage = isProduction();
+const disablePage = !isDevelopment;
```
sophstad, line 36: Since isDevelopment is a function I think `disablePage` would always evaluate to `false` which we're not ready for just yet!
```suggestion
const disablePage = !isDevelopment();
```

sophstad (changes requested):

src/pages/ProjectSettings.tsx
```diff
@@ -30,10 +30,10 @@ import { getTabTitle } from "./projectSettings/getTabTitle";
 import { ProjectSettingsTabs } from "./projectSettings/Tabs";
 import { ProjectType } from "./projectSettings/tabs/utils";
 
-const { isProduction } = environmentalVariables;
+const { isDevelopment } = environmentalVariables;
 const { validateObjectId } = validators;
 
-const disablePage = isProduction();
+const disablePage = !isDevelopment();
```
khelif96, line 36: `isDevelopment` will not allow this page to be viewable in staging. `isProduction` and `isDevelopment` only indicate if we have a compiled build or a build running on a dev server. I would recommend doing something akin to what we do with `isBeta` having an environmental variable that is added to the compile step.

khelif96 (changes requested):
It also occurred to me that this component performs a bunch of queries and initialization even if it's not going to be used. The `disablePage` fallback should live outside of this component.

sophstad:
> It also occurred to me that this component performs a bunch of queries and initialization even if it's not going to be used. The `disablePage` fallback should live outside of this component.

@khelif96 Given that this is just a stopgap to prevent users from viewing the page before we release project settings, I'd argue it's not worth the effort to change the implementation as the launch date is nearing. Good point about the `isDevelopment` implementation though -- I didn't think to test this on a local environment.

PR #1292, opened 2022-05-25
EVG-15798: Remove <PageTitle hasData />

SupaJoon:
[EVG-15798](https://jira.mongodb.org/browse/EVG-15798)

### Description 
Originally we were hiding the action buttons on the task page when the task status was null. These code changes make sure that we always show the action buttons by removing `<PageTitle hasData />`

### DIFF
--- a/src/components/TaskStatusBadge/index.tsx
+++ b/src/components/TaskStatusBadge/index.tsx
@@ -31,6 +31,10 @@ interface TaskStatusBadgeProps {
   status: string;
 }
 const TaskStatusBadge: React.VFC<TaskStatusBadgeProps> = ({ status }) => {
+  if (!status) {
+    return <></>;
+  }
+
   let displayStatus = getStatusBadgeCopy(status);
 
   if (taskStatusToCopy[status] === undefined) {
--- a/src/pages/Task.tsx
+++ b/src/pages/Task.tsx
@@ -56,14 +56,13 @@ export const Task = () => {
     displayName,
     patchNumber,
     priority,
-    status,
     annotation,
     latestExecution,
     versionMetadata,
     displayTask,
   } = task ?? {};
   const attributed = annotation?.issues?.length > 0;
-
+  const status = null;
   if (
     id === task?.id &&
     Number.isNaN(selectedExecution) &&
@@ -79,6 +78,7 @@ export const Task = () => {
   if (error) {
     return <PageDoesNotExist />;
   }
+  console.log(displayName, status);
   return (
     <PageWrapper>
       {task && (
@@ -90,7 +90,6 @@ export const Task = () => {
       )}
       <PageTitle
         loading={loading}
-        hasData={!!(displayName && status)}
         title={displayName}
         badge={
           <StyledBadgeWrapper>
--- a/src/pages/Task.tsx
+++ b/src/pages/Task.tsx
@@ -53,16 +53,17 @@ export const Task = () => {
 
   const { task, taskFiles } = data ?? {};
   const {
+    annotation,
     displayName,
+    displayTask,
+    latestExecution,
     patchNumber,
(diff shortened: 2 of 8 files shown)


src/pages/Task.tsx
```diff
@@ -56,14 +56,13 @@ export const Task = () => {
...
-    status,
     annotation,
     latestExecution,
     versionMetadata,
     displayTask,
   } = task ?? {};
   const attributed = annotation?.issues?.length > 0;
-
+  const status = null;
   if (
```
minnakt, line 66: I think passing `null` to the `TaskStatusBadge` results in an empty badge always, even if the task did have a proper status. The task shown below was successful, so ideally it should have the `SUCCEEDED` badge:

Screen Shot 2022-05-25 at 3 03 37 PM

Maybe instead we can edit `TaskStatusBadge` to return `null` if the status is `null` (or you could make it say something like "Unknown", whatever you think works best), otherwise return the normal badge?

src/utils/statuses/getStatusBadgeCopy.ts
```diff
@@ -1,7 +1,7 @@
 // &#8209; is a non-breaking hyphen used so that statuses with a hyphen do not word break when the page width shrinks
 export const getStatusBadgeCopy = (status: string) =>
```
minnakt, line 2: I think this function becomes a little confusing to read

It might be simpler if we just did an undefined check like `if (status) { // do this stuff } else { return null; }`

minnakt, line 2: Can you add a test for the case where `status` is `undefined` in the corresponding test file?

src/pages/Task.tsx
```diff
@@ -53,14 +53,14 @@ export const Task = () => {
 
   const { task, taskFiles } = data ?? {};
   const {
+    annotation,
     displayName,
+    displayTask,
+    latestExecution,
     patchNumber,
     priority,
     status,
```
minnakt, line 62: omg I love the alphabetization CHEF'S KISS

src/components/TaskStatusBadge/index.tsx
```diff
@@ -31,6 +31,10 @@ interface TaskStatusBadgeProps {
   status: string;
 }
 const TaskStatusBadge: React.VFC<TaskStatusBadgeProps> = ({ status }) => {
+  if (!status) {
+    return <></>;
+  }
+
   let displayStatus = getStatusBadgeCopy(status);
 
   if (taskStatusToCopy[status] === undefined) {
```
minnakt, line 40: We might be able to simplify `L34 ~ L46` into:

```
  if (!status || !taskStatusToCopy[status]) {
    const err = new Error(`Status '${status}' is not a valid task status`);
    reportError(err).warning();
    return null;
  }

  const displayStatus = getStatusBadgeCopy(taskStatusToCopy[status]);
```

minnakt (approved):
LGTM!!! 😇

src/components/TaskStatusBadge/index.tsx
SupaJoon: I'll like to display the status even if it's not official

src/utils/statuses/getStatusBadgeCopy.ts
```diff
@@ -1,7 +1,9 @@
...
   status
-    .split("-")
-    .map((s) => `${s[0].toLocaleUpperCase()}${s.slice(1)}`)
-    .join("-")
-    .replace("/-/g", "#8209;");
+    ? status
+        .split("-")
```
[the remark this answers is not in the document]
SupaJoon: I think this check is okay since an empty string will result in a runtime error for `s[0].toLocaleUpperCase()`

PR #1306, opened 2022-06-01
EVG-15181 Fix revert emails not being sent

khelif96:
[EVG-15181](https://jira.mongodb.org/browse/EVG-15181)

### Description 
Now tries to fetch the tagged commit before the currently checked out commit. Since we rely on git tags to determine what is deployed it should accurately track the commits between the two tags. 
Fixed some formatting now spruce deploy emails look like evergreen and cedar emails.
Added a hint for reverts.

### Screenshots
image

### Testing 
Checked out a previous tag commit that is not HEAD and attempted to send an email.

### DIFF
--- a/email.sh
+++ b/email.sh
@@ -22,19 +22,7 @@ then
     exit 1
 fi
 
-# Detect which version of sed we have available to format the email
-case "$OSTYPE" in
-  darwin*)
-    echo "OSX detected using gsed"
-    gsed -i ':a;N;$!ba;s/\n/<br \/>/g' body.txt
 
-  ;;
-  linux*)
-    echo "LINUX detected using sed"
-    sed -i ':a;N;$!ba;s/\n/<br \/>/g' body.txt
-  ;;
-  *)        echo "unknown: $OSTYPE";;
-esac
 
 # Determine which verson of evergreen is available and use that
 if ! [ -x "$(command -v evergreen)" ]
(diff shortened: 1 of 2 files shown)


email.sh
```diff
@@ -52,6 +43,13 @@ else
   EVERGREEN=evergreen
 fi
 
+# Fetch previous release tag and get the commits since that tag
+CURRENT_COMMIT_HASH=$(git rev-parse --short HEAD)
+PREVIOUS_TAG=$(git describe --abbrev=0 $CURRENT_COMMIT_HASH\^)
+# get all commits since the previous tag
+git log --no-merges $PREVIOUS_TAG..$CURRENT_COMMIT_HASH --pretty="%h %s" >> body.txt
```
bsamek, line 50: nit: I think this should be `>`, not `>>` (on the risk that somehow there's another body.txt, we should overwrite it)

bsamek, line 50: Should we really include these commits in the case of a revert?

E.g., Imagine we deploy commit C, then a few days later B, then a few days later A. We then discover we need to revert A. When we deploy B, we include the commits between C and B. Would that be confusing? That suggests these are _newly_ deployed commits, when in fact they've been in prod the whole time.

bsamek (changes requested):

email.sh
khelif96: Good point i updated the revert email to only show the currently deployed hash.
It uses the presence of a non 0 execution to trigger the revert case. 
image

bsamek (approved):

PR #1308, opened 2022-06-02
EVG-16962 Add breadcrumbs for auth checks

khelif96:
[EVG-15962](https://jira.mongodb.org/browse/EVG-15962)

### Description 
It may not be possible to tie an errored network request to a specific line of code through a stack trace. Graphql requests are technically dispatched by the Apollo client provider not from a useQuery function. (useQuery simply tells apollo provider that it wants some data and the Apollo provider handles the network request. (It does things like managing retries, updating the cache, etc). So the stack trace would never link to a line since the error does not technically originate from a given line in our code base.

Since we already leave breadcrumbs for Graphql requests i think it's fine to leave as is.

I did find a bug with how we were error checking requests for a breadcrumb they wouldn't say "Error" on the breadcrumb.

### DIFF
--- a/src/context/auth.tsx
+++ b/src/context/auth.tsx
@@ -1,6 +1,7 @@
 import React, { useContext, useReducer } from "react";
 import axios from "axios";
 import { environmentalVariables } from "utils";
+import { leaveBreadcrumb } from "utils/errorReporting";
 
 const { getLoginDomain } = environmentalVariables;
 interface AuthState {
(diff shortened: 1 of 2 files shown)


src/utils/request.ts
```diff
@@ -11,6 +11,7 @@ export const post = async (
   options: optionsType = {}
 ) => {
   try {
+    leaveBreadcrumb(`POST ${url}`, { body }, "request");
```
SupaJoon, line 14: In `authenticateIfSuccessfulLink` we call this function after we get a response. Why not do the same for `post`?

SupaJoon (approved):

src/utils/request.ts
khelif96: image

It happens automatically. 

Come to think of it this extra logging is probably a bad idea for post. We rarely use post but in the situations that we do we either send auth data or call some rest endpoint. Logging auth data including emails and passwords is a bad idea. (Although in this case it never gets hit in prod since we use okta).  I'll remove the POST logging altogether.

PR #1316, opened 2022-06-09
EVG-16805 Vite ⚡ 

khelif96:
[EVG-16805](https://jira.mongodb.org/browse/EVG-16805)

### Description 
* Replaced webpack config with vite
* Typescript checking now happens async in a separate thread so the web server is faster
* React Fast refresh. (Instant component updates when files change)
* Storybook stories are no longer included in type checking process. 
* Now uses a production build of spruce for cypress tests. 
  * This causes the tests to be 1/3 faster since there is no more dev server overhead
  * More representative of what our end users see
* Dev server start up times are approximately 1s instead of a few minutes
* Production builds take half the time
* Switch bundle analysis to use [rollup-plugin-analyzer ](https://www.npmjs.com/package/rollup-plugin-analyzer)
   * This gives us nicer charts and also incorporates it as an automatic step in the build process.
* Refactored our environmental variables.
   * Removed some unused ones from each environment.
   * Renamed them so our use cases are more clear
   * Refactored the accessor functions so it is more clear on their use cases.
   * Updated `.cmdrc.json` will be shared on slack.
* Created a custom http server to serve up and handle proxying the production bundle and any requests.
   * This behaves similarly to how production is hosted and served
* Storybook no longer works.
  * Storybook has a dependency on the existing webpack config but since we are no longer supporting it it will not run. Once we update storybook to the latest release https://jira.mongodb.org/browse/EVG-16219 we can incorporate our vite tooling. 
* Created rollup plugin to handle injecting variables into built html page. (This was a CRA feature)

**You may notice some changes to some tests  and you may ask why were some tests changed in a PR that updates the build system shouldn't functionality stay the same?** 
My only explanation is that the failing tests were a result of our own success. After migrating the app to use a production build of spruce some tests started to fail. Not because any functionality changed but because the pages were loading faster which was causing cypress to start making assertions before the data for the page had fully loaded. 
Previously parts of the page would load along side the data that was being rendered so by the time cypress started making its first assertion it had all of the data it needed. 
Now that we are serving up a production build the page loads up really quickly and cypress takes that as its queue to start running the assertions.
The fix was to add a simple assertion that checks if the page has fully loaded yet. (Which is the recommend pattern by cypress)

### Screenshots
Dev Server starting up
https://user-images.githubusercontent.com/4605522/172954532-81f8cab9-d61e-4700-983e-9b248d40c7e6.mov

Fast Refresh

https://user-images.githubusercontent.com/4605522/172954969-0651fc2d-887b-4c70-93e9-228be89d6c88.mov

### Testing 
All existing tests


src/pages/spawn/spawnHost/SpawnHostCard.stories.tsx
```diff
@@ -1,9 +1,7 @@
-import React from "react";
-import { MemoryRouter as Router } from "react-router-dom";
 import { SpawnHostCard } from "pages/spawn/spawnHost/SpawnHostCard";
 
 const host = {
-  expiration: "2020-08-21T18:00:07Z",
+  expiration: new Date("2020-08-21T18:00:07Z"),
```
SupaJoon, line 4: Is there a reason to change this from a string to an object? I think the SpawnHostCard was rendering correctly before.

src/utils/environmentalVariables.ts
```diff
@@ -1,53 +1,116 @@
...
 
+/**
+ * `getSpruceURL()` - Get the SPRUCE URL from the environment variables
+ * @returns {string} - Returns the Spruce URL
+ */
 export const getSpruceURL: () => string = () =>
   process.env.REACT_APP_SPRUCE_URL;
 
+/**
+ * `isDevelopment()` indicates if the current environment is a local development environment.
```
SupaJoon, line 43: nit: isDevelopmentBuild()

config/leafygreen-ui/emotion/index.ts
```diff
@@ -0,0 +1,29 @@
+// Shim to prevent LG's emotion package https://github.com/mongodb/leafygreen-ui/tree/main/packages/emotion from pulling in SSR dependencies.
```
SupaJoon, line 1: As there another link with more specific info related to SSR deps & shimming?

vite.config.js
```diff
@@ -0,0 +1,137 @@
...
+    alias: {
+      "@leafygreen-ui/emotion": path.resolve(
+        __dirname,
+        "./config/leafygreen-ui/emotion"
+      ),
+      ...absolutePathAliasMap,
+    },
+    extensions: [".mjs", ".js", ".ts", ".jsx", ".tsx", ".json"],
+  },
+
```
SupaJoon, line 82:
```suggestion
```

vite.config.js
```diff
@@ -0,0 +1,137 @@
...
+export default defineConfig({
+  server: {
+    proxy: {
+      "/graphql": {
+        target: "http://localhost:9090/graphql/query",
+        changeOrigin: true,
+      },
+    },
+  },
+
```
SupaJoon, line 37:
```suggestion
```

SupaJoon (approved):

.evergreen.yml
```diff
@@ -359,7 +339,8 @@ tasks:
         target: local-evergreen
     - func: yarn-install
     - func: sym-link
-    - func: yarn-start
+    - func: yarn-build
+    - func: yarn-serve
```
sophstad, line 343: 🙌

sophstad (approved):

tsconfig.json
```diff
@@ -24,15 +24,27 @@
...
   },
   "exclude": [
-    "node_modules"
+    "node_modules",
+    "vite.config.ts"
   ],
   "include": [
     "src",
     "test_utils/index.tsx",
     "test_utils/custom-queries.ts"
```
minnakt, line 43: Not a part of this PR but you might also want to change this to  `test_utils/custom-queries.js`

minnakt (approved):

tsconfig.node.json
```diff
@@ -0,0 +1,10 @@
+{
+    "compilerOptions": {
+        "composite": true,
+        "module": "esnext",
+        "moduleResolution": "node"
+    },
+    "include": [
+        "vite.config.ts"
+    ]
```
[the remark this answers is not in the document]
khelif96: Good catch i renamed to `vite.config.ts`. What linter is that i didn't get those errors?

tsconfig.json
khelif96: Great catch! updated the file type.

src/pages/spawn/spawnHost/SpawnHostCard.stories.tsx
khelif96: image
Uptime and Expiration do expect a Date

PR #1317, opened 2022-06-10
Update package.json engines field

sophstad:
### Description 
- According to [BUILD-14325](https://jira.mongodb.org/browse/BUILD-14325), evergreen uses Node v16.13.0 to test and build spruce. We also use packages with dependencies of Node 16+, so update our engines field to ensure local processes use the correct version

### DIFF
--- a/package.json
+++ b/package.json
@@ -251,7 +251,7 @@
     "worker-loader": "^3.0.7"
   },
   "engines": {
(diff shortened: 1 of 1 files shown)


khelif96 (approved):

sophstad:
I guess flakey tests are not fixed :( will reopen EVG-17049

PR #1318, opened 2022-06-10
EVG-16933: Add project settings link

sophstad:
EVG-16933

### Description 
- Add project settings link, which is hidden on prod but visible on staging, beta, and local envs. I opted to rename from "Projects" to "Project Settings" because I think this more accurately captures the pages within, but let me know if there's a good reason not to.
- When EVG-17059 is completed and project settings is released, the legacy "Projects" link will be removed and the Spruce link will be made visible on prod.

### Screenshots
image

### DIFF
--- a/src/components/Header/AuxiliaryDropdown.tsx
+++ b/src/components/Header/AuxiliaryDropdown.tsx
@@ -13,12 +13,7 @@ import { GET_SPRUCE_CONFIG } from "gql/queries";
 import { environmentalVariables } from "utils";
 import { Dropdown } from "./NavDropdown";
 
-const { getUiUrl, isBeta, isDevelopment, isStaging } = environmentalVariables;
-
-// TODO: Remove in EVG-17059
-// Project Settings should only be disabled when deployed to spruce.mongodb.com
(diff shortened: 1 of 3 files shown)


src/components/Header/AuxiliaryDropdown.tsx
```diff
@@ -3,13 +3,22 @@ import Cookies from "js-cookie";
...
 import { environmentalVariables } from "utils";
 import { Dropdown } from "./NavDropdown";
 
-const { getUiUrl } = environmentalVariables;
+const { getUiUrl, isBeta, isDevelopment, isStaging } = environmentalVariables;
+
+// TODO: Remove in EVG-17059
+// Project Settings should only be disabled when deployed to spruce.mongodb.com
+// Enable when running local dev server, or when deployed to beta or staging
+const showProjectSettings = isDevelopment() || isBeta() || isStaging();
```
khelif96, line 21: Can we package this within the environmentalVariables file? I think it's a common enough pattern that we will use down the line. I already have one use case in mind.

khelif96 (approved):

PR #1319, opened 2022-06-13
EVG-15881: Update date format

SupaJoon:
[EVG-15881](https://jira.mongodb.org/browse/EVG-15881)

### Description 
I didn't change the format for [LogMessageLine](https://github.com/evergreen-ci/spruce/blob/ab6d3de345e95f93a4ae124c713f342d42410b58/src/pages/task/taskTabs/logs/logTypes/LogMessageLine.tsx#L17) since that format is shown in the logs. For example, the timestamp for the logs shown on this [page](https://spruce.mongodb.com/task/spruce_ubuntu1604_test_patch_ab6d3de345e95f93a4ae124c713f342d42410b58_62a7591a3e8e860fc104b130_22_06_13_15_35_05/logs?execution=0&sortBy=STATUS&sortDir=ASC) come from the TaskLogs query.

### DIFF
--- a/src/components/CommitChartLabel/index.tsx
+++ b/src/components/CommitChartLabel/index.tsx
@@ -8,12 +8,10 @@ import { size, zIndex } from "constants/tokens";
 import { UpstreamProjectFragment } from "gql/generated/types";
 import { useSpruceConfig } from "hooks";
 import { ProjectTriggerLevel } from "types/triggers";
-import { string } from "utils";
-import { shortenGithash } from "utils/string";
+import { getDateCopy, shortenGithash } from "utils/string";
 import { jiraLinkify } from "utils/string/jiraLinkify";
 
 const { gray } = uiColors;
-const { shortDate } = string;
 const MAX_CHAR = 40;
 interface Props {
   githash: string;
@@ -58,7 +56,7 @@ const CommitChartLabel: React.VFC<Props> = ({
         >
           {shortenGithash(githash)}
         </StyledRouterLink>{" "}
-        <b>{shortDate(createDate)}</b>
+        <b>{getDateCopy(createDate)}</b>
       </LabelText>
       {upstreamProject && (
         <LabelText>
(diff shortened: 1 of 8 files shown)


src/pages/commits/ActiveCommits/__snapshots__/ProjectHealth.stories.storyshot
```diff
@@ -499,7 +499,7 @@ exports[`storybook Storyshots Project Health Page Actual Waterfall Page 1`] = `
             </a>
              
             <b>
-              7/13/21 2:51 PM
+              Jul 13, 2021, 2:51:30 PM
```
khelif96, line 502: Are seconds needed? I wonder if we can get rid of them?

src/components/CommitChartLabel/index.tsx
```diff
@@ -58,7 +56,7 @@ const CommitChartLabel: React.VFC<Props> = ({
         >
           {shortenGithash(githash)}
         </StyledRouterLink>{" "}
-        <b>{shortDate(createDate)}</b>
+        <b>{getDateCopy(createDate)}</b>
```
khelif96, line 59: why do we format with the user timezone below but not here?

khelif96 (commented):
I think we should be consistent about timezones and apply the timezone formatting to all of the dates.

src/components/CommitChartLabel/index.tsx
khelif96: i'm inclined to advocate for the opposite since it makes it more confusing if dates are in different timezones through out the application.

SupaJoon: Let's discuss in the design sync before moving forward

src/pages/task/executionDropdown/ExecutionSelector.tsx
```diff
@@ -60,8 +61,9 @@ export const ExecutionSelect: React.VFC<ExecutionSelectProps> = ({
             <StyledTaskStatusIcon status={singleExecution.status} />
             <StyledP1>
               Execution {executionAsDisplay(singleExecution.execution)} -{" "}
-              {shortDate(
-                singleExecution.activatedTime ?? singleExecution.ingestTime
+              {getDateCopy(
+                singleExecution.activatedTime ?? singleExecution.ingestTime,
+                { tz }
```
khelif96, line 66: One of these values ends up undefined  when the page is loading and causes spruce to crash. https://dashboard.cypress.io/projects/yshv48/runs/3502/overview/6268fcc8-01c6-4c18-b402-8e1f7a3a03a7/video

khelif96 (approved):
Everything else LGTM except the failing tests.

PR #1320, opened 2022-06-13
EVG-16550 remove host pid set events

ybrill:
[EVG-16550](https://jira.mongodb.org/browse/EVG-16550)

### Description 
Remove references to the `HOST_TASK_PID_SET` event.

### Evergreen PR 
https://github.com/evergreen-ci/evergreen/pull/5714

### DIFF
--- a/cypress/integration/host/host_events.ts
+++ b/cypress/integration/host/host_events.ts
@@ -87,10 +87,6 @@ describe("Host events", () => {
         text: "Marked as provisioned",
       },
       {
-        hostType: "host-task-pid-set",
-        text: "PID of running task set to 123",
-      },
-      {
         hostType: "host-monitor-flag",
         text: "Flagged for termination because: unreachable",
       },
--- a/src/pages/host/getHostEventString.tsx
+++ b/src/pages/host/getHostEventString.tsx
@@ -233,13 +233,6 @@ export const getHostEventString = (
           )
         </div>
       );
-    case HostEvent.HostTaskPIDSet:
-      return (
-        <div data-cy="host-task-pid-set">
-          {" "}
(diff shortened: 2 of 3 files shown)


malikchaya2 (approved):

khelif96 (commented):
Will it still be possible to hit a host that has this event on an older host from before this field was removed?

ybrill:
@khelif96 that's a good point.

It _is_ possible today, if someone loads a host from before https://github.com/evergreen-ci/evergreen/commit/9c9ad23589805c8a7c8c6b327f014b341156d4a3. Once I TTL the host events it won't be possible anymore.

I was figuring it wasn't worth breaking it up and coming back to it since it's unlikely (the hosts are > 4 years old) and it wouldn't **break**, per se, if someone did load one. It would just show the raw string, I think. 

But if you feel strongly we can hold off on this.

khelif96:
> @khelif96 that's a good point.
> 
> It _is_ possible today, if someone loads a host from before [evergreen-ci/evergreen@9c9ad23](https://github.com/evergreen-ci/evergreen/commit/9c9ad23589805c8a7c8c6b327f014b341156d4a3). Once I TTL the host events it won't be possible anymore.
> 
> I was figuring it wasn't worth breaking it up and coming back to it since it's unlikely (the hosts are > 4 years old) and it wouldn't **break**, per se, if someone did load one. It would just show the raw string, I think.
> 
> But if you feel strongly we can hold off on this.

Oh wow i hadn't realized those events are 4+ years old. This is fine with me.

khelif96 (approved):

khelif96:
the failing codegen test is safe to merge. It just means someone hasn't ran `yarn codegen` since merging in main.

PR #1321, opened 2022-06-13
EVG-16736: Update config files related to GraphQL

minnakt:
[EVG-16736](https://jira.mongodb.org/browse/EVG-16736)

### Description 
The schema files in the backend are going to change, which means that the codegen and test config files need to be updated. This PR will not pass the tests until the corresponding Evergreen PR is merged.

#### Instructions to symlink to the new schema

Run this command in your Spruce directory, which will create a symlink to the schema folder in evergreen:
```
ln -s <path to schema folder in evergreen repo> sdlschema
```

For example the evergreen repo is in my root `go` folder, so this is the command I would run:
```
ln -s ~/go/src/github.com/evergreen-ci/evergreen/graphql/schema sdlschema
```

### Screenshots
- No visible changes

### Testing 
- Ran tests using set-module [here](https://spruce.mongodb.com/version/62a7d31c3066156016432b49/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

### Evergreen PR 
- evergreen-ci/evergreen/pull/5717


.evergreen.yml
```diff
@@ -65,7 +65,7 @@ functions:
     params:
       working_dir: spruce
       script: |
-        ln -s evergreen/graphql/schema.graphql sdlschema.graphql
+        ln -s evergreen/graphql/schema sdlschema
```
khelif96, line 68: Can you also make this change to the read me.

khelif96: and possibly the gitignore if it needs it.

khelif96 (approved):
Thanks for doing this. Once this is merged can you post on the main evergreen channel so people know to make this change.

Also don't forget to make this change in the new hire guide

sophstad (changes requested):
It looks like some of the README instructions should be changed slightly:

> create a symlink from the schema.graphql file from evergreen with the spruce folder using ln -s path-to-evergreen-schema.graphql sdlschema.graphql

It looks like from this command shouldn't include a file extension anymore. This is mentioned in 2 places in the README!

PR #1323, opened 2022-06-14
EVG-17063: Check for existence before calling length on publicKeys

minnakt:
[EVG-17063](https://jira.mongodb.org/browse/EVG-17063)

### Description 
On Monday there was a [Bugsnag alert](https://app.bugsnag.com/mongodb/evergreen/errors/62a74e195a5ff1000892efb7?event_id=62a74e1900950d2c719d0000&i=sk&m=nw) in production about calling `.length` on an `undefined` variable `publicKeys`.

I think this was somewhat of a fluke because it happened right as I deployed and the [function](https://github.com/evergreen-ci/evergreen/blob/62820234635fd95a9368090507231651518b11ad/graphql/util.go#L556-L568) in Evergreen seems to return `publicKeys = []` if they don't exist. However I did modify the files so that this error wouldn't occur again.

I made a few other unrelated changes (see comments) because I noticed some other problems as I was testing that the public keys form still worked.


src/pages/spawn/spawnHost/SpawnHostModal.tsx
```diff
@@ -139,23 +138,7 @@ export const SpawnHostModal: React.VFC<SpawnHostModalProps> = ({
...
-    }
-    if (publicKeys && publicKeys.length) {
-      dispatch({
-        type: "editPublicKey",
-        publicKey: publicKeys[0],
-        savePublicKey: false,
-      });
-    }
-  }, [awsRegions, userAwsRegion, publicKeys, dispatch]);
+  }, [visible, awsRegions, publicKeys]); // eslint-disable-line react-hooks/exhaustive-deps
```
minnakt: I deleted the hook below and combined it with this one here. The information was pretty much the same so I figured it would be okay

(The deleted hook was introduced in evergreen-ci/spruce/pull/525 and appears to be specifically for spawning hosts from tasks. However I didn't really notice any difference after combining them.)

src/pages/spawn/spawnHost/SpawnHostModal.tsx
```diff
@@ -139,23 +138,7 @@ export const SpawnHostModal: React.VFC<SpawnHostModalProps> = ({
...
-    }
-    if (publicKeys && publicKeys.length) {
-      dispatch({
-        type: "editPublicKey",
-        publicKey: publicKeys[0],
-        savePublicKey: false,
-      });
-    }
-  }, [awsRegions, userAwsRegion, publicKeys, dispatch]);
+  }, [visible, userAwsRegion, awsRegions, publicKeys]); // eslint-disable-line react-hooks/exhaustive-deps
```
minnakt, line 141: I deleted the hook below and combined it with this one here. The information was pretty much the same so I figured it would be okay

(The deleted hook was introduced in evergreen-ci/spruce/pull/525 and appears to be specifically for spawning hosts from tasks. However I didn't really notice any difference after combining them.)

src/pages/spawn/spawnHost/SpawnHostModal.tsx
```diff
@@ -116,9 +116,8 @@ export const SpawnHostModal: React.VFC<SpawnHostModalProps> = ({
   const awsRegions = awsData?.awsRegions;
   const volumes = volumesData?.myVolumes ?? [];
 
-  // When the modal is opened, reset to the "default" state of the Modal contents.
+  // When the modal is opened, pre-fill parts of the form (AWS region, public key, and expiration).
   useEffect(() => {
```
minnakt, line 120: The `dispatch({ type: "reset" });` was causing a race condition with the [`useEffect`](https://github.com/evergreen-ci/spruce/blob/main/src/pages/spawn/spawnHost/spawnHostModal/SetupScriptForm.tsx#L53-L62) here when spawning a host from a task. We would initially correctly pre-fill the distro based on the URL params, but then we would end up resetting it to be empty.

Besides, I think it makes more sense to reset the form when the modal is closed, not when it is opened.

khelif96 (approved):

PR #1329, opened 2022-06-16
EVG-13640: Upgrade to Cypress v10

SupaJoon:
[EVG-13640](https://jira.mongodb.org/browse/EVG-13640)

### Description 
Upgrade to cypress 10 using this [guide](https://docs.cypress.io/guides/references/migration-guide) and use vite instead of webpack for cypress.


src/pages/configurePatch/ConfigurePatchCore.tsx
```diff
@@ -93,7 +93,7 @@ export const ConfigurePatchCore: React.VFC<Props> = ({ patch }) => {
     onCompleted(data) {
       const { schedulePatch: scheduledPatch } = data;
       dispatchToast.success("Successfully scheduled the patch");
-      navigate(getVersionRoute(scheduledPatch.versionFull.id));
+      navigate(getVersionRoute(scheduledPatch.versionFull?.id));
```
khelif96, line 96: Curious why this was added? I assume that it was necessary to fix a failing test as a result of this upgrade. 
Does this mean https://github.com/evergreen-ci/spruce/pull/1329/files#diff-6deb8e73e58a4a261d3a74d7cf885f9ee14eeefcdd1b156deb268c99e141eb1cR524 maybe didn't work the same way and this was added to fix an error?

cypress/integration/version/unscheduled_patch/configure_patch.ts
```diff
@@ -521,7 +521,7 @@ describe("Configure Patch Page", () => {
       cy.dataCy(`patch-name-input`).as("patchNameInput").clear().type(val);
       cy.dataCy("task-checkbox").first().check({ force: true });
       // TODO: Replace this with cy.intercept() once we upgrade to > Cypress 6.0.0
```
khelif96, line 523: nit. the comment is not necessary.

cypress/integration/version/unscheduled_patch/configure_patch.ts
```diff
@@ -516,14 +516,18 @@ describe("Configure Patch Page", () => {
...
-      // TODO: Replace this with cy.intercept() once we upgrade to > Cypress 6.0.0
-      cy.route2("/graphql/query", (req) => {
+      cy.intercept("/graphql/query", (req) => {
         req.reply((res) => {
-          res.body = mockedSuccessfulConfigureResponse;
+          res.body = {
+            data: {
+              schedulePatch: { versionFull: { id: unactivatedPatchId } },
+            },
+            errors: null,
```
khelif96, line 529: nit. use a variable for mockedSuccessfulConfigureResponse since we do the same thing for the error response.

package.json
```diff
@@ -173,7 +173,7 @@
     "camelcase": "^6.1.0",
     "colors": "1.4.0",
     "css-loader": "4.3.0",
-    "cypress": "5.2.0",
+    "cypress": "10.1.0",
```
khelif96, line 176: nit. can we upgrade to 10.2 since thats the latest as of now? It offers performance improvements on M1 machines and some of our team members are starting to get them.

khelif96 (commented):
This LGTM. I do think the failing tests should be addressed in this PR. The failures appear to be introduced in this PR and don't appear to be related to the retries. [Previous Cypress runs](https://dashboard.cypress.io/projects/yshv48/runs/3600/overview) don't indicate that the failing tests in this pr are flaky or affected by the retry change. 

image

The tests are also passing on main so i would be reluctant to introduce more flaky tests.

SupaJoon:
> This LGTM. I do think the failing tests should be addressed in this PR. The failures appear to be introduced in this PR and don't appear to be related to the retries. [Previous Cypress runs](https://dashboard.cypress.io/projects/yshv48/runs/3600/overview) don't indicate that the failing tests in this pr are flaky or affected by the retry change.
> 
> image
> 
> The tests are also passing on main so i would be reluctant to introduce more flaky tests.

I think addressing the flaking tests caused by the framework upgrade should happen in a separate PR since it could take a lot more experimentation/patches and observation to understand when the false positives are actually resolved.
I'm inclined to turn on retries for this PR to collect more data on flaking tests and address them later.

khelif96 (commented):
I took a look at some of the test failures and I made some notes about how we could resolve them. 

Cypress made some changes that i think were missed in this pr and could potentially be contributing to some of these weird test flakiness.

[`cypress.cookies.preserveOnce`](https://github.com/evergreen-ci/spruce/blob/43df88acb001a1815eb457e65bb885d284224f95/cypress/support/commands.ts#L31) was [deprecated](https://docs.cypress.io/guides/references/changelog#9-7-0). In favor of [`cy.session`](https://docs.cypress.io/api/commands/session)

`cy.intercept` has changed in a few other ways as well. I recommend reading some of the guides here. 
https://docs.cypress.io/guides/guides/network-requests#What-you-ll-learn
https://docs.cypress.io/guides/guides/network-requests#Use-Server-Responses
https://docs.cypress.io/guides/end-to-end-testing/working-with-graphql#What-you-ll-learn
I got the `configure_patch.ts` tests working successfully by using some of the suggested helper methods for working with graphql. 

Taking a look at some of the other test failures in your patch they do seem resolvable using a few well placed assertions.

The change in [6.1.0](https://docs.cypress.io/guides/references/changelog#6-1-0) does look like it could be a major contributing factor to some of the scrolling out of view issues we saw. 

I think it would make sense to switch the browser the tests run in to chrome instead of the default electron. Our users as well as our selves use chrome it could help with troubleshooting the bugs we have encountered.

It might be a good idea to bump the [default viewport size](https://docs.cypress.io/api/commands/viewport) cypress uses to be more in line with what the average engineer would have access to. (A standard macbook screen 13 inch would be a good one.) What do you think?

Thank you for taking this on and sorry it has dragged on so much. In the future we should make it a priority to keep our dependencies up to date more often.  So we can avoid situations like this.

SupaJoon:
> I took a look at some of the test failures and I made some notes about how we could resolve them.
> 
> Cypress made some changes that i think were missed in this pr and could potentially be contributing to some of these weird test flakiness.
> 
> [`cypress.cookies.preserveOnce`](https://github.com/evergreen-ci/spruce/blob/43df88acb001a1815eb457e65bb885d284224f95/cypress/support/commands.ts#L31) was [deprecated](https://docs.cypress.io/guides/references/changelog#9-7-0). In favor of [`cy.session`](https://docs.cypress.io/api/commands/session)
> 
> `cy.intercept` has changed in a few other ways as well. I recommend reading some of the guides here. https://docs.cypress.io/guides/guides/network-requests#What-you-ll-learn https://docs.cypress.io/guides/guides/network-requests#Use-Server-Responses https://docs.cypress.io/guides/end-to-end-testing/working-with-graphql#What-you-ll-learn I got the `configure_patch.ts` tests working successfully by using some of the suggested helper methods for working with graphql.
> 
> Taking a look at some of the other test failures in your patch they do seem resolvable using a few well placed assertions.
> 
> The change in [6.1.0](https://docs.cypress.io/guides/references/changelog#6-1-0) does look like it could be a major contributing factor to some of the scrolling out of view issues we saw.
> 
> I think it would make sense to switch the browser the tests run in to chrome instead of the default electron. Our users as well as our selves use chrome it could help with troubleshooting the bugs we have encountered.
> 
> It might be a good idea to bump the [default viewport size](https://docs.cypress.io/api/commands/viewport) cypress uses to be more in line with what the average engineer would have access to. (A standard macbook screen 13 inch would be a good one.) What do you think?
> 
> Thank you for taking this on and sorry it has dragged on so much. In the future we should make it a priority to keep our dependencies up to date more often. So we can avoid situations like this.

I Implemented all of these suggestions except replacing preserveOnce with session. That would require a large refactor since enabling the feature [causes the page to be cleared at the beginning of each test](https://docs.cypress.io/api/commands/session#Syntax).

cypress/integration/version/action_buttons.ts
```diff
@@ -88,7 +88,7 @@ describe("Action Buttons", () => {
         cy.visit(versionPath(mainlineCommit));
       });
       beforeEach(() => {
-        cy.dataCy("ellipsis-btn").click();
```
minnakt, line 91: Possibly don't need `force: true` here either

cypress/integration/projectSettings/project_settings.ts
```diff
@@ -376,9 +376,12 @@ describe("Project Settings when not defaulting to repo", () => {
     it("Should show three populated fields when navigating back from another page", () => {
       cy.dataCy("navitem-access").click();
       cy.dataCy("navitem-variables").click();
-      cy.dataCy("var-name-input").eq(0).should("have.value", "admin_var");
-      cy.dataCy("var-name-input").eq(1).should("have.value", "sample_name");
-      cy.dataCy("var-name-input").eq(2).should("have.value", "sample_name_2");
+      cy.dataCy("var-name-input").should("be.visible");
+      cy.dataCy("var-name-input")
+        .should("be.visible")
+        .and(($input) => {
```
minnakt, line 382: is it possible to combine these assertions since we have the same assertion twice?

cypress/integration/version/unscheduled_patch/configure_patch.ts
```diff
@@ -1,14 +1,23 @@
...
-
 describe("Configure Patch Page", () => {
   before(() => {
     cy.login();
   });
   beforeEach(() => {
     cy.preserveCookies();
+    cy.intercept("POST", GQL_URL, (req) => {
+      aliasQuery(req, "SchedulePatch");
+    });
```
minnakt, line 20: Not sure if you need this, I removed it and it still ran okay 🤔

minnakt (commented):
Wow the new Cypress interface is sleek 🤯

cypress/integration/spawn/host.ts
```diff
@@ -109,7 +109,10 @@ describe("Navigating to Spawn Host page", () => {
       it("Should disable 'Never expire' checkbox when max number of unexpirable hosts is met (2)", () => {
         cy.visit("/spawn/host");
         cy.contains("Spawn a host").click();
-        cy.dataCy("distro-input").click().type("{downarrow}").type("{enter}");
+        cy.dataCy("distro-input")
+          .click()
+          .type("{downarrow}", { force: true })
+          .type("{enter}", { force: true });
```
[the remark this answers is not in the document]
SupaJoon: {force: true} resolves this [test failure](https://dashboard.cypress.io/projects/yshv48/runs/3517/overview/668222d1-dc58-4c84-8bd4-307c677838fe) from [e4947f5](https://github.com/evergreen-ci/spruce/pull/1329/commits/e4947f50cb5835de9524245522ce230ecb2855c4)

cypress/integration/task/annotations.ts
```diff
@@ -40,7 +40,7 @@ describe("Task Annotation Tab", () => {
     cy.get(suspectedIssuesTable).should("have.length", 3);
 
     // add a ticket
-    cy.dataCy("add-suspected-issue-button").click();
+    cy.dataCy("add-suspected-issue-button").click({ force: true });
```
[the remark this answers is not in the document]
SupaJoon: `force:true` resolves this [test failure](https://dashboard.cypress.io/projects/yshv48/runs/845014d0-c316-4229-8db7-9481bcb9977c/test-results/19d825a2-6627-4bd0-8af9-b8355ace275e) from [1048b3f](https://github.com/evergreen-ci/spruce/pull/1329/commits/1048b3fe2c771f3d27c3afa0e292218ea5c1d03d)

cypress/integration/nav_bar.ts
```diff
@@ -15,7 +15,6 @@ describe("Nav Bar", () => {
     cy.login();
   });
   beforeEach(() => {
-    cy.server();
```
minnakt, line 18: You read my mind 😎

minnakt (approved):
It looks awesome :]

Overall LGTM, though I’d want @khelif96 to also take a look before merge

cypress/utils/mockErrorResponse.ts
```diff
@@ -16,7 +16,7 @@ export const mockErrorResponse = ({
   errorCode = "INTERNAL_SERVER_ERROR",
   path = "i am a path", // the name of the query
 }: Args) => {
-  cy.route2("/graphql/query", (req) => {
+  cy.intercept("/graphql/query", (req) => {
```
khelif96, line 19: I forgot about this method and i see that we use it in a few places. Could we update everywhere we use `cy.intercept` to use this method. Could we also update this method to behave like the intercepts defined in `configure_patch.ts`? So we have more control over what requests get intercepted.

cypress.config.ts
```diff
@@ -0,0 +1,17 @@
...
+    projectId: "yshv48",
+    reporterOptions: {
+      mochaFile: "bin/cypress/junit-[hash].xml",
+    },
+    defaultCommandTimeout: 8000,
+    supportFile: "cypress/support/e2e.ts",
+    specPattern: "cypress/integration/**/*.ts",
+    retries: {
+      runMode: 4,
+    },
```
khelif96, line 15: I'm afraid these just mask the flaky tests instead of fixing them.

khelif96 (changes requested):
Great work so far! I left some comments on how we could migrate to chrome. Cypress maintains a directory of links to chrome images we can use https://chromium.cypress.io/

khelif96:
Taking a look at how much slower the e2e_task has gotten after updating and looking at the memory usage for the task it looks like cypress v10 is [allot more expensive to run.](https://evergreen.mongodb.com/task_log_raw/spruce_ubuntu1604_e2e_test_patch_d8c2bd8b972ee96dcbbe0a26cd4bbfb14aaef726_62c89be51e2d171761fba3d6_22_07_08_21_04_38/0?type=S#L2237) (Often using over 100% of cpu)
This is especially clear when watching the videos and observing how much they stutter. 

I tried experimenting with upgrading the host to a larger one, `ubuntu1604-small` to `ubuntu1604-large` and the run times look allot better. Nearly twice as fast. https://spruce.mongodb.com/task/spruce_ubuntu1604_e2e_test_patch_d8c2bd8b972ee96dcbbe0a26cd4bbfb14aaef726_62c8a2972a60ed63416aa6a1_22_07_08_21_33_17/tests?execution=0&sortBy=STATUS&sortDir=ASC

The videos are also far more smooth. 

Moving to a larger distro seems like the way to go.

cypress/integration/taskQueue/route.ts
```diff
@@ -60,13 +62,23 @@ describe("Task Queue", () => {
...
-    cy.get(".ant-table-row-selected").contains("13").should("be.visible");
-  });
+  it(
+    "Scrolls to current task if taskId param in url",
+    { scrollBehavior: false },
+    () => {
+      cy.intercept("POST", GQL_URL, (req) => {
+        aliasQuery(req, "DistroTaskQueue");
+      });
+      cy.viewport(1920, 1080);
```
minnakt, line 72: Should we set the viewport globally in a config file somewhere? I notice we call this in `host_filtering.ts` as well

cypress/integration/taskQueue/route.ts
```diff
@@ -60,13 +62,23 @@ describe("Task Queue", () => {
...
+      cy.visit(
+        "/task-queue/osx-108/evergreen_lint_lint_service_patch_5e823e1f28baeaa22ae00823d83e03082cd148ab_5e4ff3abe3c3317e352062e4_20_02_21_15_13_48"
+      );
+      cy.wait("@gqlDistroTaskQueueQuery");
+      // eslint-disable-next-line  cypress/no-unnecessary-waiting
+      cy.wait(5000);
```
minnakt, line 78: Is there any way to get around the `cy.wait(5000)` here? Can we maybe guard against the next statement somehow?

cypress/integration/hosts/host_filtering.ts
```diff
@@ -156,98 +156,121 @@ describe("Hosts page filtering from URL", () => {
...
+      cy.preserveCookies();
+      cy.setCookie("This is an important notification", "true");
+      cy.visit(`${hostsRoute}?limit=100&page=0`);
+      cy.dataCy("hosts-table").should("be.visible");
+      cy.dataCy("hosts-table").should("not.have.attr", "data-loading", "true");
+      cy.dataCy(distroFilterIconDataCy)
+        .should("be.visible")
+        .should("not.be.disabled");
+      // eslint-disable-next-line cypress/no-unnecessary-waiting
+      cy.wait(1000);
```
minnakt, line 178: Same comment about `cy.wait(1000)`

cypress/integration/projectSettings/project_settings.ts
```diff
@@ -233,7 +233,7 @@ describe("Repo Settings", () => {
     it("Reorders the commands", () => {
       cy.dataCy("array-down-button").click();
 
-      cy.dataCy("save-settings-button").click();
+      cy.dataCy("save-settings-button").click({ force: true });
```
minnakt, line 236: I think it might work if you just do ` cy.dataCy("save-settings-button").scrollIntoView().click()` here 

(I tested on this [patch](https://spruce.mongodb.com/version/62d1b0da3066155eda46f370/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC))

cypress/integration/hosts/host_filtering.ts
```diff
@@ -156,98 +156,121 @@ describe("Hosts page filtering from URL", () => {
...
-    cy.dataCy(distroFilterIconDataCy).click();
+    it("Filters hosts with input value when Enter key is pressed", () => {
+      cy.dataCy(distroFilterIconDataCy)
+        .should("be.visible")
+        .should("not.be.disabled");
+      cy.dataCy(distroFilterIconDataCy).scrollIntoView();
+      cy.dataCy(distroFilterIconDataCy)
+        .should("be.visible")
+        .should("not.be.disabled");
+      cy.dataCy(distroFilterIconDataCy).click();
```
minnakt, line 189: Since these assertions are already checked in `beforeEach`, I think you might be able to simplify this whole block to `     cy.dataCy(distroFilterIconDataCy).scrollIntoView().click()` 

(I tested on this [patch](https://spruce.mongodb.com/version/62d1b0da3066155eda46f370/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC))

cypress/integration/version/task_filters.ts
```diff
@@ -183,7 +183,11 @@ describe("Tasks filters", () => {
   describe("Task Base Statuses select", () => {
     const urlParam = "baseStatuses";
     before(() => {
+      cy.intercept("POST", GQL_URL, (req) => {
+        aliasQuery(req, "PatchTasks");
+      });
       cy.visit(pathTasks);
+      cy.wait("@gqlPatchTasksQuery");
```
minnakt, line 190: I think it's kind of strange how we don't do this for the other filters we test on this page. Should we?

cypress/integration/hosts/host_filtering.ts
```diff
@@ -156,98 +156,121 @@ describe("Hosts page filtering from URL", () => {
...
+    filterTests.forEach(({ param, filterIconDataCy, filterValue }) => {
+      it(`Filters hosts using table filter dropdowns for ${param}`, () => {
+        cy.dataCy(filterIconDataCy).click();
+        cy.dataCy(`${filterIconDataCy}-wrapper`).should("be.visible");
+        cy.dataCy(`${filterIconDataCy}-wrapper`).within(() => {
+          if (param === statusesParam) {
+            cy.getInputByLabel("Running").click({ force: true });
+          } else if (param === currentTaskIdParam) {
+            // do this for really long text because otherwise cypress times out while typing and fails
+            const subString = filterValue.substr(0, filterValue.length - 1);
```
minnakt, line 257: nit. Replace `substr` with `substring` since `substr` is deprecated

cypress/integration/taskQueue/route.ts
khelif96: Yup i'm in support of setting a global viewport size instead of one off for tests

cypress/integration/hosts/host_filtering.ts
```diff
@@ -156,98 +156,121 @@ describe("Hosts page filtering from URL", () => {
...
+            const lastChar = filterValue.slice(-1);
+
+            cy.dataCy("input-filter")
+              .as("currentTask")
+              .invoke("val", subString)
+              .trigger("input");
```
khelif96, line 263: I learned this trick recently you can add `{delay: 0}` to `cy.type` to avoid the delay while typing it will accomplish the same affect as above but without any of the extra code. https://www.youtube.com/watch?v=JJ2o6Zceg90

cypress/integration/hosts/host_filtering.ts
```diff
@@ -156,98 +156,121 @@ describe("Hosts page filtering from URL", () => {
...
+    });
 
-  beforeEach(() => {
-    cy.preserveCookies();
-    cy.visit(`${hostsRoute}?limit=100&page=0`);
-    cy.dataCy("hosts-table").should("be.visible");
-    cy.dataCy("hosts-table").should("not.have.attr", "data-loading", "true");
-  });
+    beforeEach(() => {
+      cy.viewport(1920, 1080);
```
khelif96, line 168: This viewport should be set globally for the app. We shouldn't be setting different viewports on a case by case basis its better to be consistent.

.evergreen.yml
```diff
@@ -150,7 +150,11 @@ functions:
       working_dir: spruce
       script: |
         export PATH=/opt/nodejs/node-v16.13.0-linux-x64/bin:$PATH
-        ./node_modules/.bin/cypress run --record --key ${cypress_record_key} --reporter junit
+          cd /tmp
+          curl -L -o 'my_file.zip' 'https://www.googleapis.com/download/storage/v1/b/chromium-browser-snapshots/o/Linux_x64%2F1002917%2Fchrome-linux.zip?generation=1652398617890755&alt=media'
+          unzip my_file.zip
+          cd -
+        ./node_modules/.bin/cypress run --browser /tmp/chrome-linux/chrome --record --key ${cypress_record_key} --reporter junit
```
khelif96, line 157: I wonder if we can just run `yarn cy:run` instead of accessing the executable directly

.evergreen.yml
```diff
@@ -150,7 +150,11 @@ functions:
       working_dir: spruce
       script: |
         export PATH=/opt/nodejs/node-v16.13.0-linux-x64/bin:$PATH
-        ./node_modules/.bin/cypress run --record --key ${cypress_record_key} --reporter junit
+          cd /tmp
+          curl -L -o 'my_file.zip' 'https://www.googleapis.com/download/storage/v1/b/chromium-browser-snapshots/o/Linux_x64%2F1002917%2Fchrome-linux.zip?generation=1652398617890755&alt=media'
```
khelif96, line 154: This should probably be an expansion. You can add it to project variables.

cypress.config.ts
```diff
@@ -0,0 +1,14 @@
+import { defineConfig } from "cypress";
+
+export default defineConfig({
+  e2e: {
+    baseUrl: "http://localhost:3000",
+    projectId: "yshv48",
+    reporterOptions: {
+      mochaFile: "bin/cypress/junit-[hash].xml",
+    },
+    defaultCommandTimeout: 8000,
```
khelif96, line 10: Can we bump `defaultCommandTimeout` back down to its default?

cypress/integration/hosts/host_filtering.ts
khelif96: I agree with @minnakt  using `cy.wait` is an anti pattern and isn't recommended by cypress. https://docs.cypress.io/guides/references/best-practices#Unnecessary-Waiting

cypress/integration/projectSettings/project_settings.ts
```diff
@@ -376,9 +376,22 @@ describe("Project Settings when not defaulting to repo", () => {
...
+          $inputs.eq(0).val(),
+          $inputs.eq(1).val(),
+          $inputs.eq(2).val(),
+        ]);
+        const copy = new Set(["admin_var", "sample_name", "sample_name_2"]);
+        expect(inputVals.size).to.equal(copy.size);
+        expect(
+          Array.from(inputVals).every((value) => copy.has(value))
+        ).to.equal(true);
+      });
```
khelif96, line 394: [`.each()`](https://docs.cypress.io/api/commands/each#Syntax) seems suited for this use case. 
I am confused as to why these tests would start to flake however?

cypress/integration/hosts/host_filtering.ts
SupaJoon: I only meant to indent this code and rather change the test in a separate PR.

cypress/integration/projectSettings/project_settings.ts
SupaJoon: We need to wrap the input values in a set to make sure they are each accounted for. We could use `.each` but we would need to create an outside set or counter to make sure all 3 values exist in the tested inputs. I think it's better to use `.should` and have the set logic contained in there.

cypress/integration/version/task_filters.ts
SupaJoon: cypress was having issues toggling the table filter so the lines above make sure the table data is loaded. I don't think we need to do this for the other filters since they didn't have the same flake condition

.evergreen.yml
SupaJoon: I tried in this commit but didn't get the correct result. I made sure to update the variable name in the settings legacy ui page

cypress/integration/hosts/host_filtering.ts
minnakt: I was able to get these tests to pass on this [patch](https://spruce.mongodb.com/task/spruce_ubuntu1604_e2e_test_patch_6c09c39220d1a3d8444d8782696e5d5cba647be3_62e03962c9ec445e89bb1f32_22_07_26_18_58_51/tests?execution=0&sortBy=STATUS&sortDir=ASC) by waiting all of the queries and guarding all of the popovers and inputs. I think waiting all of the queries is important for some reason but I don't know really know why it allows them to pass 🤔

cypress/integration/hosts/host_filtering.ts
SupaJoon: I set the viewport to 1920, 1080 in the cypress config but set it the default for the variantHistory spec where the pagination test relies on the original width

cypress/integration/hosts/host_filtering.ts
SupaJoon: It makes sense to guard any queries that could cause the table to rerender and cause the icon to detach. It's possible that one of the queries waited on prevented cypress from accessing the icon before it detaches but it's also possible that waiting on queries caused the timeout to increase like cy.wait(1000)

.evergreen.yml
khelif96: It doesn't look like the `run-cypress-test` command even ran in that commit. You may have pushed a new commit to it https://evergreen.mongodb.com/lobster/evergreen/task/spruce_ubuntu1604_e2e_test_patch_321db0bd4b1817ce52a0dbd9e335339ee210dc13_62debe82d6d80a73df880aea_22_07_25_16_02_11/0/task#bookmarks=0%2C733&f~=100~run-cypress-tests&l=1

But did in a subsequent [commit](https://spruce.mongodb.com/task/spruce_ubuntu1604_e2e_test_patch_6c09c39220d1a3d8444d8782696e5d5cba647be3_62debf24a4cf476a8835c10c_22_07_25_16_04_54/logs?execution=0) before you reverted. The tests seemed to run before the spot host was reaped.

cypress/integration/hosts/host_filtering.ts
khelif96: Cool can you open a ticket for that?

cypress.config.ts
```diff
@@ -0,0 +1,18 @@
+import { defineConfig } from "cypress";
+
+export default defineConfig({
+  e2e: {
+    baseUrl: "http://localhost:3000",
+    projectId: "yshv48",
+    reporterOptions: {
+      mochaFile: "bin/cypress/junit-[hash].xml",
+    },
+    defaultCommandTimeout: 15000,
```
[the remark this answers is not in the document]
SupaJoon: I changed it to the default

cypress/integration/hosts/host_filtering.ts
SupaJoon: https://jira.mongodb.org/browse/EVG-17351

.evergreen.yml
SupaJoon: The subsequent commit passes because the evergreen.yml  [changes](https://spruce.mongodb.com/version/62debf24a4cf476a8835c10c/changes?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)  does include the expansion. Here are two patches that fail due to the expansion 
[[47a196b](https://github.com/evergreen-ci/spruce/pull/1329/commits/47a196b4d0dad29aa0e6e5498f291e5e20bd13b1)](https://spruce.mongodb.com/task/spruce_ubuntu1604_e2e_test_patch_6c09c39220d1a3d8444d8782696e5d5cba647be3_62debf24a4cf476a8835c10c_22_07_25_16_04_54/logs?execution=1)
[[25cfb7a](https://github.com/evergreen-ci/spruce/pull/1329/commits/25cfb7a1ce12886a8b46efe0ee6e25bc4c2ba964)](https://spruce.mongodb.com/task/spruce_ubuntu1604_e2e_test_patch_312cde48481a5a199aadaf4c24a1946c45892418_62e19bc4d1fe0746b4d986db_22_07_27_20_10_46/logs?execution=0)

SupaJoon: The 2nd execution of the subsequent commit failed because the chrome download wasn't found. This happened again in this [patch](https://spruce.mongodb.com/task/spruce_ubuntu1604_e2e_test_patch_312cde48481a5a199aadaf4c24a1946c45892418_62e19bc4d1fe0746b4d986db_22_07_27_20_10_46/logs?execution=0) but the latest commit seems to be fine.

cypress/integration/projectSettings/project_settings.ts
```diff
@@ -402,9 +390,22 @@ describe("Project Settings when not defaulting to repo", () => {
...
```
khelif96, line 408: Final nit are we sure these weren't failing as a result of the laggy cypress tests. afaict these should always be sorted.

khelif96 (approved):
Thank you doing this! 
LGTM
mod the comments!

cypress/integration/spawn/host.ts
```diff
@@ -109,7 +99,10 @@ describe("Navigating to Spawn Host page", () => {
...
```
minnakt, line 105:
```suggestion
        cy.dataCy("distro-input")
          .click()
          .type("{downarrow}")
          .type("{enter}");
```

cypress/integration/task/annotations.ts
```diff
@@ -49,7 +46,8 @@ describe("Task Annotation Tab", () => {
 
     // delete the added ticket
     cy.dataCy("A-New-Ticket-delete-btn").click();
-    cy.get(popconfirmYesClassName).click();
+    cy.get(popconfirmYesClassName).should("exist").should("not.be.disabled");
+    cy.get(popconfirmYesClassName).click({ force: true });
```
minnakt, line 50:
```suggestion
    cy.get(popconfirmYesClassName).should("exist").should("not.be.disabled");
    cy.get(popconfirmYesClassName).click();
```

cypress/integration/version/action_buttons.ts
```diff
@@ -86,14 +78,11 @@ describe("Action Buttons", () => {
     describe("Version dropdown options", () => {
       before(() => {
         cy.visit(versionPath(mainlineCommit));
+        cy.dataCy("ellipsis-btn").click({ force: true });
```
minnakt, line 81:
```suggestion
        cy.dataCy("ellipsis-btn").click();
```

cypress/integration/version/task_duration.ts
```diff
@@ -42,7 +34,7 @@ describe("Task Duration Tab", () => {
         `duration=DESC&page=0&statuses=started`
       );
       // Clear status filter.
-      cy.dataCy("status-filter-popover").click();
+      cy.dataCy("status-filter-popover").click({ force: true });
```
minnakt, line 37:
```suggestion
      cy.dataCy("status-filter-popover").click();
```

cypress/integration/banners/slack_notification_banner.ts
```diff
@@ -34,6 +30,11 @@ describe("Slack notification banner", () => {
   });
 
   it("does not show up after user has entered their username and clicks 'save'", () => {
+    cy.clearCookie("has-closed-slack-banner");
+
+    cy.visit("/version/5ecedafb562343215a7ff297/tasks");
+
+    cy.dataCy(slackNotificationBanner).should("be.visible");
```
minnakt, line 37:
```suggestion
    cy.dataCy(slackNotificationBanner).should("exist");
```

minnakt (approved):
All of my suggestions were tested twice in this [patch](https://spruce.mongodb.com/version/62e81254a4cf471451045921/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)!

.evergreen.yml
```diff
@@ -150,8 +150,12 @@ functions:
       working_dir: spruce
       script: |
         export PATH=/opt/nodejs/node-v16.13.0-linux-x64/bin:$PATH
-        ./node_modules/.bin/cypress run --record --key ${cypress_record_key} --reporter junit
-
+        cd /tmp
+        curl -L -o "my_file.zip" "${chrome_download_url}"
+        unzip my_file.zip
+        cd -
```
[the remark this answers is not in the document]
SupaJoon: - represents the previous working directory

cypress/integration/banners/slack_notification_banner.ts
SupaJoon: I think `be.visible` is better in this test since it's more specific than `exist`. Cypress makes the distinction between the two in v6. It's the first breaking change in this changelog: https://docs.cypress.io/guides/references/changelog#6-0-0

cypress/integration/projectSettings/project_settings.ts
SupaJoon: I reverted the test and it passed a few times. If it fails because of sort order we will be able to see it in the video.

PR #1335, opened 2022-06-22
EVG-17051 Remove single quotes from deploy emails

khelif96:
[EVG-17051](https://jira.mongodb.org/browse/EVG-17051)

### Description 
Removes single quotes from strings and replaces them with an html character code equivalent. This prevents us from accidentally escaping the bash command to send an email if there is a single quote in the email message.

### Testing 
Tested both sed and gsed commands on linux and osx respectively.

### DIFF
--- a/email.sh
+++ b/email.sh
@@ -82,8 +82,6 @@ case "$OSTYPE" in
   *)        echo "unknown: $OSTYPE";;
 esac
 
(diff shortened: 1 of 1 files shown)


bsamek (changes requested):
ignoring for now after slack discussion about gsed

bsamek (approved):

PR #1338, opened 2022-06-23
EVG-17058: Use Spruce Form for notifications

minnakt:
[EVG-17058](https://jira.mongodb.org/browse/EVG-17058)

### Description 
We had a really complicated way of keeping track of state changes for Notifications. This PR rewrites the Notifications to use Spruce Form which will handle a lot of the state management for us and make the code easier to read and change.

Other changes:
- Upgraded `antd` to`4.20.0`. This is not the latest version, because it seems there is a small focus ref bug in the Table filter dropdown which affects the sorters.
- Fixed `TimePicker` crashing by changing import statements
- Changed Jest config to accommodate for new import statements

Other notable things:
- The regex selector does NOT have full functionality in terms of disabling inputs that have already been selected. I spoke to Sophie about this component and it seems pretty complicated and would better be explored in a different ticket. To clarify, the regex selectors are functional and useable, they just don't mimic the same conditional disabling behavior we have on production right now. I added a description in the meantime to inform the user how they can use the regex selectors.
- Skipped some e2e tests related to regex selector functionality I can't implement right now.
- I created a custom `AntdSelect` widget because we are so far behind on LG Modal and Select versions that it's just not feasible to use the LG Select.

### Screenshots

https://user-images.githubusercontent.com/47064971/175362780-11f60b18-66d2-43bb-aec7-2b5af9436ccd.mov

### Testing 
- Updated e2e tests


src/pages/projectSettings/tabs/NotificationsTab/getFormSchema.tsx
```diff
@@ -12,53 +16,86 @@ export const getFormSchema = (
...
-    },
-    subscriptions: {
-      "ui:placeholder": "No subscription are defined.",
-      "ui:addButtonText": "Add Subscription",
-      "ui:orderable": false,
-      items: {
-        "ui:field": "subscriptionField",
+      subscriptions: {
+        "ui:placeholder": "No subscription are defined.",
+        "ui:addButtonText": "Add Subscription",
```
malikchaya2, line 85: The title no longer shows up
image

how it used to be: 
image

do we want to add it back in now, or do it in a separate ticket?

src/components/SpruceForm/customFormats.ts
```diff
@@ -1,11 +1,26 @@
 import { validators } from "utils";
 
-const { validateURL, validateJiraURL } = validators;
+const {
```
malikchaya2, line 3: image
this shows an error despite being selected

src/types/triggers.ts
```diff
@@ -24,35 +24,78 @@ export enum TriggerType {
...
+  VERSION_FAILS = "version-fails",
+  VERSION_SUCCEEDS = "version-succeeds",
+  VERSION_EXCEEDS_DURATION = "version-exceeds-duration",
+  VERSION_RUNTIME_CHANGE = "version-runtime-change",
+  BUILD_VARIANT_FINISHES = "build-variant-finishes",
+  BUILD_VARIANT_FAILS = "build-variant-fails",
+  BUILD_VARIANT_SUCCEEDS = "build-variant-succeeds",
+}
+
+export enum ProjectTriggers {
```
malikchaya2, line 48: this may be an old bug, but these two should not be on the project page
image

src/components/SpruceForm/customFormats.ts
malikchaya2: same here:
image

src/components/Notifications/form/notification.ts
```diff
@@ -0,0 +1,196 @@
...
+                title: "",
+                required: ["urlInput"],
+                properties: {
+                  urlInput: {
+                    type: "string" as "string",
+                    title: "Webhook URL",
+                    minLength: 1,
+                  },
+                  secretInput: {
+                    type: "string" as "string",
```
malikchaya2, line 96: image
Why is this disabled?

malikchaya2 (changes requested):
Great work 🎉

src/components/Notifications/form/notification.ts
minnakt: this field should always have a default / never be editable because it's generated but I'm not really sure how they generate the secret on the legacy UI 🤔

minnakt: I think that would be best 😭 Here's the ticket: [EVG-17181](https://jira.mongodb.org/browse/EVG-17181) and I'll also leave a comment about it in the code

src/components/SpruceForm/Widgets/LeafyGreenWidgets.tsx
```diff
@@ -45,11 +45,14 @@ export const LeafyGreenTextInput: React.VFC<
     optional,
     warnings,
   } = options;
+
   const hasError = !!rawErrors?.length;
+  const deduplicatedErrors = Array.from(new Set(rawErrors));
```
minnakt, line 50: For some reason I keep getting duplicated errors on the text inputs, again _only_ on the Project Settings page, which is why I added this... I'm not sure if this a RJSF bug or I'm doing something wrong?

Screen Shot 2022-06-24 at 3 40 27 PM

src/components/SpruceForm/errors.ts
```diff
@@ -4,6 +4,11 @@ export const transformErrors = (errors: AjvError[]) =>
   errors
     .map((error) => {
       switch (error.name) {
+        case "enum":
+          return {
+            ...error,
+            message: "",
+          };
```
minnakt, line 11: I don't think `ui:hideError` works for what I need it to do, but adding this handled the selection errors that Chaya brought up. The errors that fall under this case are the `should be equal to one of the allowed values` errors

src/pages/projectSettings/tabs/NotificationsTab/SubscriptionField.tsx
```diff
@@ -23,89 +15,18 @@ const getSubscriberType = (subscriber: Subscriber) => {
...
-        />
-        <br />
-        {/* todo: display properly */}
-        <strong>subscriberType: </strong> {subscriberType} <br />
-        <strong>subscriber: </strong> {subscriberName} <br />
-      </Body>
-    </StyledExpandableCard>
+    <Body>
+      {/* TODO in EVG-16971: display properly */}
+      <strong>subscriberType: </strong> {subscriberType} <br />
```
malikchaya2, line 28: Let's remove this now that the subscriber is being displayed properly.

malikchaya2 (approved):
LGTM with nit. 
🎉

sophstad (approved):

malikchaya2 (approved):

PR #1341, opened 2022-06-24
EVG-17181  Add generate webhook secret function on Spruce 

malikchaya2:
[EVG-17181 ](https://jira.mongodb.org/browse/EVG-17181 )

### Description 
This adds a function to generate webhook secrets which is needed on the project page.

### DIFF
--- a/src/pages/projectSettings/tabs/NotificationsTab/SubscriptionField.tsx
+++ b/src/pages/projectSettings/tabs/NotificationsTab/SubscriptionField.tsx
@@ -10,3 +10,16 @@ export const SubscriptionField: Field = ({ formData }) => {
     </Body>
   );
 };
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export const generateWebhookSecret = () => {
+  let text = "";
+  const possible =
+    "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
+
+  for (let i = 0; i < 64; i++) {
+    text += possible.charAt(Math.floor(Math.random() * possible.length));
+  }
+
+  return text;
+};


src/pages/projectSettings/tabs/NotificationsTab/SubscriptionField.tsx
```diff
@@ -10,6 +10,19 @@ import { ResourceType, StringMap, Trigger } from "types/triggers";
 import { toSentenceCase } from "utils/string";
 import { SubscriptionSelect } from "./NotificationsRow";
 
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const generateWebhookSecret = () => {
```
SupaJoon, line 14: We may want to generate this in evergreen if it's private information or if we want to guarantee unique keys

minnakt, line 14: Given Arjun's comment, maybe someone on app should also review?

Also, do you want to hook this function up to the form, or is the goal of this ticket just to make the function available?

ablack12: Hmm I guess I'm not sure what the use of the webhook secret is. Do users need to copy this to use elsewhere? I think it would be okay to push improving this off to a new ticket anyway, since no one has a problem with how it is now, and we could investigate maybe just hiding it completely.

malikchaya2: sounds good. https://jira.mongodb.org/browse/EVG-17253

malikchaya2: This is ready for re-review. Please see the comment below.

PR #1346, opened 2022-06-29
EVG-16934: Reduce frequency of Tab event and refetch query on return to tab

minnakt:
[EVG-16934](https://jira.mongodb.org/browse/EVG-16934)

### Description 
This PR makes it so that the `Tab Active`/`Tab Not Active` events appear once per tab instead of once per GraphQL query. 

It also makes it so that polled queries are refetched when a user returns to a tab using the `refetch` function that comes from Apollo's `useQuery` hook. From the [docs](https://www.apollographql.com/docs/react/data/queries/#refetching):
```javascript
You can optionally provide a new variables object to the refetch function. If you avoid passing a variables
object and use only refetch(), the query uses the same variables that it used in its previous execution.
```

### Screenshots
- No visible changes

### Testing 
- Added test for refetch functionality in `usePolling.test.tsx`


src/hooks/useNetworkStatus.ts
```diff
@@ -1,17 +1,39 @@
...
-export const useNetworkStatus = () => {
+export const useNetworkStatus: useNetworkStatusType = (
+  sendAnalytics = false
+) => {
   const [isOnline, setIsOnline] = useState(true);
+  const { sendEvent } = usePollingAnalytics();
+
+  const sendOfflineEvent = () => {
+    if (sendAnalytics) {
+      sendEvent({ name: "Tab Not Active", status: "offline" });
```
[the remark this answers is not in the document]
khelif96: Do we need analytics for offline events anyways?

src/analytics/polling/usePollingAnalytics.ts
```diff
@@ -2,7 +2,7 @@ import { addPageAction, Properties, Analytics } from "analytics/addPageAction";
 import { useGetUserQuery } from "analytics/useGetUserQuery";
 
 type Action =
-  | { name: "Tab Active" }
+  | { name: "Tab Active"; status: string }
```
khelif96, line 5: Would an enum make sense instead?

src/analytics/polling/usePollingAnalytics.ts
```diff
@@ -2,8 +2,8 @@ import { addPageAction, Properties, Analytics } from "analytics/addPageAction";
 import { useGetUserQuery } from "analytics/useGetUserQuery";
```
minnakt, line 2: I'm not sure if I should rename this file to something different 🙂

khelif96: useActivityAnalytics?

khelif96 (approved):

PR #1359, opened 2022-07-05
EVG-17191: Remove cedar test results setting

sophstad:
EVG-17191

### Description 
- Cedar is now used by default, so remove it as a configurable setting

### Testing 
- Update unit tests


khelif96 (changes requested):
Changes make sense to me but we should probably hold off on this. Cedar is the default for everything except Mongo projects.

khelif96 (approved):

PR #1370, opened 2022-07-15
EVG-17308 Fix the project settings menu headers 

malikchaya2:
[EVG-17308](https://jira.mongodb.org/browse/EVG-17308)

### Description 
The Repo was getting lost with the project group filtering, making the headers not clickable. I switched it back to using the repoRefId that's already in each project in the project group. 

### Testing 
Tested on staging. This doesn't work for certain repos (like grip), but I'm guessing it's more about the test data and configuration.


src/components/projectSelect/index.tsx
```diff
@@ -104,8 +104,10 @@ export const ProjectSelect: React.VFC<ProjectSelectProps> = ({
           projects={projectGroup.projects}
           name={projectGroup.groupDisplayName}
           onClick={onClick}
-          repoIdentifier={projectGroup?.repo?.id}
-          canClickOnRepoGroup={isProjectSettingsPage && projectGroup?.repo?.id}
```
sophstad, line 108: It seems like a more straightforward approach to include the `projectGroup.repo` in the structure returned on line 164. In principle it seems like we should be able to show a repo group that has no projects (even though that's unlikely to happen in practice with our current design -- but you never know!)

malikchaya2: done

sophstad (commented):
We should add a Cypress test for clicking on a repo and project in project settings! Sorry for not mentioning in my initial review

malikchaya2:
> We should add a Cypress test for clicking on a repo and project in project settings! Sorry for not mentioning in my initial review

done

cypress/integration/projectSettings/project_settings.ts
```diff
@@ -11,6 +11,30 @@ const project = "spruce";
...
+
+  it("Headers are clickable", () => {
+    cy.dataCy("project-select").should("be.visible");
+    cy.dataCy("project-select").click();
+    cy.dataCy("project-select-options").should("be.visible");
+    cy.dataCy("project-select-options")
+      .find("div")
+      .contains("evergreen-ci/evergreen")
+      .click();
+    cy.location().should((loc) => expect(loc.pathname).to.not.eq(destination));
```
sophstad, line 34: We can probably also assert that this links to the repo ID saved above!

sophstad (approved):

PR #1371, opened 2022-07-15
EVG-17299: Make emptyValue default to empty string

minnakt:
[EVG-17299](https://jira.mongodb.org/browse/EVG-17299)

### Description 
The `emptyValue` (i.e. the value for an empty text input or text area) is set to `undefined` rather than an empty string, which causes issues with saving the page. 

For example, suppose that a user starts filling out an optional field but then changes their mind and deletes it. Even though the field is optional, they can no longer save the page because the value is `undefined`. This behavior can be verified on staging and in the video below.

https://user-images.githubusercontent.com/47064971/179265566-9730fc24-0f7f-4e04-b65f-85c3477e1422.mov

This PR fixes that issue by setting the default `emptyValue` to be an empty string.

### Other issues

We currently overwrite the `minLength` error message to be an empty string. It turns out that this causes problems when used in conjuction with the `validate` function because of how the JSON schema validation and custom validation errors are merged. In the source code [here](https://github.com/rjsf-team/react-jsonschema-form/blob/851c03aa94bb51c0b835222e2fabe69a64aca59f/packages/core/src/validate.js#L76) you can see that falsy messages are not preserved in the errors, which means our `minLength` error messages get ignored.

To get around this, Sophie and I decided to set the `minLength` error message to be the text string `"invisible"` and filter out that error message on the widget level. Doing this keeps the error in the form state but prevents the error from showing up visually. This is consistent with the current behavior of `minLength` validation.

### Testing 
- Added test for Text Input and Text Area widgets in `SpruceForm.test.tsx`
- For manual testing, you can test my branch locally and compare it to spruce-staging


src/components/SpruceForm/SpruceForm.test.tsx
```diff
@@ -124,3 +324,46 @@ const basicForm = {
...
+        type: "string" as "string",
+        title: "Text Area",
+        default: "",
+        minLength: 1,
+      },
+    },
+  },
+  uiSchema: {
+    textArea: {
+      "ui:widget": widgets.TextareaWidget,
```
sophstad, line 364: [nit] You can simply refer to this as `"textarea"` like we do [here](https://github.com/evergreen-ci/spruce/blob/6c09c39220d1a3d8444d8782696e5d5cba647be3/src/components/SpruceForm/SpruceForm.stories.tsx#L99) because we override the default widgets [here](https://github.com/evergreen-ci/spruce/blob/6c09c39220d1a3d8444d8782696e5d5cba647be3/src/components/SpruceForm/Widgets/index.tsx#L15) (there are other places in the codebase where this change could be made too)
```suggestion
      "ui:widget": "textarea",
```

src/components/SpruceForm/errors.ts
```diff
@@ -28,7 +28,7 @@ export const transformErrors = (errors: AjvError[]) =>
         case "minLength":
           return {
             ...error,
-            message: "",
+            message: "invisible",
```
sophstad, line 31: [nit] Perhaps we could use a string enum to reduce the magic string-ness of this admittedly hacky solution 😄

src/components/SpruceForm/Widgets/LeafyGreenWidgets.tsx
```diff
@@ -38,19 +48,20 @@ export const LeafyGreenTextInput: React.VFC<
...
   } = options;
 
-  const hasError = !!rawErrors?.length;
+  const errors = filterInvisibleErrors(rawErrors);
+  const hasError = !!errors?.length;
 
   // Deduplicate errors due to a bug when using oneOf dependencies.
   // https://github.com/rjsf-team/react-jsonschema-form/issues/1590
-  const deduplicatedErrors = Array.from(new Set(rawErrors));
+  const deduplicatedErrors = Array.from(new Set(errors));
```
sophstad, line 64: What do you think of making one util function to process `rawErrors`? It seems like there's a chance that duplicate errors could be encountered in other widgets too so it seems like there's no harm in deduplicating processing all input types uniformly

src/components/SpruceForm/Widgets/LeafyGreenWidgets.tsx
```diff
@@ -23,6 +23,16 @@ const { yellow } = uiColors;
 
 const isNullish = (val: any) => val === null || val === undefined;
 
+/**
+ * "Invisible" errors are errors that we want to affect formState (e.g. preventing submission) but
+ * not show visibly on the UI. This function filters out invisible errors so that they do not affect
+ * the visual appearance of the form elements.
+ * @param errors - array of error messages
+ * @returns error messages array with "invisible" errors removed
+ */
```
sophstad, line 32: Maybe worth noting here that this bug occurs specifically when using a `validate` function just so that fact is noted somewhere in code.

sophstad (approved):

src/components/SpruceForm/Widgets/utils.ts
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Returns true if a given value is null or undefined.
+ */
+export const isNullish = (val: any) => val === null || val === undefined;
+
+/**
+ * Processes errors by removing "invisible" and duplicate errors.
+ */
+export const processErrors = (errors: string[]) =>
```
sophstad, line 11: [nit] Since we always use the result of this function to calculate `hasErrors` and an error string, we could directly return an object with those fields

sophstad (approved):

PR #1373, opened 2022-07-18
EVG-17326 Skunkworks - Save codegen diff to evergreen task

khelif96:
[EVG-17326](https://jira.mongodb.org/browse/EVG-17326)

### Description 
Thought it would be useful for reviewers to see the diff of why the codegen task broke. 

### Testing 
Example task
https://spruce.mongodb.com/task/spruce_ubuntu1604_check_codegen_patch_2a281b272efeb91c38b653e4ccbc1e63b1266835_62d596903e8e86231992ee3c_22_07_18_17_21_26/logs?execution=0
Example diff
https://mciuploads.s3.amazonaws.com/spruce/spruce_ubuntu1604_check_codegen_patch_2a281b272efeb91c38b653e4ccbc1e63b1266835_62d596903e8e86231992ee3c_22_07_18_17_21_26/codegen/codegen.diff

### DIFF
--- a/.evergreen.yml
+++ b/.evergreen.yml
@@ -234,6 +234,20 @@ functions:
       content_type: text/html
       permissions: public-read
       display_name: storybook-
+  
+  attach-codegen-diff:
+    command: s3.put
+    type: system
+    params:
+      aws_key: ${aws_key}
+      aws_secret: ${aws_secret}
+      local_files_include_filter:
+        - "spruce/tmp/codegen.diff"
+      remote_file: spruce/${task_id}/codegen/
+      bucket: mciuploads
+      content_type: text/plain
+      permissions: public-read
+      display_name: "codegen diff"
     
   setup-credentials:
     command: shell.exec
@@ -394,3 +408,4 @@ post:
   - func: attach-cypress-videos
   - func: attach-server-logs
   - func: attach-source-map
+  - func: attach-codegen-diff
--- a/.evergreen.yml
+++ b/.evergreen.yml
@@ -247,7 +247,6 @@ functions:
       bucket: mciuploads
       content_type: text/plain
       permissions: public-read
-      display_name: "codegen diff"
     
   setup-credentials:
     command: shell.exec
--- a/.evergreen.yml
+++ b/.evergreen.yml
@@ -242,7 +242,7 @@ functions:
       aws_key: ${aws_key}
       aws_secret: ${aws_secret}
       local_files_include_filter:
-        - "spruce/tmp/codegen.diff"
+        - "spruce/bin/codegen.diff"
       remote_file: spruce/${task_id}/codegen/
       bucket: mciuploads
       content_type: text/plain
--- a/check_codegen.sh
+++ b/check_codegen.sh
@@ -1,14 +1,30 @@
 # This is a script which checks if yarn codegen was run it used by the check_codegen evergreen task
-BEFORE=$(git status --porcelain | grep src/gql/generated/types.ts)
-yarn codegen
-AFTER=$(git status --porcelain | grep src/gql/generated/types.ts)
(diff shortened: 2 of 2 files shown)


check_codegen.sh
```diff
@@ -1,12 +1,29 @@
...
+  then
+      echo "Directory /bin already exists." 
+  else
+      mkdir bin
+  fi
+  git diff $types_file >> bin/codegen.diff
   exit 1
 else
-  echo "src/gql/types.ts is up to date"
-  exit 0
```
khelif96, line 11: The script was always exiting 0 despite the `yarn codegen` task exiting 1 which would make the task misleadingly succeed. I removed that default exit 0 condition and added a guard above to exit if the script ever fails. cc @minnakt @sophstad

check_codegen.sh
```diff
@@ -1,12 +1,29 @@
...
+# path to generated types
+types_file="src/gql/generated/types.ts"
+{
+  BEFORE=$(git status --porcelain | grep "$types_file")
+  yarn codegen
+  AFTER=$(git status --porcelain | grep "$types_file")
+} || {
+  echo "script failed"
+  exit 1
+}
```
khelif96, line 12: bash does not have a concept of `try/catch` so this is the next best thing.  :/

minnakt, line 12: Should this be failing like it does on CI?

I changed this part of the script to this, which seemed to produce the correct results:

```suggestion
BEFORE=$(git status --porcelain | grep "$types_file")
yarn codegen || {
  echo "script failed"
  exit 1
}
AFTER=$(git status --porcelain | grep "$types_file")
```

[Passing Patch](https://spruce.mongodb.com/task/spruce_ubuntu1604_check_codegen_patch_16a6c01c5435829171c37bab8a45fb34f2dae505_62d856760ae60666474ddede_22_07_20_19_24_46/logs?execution=0)   |   [Failing Patch](https://spruce.mongodb.com/task/spruce_ubuntu1604_check_codegen_patch_16a6c01c5435829171c37bab8a45fb34f2dae505_62d857afc9ec4462ba4a6df9_22_07_20_19_29_56/logs?execution=0)

This [approach](https://github.com/minnakt/spruce/blob/alternate_codegen/check_codegen.sh) might also work 🤔 Not sure what's best

.evergreen.yml
```diff
@@ -234,6 +234,20 @@ functions:
...
+    params:
+      aws_key: ${aws_key}
+      aws_secret: ${aws_secret}
+      local_files_include_filter:
+        - "spruce/bin/codegen.diff"
+      remote_file: spruce/${task_id}/codegen/
+      bucket: mciuploads
+      content_type: text/plain
+      permissions: public-read
+      display_name: "codegen diff"
```
minnakt, line 250: Is there any way we can get this to say `codegen diff` only?

Screen Shot 2022-07-20 at 2 56 59 PM

khelif96: I could remove the display name so it would only say `codegen.diff`

check_codegen.sh
```diff
@@ -1,12 +1,27 @@
 # This is a script which checks if yarn codegen was run it used by the check_codegen evergreen task
-BEFORE=$(git status --porcelain | grep src/gql/generated/types.ts)
-yarn codegen
-AFTER=$(git status --porcelain | grep src/gql/generated/types.ts)
+
+# path to generated types
+types_file="src/gql/generated/types.ts"
+BEFORE=$(git status --porcelain | grep "$types_file")
+yarn codegen || {
+  echo "script failed"
```
minnakt, line 7: nit. It might be nice if this was more explicit like `"yarn codegen failed to run"`, but it's up to you!

.evergreen.yml
khelif96: done

minnakt (approved):
:shipit:

PR #1376, opened 2022-07-26
EVG-16708 Add allowlist for env variables in html

khelif96:
[EVG-16708](https://jira.mongodb.org/browse/EVG-16798)

### Description 
some env variables weren't properly being injected into the static html file that is generated.


minnakt (approved):
approve :shipit:

PR #1381, opened 2022-07-28
EVG-17287: Misc plugins page updates

sophstad:
EVG-17287

### Description 
- Update Plugins page validation to use less custom function and more schema rules. Also use dependencies instead of a custom hide function to control rendering of radio box fields
- On General tab, use minLength along with required. Also, don't require display name or config file (these are valid when empty according to legacy)

### DIFF
--- a/src/pages/projectSettings/tabs/PluginsTab/PluginsTab.tsx
+++ b/src/pages/projectSettings/tabs/PluginsTab/PluginsTab.tsx
@@ -1,5 +1,6 @@
 import { useMemo } from "react";
 import { SpruceForm } from "components/SpruceForm";
+import { ValidateProps } from "components/SpruceForm/types";
 import { ProjectSettingsTabRoutes } from "constants/routes";
 import {
   usePopulateForm,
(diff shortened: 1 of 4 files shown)


src/pages/projectSettings/tabs/PluginsTab/PluginsTab.tsx
```diff
@@ -52,10 +51,7 @@ const validate = (formData, errors) => {
   const searchProject = formData?.buildBaronSettings?.ticketSearchProjects;
 
   // if a search project is defined, a create project must be defined, and vice versa
-  const searchProjectDefined =
-    searchProject?.length !== 0 &&
-    searchProject?.[0]?.searchProject !== undefined &&
-    searchProject?.[0]?.searchProject.trim() !== "";
+  const searchProjectDefined = searchProject?.length !== 0;
```
khelif96, line 54: can searchProject be undefined? 
`(undefined !== 0) === true`

khelif96 (approved):

malikchaya2 (approved):

src/pages/projectSettings/tabs/PluginsTab/PluginsTab.tsx
sophstad: No, it will always [default to `[]`](https://github.com/evergreen-ci/spruce/blob/d3b68d1821f8500ae81d6663578368f2beef143b/src/pages/projectSettings/tabs/PluginsTab/transformers.ts#L28-L31)

I also added typing to the validate function!

PR #1397, opened 2022-08-03
EVG-17411 Bump @types/node to match required node version

khelif96:
[EVG-17411](https://jira.mongodb.org/browse/EVG-17411)

### Description 
Bumps types/node to have the same major version that we require to build spruce.

### DIFF
--- a/package.json
+++ b/package.json
@@ -149,7 +149,7 @@
     "@types/js-cookie": "^3.0.2",
     "@types/lodash.debounce": "4.0.7",
     "@types/new-relic-browser": "0.1118.0",
-    "@types/node": "14.14.10",
+    "@types/node": "^16.11.47",
     "@types/react": "17.0.0",
     "@types/react-dom": "17.0.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -149,7 +149,7 @@
     "@types/js-cookie": "^3.0.2",
     "@types/lodash.debounce": "4.0.7",
     "@types/new-relic-browser": "0.1118.0",
-    "@types/node": "14.14.10",
+    "@types/node": "^16.11.47",
```
sophstad, line 152: It looks like we actually require >= 16.13.x
https://github.com/evergreen-ci/spruce/blob/eac039ab65712d6160aad369ce2dab17b5ff3526/package.json#L234

sophstad (changes requested):

package.json
khelif96: it doesn't exist :/ 
https://releasealert.dev/npm/@types/node
16.11.47 is the latest release for node version 16.

sophstad (approved):

PR #1402, opened 2022-08-09
EVG-17423, EVG-17391: Remove default to repo button on Plugins & Notifications tabs

sophstad:
EVG-17423, EVG-17391 (sorry @minnakt I stole your ticket since I was making this change for Plugins)

### Description 
Remove "Default to Repo" buttons on pages that don't support the functionality for all of their fields

### Screenshots
image
image

### Testing 
- Added integration test


src/pages/projectSettings/HeaderButtons.tsx
```diff
@@ -126,22 +131,23 @@ export const HeaderButtons: React.VFC<Props> = ({ id, projectType, tab }) => {
...
-          <DefaultSectionToRepoModal
-            handleClose={() => setDefaultModalOpen(false)}
-            open={defaultModalOpen}
-            projectId={id}
-            section={mapRouteToSection[tab]}
-          />
-        </>
-      )}
+      {projectType === ProjectType.AttachedProject &&
+        !defaultToRepoDisabled.has(tab) && (
```
minnakt, line 135: I think it might be slightly more readable if we stored this in a variable above like

`const canDefaultToRepo = !defaultToRepoDisabled.has(tab)`

and then did 
```
{ projectType === ProjectType.AttachedProject && canDefaultToRepo && ... }
```

src/pages/projectSettings/HeaderButtons.tsx
```diff
@@ -26,6 +26,11 @@ import { formToGqlMap } from "./tabs/transformers";
 import { FormToGqlFunction, WritableTabRoutes } from "./tabs/types";
 import { ProjectType } from "./tabs/utils";
 
+const defaultToRepoDisabled: Set<WritableTabRoutes> = new Set([
+  ProjectSettingsTabRoutes.Notifications,
+  ProjectSettingsTabRoutes.Plugins,
+]);
```
minnakt, line 32: possibly consider moving to `tabs/types`? That's where `readOnlyTabs` exist, which seems kind of similar

minnakt (approved):
wonderful 🌺 and you are simply a queen for including my ticket

src/pages/projectSettings/HeaderButtons.tsx
sophstad: I considered that, but I felt like this property is very specific to the component and won't be reused, so decided to include it in the component file itself. 😺

sophstad:
Merging despite check_codegen failure (an evergreen commit that removed some GraphQL fields was reverted, but since they'll be removed once again we can ignore the changes)

PR #1405, opened 2022-08-11
EVG-17491: Make git tag user/team fields not orderable

sophstad:
EVG-17491

### Description

- Authorized users/teams for git tags should not be able to be reordered

### Screenshots

Field no longer has arrows next to each array item:
image


minnakt (approved):
:shipit:

PR #1413, opened 2022-08-22
EVG-15841 Add ability to specify a custom date format 

khelif96:
EVG-15841

### Description
Adds an option in the profile settings to allow users to specify a custom date format which all dates will be rendered with. 
* Refactor the profile tab to use SpruceForm as a bonus.
* Adds a bunch of custom formats.
* Add a hook to fetch the data necessary to specify a date format. 

I opted not to allow users to specify a custom format since there is a risk of using an invalid format string which would result in `date-fns` throwing an error instead i supplied a variety of options.

### Screenshots
image

image
image
image

### Evergreen PR
https://github.com/evergreen-ci/evergreen/pull/5887

### DIFF
--- a/src/constants/fieldMaps.ts
+++ b/src/constants/fieldMaps.ts
@@ -172,6 +172,8 @@ export const awsRegions = [
   },
 ];
 
+export const dateFormats = ["MM/DD/YYYY", "DD/MM/YYYY", "YYYY/MM/DD"];
+
 export const notificationFields = {
   patchFinish: "Patch finish",
   patchFirstFailure: "Patch first task failure",
--- a/src/constants/fieldMaps.ts
+++ b/src/constants/fieldMaps.ts
@@ -1,3 +1,5 @@
+import { getDateCopy } from "utils/string";
+
 export const timeZones = [
   {
     str: "Coordinated Universal Time",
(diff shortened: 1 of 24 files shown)


src/constants/fieldMaps.ts
```diff
@@ -172,6 +174,51 @@ export const awsRegions = [
   },
 ];
 
+export const dateFormats = [
+  {
+    value: "MM-dd-yyyy",
+    str: `MM-dd-yyyy - ${getDateCopy("08/31/2022", {
+      dateFormat: "MM-dd-yyyy",
```
sophstad, line 181: IMO it may be a bit easier to parse this dropdown if we call this function with `dateOnly` since time format is not being addressed here. But your call!

src/constants/fieldMaps.ts
```diff
@@ -172,6 +174,51 @@ export const awsRegions = [
   },
 ];
 
+export const dateFormats = [
```
sophstad, line 177: What about something like this?

```js
export const dateFormats = [...listOfDateFormatStrings].map((format) => ({
  value: format,
  str: `${format} - ${getDateCopy("08/31/2022", {
    dateFormat: format,
  })}`,
}));
```

Lots of copy/pasted strings that we could eliminate!

sophstad (changes requested):

src/constants/fieldMaps.ts
khelif96: Great idea!

src/utils/string/index.ts
```diff
@@ -115,14 +116,18 @@ export const getDateCopy = (
     return "";
   }
   const { tz, dateOnly, omitSeconds } = options || {};
-  const dateFormat = dateOnly
-    ? "MMM d, yyyy"
-    : `MMM d, yyyy, h:mm${omitSeconds ? "" : ":ss"} aa`;
+  let { dateFormat } = options || {};
+  if (!dateFormat) {
+    dateFormat = "MMM d, yyyy";
+  }
```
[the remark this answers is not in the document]
khelif96: So dateFormat can sometimes be an empty string in the db which would cause that statement to not work.

src/pages/preferences/preferencesTabs/ProfileTab.tsx
```diff
@@ -96,45 +95,74 @@ export const ProfileTab: React.VFC = () => {
...
+                  "ui:placeholder": "Enter your GitHub username",
+                },
+              },
+              dateFormat: {
+                "ui:placeholder": "Select a date format",
+              },
+            }}
+            schema={{
+              properties: {
+                githubUser: {
```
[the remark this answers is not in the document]
khelif96: the graphql mutation expects an object with the shape 
```
githubUser: {
 lastKnownAs: "thing"
}
```

I did this to avoid transforming it later on. I don't think it has any visual change.

src/constants/fieldMaps.ts
```diff
@@ -172,6 +174,24 @@ export const awsRegions = [
...
+  "MM-dd-yyyy",
+  "dd-MM-yyyy",
+  "yyyy-MM-dd",
+  "MM/dd/yyyy",
+  "dd/MM/yyyy",
+  "yyyy/MM/dd",
+  "MMM d, yyyy",
+];
+
+export const dateFormats = [...listOfDateFormatStrings].map((format) => ({
```
sophstad, line 187: hoisted by my own petard
```suggestion
export const dateFormats = listOfDateFormatStrings.map((format) => ({
```

sophstad (approved):

PR #1415, opened 2022-08-24
EVG-17584: Include subscription ID on save

sophstad:
EVG-17584

### Description

- Include subscription ID on save so that subscriptions are updated rather than recreated

### Screenshots

Previously when updating the "notify on build failure" setting on the Notifications tab, all subscriptions would be recreated:
image

Now these objects are persisted instead of replaced on save:
image

And their fields can be updated in isolation:
image

### Testing

- Updated unit tests
- Local testing


malikchaya2 (approved):
🎉

khelif96 (approved):

PR #1432, opened 2022-09-14
EVG-17576 Add Priority to task-queue page

khelif96:
EVG-17576

### Description
Adds priority column to task-queue page
### Screenshots
image


ablack12 (approved):
Zoom zoom!!

bynn (approved):
Wheee

khelif96:
Failing tests are the types which i bumped i think this is still caused by an evergreen bug where it isn't using the correct  version of evergreen.

PR #1443, opened 2022-09-21
EVG-17352: Remove polling on getVersion lazy query

minnakt:
EVG-17352

### Description
`@apollo/client` changed the behavior of `startPolling` and `stopPolling`. It used to be that they would be `undefined` for lazy queries if the query hadn't been called yet, but now they are always defined. This was causing the `getVersion` to be called on component mount.

I ended up removing the polling for the `getVersion` lazy query because it didn't seem necessary.

Also refactored some usage of `usePageTitle` because it helps make `Version.tsx` look cleaner.

### Screenshots
- No visible changes

### Testing
- N/A

### DIFF
--- a/src/pages/Version.tsx
+++ b/src/pages/Version.tsx
@@ -11,7 +11,6 @@ import {
   PageSider,
 } from "components/styles";
 import VersionTaskPageBreadcrumbs from "components/VersionTaskPageBreadcrumbs";
-import { pollInterval } from "constants/index";
 import { commitQueueAlias } from "constants/patch";
 import { getCommitQueueRoute, getPatchRoute } from "constants/routes";
 import { useToastContext } from "context/toast";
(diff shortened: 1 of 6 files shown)


src/pages/Version.tsx
```diff
@@ -97,14 +89,8 @@ export const VersionPage: React.VFC = () => {
     },
   });
 
-  // This resets the pages states to force showing the loading state when the page is changed on navigation
-  useEffect(() => {
-    setIsLoadingData(true);
-    setRedirectURL(undefined);
-  }, [id]);
```
minnakt, line 104: I deleted this hook. Not sure if it's important but it didn't seem to do anything

khelif96 (changes requested):
Please ignore this review my computer is being dumb or maybe i am idk at this point

khelif96 (approved):

PR #1445, opened 2022-09-21
EVG-17932 Redirect lobster links to parsley

khelif96:
EVG-17932

### Description
Lobster links will now redirect to parsley on non prod deployments.
Note. This is temporary during development for the production release we will update the links on the evergreen side and return the appropriate link. 

### Testing
Didn't write any tests since this is only used for internal testing


minnakt (approved):
Don't forget to post an updated `.cmdrc` file!

PR #1447, opened 2022-09-30
EVG-14409 secrets

ybrill:
[EVG-14409](https://jira.mongodb.org/browse/EVG-14409)

### Description
Fetch the env file from AWS Secrets Manager locally and in CI.
Fetching locally assumes developers have the AWS CLI on their path and have a section added to their ~/.aws/config. The required config is static and can be documented somewhere (the dev guide?). Locally, it can be invoked as
```
yarn fetch-config --profile <profile> --login-profile <login-profile> -s <secret-id>
``` 
If a reviewer wants to try it out I can provide them with the config and parameters.

I thought about a separate subcommand to replace the secret, but I think we might want that to be harder to do since we wouldn't want anyone to invoke it by mistake. A procedure for rotating the secret with the AWS CLI and/or browser console would be documented somewhere (the dev guide again?).

### Testing
I created [a patch](https://spruce.mongodb.com/version/6337001c5623432c922e2bbb/tasks) that retrieves the secret and prints it (["this is the secret"](https://evergreen.mongodb.com/task_log_raw/spruce_ubuntu1604_check_credentials_patch_35616394971d2de6993c4d4f9f14095301b98f02_6337001c5623432c922e2bbb_22_09_30_14_41_37/0?type=T#L160) was the value at the time this ran).


cli-scripts/get-credentials.js
```diff
@@ -0,0 +1,113 @@
+const fs = require("fs/promises");
+const path = require("path");
+const util = require("util");
+const exec = util.promisify(require("child_process").exec);
```
khelif96, line 4: Since it looks like you are awaiting allot of these functions anyways you can use the synchronous version of these and remove all of the awaits and asyncs it will make the code more readable.

https://nodejs.org/api/child_process.html#child_processexecsynccommand-options
https://nodejs.org/api/fs.html#fs_fs_writefilesync_file_data_options

khelif96 (commented):
Thanks for doing this! 

I wasn't able to test this out locally because i don't think i have permissions to access secrets manager but here is my initial comment about the script. I will probably have more to say after we sort out the permissions.

ybrill:
Of note is that users will need to be added to a MANA guild to have access to the secret.

cli-scripts/get-credentials.js
ybrill: I've switched to the synchronous versions of those.

It didn't look to me like there was a synchronous version of the AWS calls so I left those be. 
Please bear with me if I'm wrong about that (I'm new at this 😄)

cli-scripts/get-credentials.js
```diff
@@ -0,0 +1,100 @@
...
+        };
+    }
+
+    if (!("SecretString" in data)) {
+        throw {
+            message: `unexpected response from AWS:\n${data}`
+        }
+    }
+
+    return data.SecretString;
```
khelif96, line 74: vars are global so we try not to use them.
```suggestion
    let data;
    try {
       data = await smClient.send(new GetSecretValueCommand({ SecretId: secretId }));
    } catch (error) {
        throw {
            message: "fetching config file from AWS",
            error: error
        };
    }

    if (!("SecretString" in data)) {
        throw {
            message: `unexpected response from AWS:\n${data}`
        }
    }

    return data.SecretString;
```

cli-scripts/get-credentials.js
```diff
@@ -0,0 +1,100 @@
...
+}
+
+function writeCredentialsFile(secret) {
+    try {
+        var envFile = path.join(__dirname, "../env", ".cmdrc.json");
```
khelif96, line 79:
```suggestion
        const envFile = path.join(__dirname, "../env", ".cmdrc.json");
```

cli-scripts/get-credentials.js
```diff
@@ -0,0 +1,100 @@
...
+    }
+    return fromIni({ profile: profile });
+}
+
+async function getSecret(credentials, region, secretId) {
+    console.log("Retrieving secret from AWS");
+
+    const smClient = new SecretsManagerClient({ credentials: credentials, region: region });
+    try {
+        var data = await smClient.send(new GetSecretValueCommand({ SecretId: secretId }));
```
khelif96, line 60: Same comment about var

cli-scripts/get-credentials.js
```diff
@@ -0,0 +1,100 @@
+const fs = require("fs");
+const path = require("path");
+const { execSync } = require("child_process");
+const { program } = require("commander");
```
khelif96, line 4: Cool package! Didn't know it existed

khelif96:
One thing i would like to iron out before merging this is how to get everyone on the team permissions to access secrets manager. This would be something everyone on evergreen would need to have access too. Since everyone will need it to deploy at some point and/or eventually use locally. Would it make sense to work with build to get permissions for secrets manager or at least some instance of it to the standard aws account? @ybrill

cli-scripts/get-credentials.js
ybrill: done

cli-scripts/get-credentials.js
ybrill: done
I moved this line out of the try/catch so we could use it in the catch too (I _think_ the catch is a different block?).

ybrill:
@khelif96 I asked Jason Walsh if we could incorporate the Okta group in the existing evergreen guild. 
I'll keep you posted.

ybrill:
@khelif96 this is done!
Everyone on the wider evergreen team should have automatic access now.

cli-scripts/get-credentials.js
ybrill: I think this was the same var? In any case the script no longer has any `var`s

cli-scripts/get-credentials.js
ybrill: thank you!

khelif96:
Temporarily removed my review request pending the changes we discussed.

ybrill:
Declaring bankruptcy on this since we haven't gotten traction with IT to make the changes we needed.
Also, the company might be getting an officially sanctioned method that we can use to accomplish this.

PR #1461, opened 2022-10-05
EVG-13709 Refactor breadcrumbs to show version message on task page

khelif96:
EVG-13709

### Description
Did some refactoring of the breadcrumbs to separate some of the logic for each task and version page.
Task pages now show the commit message.

### Screenshots
image


src/components/Breadcrumbs/index.tsx
```diff
@@ -16,51 +18,58 @@ interface BreadcrumbsProps {
...
+    >
+      {!to ? (
+        <Body {...rest}>{message}</Body>
+      ) : (
+        <Body {...rest}>
+          <StyledRouterLink to={to} onClick={onClick}>
+            {message}
+          </StyledRouterLink>
+        </Body>
+      )}
```
minnakt, line 62: I would swap these since it's easier `to` parse to than `!to` (the lint is probably wrong sorry)

```suggestion
      {to ? (
              <Body {...rest}>
          <StyledRouterLink to={to} onClick={onClick}>
            {message}
          </StyledRouterLink>
        </Body>
      ) : (
        <Body {...rest}>{message}</Body>
      )}
```

src/components/Breadcrumbs/index.tsx
```diff
@@ -16,51 +18,58 @@ interface BreadcrumbsProps {
...
     {breadcrumbs.map((bc, index) => (
-      <Fragment
-        breadcrumb={bc}
-        key={`breadCrumb-${bc.text}`}
-        active={breadcrumbs.length - 1 === index}
-      />
+      <Fragment key={`breadCrumb-${bc.text}`}>
+        <BreadcrumbFragment breadcrumb={bc} />
+        {breadcrumbs.length - 1 !== index && (
+          <Icon data-cy="breadcrumb-chevron" glyph="ChevronRight" />
```
minnakt, line 24: I think adding some spacing by styling the icon with `margin: 0 ${size.xxs};` would be nice

src/components/Breadcrumbs/index.tsx
```diff
@@ -16,51 +18,58 @@ interface BreadcrumbsProps {
...
-const StyledP1 = styled(P1)`
-  display: inline-block;
-`;
-
 const Container = styled.div`
   display: flex;
   align-items: center;
   > * {
     margin-bottom: 0;
   }
```
minnakt, line 72: I don't think we need the `> *` selector. But we do need some extra margin on the bottom! (page looks weird when browser window is small)

```suggestion
    margin-bottom: ${size.l};
```

src/components/Breadcrumbs/index.tsx
```diff
@@ -16,51 +18,58 @@ interface BreadcrumbsProps {
...
-    <StyledP1 {...rest}>
-      <StyledRouterLink to={to} onClick={onClick}>
-        {text}
-      </StyledRouterLink>
-    </StyledP1>
+  const shouldTrimMessage = text?.length > 25;
+  const message = shouldTrimMessage ? trimStringFromMiddle(text, 25) : text;
+  return (
+    <ConditionalWrapper
+      condition={shouldTrimMessage}
```
minnakt, line 42: Does this work? `trimStringFromMiddle` should leave string unchanged if trimming it would have no effect

```suggestion
  const trimmedText = trimStringFromMiddle(text ?? "", 25);
  return (
    <ConditionalWrapper
      condition={trimmedText !== text}
```

src/pages/version/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+  ];
+
+  return (
+    <Container>
+      <Breadcrumbs breadcrumbs={breadcrumbs} />
+    </Container>
+  );
+};
+
+const Container = styled.div`
```
minnakt, line 64: nit. maybe we don't need the container?

src/pages/task/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+  ];
+
+  return (
+    <Container>
+      <Breadcrumbs breadcrumbs={breadcrumbs} />
+    </Container>
+  );
+};
+
+const Container = styled.div`
```
minnakt, line 65: same comment

cypress/integration/breadcrumbs.ts
```diff
@@ -49,9 +52,9 @@ describe("Viewing a mainline commit", () => {
   it("Shows the commit hash", () => {
     cy.dataCy("bc-version").should("include.text", "5e823e1");
   });
-  it("Clicking the commit hash breadcrumb from a task should take you to that version", () => {
-    cy.dataCy("bc-version").should("include.text", "5e823e1");
```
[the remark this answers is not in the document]
khelif96: the click handler has been moved from the version to the message now so clicking on the version shouldn't do anything.

src/components/Breadcrumbs/Breadcrumbs.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+  it("breadcrumbs with long text should be collapsed and viewable with a tooltip", async () => {
+    const longMessage = "some really long string that could be a patch title";
+    const breadcrumbs = [{ text: longMessage }];
+    render(<Breadcrumbs breadcrumbs={breadcrumbs} />);
+    expect(screen.queryByText(longMessage)).not.toBeInTheDocument();
+
+    expect(
+      screen.getByText(trimStringFromMiddle(longMessage, 25))
+    ).toBeInTheDocument();
+    userEvent.hover(screen.getByText(trimStringFromMiddle(longMessage, 25)));
```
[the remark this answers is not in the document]
khelif96: I'm a little confused by what you mean? I'm not sure what compiled test runner code you are referring to? 


This test is asserting that we do trim longer strings and show a tooltip the actual text contents aren't what is being tested. That is handled by `trimStringFromMiddle` which is thoroughly tested on its own. I added an additional test to test the inverse condition when the text content is short.

Is that what you were referring to or is it something else?

src/components/Breadcrumbs/index.tsx
```diff
@@ -16,51 +19,61 @@ interface BreadcrumbsProps {
...
-interface FragmentProps {
+interface BreadcrumbFragmentProps {
   breadcrumb: Breadcrumb;
-  active: boolean;
 }
-const Fragment: React.VFC<FragmentProps> = ({ breadcrumb, active }) => {
+const BreadcrumbFragment: React.VFC<BreadcrumbFragmentProps> = ({
+  breadcrumb,
+}) => {
   const { text, to, onClick, ...rest } = breadcrumb;
```
minnakt, line 38: I think you still have to guard against the `text` being `undefined`. 

Screen Shot 2022-10-11 at 9 09 12 AM


I think this would work: 
```suggestion
  const { text = "", to, onClick, ...rest } = breadcrumb;
```

src/pages/task/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,59 @@
+import { useBreadcrumbAnalytics } from "analytics";
```
minnakt, line 1: should the breadcrumbs folder name be lowercase?

Screen Shot 2022-10-11 at 9 16 03 AM

src/components/Breadcrumbs/Breadcrumbs.test.tsx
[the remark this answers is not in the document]
khelif96: ahh ok yeah that makes sense this only imports app code in jest tests we totally shouldn't do that for e2e tests

src/pages/task/Breadcrumbs/index.tsx
khelif96: I wish we went with the same pattern we use for the components folder capital folder names and all the logic in the folder. :/

 But i guess i can lower case it now for consistency but we WILL change it in the future.

src/pages/task/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  const commitBreadcrumb = {
+    text: shortenGithash(revision),
+    "data-cy": "bc-version",
+  };
+
+  const breadcrumbs: Breadcrumb[] = [
+    breadcrumbRoot,
+    messageBreadcrumb,
+    isPatch ? patchBreadcrumb : commitBreadcrumb,
```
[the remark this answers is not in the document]
khelif96: yeah its a little confusing tbh i'm not sure if we should just omit the patchBreadcrumb or not it does technically contain different info since it shows the patch number or commit hash which isn't otherwise visible.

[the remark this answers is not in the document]
minnakt: could you combine the `messageBreadcrumb` + `patchBreadcrumb` into one thing?

src/pages/task/Breadcrumbs/index.tsx
minnakt: do we have a ticket for this? if not, would you mind writing a ticket 🙏

khelif96: https://jira.mongodb.org/browse/EVG-18051 Just opened!

src/components/Breadcrumbs/Breadcrumbs.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+      expect(screen.getByDataCy("breadcrumb-tooltip")).toBeInTheDocument();
+    });
+    expect(screen.getByText(longMessage)).toBeInTheDocument();
+  });
+  it("should not display a tooltip if the text is short", async () => {
+    const shortMessage = "short";
+    const breadcrumbs = [{ text: shortMessage }];
+    render(<Breadcrumbs breadcrumbs={breadcrumbs} />);
+    expect(screen.getByText(shortMessage)).toBeInTheDocument();
+    userEvent.hover(screen.getByText(shortMessage));
```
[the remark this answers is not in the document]
khelif96: We are on `@testing-library/user-event@12.5.0` .setup was introduce is 14 :(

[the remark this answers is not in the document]
minnakt: my reaction upon learning this information: 😦😧😨

minnakt (approved):
lgtm I think the breadcrumb hierarchy makes a lot more sense now! 🍞

PR #1463, opened 2022-10-12
EVG-17080: Fix e2e failure caused by patch trigger alias

minnakt:
EVG-17080

### Description
(Disclaimer: I don’t know if this is the correct solution. But I ran this test in a [patch](https://spruce.mongodb.com/task/spruce_ubuntu1604_e2e_test_patch_126f75e0b9637086da6925747c8cd7c43c6169cd_63457fbf30661531116aee10_22_10_11_14_37_57/tests?execution=19&sortBy=STATUS&sortDir=ASC) 20 times and the only time it failed was because of an unrelated test.)

This PR makes it so that all toasts in `project_settings.ts` are automatically closed.

I noticed that there would always be a toast covering the `Add to GitHub Trigger Alias` checkbox. Because the toast is covering the checkbox, it's impossible to confirm in the Cypress video whether it has been properly checked or not. I was thinking that the toast might be causing us to sometimes check the checkbox, and sometimes not.

In any case, I think it is a good idea to automatically close the toasts. A lot of the toasts in the project settings say the same thing (e.g. `Successfully updated project`), so we could potentially get false positives because a toast is persisting for several seconds. 

### Screenshots
- No visible changes.

### Testing
- N/A

### DIFF
--- a/cypress/integration/projectSettings/project_settings.ts
+++ b/cypress/integration/projectSettings/project_settings.ts
@@ -46,7 +46,7 @@ describe("Access page", () => {
       .first()
       .type("admin");
     cy.dataCy("save-settings-button").should("be.enabled").click();
-    cy.validateToast("success", "Successfully updated project");
+    cy.validateToast("success", "Successfully updated project", true);
     cy.visit(destination);
     cy.get("label")
       .contains("Private")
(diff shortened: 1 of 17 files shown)


cypress/integration/projectSettings/project_settings.ts
```diff
@@ -46,7 +46,7 @@ describe("Access page", () => {
       .first()
       .type("admin");
     cy.dataCy("save-settings-button").should("be.enabled").click();
-    cy.validateToast("success", "Successfully updated project");
+    cy.validateToast("success", "Successfully updated project", true);
```
sophstad, line 49: Nice work, this makes a lot of sense!

Did you consider updating the `validateToast` command to close by default? It seems like that could help avoid this issue in the future since it proved a bit mysterious.

minnakt: great suggestion! done! 😼

sophstad (approved):
💯

PR #1468, opened 2022-10-17
EVG-17488: Migrate Modal

SupaJoon:
[EVG-17488](https://jira.com/browse/EVG-17488)

The Migrate Modal is the same as the Spawn Host Modal but the volume section is hidden and prefilled. Open a Migrate Modal here: http://localhost:3000/spawn/volume
Screen Shot 2022-10-19 at 3 25 47 PM


src/pages/spawn/spawnHost/SpawnHostModal.tsx
```diff
@@ -177,20 +197,31 @@ export const SpawnHostModal: React.VFC<SpawnHostModalProps> = ({
...
+      spawnHostMutation({
+        variables: { SpawnHostInput: mutationInput },
+      });
+    }
   };
-
+  const loadingSubmit = loadingSpawnHost || loadingMigration;
   return (
     <DisplayModal
       title="Spawn New Host"
```
khelif96, line 227: The Title should be Migrate Volume

src/pages/spawn/spawnHost/SpawnHostModal.tsx
```diff
@@ -119,7 +124,21 @@ export const SpawnHostModal: React.VFC<SpawnHostModalProps> = ({
     },
     refetchQueries: ["MyHosts", "MyVolumes", "GetMyPublicKeys"],
```
khelif96, line 125: One thing i noticed is that all the queries for this modal are loaded on page load despite the spawn host modal never being opened. We can avoid this using the [`skip`](https://www.apollographql.com/docs/react/api/react/hooks#skip) command on the queries.  Particularly GET_DISTROS, GET_MY_VOLUMES, GET_MY_PUBLIC_KEYS,

src/pages/spawn/spawnHost/SpawnHostModal.tsx
```diff
@@ -217,12 +248,15 @@ export const SpawnHostModal: React.VFC<SpawnHostModalProps> = ({
...
+          disabled={
+            !validateSpawnHostForm(formState, !!migrateVolumeId) ||
+            loadingSubmit
+          } // @ts-expect-error
           onClick={spawnHost}
           variant={Variant.Primary}
           key="spawn_host_button"
         >
-          {loadingSpawnHost ? "Spawning Host" : "Spawn"}
+          {loadingSubmit ? "Spawning Host" : "Spawn"}
```
khelif96, line 259: The designs actually prompt you to hit next which directs you to a confirmation modal. 
image
image

src/pages/spawn/spawnVolume/spawnVolumeTable/SpawnVolumeTableActions.tsx
```diff
@@ -1,38 +1,53 @@
...
-      data-cy={`trash-${volume.displayName || volume.id}`}
-      volume={volume}
-    />
-    {volume.host ? (
-      <UnmountBtn
-        data-cy={`unmount-${volume.displayName || volume.id}`}
+export const SpawnVolumeTableActions: React.VFC<Props> = ({ volume }) => {
+  const { data: myHostsData } = useQuery<MyHostsQuery, MyHostsQueryVariables>(
+    GET_MY_HOSTS
+  );
```
khelif96, line 20: Although apollo is smart enough to not duplicate this query can we still move it up and pass it in as a prop.

src/pages/spawn/spawnVolume/spawnVolumeTable/spawnVolumeTableActions/MigrateBtn.tsx
```diff
@@ -0,0 +1,32 @@
...
+        size={Size.XSmall}
+        data-cy={`migrate-btn-${volume.displayName || volume.id}`}
+        onClick={(e) => {
+          e.stopPropagation();
+          setOpenModal(true);
+        }}
+      >
+        Migrate
+      </Button>
+      <SpawnHostModal
```
khelif96, line 25: We should probably extract the shared logic into a separate component and have a spawn host component and migrate component that abstracts the shared logic.

khelif96 (changes requested):

src/pages/spawn/spawnVolume/spawnVolumeTable/spawnVolumeTableActions/MigrateBtn.tsx
sophstad: +1, would allow for more flexibility in the future and also facilitate converting the `EditSpawnHostModal` component to use a SpruceForm. There are a few other changes to make for the migration modal:

- Only allow selection of virtual workstation distros as per our Slack discussion
- Change the modal heading to match designs
- Add some modal help text: "Migrate this home volume to a new host. Upon successful migration, the unused host will be scheduled to expire in 24 hours."

src/pages/spawn/spawnVolume/spawnVolumeTable/SpawnVolumeTableActions.tsx
```diff
@@ -1,38 +1,53 @@
...
+  return (
+    <FlexRow>
+      <DeleteVolumeBtn
+        data-cy={`trash-${volume.displayName || volume.id}`}
         volume={volume}
       />
-    ) : (
-      <MountBtn
-        data-cy={`mount-${volume.displayName || volume.id}`}
+      {showMigrateBtn && <MigrateBtn volume={volume} />}
```
sophstad, line 32: What do you think about hiding the mount/unmount buttons for Virtual Workstations and instead showing just the "Migrate" button? It seems to me like they are basically enabling the same functionality for virtual workstations vs standard hosts (i.e. the ability to move your volume)

src/pages/spawn/spawnHost/SpawnHostModal.tsx
```diff
@@ -119,7 +124,21 @@ export const SpawnHostModal: React.VFC<SpawnHostModalProps> = ({
     },
     refetchQueries: ["MyHosts", "MyVolumes", "GetMyPublicKeys"],
   });
-
+  const [migrateVolumeMutation, { loading: loadingMigration }] = useMutation<
+    MigrateVolumeMutation,
+    MigrateVolumeMutationVariables
+  >(MIGRATE_VOLUME, {
+    onCompleted() {
+      dispatchToast.success("Migrated volume onto host.");
```
sophstad, line 132: A successful response from the GraphQL mutation only indicates that the migration job was added to the queue. Something like this would be more accurate/helpful:

"Volume migration has been scheduled. A new host will be spawned and accessible on your Hosts page"

src/pages/spawn/spawnHost/SpawnHostModal.tsx
sophstad: Agreed, another approach I use sometimes is conditionally rendering the modal based on the open state in the parent component. E.g.

```js
{open && <Modal {...props} />
```

This has the added benefit of resetting any form states upon closing the modal (which is generally expected behavior for modals IMO) since the component is removed from the DOM.

sophstad (changes requested):

src/components/Spawn/spawnHostModal/useVirtualWorkstationDefaultExpiration.ts
```diff
@@ -0,0 +1,30 @@
+import { useEffect } from "react";
+import { usePrevious } from "hooks";
+import { getDefaultExpiration } from "./getFormSchema";
+
+export const useVirtualWorkstationDefaultExpiration = ({
```
sophstad, line 5: During a migration, we should permit users to create a new expirable host even if they already have reached `unexpirableHostsPerUser`, since the current host will be set to expire shortly. I think this hook may be a good place to handle that!

sophstad (changes requested):

src/components/Spawn/spawnHostModal/useVirtualWorkstationDefaultExpiration.ts
khelif96: hmm. I would be cautious about this assumption/feature. This seems like an easy way to run into a scenario where a user can spawn unlimited virtual workstations by just creating new volumes and then updating the expiration time on the old hosts. Unless I am missing something?

src/pages/spawn/spawnVolume/spawnVolumeTable/spawnVolumeTableActions/migrateBtn/MigrateVolumeModal.tsx
```diff
@@ -0,0 +1,166 @@
...
+      onCancel={
+        submitClickCount === 0
+          ? () => setOpen(false)
+          : () => setSubmitClickCount(0)
+      }
+    >
+      <Body>
+        Migrate this home volume to a new host. Upon successful migration, the
+        unused host will be scheduled to expire in 24 hours.
+      </Body>
```
khelif96, line 153: Is this supposed to show up twice? 
image

image

src/pages/spawn/spawnVolume/spawnVolumeTable/spawnVolumeTableActions/migrateBtn/MigrateVolumeModal.tsx
```diff
@@ -0,0 +1,166 @@
...
+  open: boolean;
+  setOpen: React.Dispatch<React.SetStateAction<boolean>>;
+}
+
+export const MigrateVolumeModal: React.VFC<MigrateVolumeModalProps> = ({
+  open,
+  setOpen,
+  volume,
+}) => {
+  const [submitClickCount, setSubmitClickCount] = useState(0);
```
khelif96, line 35: I'm not a big fan of this pattern.  I think it might make sense to use something like `useReducer` and treat this more like a state machine instead of just counting the number of times the user clicked the submit button.

`useReducer` would be more verbose about what is actually happening to the modal state and would make the code more understandable since you would be able to correspond a dispatched action to the state of the modal instead of relying on an index.

src/pages/spawn/spawnHost/spawnHostButton/SpawnHostModal.tsx
```diff
@@ -0,0 +1,139 @@
...
+  setOpen,
+}) => {
+  const dispatchToast = useToastContext();
+  const spawnAnalytics = useSpawnAnalytics();
+
+  // Handle distroId, taskId query param
+  const { search } = useLocation();
+  const queryParams = parseQueryString(search);
+  const taskIdQueryParam = getString(queryParams.taskId);
+  const distroIdQueryParam = getString(queryParams.distroId);
```
khelif96, line 43: Seeing this makes me excited to bring in the `useQueryParam` hooks we have In parsley.

khelif96 (changes requested):
This is looking good so far just a few more suggestions and questions

src/components/Spawn/spawnHostModal/useVirtualWorkstationDefaultExpiration.ts
sophstad: Good point -- I don't think we need to assume super malicious user intent (especially since the back end doesn't even prevent this) but we can make it a little tougher to spawn infinite hosts. We could permit this only on the migration modal if the user has unexpiredHostCount <= unexpirableHostsPerUser (as opposed to just <), thereby allowing them to create 1 more unexpirable host but not opening it up entirely. cc @SupaJoon

src/pages/spawn/spawnVolume/spawnVolumeTable/spawnVolumeTableActions/MigrateBtn.tsx
```diff
@@ -0,0 +1,34 @@
...
+interface Props {
+  volume: MyVolume;
+}
+
+export const MigrateBtn: React.VFC<Props> = ({ volume }) => {
+  const [openModal, setOpenModal] = useState(false);
+
+  return (
+    <>
+      <Button
```
sophstad, line 15: This button should be disabled while the volume is migrating
image

src/components/Spawn/spawnHostModal/useVirtualWorkstationDefaultExpiration.ts
sophstad: Ah, and we prevent editing a host to be unexpirable if 2 already exist :)
image

sophstad (approved):
Agree with Mohamed's notes about the modal state but otherwise looks great! Exciting 😁

src/pages/spawn/spawnVolume/spawnVolumeTable/spawnVolumeTableActions/migrateBtn/migrateVolumeModal/reducer.ts
```diff
@@ -0,0 +1,37 @@
...
+        page: state.page === Page.First ? Page.Second : Page.First,
+      };
+    case "resetPage":
+      return { ...state, page: Page.First };
+    case "setForm":
+      return { ...state, form: action.payload };
+    case "resetForm":
+      return { ...state, form: {} };
+    default:
+      throw new Error();
```
khelif96, line 29: Lets add an error message here just so it shows up in bugsnag.
```suggestion
      throw new Error(`Unknown reducer action ${action}`);
```

khelif96 (approved):
Great work on this I'm excited to see it released!

khelif96:
Try merging main to fix the failures. If not can you run a set-module patch to ensure there are no failures on your patch.

PR #1471, opened 2022-10-20
EVG-17248: Surface downstream failures on patch page

minnakt:
EVG-17248

### Description
This PR shows a badge on the Downstream Tasks tab indicating how many child patches have failed. Additionally, failed child patches will be automatically expanded.

I made the badge indicate how many child patches failed rather than how many tasks failed. I didn't think it was really worth querying for the tasks because it would have to go through the expensive `VersionTasks` pipeline.

### Screenshots
Screen Shot 2022-11-04 at 3 16 22 PM

### Testing
- Added tests in `version/downstream_tasks.ts`


minnakt:
evergreen retry

cypress/integration/version/routes.ts
```diff
@@ -59,19 +59,25 @@ describe("Version route", () => {
     });
 
     it("Lists the patch's build variants", () => {
-      cy.dataCy("patch-build-variant").within(
-        ($variants) => Array.from($variants).length > 0
-      );
+      cy.dataCy("build-variants").within(() => {
+        cy.dataCy("patch-build-variant").within(
+          // @ts-expect-error
+          ($variants) => Array.from($variants).length > 0
```
khelif96, line 65: Why do we expect error here? Is the typing wrong?

src/components/GroupedTaskStatusBadge/index.tsx
```diff
@@ -1,45 +1,35 @@
...
-  status: TaskStatus;
+interface GroupedTaskStatusBadgeProps {
   count: number;
   onClick?: () => void;
+  status: TaskStatus;
   statusCounts?: { [key: string]: number };
   versionId: string;
-  variant?: string;
-  preserveSorts?: boolean;
+  queryParamsToPreserve?: GetVersionRouteOptions;
```
khelif96, line 20: Is this still necessary now that we no longer link from the side panel?

src/pages/Version.tsx
```diff
@@ -31,10 +31,12 @@ import { PageDoesNotExist } from "pages/404";
...
-import { BuildVariants } from "./version/BuildVariants";
-import { ActionButtons } from "./version/index";
-import { Metadata } from "./version/Metadata";
-import { Tabs } from "./version/Tabs";
+import {
+  ActionButtons,
+  BuildVariants,
+  Metadata,
+  VersionTabs,
+} from "./version/index";
```
khelif96, line 39: an import from ./version should work the same way?
```suggestion
} from "./version";
```

src/pages/version/BuildVariants.tsx
```diff
@@ -44,66 +44,99 @@ export const BuildVariants: React.VFC = () => {
...
-                onClick={() =>
-                  sendEvent({
-                    name: "Click Build Variant Grid Link",
-                  })
-                }
-                data-cy="build-variant-display-name"
-              >
-                {displayName}
-              </StyledRouterLink>
+        <div data-cy="build-variants">
```
khelif96, line 47: Same comment as above.

cypress/integration/projectSettings/project_settings.ts
```diff
@@ -28,21 +28,22 @@ describe("Access page", () => {
...
   });
 
   it("Changing settings and clicking the save button produces a success toast and the changes are persisted", () => {
-    cy.get("label").contains("Private").click();
+    cy.getInputByLabel("Private").parent().click();
     cy.getInputByLabel("Private").should("have.attr", "aria-checked", "true");
-    cy.get("label").contains("Unrestricted").click();
+
+    cy.getInputByLabel("Unrestricted").parent().click();
     cy.getInputByLabel("Unrestricted").should(
```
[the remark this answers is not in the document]
minnakt: Hmm, I don't see these changes on my branch. It's possible you reviewed an old version of this PR 🙂

src/pages/Version.tsx
minnakt: It doesn't work 😟... I think it might be getting confused between `Version` and `version`

ablack12 (changes requested):
If we're going to have the badge show number of failing downstream patches instead of number of downstream tasks, then I think the name of tab needs to be Downstream Patches instead. cc @khelif96

src/components/GroupedTaskStatusBadge/index.tsx
minnakt: I kind of like this pattern better, since it's more flexible 😅 I originally wrote the `preserveSorts` stuff and I don't really like it

khelif96:
> If we're going to have the badge show number of failing downstream patches instead of number of downstream tasks, then I think the name of tab needs to be Downstream Patches instead. cc @khelif96
> 
> Alternatively, it may be worth doing an expensive query (although I didn't look into the details here) since we'd only need to do this in the failing case but I'm not sure what pipeline this is

The name change makes sense to me. The expensive query is doable as well but I think the name change is more clear.

ablack12:
Actually maybe Downstream Projects? Downstream Patches may be unclear to users since it's displayed as one patch; that's more internal knowledge.

khelif96 (approved):

PR #1477, opened 2022-10-25
EVG-17477 Update placeholder and error text

malikchaya2:
EVG-17477 

### Description
Just updating some text to include memberID
image

### DIFF
--- a/src/components/Notifications/form/notification.ts
+++ b/src/components/Notifications/form/notification.ts
@@ -162,7 +162,7 @@ export const getNotificationSchema = (
       "ui:data-cy": "jira-comment-input",
     },
     slackInput: {
-      "ui:placeholder": "@user or #channel",
+      "ui:placeholder": "MEMBERID, @user or #channel",
       "ui:data-cy": "slack-input",
     },
     emailInput: {
--- a/src/components/Notifications/form/notification.ts
+++ b/src/components/Notifications/form/notification.ts
@@ -162,7 +162,9 @@ export const getNotificationSchema = (
       "ui:data-cy": "jira-comment-input",
     },
     slackInput: {
-      "ui:placeholder": "MEMBERID, @user or #channel",
+      "ui:description":
+        "Notifications can be sent to a Slack channel, @user, or member ID represented as an alphanumeric string.",
+      "ui:placeholder": "#channel, @user, or MEMBERID",
       "ui:data-cy": "slack-input",
     },
     emailInput: {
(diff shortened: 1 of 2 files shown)


src/components/SpruceForm/errors.ts
```diff
@@ -74,7 +74,8 @@ export const transformErrors = (errors: AjvError[]) =>
             case "validSlack":
               return {
                 ...error,
-                message: "Value should be a valid Slack username or channel.",
+                message:
+                  "Value should be a valid Slack memberID, Slack username or channel.",
```
minnakt, line 78:
```suggestion
                  "Value should be a valid Slack member ID, Slack username or channel.",
```

src/components/SpruceForm/errors.ts
```diff
@@ -74,7 +74,8 @@ export const transformErrors = (errors: AjvError[]) =>
             case "validSlack":
```
minnakt, line 74: Because we have this function `validSlack`, I don't think users will actually be able to submit a subscription using a Slack member ID. Do you want to try fixing it in this ticket (which will require updating the regex expression found [here](https://github.com/evergreen-ci/spruce/blob/2a281b272efeb91c38b653e4ccbc1e63b1266835/src/utils/validators/index.ts#L46))? Or if you want to do it separately, that's ok too :)

src/components/Notifications/form/notification.ts
```diff
@@ -162,7 +162,7 @@ export const getNotificationSchema = (
       "ui:data-cy": "jira-comment-input",
     },
     slackInput: {
-      "ui:placeholder": "@user or #channel",
+      "ui:placeholder": "MEMBERID, @user or #channel",
```
sophstad, line 165:
```suggestion
      "ui:placeholder": "Member ID, @user or #channel",
```

malikchaya2: I wasn't sure about this one. We tend to display them with hints for the format, such as @user and #channel. The member ID will look something like this: UJKMY34TL. I was trying to find a way to hint that to the user. Any ideas?

src/components/SpruceForm/errors.ts
malikchaya2: good point, I'll create a ticket for it.

src/components/Notifications/form/notification.ts
sophstad: Oooh, that sounds like a good candidate for some description text maybe? Something like `Notifications can be sent to a Slack channel, @user, or member ID represented as an alphanumeric string.`?

minnakt (approved):
Awesome! Personally, I feel okay about the hint being `MEMBERID` ^^

PR #1482, opened 2022-10-28
EVG-18145: Upgrade LG Badge

sophstad:
EVG-18145

### Description

- I opted to get rid of our own `StyledBadge` component, which made the letters lowercase. I noticed that our Figma always uses the badge as designed, so we may as well standardize. Let me know if you don't like the new look though 👀  

### Screenshots

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
     "@emotion/eslint-plugin": "11.0.0",
     "@emotion/react": "11.1.4",
     "@emotion/styled": "11.0.0",
-    "@leafygreen-ui/badge": "4.0.5",
+    "@leafygreen-ui/badge": "7.0.0",
     "@leafygreen-ui/banner": "3.0.9",
     "@leafygreen-ui/button": "12.0.5",
     "@leafygreen-ui/card": "5.1.3",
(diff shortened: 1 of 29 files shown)


sophstad:
evergreen retry

package.json
```diff
@@ -62,7 +62,7 @@
     "@emotion/eslint-plugin": "11.0.0",
     "@emotion/react": "11.1.4",
     "@emotion/styled": "11.0.0",
-    "@leafygreen-ui/badge": "4.0.5",
+    "@leafygreen-ui/badge": "7.0.0",
     "@leafygreen-ui/banner": "3.0.9",
```
minnakt, line 66: (placement of comment is random) I think it might make sense to also remove the excess padding on the badges that show under Failure Details. Here's the [patch on staging](http://localhost:3000/task/evg_coverage_coverage_patch_8f1a1622b544421afa3f29fb7d80376101bb28c8_5f6babf297b1d36af9f74b05_21_05_26_20_26_39/annotations?execution=0) I was looking at for this. I think the styling exists in two places, `BBComponents.tsx` and `AnnotationTicketRow.tsx`.

Screen Shot 2022-10-31 at 3 00 00 PM

minnakt (approved):
thank you for fixing the spelling of accordion 🙏

src/components/TaskStatusBadge/__snapshots__/TaskStatusBadge.stories.storyshot
```diff
@@ -8,7 +8,7 @@ exports[`storybook Storyshots Components/Task Status Badges Default 1`] = `
     className="css-11yt18y-Wrapper e11i4bjx0"
```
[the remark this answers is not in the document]
sophstad: loves it!

package.json
sophstad: Good catch! I think the `StyledBadge` margin in `AnnotationTicketRow.tsx` only applies spacing between badges but I'll update `BBComponents` 👍

PR #1483, opened 2022-10-31
EVG-18142 Expand the validateSlack regex 

malikchaya2:
EVG-18142

### Description
This expands the regex check to allow users to specify a memberId. I can't restrict the regex further because according to slack [documentation](https://api.slack.com/types/user), "We recommend considering the format of the string to be an opaque value, and not to rely on a particular structure."

### Testing
[Tested here](https://regex101.com/r/LxrDtf/1)


src/utils/validators/index.ts
```diff
@@ -43,7 +43,8 @@ const validatePercentage = (percent: string) => {
   return true;
 };
 
-const validateSlack = (v: string): boolean => v.match("(#|@).+") !== null;
+const validateSlack = (v: string): boolean =>
+  v.match("(#|@).+|(^S+$)") !== null;
```
minnakt, line 47: Not sure why we didn't already have them, but could you add some tests for this function in `validators.test.ts`? these tests are really easy to write and should only take like 2 minutes 🥺

malikchaya2: done

malikchaya2: I also narrowed down the regex to the more general one

src/utils/validators/index.ts
```diff
@@ -43,8 +43,10 @@ const validatePercentage = (percent: string) => {
   return true;
 };
 
-const validateSlack = (v: string): boolean =>
-  v.match("(#|@).+|(^S+$)") !== null;
+const validateSlack = (v: string): boolean => {
```
khelif96, line 46: Would you mind adding some test cases to `validators.test.ts`

cypress/integration/subscription_modal.ts
```diff
@@ -67,7 +67,7 @@ const testSharedSubscriptionModalFunctionality = (
       it("has an invalid slack username", () => {
         selectAntdOption("notification-method-select", "Slack");
         cy.dataCy("slack-input").clear();
-        cy.dataCy("slack-input").type("sart");
+        cy.dataCy("slack-input").type("sa rt");
```
minnakt, line 70: I think it would make sense if `sart` continued to fail the regex, considering the member ID is all caps (plus numbers) and we even indicate that in the hint. I think one regex that would work is `/(#|@).+|(^[A-Z0-9]+$)/`, not sure if it can be simplified further.

minnakt (approved):
LGTM, but would prefer if we could narrow down the regex a little more (see comment)

cypress/integration/subscription_modal.ts
malikchaya2: I know, I also wish we could narrow it down more. The problem is that while they seem to be all caps right now, slack seems to be suggesting against relying on it, as they provide no guarantees for the format. 

 "We recommend considering the format of the string to be an opaque value, and not to rely on a particular structure."

PR #1486, opened 2022-11-01
EVG-18154: Upgrade tabs

sophstad:
EVG-18154

### Description

- Upgrade tabs. Our previous approach to StyledTabs started throwing a type error, so I opted to go with a more lightweight component to apply padding beneath tabs. If this styling breaks due to an LG change, it will be an easy fix and wouldn't dramatically affect the look of Spruce, so I think it's preferable to a more complex component.
- Applied some other small fixes to standardize styling/spacing below tabs

### Screenshots

image


src/components/styles/StyledTabs.tsx
```diff
@@ -1,23 +1,10 @@
...
-          </Tab>
-        )
-    )}
-  </Tabs>
-);
-
-const PaddedContainer = styled.div`
-  margin-top: 12px;
+// @ts-expect-error
+export const StyledTabs = styled(Tabs)`
```
minnakt, line 6: You can get rid of the `ts-expect-error` using this type

```
export type TabsType = React.ComponentType<
  Omit<React.ComponentProps<typeof Tabs>, "as">
>;
```

minnakt (approved):
they look so small and cuteeeeeee huge fan

PR #1498, opened 2022-11-02
EVG-18060: Allow searching by repo name in ProjectSelect

minnakt:
EVG-18060

### Description
This PR makes it so that you can search by the repo name in the `ProjectSelect` used on the Project Settings pages. You can also click on the repo option.

You will also be able to search by repo name in the `ProjectSelect` used on the Project Health page, but the option will not be clickable.

### Screenshots

Screen Shot 2022-11-02 at 4 56 05 PM

### Testing
- Reorganized tests in `ProjectSelect.test.tsx`
- Added test for searching by repo name and checking clickability of the option
- Removed test case `"should toggle dropdown when clicking on it"` because that is double testing the functionality of `SearchableDropdown`, which is not necessary.
- e2e test failures are unrelated, the PR ran against the wrong Evergreen commit


sophstad (approved):
Meh looks like your commit is stuck using a broken version of evergreen -_- but LGTM, looks like any tests this change would touch are passing.

minnakt:
evergreen retry

PR #1500, opened 2022-11-03
EVG-17490: Show guide cue for Migrate button first time experience

SupaJoon:
[EVG-17490](https://jira.com/browse/EVG-17490)

### Description
Show the guide cue before the user interacts with the Migrate button

### Screenshots
Screen Shot 2022-11-03 at 4 13 07 PM


khelif96:
The commit history on this pr looks really messed up would you mind rebasing before we review it? it's difficult to tell which changes are yours and which are from the messed up commit history.

minnakt (commented):
it's cool that we went with the guide cue instead of the banner ^^

cypress/integration/spawn/volume.ts
```diff
@@ -184,26 +184,37 @@ describe("Navigating to Spawn Volume page", () => {
   });
 
   describe("Migrate Modal", () => {
-    it("open the Migrate modal and spawn a host", () => {
+    const cueCopy = "You can now migrate your home volume to a new spawn host!";
```
minnakt, line 187: In this case I think it's better to use `data-cy` then, since the dialog role is too ambiguous. I'm not really a huge fan of looking at the text to check that the guide cue is gone. I checked and it looks like `data-cy` will properly be applied to the element:

Screen Shot 2022-11-04 at 3 38 50 PM

package.json
```diff
@@ -71,10 +71,11 @@
     "@leafygreen-ui/confirmation-modal": "2.2.2",
     "@leafygreen-ui/emotion": "^4.0.0",
     "@leafygreen-ui/expandable-card": "0.9.1",
+    "@leafygreen-ui/guide-cue": "^2.0.1",
     "@leafygreen-ui/icon": "11.4.0",
     "@leafygreen-ui/icon-button": "9.1.7",
     "@leafygreen-ui/inline-definition": "2.0.6",
-    "@leafygreen-ui/leafygreen-provider": "2.1.3",
+    "@leafygreen-ui/leafygreen-provider": "^3.0.1",
```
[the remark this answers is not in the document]
minnakt: reminder to pin (let's discuss more in the meeting!)

src/pages/spawn/spawnVolume/SpawnVolumeTable.tsx
```diff
@@ -24,42 +25,54 @@ export const SpawnVolumeTable: React.VFC<SpawnVolumeTableProps> = ({
...
+    const volumesCopy = [...volumes];
+    volumesCopy.sort(sortByHost);
+    const firstMigrateableVolumeId =
+      volumesCopy.find((v) => v.homeVolume)?.id ?? "";
+    return volumes.map((v) => ({
+      ...v,
+      showMigrateBtnCue:
+        v.id === firstMigrateableVolumeId &&
+        Cookies.get(SEEN_MIGRATE_GUIDE_CUE) !== "true",
+    }));
```
minnakt, line 38: (nothing to change, just a random comment) I feel like we could skip all of this work if `SEEN_MIGRATE_GUIDE_CUE` is true, but I also don't think it's a big deal since people don't usually have that many volumes

minnakt (approved):
yay! 🎉

PR #1501, opened 2022-11-04
EVG-14911: Indicate that logs shown on task pages are log tails

minnakt:
EVG-14911

### Description
This PR implements a small redesign for the log headings on the task details pages. The redesign is meant to indicate to new users that the logs they see on the task pages are just log tails, and not the entire log.

### Screenshots

Screen Shot 2022-11-04 at 10 18 05 AM

### Testing
- Updated `task_logs.ts` (Cypress)


src/pages/task/taskTabs/logs/LogTypes.tsx
```diff
@@ -192,8 +194,8 @@ const useRenderBody: React.VFC<{
   const taskAnalytics = useTaskAnalytics();
   const updateQueryParams = useUpdateURLQueryParams();
   const noLogs = !!((error && !data) || !data.length);
-  const onChangeLog = (event: React.ChangeEvent<HTMLInputElement>): void => {
-    const nextLogType = event.target.value as LogTypes;
+  const onChangeLog = (value: string): void => {
```
sophstad, line 197: > Because we are on an old component version, it is a little wonky, but not unusable by any means.

I actually am not sure that this is why! I believe the animation is a little messed up by how the `currentLog` state is managed. I noticed that if you console.log `currentLog`, clicking on a new log type prints/rerenders the initial log type before printing the new type. Maybe that's related to how the query param hook works? Regardless, when I set up a simpler state with the component, it transitions smoothly. E.g.

```js
const [currentLog, setCurrentLog] = useState("all");

return (
  <SegmentedControl onChange={setCurrentLog} value={currentLog}>
    {...}
  </SegmentedControl>
);
```

Not sure this is worth the effort to fix but I wanted to flag! Maybe you have some tricks up your sleeve 😉

minnakt: You are so right, the fool doth think he is wise 😔

I ended up moving the LogHeader (i.e. the segmented control options, the buttons) up to the parent component because otherwise they re-render with every log type change, which I don't think makes a lot of sense.

src/pages/task/taskTabs/logs/LogTypes.tsx
```diff
@@ -174,32 +172,20 @@ const useRenderBody: React.VFC<{
...
-      name: "Select Logs Type",
-      logsType: nextLogType,
-    });
-  };
+  // Update the value of noLogs in the parent component.
+  useEffect(() => {
+    setNoLogs(noLogs);
+  }, [setNoLogs, noLogs]);
 
   let body = null;
```
sophstad, line 190: [nit] We may as well just return each of these JSX chunks rather than assign them to `body`

src/pages/task/taskTabs/logs/LogTypes.tsx
```diff
@@ -174,32 +172,20 @@ const useRenderBody: React.VFC<{
...
-    updateQueryParams({ [QueryParams.LogType]: nextLogType });
-    taskAnalytics.sendEvent({
-      name: "Select Logs Type",
-      logsType: nextLogType,
-    });
-  };
+  // Update the value of noLogs in the parent component.
+  useEffect(() => {
+    setNoLogs(noLogs);
+  }, [setNoLogs, noLogs]);
```
sophstad, line 188: All of the information needed to calculate `noLogs` is present in the parent component, might it make more sense to set that variable there rather than in this hook?

sophstad (approved):
so smoooooth 🧈

src/pages/task/taskTabs/logs/LogTypes.tsx
minnakt: I think I need the results of the query to determine if there are no new logs (i.e. `data_from_query.length === 0`) 😔 It could also be done through Apollo's `onCompleted` or `onError`, though that becomes a little verbose 😅

I also feel like maybe it could be done through a larger refactor 🤔 I explored it a little but it was too big for this PR. I will think about it more and maybe open a ticket for it

sophstad: Oh yes I must've misread! sorry, sgtm 👍

PR #1503, opened 2022-11-07
EVG-18174: Upgrade Menu

sophstad:
EVG-18174

### Description

- Another easy one, but I think the upgrade introduced one small type regression

### Screenshots

image


src/components/Header/NavDropdown.tsx
```diff
@@ -33,6 +33,9 @@ const DropdownItem: React.VFC<DropdownItemType> = ({
 }) => (
   <MenuItem
     as={to && Link}
+    // LG typing should permit props associated with the `as`
+    // component, but right now it doesn't. ¯\_(ツ)_/¯
+    // @ts-expect-error
     to={to}
```
khelif96, line 39: Can we raise this with LG and link a ticket here. I feel like this should work

PR #1505, opened 2022-11-08
EVG-18178: Upgrade text input & area

sophstad:
EVG-18178

### Description

- Remove one last fixed-width badge (the rest were updated in #1482)

### Screenshots

image


sophstad:
evergreen retry

minnakt (approved):
LGTM, though I see some console errors which seems to happen when text inputs don't have a label:

Screen Shot 2022-11-08 at 1 45 59 PM

I think it might get fixed in this ticket [LG-2517](https://jira.mongodb.org/browse/LG-2517), so we can try upgrading this library again when it gets released

sophstad:
> LGTM, though I see some console errors which seems to happen when text inputs don't have a label:
> 
> Screen Shot 2022-11-08 at 1 45 59 PM
> 
> I think it might get fixed in this ticket [LG-2517](https://jira.mongodb.org/browse/LG-2517), so we can try upgrading this library again when it gets released

Good idea, I think I'll just put this PR on ice until that fix is released since there's already a PR open!

PR #1506, opened 2022-11-08
EVG-18172: Upgrade Icon & Icon Button

sophstad:
EVG-18172

### Description

- Upgrade Icon & Icon button, no API changes


src/components/EditableTagField/__snapshots__/EditableTagField.stories.storyshot
```diff
@@ -111,7 +111,7 @@ exports[`storybook Storyshots Components/Editable Tags Field Editable Tag Field
         <button
```
minnakt, line 111: Strangely the grouped badges are messed up for me when I look at a version, but not on the My Patches page

Screen Shot 2022-11-08 at 2 15 50 PM

src/components/EditableTagField/__snapshots__/EditableTagField.stories.storyshot
```diff
@@ -111,7 +111,7 @@ exports[`storybook Storyshots Components/Editable Tags Field Editable Tag Field
...
           aria-disabled={false}
```
minnakt, line 112: 😭 I don't know why it looks like this

Screen Shot 2022-11-08 at 2 18 15 PM

minnakt (approved):
changes look good to me! I noticed some visual regressions that may have been caused along the way so I just wanted to document them but they can be fixed some other time!

(also there is a potential for my computer to be messed up so if it doesn't look like this for you please ignore my comments 🙏)

sophstad:
> changes look good to me! I noticed some visual regressions that may have been caused along the way so I just wanted to document them but they can be fixed some other time!

Thanks for spotting! I made a general cleanup ticket because I've found that some visual discrepancies are fixed by other upgrades. I think it may be helpful to make sure we do a sweep at the end! EVG-18234

PR #1509, opened 2022-11-09
EVG-18226: Fix destructuring error in task action buttons

minnakt:
[EVG-18226](https://jira.mongodb.org/browse/EVG-18226)

### Description
It seems like we occasionally encounter a destructuring error because we assume the task is always defined, when it's possible the task hasn't been fully fetched yet.

### Screenshots
- N/A

### Testing
- Manually set the task to be `undefined` and checked that no errors would be thrown


minnakt:
evergreen retry

khelif96 (approved):

PR #1510, opened 2022-11-11
EVG-18125,EVG-18126: Update Modal, Confirmation Modal, & Select

sophstad:
EVG-18125, EVG-18126

### Description

Sorry this is so big 🥵 
- Convert all instances of Antd modals to LeafyGreen modals
  - Ensure DatePicker is clickable on all of these modals by adding a `getPopupContainer` prop
- Convert execution dropdown to LeafyGreen select
- Convert antd Select SpruceForm widget to LeafyGreen (this necessitates adding `allowDeselect: false` to mimic antd's default behavior)
  - Note that this conditional enumDisabled feature is still broken. I've punted it to EVG-NNNNN
  image
- I previously converted InlineCode links to use `href`; convert back to `onClick` using React Router's `navigate()` so that we can leverage the router instead of raw hrefs.


src/components/Spawn/spawnHostModal/getFormSchema.tsx
```diff
@@ -502,26 +501,19 @@ export const getFormSchema = ({
...
-              ? widgets.RadioBoxWidget
-              : "hidden",
-          },
-          volumeSelect: {
-            "ui:widget": isVirtualWorkstation ? AntdSelect : "hidden",
-            "ui:allowDeselect": false,
-            "ui:data-cy": "volume-select",
-            "ui:disabledEnums": (volumes || [])
-              .filter((v) => !!v.hostID)
-              .map((v) => v.id),
```
sophstad, line 518: Remove this line because we already filter out the volumes that have a host ID in L366 so there's no need to disable them here.

sophstad:
Split into #1526 and #1532

PR #1523, opened 2022-11-17
EVG-18175: Upgrade Tooltip and Popover

sophstad:
EVG-18175

### Description

- No API changes. Visually looks pretty much the same too!

### Screenshots

image
image

### Testing
- Unit test flakiness tracked in EVG-18333


package.json
```diff
@@ -79,7 +79,7 @@
     "@leafygreen-ui/menu": "18.0.0",
     "@leafygreen-ui/modal": "6.1.2",
     "@leafygreen-ui/palette": "3.4.4",
-    "@leafygreen-ui/popover": "7.2.2",
+    "@leafygreen-ui/popover": "11.0.0",
     "@leafygreen-ui/radio-box-group": "12.0.0",
     "@leafygreen-ui/radio-group": "10.0.0",
```
minnakt, line 84: Dark purple doesn't look too good inside the tooltip, sadly... we can revisit it later 😔
Screen Shot 2022-11-18 at 4 57 16 PM

minnakt, line 84: This one random tooltip is in `darkMode`, would you mind taking the prop off for it 😅

Screen Shot 2022-11-18 at 5 01 49 PM

package.json
```diff
@@ -79,7 +79,7 @@
...
     "@leafygreen-ui/segmented-control": "5.0.0",
```
minnakt, line 85: (writing this mostly as a reminder to myself) The new tooltips make hovering on the waterfall icons pretty funky, it's possible we need to reduce the delay

minnakt (approved):

package.json
khelif96: Can we ticket this? Cailin also mentioned that highlighting tooltips on the waterfall is too slow for her liking. So if it's even slower we should address it before releasing it.

package.json
sophstad: Good catch—a lot of these icons actually get pretty messed up with the dark background:
image
I'm going to apply dark mode here as a workaround
image

package.json
sophstad: done EVG-18374

PR #1528, opened 2022-11-22
EVG-18171: Update LG Emotion semver

SupaJoon:
[EVG-18171](jira.com/browse/EVG-18171)

### Description
Updating the semver didn't result in a new installation

### DIFF
--- a/src/components/EditableTagField/__snapshots__/EditableTagField.stories.storyshot
+++ b/src/components/EditableTagField/__snapshots__/EditableTagField.stories.storyshot
@@ -30,11 +30,11 @@ exports[`storybook Storyshots Components/Editable Tags Field Editable Tag Field 
               className="leafygreen-ui-lzja97"
             >
               <div
-                className="leafygreen-ui-1i2djk5"
+                className="leafygreen-ui-1cen7zh leafygreen-ui-1ago99h"
               >
                 <input
                   autoComplete="on"
-                  className="leafygreen-ui-1il5lef"
+                  className="leafygreen-ui-pppkef leafygreen-ui-exxjz2"
                   data-cy="user-tag-key-field"
                   disabled={false}
                   id="tag_key_undefined"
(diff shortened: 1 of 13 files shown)


khelif96 (approved):
LGTM looks like this needs to update snapshots!

package.json
```diff
@@ -69,7 +69,7 @@
     "@leafygreen-ui/checkbox": "11.0.1",
     "@leafygreen-ui/code": "13.0.1",
     "@leafygreen-ui/confirmation-modal": "2.2.2",
-    "@leafygreen-ui/emotion": "^4.0.0",
+    "@leafygreen-ui/emotion": "4.0.3",
```
minnakt, line 72: Seems good to me, though we should double check that this console warning goes away after upgrading all of the LeafyGreen packages. (This warning used to exist in Parsley, but it went away after we updated LG components.)

Screen Shot 2022-11-22 at 5 19 56 PM

minnakt (approved):

minnakt:
Oop one more thing -- I think you need to run `yarn`, I noticed that the generated `yarn.lock` file should have changed in this PR

minnakt:
> It's weird that the snapshot updated. If you look at the yarn.lock diff that generated from `yarn`, you'll see that there was no change in the emotion sha. 4.0.3 is the emotion installation in main and in this PR.

I think you could be running into [LG-2179](https://jira.mongodb.org/browse/LG-2179). LeafyGreen temporarily had a function that would generate unique classnames at runtime, meaning the CSS classnames change on every iteration.

We are on `leafygreen-ui/lib@9.1.0` but according to the [CHANGELOG](https://github.com/mongodb/leafygreen-ui/blob/main/packages/lib/CHANGELOG.md) it was only entirely removed in `10.0.0`. You can temporarily mock it out by copying the code in Parsley's `storyshots.test.ts`.

PR #1530, opened 2022-11-22
EVG-16249 Fix error grouping.

khelif96:
EVG-16249

### Description

Updated deploy script to preserve previous source maps so we should start seeing fewer minimized code errors that result from our project.

Passes errors into bugsnag report errors.

Graphql errors are now grouped based on the error message. This should hopefully lead to more useful grouping and reduce the risk of unrelated errors being grouped together.
image


minnakt:
I wonder if it makes sense to change the function signature of `reportError` so that it throws an TS error if we don't pass in a proper error object? 🤔 (so that it doesn't happen again)

khelif96:
> I wonder if it makes sense to change the function signature of `reportError` so that it throws a TS error if we don't pass in a proper error object? 🤔 (so that it doesn't happen again)

afaict this is typed correctly according to the bugsnag types. 
image

image

minnakt:
I guess I'm wondering if we should be stricter with the typing of `CustomBugnagError` and just make it 

```
type CustomBugsnagError = Error & {
  metadata?: any;
};
```

This seems to not throw any TS errors, and also prevent us from making similar mistakes in the future. Tbh I'm a little confused how TS doesn't complain originally since we add a metadata field which doesn't agree with any of the types defined for `NotifiableError` 🤔

sophstad (commented):
I kinda still agree with @minnakt about augmenting the Bugsnag error type, any thoughts on that?

scripts/app-version.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
```
minnakt, line 2: This might be a silly question, but are we calling this script anywhere? On Parsley, we call this script in a few different places, but I can't find any references to this script on Spruce

minnakt (approved):

scripts/app-version.sh
khelif96: This was actually a great question I was using the wrong app-version script. I just updated it. Nice catch!

khelif96:
> type CustomBugsnagError = Error & {
>   metadata?: any;
> };

done

PR #1531, opened 2022-11-23
EVG-13829: Support adding personal subscriptions on projects

minnakt:
EVG-13829

### Description
This PR makes it so that a user can add personal subscriptions onto projects. Note that these notifications differ from the Project Settings notifications because these notifications belong to a single person, while the Project Settings notifications belong to the project.

We don't have a mechanism for showing a user's subscriptions on Spruce so the easiest way to test this is to add the notification and then check the [legacy notifications page](https://evergreen.mongodb.com/notifications) to see what subscriptions are associated with your user.

Other things:
* Fixed some inaccurate things in the form schema. In particular, the `ANY_TASK_FAILS` trigger was missing its associated regex selectors.
* Added `ANY_VERSION_SUCCEEDS` and `ANY_BUILD_SUCCEEDS` events as triggers that can be used on waterfall subscriptions.
* Removed the `allowedSelectors` field as it seems we only allow the specific selectors defined [here](https://github.com/evergreen-ci/evergreen/blob/main/model/event/subscriptions.go#L357-L368).
* Fixed [EVG-18363](https://jira.mongodb.org/browse/EVG-18363)

### Screenshots
https://user-images.githubusercontent.com/47064971/203637432-53018342-9a3f-4cb2-8133-574ebd76c62d.mov

### Testing
- Added test file `utils.test.ts`
- Added Cypress test in `subscription_modal.ts`


minnakt:
If testing the Project Settings page on staging, I don't recommend using the triggers `"Any Version Fails"` or `"Any Version Finishes"` because the page will completely break. It seems like Chaya is doing testing on [EVG-14528](https://jira.mongodb.org/browse/EVG-14528) that causes these two triggers to break if her changes are deployed on staging.

khelif96 (approved):

PR #1535, opened 2022-11-30
EVG-18250: Upgrade Expandable Card

sophstad:
EVG-18250

### Description

- Upgrade expandable card, which is used only in project settings

### Screenshots

image

### DIFF
--- a/src/components/SpruceForm/FieldTemplates.tsx
+++ b/src/components/SpruceForm/FieldTemplates.tsx
@@ -152,7 +152,7 @@ const ArrayItem: React.VFC<
     <StyledExpandableCard
       defaultOpen={!isDisabled}
       data-cy="expandable-card"
-      // Override LeafyGreen's string typing for title so we can include buttons.
+      // Override LeafyGreen's string typing for title so we can include buttons. (LG-2193)
       /* @ts-expect-error */
       title={
         <>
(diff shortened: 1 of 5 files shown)


src/components/SpruceForm/FieldTemplates.tsx
```diff
@@ -145,12 +145,15 @@ const ArrayItem: React.VFC<
       disabled={isDisabled}
       leftGlyph={<Icon glyph="Trash" />}
       data-cy="delete-item-button"
+      size="small"
```
minnakt, line 148: I'm not sure, but this change may have affected the button size in array items (maybe it will be handled by the button PR)

Screen Shot 2022-11-30 at 2 14 51 PM

src/components/SpruceForm/FieldTemplates.tsx
```diff
@@ -145,12 +145,15 @@ const ArrayItem: React.VFC<
...
     />
   );
   return useExpandableCard ? (
     <StyledExpandableCard
       defaultOpen={!isDisabled}
       data-cy="expandable-card"
+      // Override LeafyGreen's string typing for title so we can include buttons.
+      /* @ts-expect-error */
```
minnakt, line 156: Maybe link this here [LG-2193](https://jira.mongodb.org/browse/LG-2193) 🐰

minnakt (approved):
looks good!!

src/components/SpruceForm/FieldTemplates.tsx
sophstad: Good find 🕵️‍♀️

src/components/SpruceForm/FieldTemplates.tsx
sophstad: Thanks for catching — I bet it was messed up by updating text-input, but I'm going to note that it should be fixed in EVG-18234 in case the button update changes things again.

PR #1542, opened 2022-12-01
EVG-14528 Add family trigger types

malikchaya2:
[EVG-14528 ](https://jira.mongodb.org/browse/EVG-14528)

### Description
Add trigger family types. 
This was reverted because projectTriggers[triggerEnum].extraFields was crashing since they have no extra fields.  Once the evergreen PR is merged, the existing "outcome" triggers will be changed to "family-outcome" for versions and patches. 

### Screenshots
No user facing changes

### Evergreen PR 
[6090](https://github.com/evergreen-ci/evergreen/pull/6090)


src/pages/projectSettings/tabs/NotificationsTab/transformers.ts
```diff
@@ -57,7 +57,7 @@ const getExtraFields = (
   if (!triggerData) return {};
 
   const extraFields = {};
-  projectTriggers[triggerEnum].extraFields.forEach((e) => {
+  projectTriggers[triggerEnum]?.extraFields.forEach((e) => {
```
khelif96, line 60: This is more indicative that you may have forgotten to add these fields here. https://github.com/evergreen-ci/spruce/blob/2a281b272efeb91c38b653e4ccbc1e63b1266835/src/constants/triggers.ts#L206

I would imagine these new fields would need the same dependencies.

khelif96 (changes requested):

src/pages/projectSettings/tabs/NotificationsTab/transformers.ts
malikchaya2: It is true that if I added it there then extra fields would be defined. However, I don’t want them to be selectable on the users end. All I want to do right now is make the evergreen change deployable without crashing the UI so that I can then come back to the UI and change what is now “outcome” to “family-outcome” for versions. 

All this change does is it makes it not crash if “family-outcome” exists in the db and it’s trying to display it. It does not yet handle being able to create notifications like that. It’s essentially like a “read-only” change.

malikchaya2: I do some converting in my evergreen PR which makes it work until I can send them properly (which I can’t do right now because the evergreen changes are needed first).

khelif96 (approved):

PR #1553, opened 2022-12-02
EVG-18234: Fix style bugs

sophstad:
EVG-18234

### Description

- Fix style bugs from the upgrade process
- Make header layout more consistent across task and version pages
- Also remove heap size increase — it seems like we're not seeing OOM errors anymore!


src/components/Breadcrumbs/index.tsx
```diff
@@ -66,10 +74,10 @@ const BreadcrumbFragment: React.VFC<BreadcrumbFragmentProps> = ({
   );
 };
 
-const Container = styled.div`
+const Container = styled.nav`
   display: flex;
   align-items: center;
-  margin-bottom: ${size.l};
+  margin-bottom: ${size.s};
```
minnakt, line 80: The larger margin is somewhat necessary due to the headers being squashed when the browser window is smaller 😖

Screen Shot 2022-12-02 at 12 06 24 PM

src/components/Breadcrumbs/index.tsx
```diff
@@ -66,10 +74,10 @@ const BreadcrumbFragment: React.VFC<BreadcrumbFragmentProps> = ({
...
 `;
 
 const PaddedIcon = styled(Icon)`
```
minnakt, line 83: (placement of comment is random) Could we also prevent button text from overflowing? 🤔

Screen Shot 2022-12-02 at 12 07 10 PM

src/components/SpruceForm/FieldTemplates.tsx
```diff
@@ -20,6 +20,10 @@ import ElementWrapper from "./ElementWrapper";
 
 const { gray } = palette;
 
+// Total pixel count above a text field with a label. Used to align buttons to the
+// top of the text box itself.
+const labelOffset = "24px";
```
minnakt, line 25: I think we can set this to our token!

src/components/TabLabelWithBadge.tsx
```diff
@@ -14,8 +15,13 @@ export const TabLabelWithBadge: React.VFC<Props> = ({
...
+    <StyledBadge data-cy={dataCyBadge} variant={badgeVariant}>
       {badgeText}
-    </Badge>
+    </StyledBadge>
   </>
 );
+
+const StyledBadge = styled(Badge)`
+  // Fix height to be consistent with text-only tabs
+  height: 16px;
```
minnakt, line 26: this could be replaced by one of our tokens!

src/components/styles/StyledTabs.tsx
```diff
@@ -4,7 +4,7 @@ import { size } from "constants/tokens";
 import { TabsType } from "types/leafygreen";
 
 export const StyledTabs = styled<TabsType>(Tabs)`
-  [role="tablist"] {
+  > div > [role="tablist"] {
```
minnakt, line 7: Just a question: what's the meaning of this change? is it more correct?

src/pages/version/taskDuration/TaskDurationRow.tsx
```diff
@@ -73,21 +74,23 @@ TaskDurationRow.displayName = "Row";
 const calculateBarWidth = (value: number, max: number) =>
```
minnakt, line 74: oh no... this may be another ticket. 😔 I see that the display tasks won't expand anymore in this table...

Screen Shot 2022-12-02 at 12 25 44 PM

src/components/SpruceForm/FieldTemplates.tsx
```diff
@@ -205,7 +209,7 @@ const OrderControls = styled.div<{ topAlignDelete: boolean }>`
   display: flex;
   flex-direction: column;
   margin-right: ${size.s};
-  margin-top: ${({ topAlignDelete }) => (topAlignDelete ? "0px" : "20px")};
+  margin-top: ${({ topAlignDelete }) => (topAlignDelete ? "0px" : labelOffset)};
 
   > :not(:last-of-type) {
```
minnakt, line 214: (placement of comment is random) For me, `SegmentedControl` still looks strange in some areas of project settings. This is on the patch alias page:

Screen Shot 2022-12-02 at 12 26 41 PM

src/pages/commits/InactiveCommits/index.tsx
```diff
@@ -193,7 +193,6 @@ const InactiveCommitLine = styled.div`
 `;
 
 const TooltipContainer = styled.div`
-  width: 300px;
```
minnakt, line 196: We actually apply a `width` to a lot of tooltips in the codebase as a workaround for old LeafyGreen problems. I think you'll be able to find a lot of them by searching for `StyledTooltip`. It might be worth seeing in this PR or in another ticket if we can now remove those `width`s!

minnakt (changes requested):

src/components/styles/StyledTabs.tsx
sophstad: Segmented Control also has `role="tablist"`, so this rule was causing funky styling in those components. Using the direct child selectors just allows us to be more specific and only apply a margin directly below the tab component.

src/pages/version/taskDuration/TaskDurationRow.tsx
sophstad: EVG-18449

src/pages/commits/InactiveCommits/index.tsx
```diff
@@ -192,12 +191,9 @@ const InactiveCommitLine = styled.div`
   border: 1px dashed ${gray.light1};
 `;
 
-const TooltipContainer = styled.div`
+// @ts-expect-error
+const StyledTooltip = styled(Tooltip)`
```
minnakt, line 195: It seems like the `(see more commits...)` line got uncentered

Screen Shot 2022-12-05 at 2 24 59 PM

src/pages/task/ActionButtons.tsx
```diff
@@ -323,7 +323,7 @@ export const ActionButtons: React.VFC<Props> = ({
             setIsVisibleModal(true);
           }}
         >
-          Notify Me
+          Notify me
```
minnakt, line 326: there's another `"Notify Me"` for the patch buttons, did you want to update that too?

minnakt (approved):
lgtm! 🎉 I still kind of don't like that we have random tooltips in `darkMode`, so maybe we can ask design some time how to modify this. It could be as simple as adding a background or outlining the icons 🤔

PR #1557, opened 2022-12-05
EVG-18374: Fix task icon tooltip hover slowness on Project Health

SupaJoon:
[EVG-18374](https://jira.mongodb.org/browse/EVG-18374)
### Description
We experienced tooltip slowness because the tooltip visibility was enabled within a timeout. The `timeout` variable was being reset during every render which made setTimeout ineffective.

### DIFF
--- a/src/pages/commits/ActiveCommits/buildVariantCard/WaterfallTaskStatusIcon.tsx
+++ b/src/pages/commits/ActiveCommits/buildVariantCard/WaterfallTaskStatusIcon.tsx
@@ -57,6 +57,14 @@ export const WaterfallTaskStatusIcon: React.VFC<
     }
   }, [enabled]);
 
+  useEffect(
+    () => () => {
+      if (timeout) {
+        clearTimeout(timeout);
+      }
+    },
+    [] // eslint-disable-line react-hooks/exhaustive-deps
+  );
   return (
     <Tooltip
       align="top"


src/pages/commits/ActiveCommits/buildVariantCard/WaterfallTaskStatusIcon.tsx
```diff
@@ -39,24 +40,23 @@ export const WaterfallTaskStatusIcon: React.VFC<
...
-    setEnabled(false);
+  useEffect(() => {
     if (timeout) {
       clearTimeout(timeout);
     }
-  };
+    if (enabled) {
+      injectGlobalStyle(identifier);
+      timeout = setTimeout(() => {
+        // Only query failing test names if the task has failed.
```
khelif96, line 50:
```suggestion
      timeout = setTimeout(() => {
         injectGlobalStyle(identifier);
        // Only query failing test names if the task has failed.
```

khelif96 (approved):
Great catch! Left one small improvement so we don't run the animation every time since it is really slow!

src/pages/commits/ActiveCommits/buildVariantCard/WaterfallTaskStatusIcon.tsx
khelif96: My implication was that since injectingGlobalStyle is slow it might make sense to throttle it to only if a user focuses on a task. But you are right it does lead to a worse experience. So I'm ok with your suggestion!

PR #1559, opened 2022-12-06
EVG-14528 Display the trigger without the 'family' prefix

malikchaya2:
[EVG-14528](https://jira.mongodb.org/browse/EVG-14528)

### Description
If a trigger is of type family-outcome, it should still say "outcome" since that is more understandable to users. 

### Screenshots
image

### Testing
Tested locally

### Evergreen PR
[6105](https://github.com/evergreen-ci/evergreen/pull/6105)


minnakt (approved):
I'll add a test for this in my PR since tests don't exist on the main branch yet!

PR #1560, opened 2022-12-06
EVG-18272: Adjust CSS for project settings side nav

minnakt:
EVG-18272

### Description
This PR just adjusts some CSS. Users reported that because the `SideNav` was so small, it was hard to scroll and read through the projects in the `SearchableDropdown`.

### Screenshots
(Screenshots are from beta to show that the changes will work with LeafyGreen updates.)

`SideNav`        |  `SearchableDropdown`
Screen Shot 2022-12-06 at 4 27 40 PM  |  Screen Shot 2022-12-06 at 4 28 02 PM

### DIFF
--- a/src/pages/ProjectSettings.tsx
+++ b/src/pages/ProjectSettings.tsx
@@ -110,7 +110,7 @@ export const ProjectSettings: React.VFC = () => {
 
   return (
     <ProjectSettingsProvider>
-      <SideNav aria-label="Project Settings">
+      <SideNav aria-label="Project Settings" widthOverride={250}>
         <ButtonsContainer>
           <ProjectSelect
             selectedProjectIdentifier={projectLabel}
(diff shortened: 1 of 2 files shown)


src/pages/ProjectSettings.tsx
```diff
@@ -110,7 +110,7 @@ export const ProjectSettings: React.VFC = () => {
 
   return (
     <ProjectSettingsProvider>
-      <SideNav aria-label="Project Settings">
+      <SideNav aria-label="Project Settings" widthOverride={250}>
```
khelif96, line 113: Should we make this consistent everywhere else? (Maybe a question for design)

minnakt: I think the only reason the project settings page needs it is because of the dropdown. The project settings page also has longer section titles. It would not look good on our other SideNavs (see how it just takes away from the table space):

Screen Shot 2022-12-06 at 4 58 54 PM

khelif96 (approved):

PR #1565, opened 2022-12-07
EVG-18449: Fix nested rows in task duration table

sophstad:
EVG-18449

### Description

- Updates to LeafyGreen Table prevent nested rows from being wrapped in parent components. To get around this limitation, render execution tasks directly as `Row` and `Cell` components. This should be fixed by LeafyGreen's table overhaul [LG-2231](https://jira.mongodb.org/browse/LG-2231). See [here](https://mongodb.slack.com/archives/CFFA5BS79/p1670426093909309?thread_ts=1651780191.082089&cid=CFFA5BS79) for more details.

### Testing

- Add unit test

### DIFF
--- a/src/pages/version/taskDuration/TaskDurationRow.tsx
+++ b/src/pages/version/taskDuration/TaskDurationRow.tsx
@@ -12,12 +12,13 @@ import {
 import { size } from "constants/tokens";
 import { VersionTaskDurationsQuery } from "gql/generated/types";
 import { TaskStatus } from "types/task";
+import { Unpacked } from "types/utils";
 import { string } from "utils";
 
 const { msToDuration } = string;
 
 interface RowProps {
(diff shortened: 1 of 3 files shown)


src/pages/version/taskDuration/TaskDurationRow.tsx
```diff
@@ -29,12 +30,13 @@ export const TaskDurationRow: React.VFC<RowProps> = forwardRef(
     ref: React.Ref<HTMLTableRowElement>
   ) => {
```
minnakt, line 31: I noticed that there's an inner scroll bar on the tasks duration table if there are a lot of display tasks. For example, on any `mongodb-mongo-master` commit I can see a scroll bar in the table, which doesn't happen on the currently deployed version of Spruce. Can check if we can get rid of this duplicate scroll bar? 🤔

PR #1566, opened 2022-12-08
EVG-18261: Update Spruce logo

sophstad:
EVG-18261

### Description

- I opted not to do a poll because imo this option definitely looks best for our header logo... but let me know if you think we should get some more opinions! Very open to thoughts.

### Screenshots

image

I was thinking a color variant like this could be fun for our Slack bot etc (updated):
evg-icon


khelif96 (approved):
Code Changes LGTM assuming the only future change will be a new logo

khelif96:
Could you also update the spruce logo in parsley alongside these changes/ or open a ticket to do so.

PR #1569, opened 2022-12-09
EVG-17958: Update project cookie when visiting project pages

minnakt:
EVG-17958

### Description
This PR updates the project cookie when visiting the project patches page or the project waterfall page. I made it so that the cookie is only updated if the GraphQL query is successful, because it would be kind of sad if I accidentally typed `spruuce` into the URL and my cookie was automatically set to `spruuce`.

### Screenshots
- No visible changes.

### Testing
- Added test in `project/patches.ts` (Cypress)
- Added test in `projectHealth/route.ts` (Cypress)

### DIFF
--- a/cypress/integration/project/patches.ts
+++ b/cypress/integration/project/patches.ts
@@ -1,6 +1,7 @@
-const route = "/project/evergreen/patches";
-
 describe("Project Patches Page", () => {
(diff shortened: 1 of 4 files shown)


minnakt:
Closing pull request as this does not work.

PR #1574, opened 2022-12-12
EVG-18546 Switch version outcomes to family 

malikchaya2:
[EVG-18546](https://jira.mongodb.org/browse/EVG-18546)
### Description
Right now, [evergreen converts](https://github.com/evergreen-ci/evergreen/blob/main/rest/data/subscription.go#L63) any incoming subscriptions that are a Version type and on an outcome to the family equivalent. Ideally, they should be sent in the family format. 

### Testing
Existing tests

### Evergreen PR
[6131](https://github.com/evergreen-ci/evergreen/pull/6131)

### DIFF
--- a/src/constants/triggers.ts
+++ b/src/constants/triggers.ts
@@ -133,19 +133,19 @@ export const taskTriggers: Trigger = {
 // VERSION TRIGGERS
 export const versionTriggers: Trigger = {
   [VersionTriggers.VERSION_FINISHES]: {
-    trigger: TriggerType.OUTCOME,
+    trigger: TriggerType.FAMILY_OUTCOME,
     label: "This version finishes",
     resourceType: ResourceType.VERSION,
     payloadResourceIdKey: "id",
   },
   [VersionTriggers.VERSION_FAILS]: {
-    trigger: TriggerType.FAILURE,
+    trigger: TriggerType.FAMILY_FAILURE,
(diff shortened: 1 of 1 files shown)


src/constants/triggers.ts
```diff
@@ -133,19 +133,19 @@ export const taskTriggers: Trigger = {
...
   [VersionTriggers.VERSION_FINISHES]: {
-    trigger: TriggerType.OUTCOME,
+    trigger: TriggerType.FAMILY_OUTCOME,
     label: "This version finishes",
     resourceType: ResourceType.VERSION,
     payloadResourceIdKey: "id",
   },
   [VersionTriggers.VERSION_FAILS]: {
-    trigger: TriggerType.FAILURE,
+    trigger: TriggerType.FAMILY_FAILURE,
```
minnakt, line 142: Just for my understanding, is it true that the `family-` triggers only apply to entire versions? (i.e. nothing to do with tasks, build variants)

minnakt (approved):
lgtm as long as the unit tests are fixed before merging

src/constants/triggers.ts
malikchaya2: yes, exactly! If someone wants to be notified when their entire patch is succeeded, failed, done, etc, then we only want to notify them when the entire patch AND the child patches are done.

PR #1573, opened 2022-12-09
EVG-18520 Migrate all modals to lg modals

khelif96:
EVG-18520

### Description
Migrates all of the antd modals I could find to their leafygreen equivalents. 
Also move some (not all) antd selects to lg selects
### Screenshots
Most were a 1:1 change. except for subscriptions. Here are a few samples
Before:
Subscriptions
image
EditSpawnHost
image
SpawnVolume
image
Notify
image

After:
Subscriptions
image
EditSpawnHost
image
SpawnVolume
image
Notify
image

PR #1578, opened 2022-12-13
EVG-18544: Disable region selector on migration modal

sophstad:
EVG-18544

### Description

- Disable region selector during a migration. The field is populated using the volume's current availability zone.
- Note that this may have some merge conflicts with #1573

### Screenshots

image

### Testing

- Update Cypress


src/pages/spawn/utils.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * AZToRegion takes an availability zone and returns its region name.
+ * An AWS region is just the availability zone minus the final letter.
+ * This rule is defined in evergreen:
+ * https://github.com/evergreen-ci/evergreen/blob/299bf20c8da9e353b1097183672cd5169ff3447d/cloud/ec2_util.go#L76-L80
```
minnakt, line 5: Based on the references in Evergreen, I get the impression that this function is more of an internal function to create spawn hosts or volumes rather than something to surface to the users. 🤔 I could be wrong though!!!

minnakt (approved):
i had some questions but they may not matter, so approving!

src/pages/spawn/spawnVolume/spawnVolumeTable/spawnVolumeTableActions/migrateButton/MigrateVolumeModal.tsx
```diff
@@ -69,6 +70,7 @@ export const MigrateVolumeModal: React.VFC<MigrateVolumeModalProps> = ({
     distros,
     isMigration: true,
     isVirtualWorkstation: !!form?.distro?.isVirtualWorkstation,
+    userAwsRegion: AZToRegion(volume.availabilityZone),
```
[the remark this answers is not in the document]
sophstad: The region is used to configure the new host that is spun up, and hosts only accept a region (i.e. no extra letter) as part of their configuration. I actually tried it this way and the new host failed to spawn!

src/pages/spawn/utils.ts
sophstad: It is internal, but the purpose is to convert the volume's existing availability zone to a region that is assigned to a host. Interestingly we never expose a host's region in evergreen, only its availability zone (which is not a valid param for configuring a host). Selecting a region is pretty similar to the spawn host modal though, which is really what the migrate volume modal is (it's really not a volume configuration modal since the volume already exists). Let me know if you still feel like details are a little murky!!

PR #1582, opened 2022-12-14
EVG-15703: Restart only failed execution tasks

sophstad:
EVG-15703

### Description

- For a display task, provide the option to restart only unsuccessful tasks if it has execution tasks that have not completed successfully
  - If all execution tasks were successful, just use the normal "Restart" button with no menu

### Screenshots

image

### Testing

- I added an integration test for the case where all execution tasks have succeeded. We don't have the right data right now for the other case 😢 but you can check out [my patch](http://localhost:3000/task/evg_ubuntu1604_display_sophie_display_task_patch_d9ab917dceb5cc9eb2fd7b42aa1a92f165811ed8_639a2dcb9dbe326f222e00f5_22_12_14_20_11_13/execution-tasks?execution=1&sorts=STATUS%3AASC) via `yarn staging` 

### Evergreen PR

https://github.com/evergreen-ci/evergreen/pull/6129


src/pages/task/ActionButtons.tsx
```diff
@@ -37,19 +38,22 @@ import {
...
   initialPriority?: number;
+  isDisplayTask: boolean;
   isExecutionTask: boolean;
   task: GetTaskQuery["task"];
 }
 
 export const ActionButtons: React.VFC<Props> = ({
   initialPriority = 1,
-  task,
+  isDisplayTask,
```
khelif96, line 54: If we are passing task in any way why do we need to calculate isDisplayTask independently and pass it in as well? It seems like it can be calculated below as well.

khelif96 (changes requested):

src/pages/task/ActionButtons.tsx
sophstad: It could be calculated here for sure — but I decided to lift up `isDisplayTask` into the `Task` component so it would only be computed once. That way we can make sure the same logic is being used wherever we determine if a task is a display task. Open to repeating that logic in `ActionButtons` and `TaskTabs` if you think it makes more sense though.

khelif96: Great call! I missed that it was also being passed to TaskTabs

khelif96 (approved):

PR #1586, opened 2022-12-19
EVG-18604: enable New Relic for Spruce beta

SupaJoon:
[EVG-18604](https://jira.com/browse/EVG-18604)


minnakt:
I wonder if this should go into the `lg-updates` branch instead? I think we only want this while beta testing the `lg-updates` branch, and we might want to remove it it before merging `lg-updates` into `main`.

minnakt (approved):

PR #1588, opened 2022-12-19
EVG-18495 Use IntersectionObserver to performantly render task icons on project health page

khelif96:
EVG-18495

### Description
The project health page can sometimes get into a state where the page needs to render thousands of Icons. This is an extremely expensive operation that causes the page to lag heavily due to the need to rerender each of those icons upon interactions. This pr adds a component called `VisibilityContainer` (Name subject to change). This component leverages the [`IntersectionObserver`](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) API to only render these Icons if they are visible. 

This greatly improves performance on the page, especially on initial renders as well as when using the icon hover features
In the below example, I render several thousand icons and try to use the hover feature. You can see a noticeable difference in the hover performance.
### Screenshots

Before:
https://user-images.githubusercontent.com/4605522/208544846-c8c01726-0c96-4053-9680-1146498898a5.mov

After:
https://user-images.githubusercontent.com/4605522/208544851-af006b06-c63b-46ac-80ca-886f0fc439b1.mov

### Testing
Wrote some unit tests but needed to mock IntersectionObserver since it is not available in the jsdom environment. 

Tested against prod mainline commits waterfall
Created sample playground in storybook


src/components/VisibilityContainer/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+  const containerRef = useRef(null);
+  const [isVisible, setIsVisible] = useState(false);
+
+  useEffect(() => {
+    const currentRef = containerRef.current;
+    const observer = new IntersectionObserver(([entry]) => {
+      setIsVisible(entry.isIntersecting);
+    });
+    observer.observe(currentRef);
+    return () => observer.unobserve(currentRef);
```
minnakt, line 21: can we `disconnect` or does it need to be `unobserve`?

src/components/VisibilityContainer/index.tsx
```diff
@@ -0,0 +1,27 @@
+import React, { useRef, useEffect, useState } from "react";
```
minnakt, line 1: nit. remove `React`

src/pages/commits/ActiveCommits/ProjectHealth.stories.tsx
```diff
@@ -39,36 +57,47 @@ const buildVariantUpdateLength = (
...
-    const newTasks = tasks.slice(0, taskCount);
-    return newTasks;
+    newTasks = tasks.slice(0, taskCount);
   }
+  return newTasks.map((task, index) => ({
+    ...task,
+    displayName: `${task.displayName} ${index}`,
+    id: `${task.id} ${index}`,
+  }));
 };
```
minnakt, line 101: nit. Can we fix this weird numbering situation? (i.e. multiple numbers after the build variant name and task name)

Screen Shot 2022-12-20 at 3 08 19 PM

src/components/VisibilityContainer/index.tsx
khelif96: Sure, I don't see much specific guidance on which one I should use so disconnect works for me as well.

src/pages/commits/ActiveCommits/ProjectHealth.stories.tsx
khelif96: weird.

src/components/VisibilityContainer/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+  const [isVisible, setIsVisible] = useState(false);
+
+  useEffect(() => {
+    const currentRef = containerRef.current;
+    const observer = new IntersectionObserver(([entry]) => {
+      setIsVisible(entry.isIntersecting);
+    });
+    observer.observe(currentRef);
+    return () => observer.unobserve(currentRef);
+  }, []);
```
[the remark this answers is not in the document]
khelif96: Why not :)

src/pages/commits/ActiveCommits/ProjectHealth.stories.tsx
```diff
@@ -22,80 +41,131 @@ export const ActualWaterfallPage = ({
   hasFilters,
 }) => {
   const updatedVersions = versions.map((version) =>
-    formatVersion(version, buildVariantCount, taskCount)
+    formatVersion(version, buildVariantCount, taskCount, hasTaskFilter)
```
minnakt, line 44: Is it possible to tweak this function so that there isn't whitespace when `hasTaskFilter` is `false`?

Screen Shot 2022-12-22 at 1 02 22 PM

src/components/VisibilityContainer/index.tsx
```diff
@@ -0,0 +1,23 @@
+import { useRef, useState } from "react";
+import useIntersectionObserver from "hooks/useIntersectionObserver";
+
+interface VisibilityContainerProps {
+  children: React.ReactNode;
+}
+/**
+ * `VisibilityContainer` is a component that will only render its children when it is visible in the viewport.
```
minnakt, line 8: I'm not sure how much this would differ in production, but if there's a lot of content,  there's a noticeable lag between when it actually loads in. Is that expected? (this would probably only affect large projects, but I only checked on dev since beta is occupied right now)

https://user-images.githubusercontent.com/47064971/209198264-7d9e3d94-0bd6-43ea-bdbe-44a37be4c643.mov

khelif96: This is just a result of rendering so many icons. Currently, this lag exists but it all happens on the initial render as opposed to when scrolling. In most scenarios, it shouldn't be noticeable unless they open a build variant with hundreds of task icons.

src/pages/commits/ActiveCommits/ProjectHealth.stories.tsx
khelif96: I tried but couldn't figure out an elegant solution. :/

minnakt: that's okay, I kind of hope the design changes on this page anyways (Like with the build variants listed on the left side like Evergreen), which would help reduce the complexity in the story !

minnakt (approved):
cool :]

PR #1589, opened 2022-12-20
EVG-18549: Upgrade to Cypress v12

SupaJoon:
[EVG-18549
](htttps://jira.com/browse/EVG-18549)

### DIFF
--- a/cypress/integration/host/host_events.ts
+++ b/cypress/integration/host/host_events.ts
@@ -4,9 +4,8 @@ const pathWithEvents = "/host/i-0f81a2d39744003dd";
 
 describe("Host events", () => {
   beforeEach(() => {
-    cy.session("log in and set pagination", () => {
-      cy.login();
-      window.localStorage.setItem("recentPageSize", "20");
+     cy.window().then((win) => {
+         win.localStorage.setItem("recentPageSize", "20");
     });
   });
   it("host events display the correct text", () => {


cypress/integration/host/host_events.ts
```diff
@@ -102,7 +108,7 @@ describe("Host events", () => {
     });
   });
 
-  it("host events with logs display the correct text and the logs get displayed when available", () => {
+  it.only("host events with logs display the correct text and the logs get displayed when available", () => {
```
khelif96, line 111: it.only

cypress/support/commands.ts
```diff
@@ -5,9 +5,9 @@ const user = {
   username: "admin",
   password: "password",
 };
-Cypress.Cookies.defaults({
-  preserve: TOAST_COOKIE,
-});
+// Cypress.Cookies.defaults({
+//   preserve: TOAST_COOKIE,
+// });
```
khelif96, line 10: Why are these being commented out?

cypress/support/commands.ts
```diff
@@ -62,13 +62,7 @@ Cypress.Commands.add("login", () => {
...
 
 /* preserveCookies */
-Cypress.Commands.add("preserveCookies", () => {
-  Cypress.Cookies.preserveOnce(
-    LOGIN_COOKIE,
-    "mci-session",
-    "mci-project-cookie"
-  );
-});
+Cypress.Commands.add("preserveCookies", () => {});
```
khelif96, line 65: likewise if its a no op we should remove it.

cypress/integration/host/host_events.ts
```diff
@@ -1,8 +1,14 @@
 import { clickOnPageSizeBtnAndAssertURLandTableSize } from "../../utils";
 
-const pathWithEvents = `/host/i-0f81a2d39744003dd`;
+const pathWithEvents = "/host/i-0f81a2d39744003dd";
 
 describe("Host events", () => {
+  beforeEach(() => {
+    cy.session("log in and set pagination", () => {
+      cy.login();
```
khelif96, line 8: login should happen by default.
Why are we using cy.session?

cypress/integration/host/host_events.ts
```diff
@@ -1,8 +1,14 @@
...
+      window.localStorage.setItem("recentPageSize", "20");
```
khelif96, line 9: Why is this now necessary?

khelif96 (changes requested):

cypress/support/commands.ts
minnakt: If these functions are being removed, remember to also remove them from `e2e.ts` which is the corresponding file where they are declared

cypress/integration/host/host_events.ts
```diff
@@ -1,8 +1,14 @@
...
+    });
+  });
```
minnakt, line 11: Based on my testing, I think this would also work and would avoid from having to login again. Not sure which is the better approach, but just suggesting an alternative :]

```suggestion
  beforeEach(() => {
     cy.window().then((win) => {
         win.localStorage.setItem("recentPageSize", "20");
    });
  });
```

src/utils/url/index.ts
```diff
@@ -49,7 +49,7 @@ export const removeQueryParam = (param: string | string[], value: string) => {
 export const getDefaultPageSize = () => {
   const pageSizeFromLocalStorage: number = parseInt(
     localStorage.getItem(RECENT_PAGE_SIZE_KEY),
-    DEFAULT_PAGE_SIZE
+    10
```
minnakt, line 52: nice!

khelif96 (changes requested):

src/utils/url/index.ts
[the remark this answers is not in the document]
minnakt: I think it's more correct not to use the constant. The second argument to `parseInt` is the radix, which is typically always 10 when we use it, and doesn't have any specific ties to the `DEFAULT_PAGE_SIZE`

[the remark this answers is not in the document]
khelif96: Oh snap I totally missed that

cypress/integration/host/host_events.ts
khelif96: Bumping for consistency!

khelif96 (approved):
LGTM Great job! mod @minnakt's  [comment](https://github.com/evergreen-ci/spruce/pull/1589#discussion_r1065059553)

PR #1590, opened 2022-12-20
EVG-18615 An Evergreen miracle

khelif96:
EVG-18615

### Description
It's gonna be a white Christmas


src/components/Header/Navbar.tsx
```diff
@@ -137,3 +138,8 @@ const secondaryStyle = css`
 const SecondaryLink = styled(StyledLink)`
   ${secondaryStyle}
 `;
+
+const ChristmasTreeIcon = styled.img`
+  height: 32px;
+  width: 32px;
```
minnakt, line 144: It's a tad small; I think you could size up to `44px`. 

If you want it to be even bigger, you can add `position: relative` and move the icon up a little. Example:

Screen Shot 2022-12-20 at 11 38 05 AM

minnakt (approved):

PR #1591, opened 2022-12-21
EVG-18623: Wrap task and version links on task queue page

minnakt:
EVG-18623

### Description
This PR makes it so that task links are wrapped and don't overflow into the next column of the Task Queue table.

### Screenshots
Before        |  After
Screen Shot 2022-12-21 at 11 56 19 AM | Screen Shot 2022-12-21 at 11 55 57 AM

### DIFF
--- a/src/pages/taskQueue/TaskQueueTable.tsx
+++ b/src/pages/taskQueue/TaskQueueTable.tsx
@@ -7,7 +7,7 @@ import { Table } from "antd";
 import { ColumnProps } from "antd/es/table";
 import { useParams, useLocation } from "react-router-dom";
 import { useTaskQueueAnalytics } from "analytics";
-import { StyledRouterLink } from "components/styles";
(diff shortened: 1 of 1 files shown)


src/pages/taskQueue/TaskQueueTable.tsx
```diff
@@ -87,7 +87,7 @@ export const TaskQueueTable = () => {
                 taskQueueAnalytics.sendEvent({ name: "Click Task Link" })
               }
             >
-              {displayName}
+              <WordBreak>{displayName}</WordBreak>
```
khelif96, line 90: Can you make the same change to version ids I commented on the ticket.

khelif96 (approved):

src/pages/taskQueue/TaskQueueTable.tsx
minnakt: sure thing

PR #1592, opened 2022-12-21
EVG-12434 Expose message about task group restart status

khelif96:
EVG-12434

### Description
Expose a message alerting users a task is pending a restart when its task group has finished. (Open to suggestions on copy)
### Screenshots
Adds the following message to task metadata
image

### Evergreen PR
https://github.com/evergreen-ci/evergreen/pull/6154


src/pages/task/metadata/index.tsx
```diff
@@ -277,6 +278,12 @@ export const Metadata: React.VFC<Props> = ({
           ))}
         </span>
       ) : null}
+      {resetWhenFinished && (
+        <P2 data-cy="reset-when-finished-container">
```
minnakt, line 282: nit. the `depends-on-container` is only named that way because it can contain multiple dependent tasks

```suggestion
        <P2 data-cy="reset-when-finished">
```

src/pages/task/metadata/index.tsx
```diff
@@ -277,6 +278,12 @@ export const Metadata: React.VFC<Props> = ({
           ))}
         </span>
       ) : null}
+      {resetWhenFinished && (
```
minnakt, line 281: about placement: I'm not sure if this should fall below the `Depends On` section. The `Depends On` section could be very long. It might be better it we moved where this message displays

src/pages/task/metadata/index.tsx
```diff
@@ -277,6 +278,12 @@ export const Metadata: React.VFC<Props> = ({
...
+        <P2 data-cy="reset-when-finished-container">
+          This task is scheduled to restart when all of the tasks in its group
+          have finished.
```
minnakt, line 284: lol I know we say "task" a lot in this sentence but I feel like saying `task group` might be a little clearer (other stuff is just suggestion, I don't feel strongly about it)

```suggestion
          This task will restart when all of the tasks in its task group
          have finished running.
```

minnakt (approved):

ablack12 (approved):
Verbage LGTM

PR #1609, opened 2023-01-10
EVG-17958: Modify Navbar to enable better routing system

minnakt:
EVG-17958

### Description
This PR modifies the `NavBar` component so that it calls the `useParams` hook to get the project from the URL, and then uses that project for the links. If no project exists in the URL, the links will fall back to the recent project cookie. If no cookie value exists, the links will fall back to the project defined in the Spruce config. This is similar to how the active project is determined in Evergreen ([link](https://github.com/evergreen-ci/evergreen/blob/2fa031e6a1781e9d3330244a5c86bf8a24e12963/service/middleware.go#L318-L320)).

I made these changes because it's rather confusing to update the cookie in several places across the codebase and have that affect the links in the `NavBar`. Note that the only time the cookie is actually used is to visit the `/commits` page without a project in the URL, so it has limited usage.

#### Explanation of how it works in Evergreen
- Evergreen has the project select dropdown in the nav bar, so there's no visual disconnect like in Spruce. Because you can change the project in the nav bar, you can also update the links in the nav bar at the same time.
- Evergreen only sets the recent project cookie in one place ([link](https://github.com/evergreen-ci/evergreen/blob/2fa031e6a1781e9d3330244a5c86bf8a24e12963/service/middleware.go#L402)), which I assume runs every time the URL updates.

### Screenshots
- N/A

### Testing
- Added Cypress tests


src/components/Content/index.tsx
```diff
@@ -60,56 +67,57 @@ export const Content: React.VFC = () => {
...
-          element={<UserPatchesRedirect />}
-        />
-        <Route path={routes.variantHistory} element={<VariantHistory />} />
-        <Route path={routes.version} element={<VersionPage />}>
-          <Route path={tab} element={null} />
+        <Route element={<Layout />}>
+          <Route path="*" element={<PageDoesNotExist />} />
+          <Route path="/" element={<MyPatches />} />
+          <Route path={routes.commits} element={<Commits />}>
+            <Route path=":projectId" element={null} />
```
sophstad, line 75: Does this param actually refer to the project ID or the identifier? We've used them interchangeably so far, but EVG-16434 exists to address this. AFAIK most routes use the more user-friendly identifier but I may be mixed up.

sophstad (commented):
Wow this is finally working how I'd expect it to 😍 one question for ya

src/components/Content/index.tsx
minnakt: technically `identifier` (or maybe `projectIdentifier`?) is more correct haha 😅
I'll rename the paths, it's an easy fix (^^) thanks Sophie!

src/constants/routes.ts
```diff
@@ -69,18 +69,18 @@ export const routes = {
   myPatches: `${paths.user}/${PageNames.Patches}`,
   patch: `${paths.patch}/:id`,
   preferences: paths.preferences,
-  projectPatches: `${paths.project}/:id/${PageNames.Patches}`,
-  projectSettings: `${paths.project}/:identifier/${PageNames.Settings}`,
+  projectPatches: `${paths.project}/:projectIdentifier/${PageNames.Patches}`,
+  projectSettings: `${paths.project}/:projectIdentifier/${PageNames.Settings}`,
```
sophstad, line 73: This is a somewhat annoying route because the identifier here could be for a project _or_ repo (the previously used `identifier` could signify either). But I think there's benefit to keeping the path segment consistent in our routes so the change makes sense.

sophstad (approved):
cool (◕‿◕✿)

src/constants/routes.ts
minnakt: Ohh that is a good point, I did not consider this 😔 I think it's okay for now for the same reason we call it "Project Settings", yet it also technically contains repo settings

we can change it later too! :]

PR #1610, opened 2023-01-10
EVG-18664: Run Cypress on Electron browser

sophstad:
EVG-18664

### Description

- Run Cypress on its native Electron browser. I'll remove `chrome_download_url` from project settings once this is merged.
- I think this change could be made in Parsley too—any objections?
- Remove test for job logs page because it broke but also the page is broken 🤪 new tests will be introduced as part of [PM-3139](https://jira.mongodb.org/browse/PM-3139)
- Several tests were missing `limit` parameters on their URLs; since we explicitly expect 10 items on these pages, I added the limit to the `cy.visit` calls.


khelif96:
Making this change on parsley as well makes sense to me!

minnakt (approved):
yay!

PR #1611, opened 2023-01-10
EVG-18329: Remove null handling logic from LeafyGreen Select widget

minnakt:
EVG-18329

### Description
This PR modifies the LeafyGreen Select widget to fix an issue on the Patch Alias page, where it wasn't possible to return to the `Select Event...` option after selecting a different option.

### Screenshots
Current behavior:

Screen Shot 2023-01-11 at 10 22 29 AM 

New behavior:

https://user-images.githubusercontent.com/47064971/211680128-ab6a8f23-2728-404d-9562-efedc61eadf7.mov

### Testing
- Added tests in `project_settings.ts` (Cypress)


src/components/SpruceForm/Widgets/LeafyGreenWidgets.tsx
```diff
@@ -184,7 +184,7 @@ export const LeafyGreenSelect: React.VFC<
           disabled={isDisabled}
           value={value}
           {...labelProps}
-          onChange={(v) => onChange(v === "" ? null : v)}
```
minnakt, line 187: I don' think it makes sense for the value to ever be `null` because the widget expects a string, which causes the errors

sophstad (approved):
that's amore

PR #1612, opened 2023-01-11
EVG-18710: Add font-smoothing property

sophstad:
EVG-18710

### Description

- Apply [font smoothing](https://css-tricks.com/understanding-web-fonts-getting/#aa-font-smoothing), which is used by Atlas and I think improves legibility of our sans serif font.
- Preview on [beta](https://spruce-beta.corp.mongodb.com/)

### DIFF
--- a/src/components/styles/GlobalStyles.tsx
+++ b/src/components/styles/GlobalStyles.tsx
@@ -40,6 +40,8 @@ const globalStyles = css`
     font-family: ${fontFamilies.default};
     font-size: 13px;
     margin: 0;
+    -webkit-font-smoothing: antialiased; /* Chrome, Safari */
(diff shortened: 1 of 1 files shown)


minnakt (approved):
I like this change overall, especially how the links became less bold on the My Patches page!

Maybe the only thing I don't like as much is how it emphasizes the stick-ness of the serif fonts, but we win some, lose some 😅

Screen Shot 2023-01-11 at 1 29 19 PM

sophstad:
> Maybe the only thing I don't like as much is how it emphasizes the stick-ness of the serif fonts, but we win some, lose some 😅
> 
> Screen Shot 2023-01-11 at 1 29 19 PM

Agree, I wish serif headings used the "Medium" weight font instead of "Regular." Alas!

PR #1613, opened 2023-01-11
EVG-15943: Replace patch alias regex input with area

SupaJoon:
[EVG-15943](https://jira.mongodb.org/browse/EVG-15943)
Check it out on this page: http://localhost:3000/project/spruce/settings/patch-aliases
### Screenshots
Screen Shot 2023-01-11 at 1 02 20 PM


sophstad (changes requested):

src/pages/projectSettings/tabs/utils/alias.ts
```diff
@@ -1,7 +1,12 @@
+import { css } from "@emotion/react";
 import { AccordionFieldTemplate } from "components/SpruceForm/FieldTemplates";
 import widgets from "components/SpruceForm/Widgets";
 import { ProjectAlias, ProjectAliasInput } from "gql/generated/types";
 
+const textAreaCSS = css`
+  width: 374px;
+`;
```
minnakt, line 8: If we have to use wrapper CSS, then I think this might be more accurate:

```suggestion
const textAreaCSS = css`
  box-sizing: border-box;
  max-width: 400px;
`;
```

where `max-width` matches the value defined under `MaxWidthContainer` in the LeafyGreen widgets file.

Alternatively, and might be a better approach, would be to use the `MaxWidthContainer` to wrap the LeafyGreen TextArea widget. This would have to be conditional though, as we don't always want to put a max width container on this widget.

sophstad (approved):

PR #1620, opened 2023-01-20
EVG-17632 Set up redirect from waterfall to commits route

khelif96:
EVG-17632

### Description
Simple redirect from waterfall page to commits.
Tiny refactor to move "redirect" routes to their own obj


src/constants/routes.ts
```diff
@@ -53,10 +53,16 @@ const paths = {
...
+  waterfall: "/waterfall",
   variantHistory: "/variant-history",
   taskHistory: "/task-history",
   jobLogs: "/job-logs",
 };
+export const redirectRoutes = {
+  projectSettingsRedirect: paths.projects,
+  userPatchesRedirect: `${paths.user}/:id`,
+  waterfallRedirect: `${paths.waterfall}/:projectIdentifier`,
+};
```
minnakt, line 65: Just pointing out something that might have gone unnoticed, but since these are now under `redirectRoutes` you could remove the trailing `"Redirect"` for each of the fields (e.g. `redirectRoutes.userPatchesRedirect` => `redirectRoutes.userPatches`). Ultimately it's up to your personal preference! 🧑‍🍳

minnakt (approved):
nice and straightforward refactor!

PR #1622, opened 2023-01-23
EVG-17684: Load additional project settings events

sophstad:
EVG-17684

### Description

- Allow users to load additional project settings events via "Load More" button at the bottom of the event log page. I used [Apollo's `fetchMore` function](https://www.apollographql.com/docs/react/pagination/core-api/#the-fetchmore-function) which is kinda cool!
- Disable caching for some fields nested within `ProjectEventSettings`. This problem had previously been fixed in #1401 by disabling caching altogether using `fetchPolicy: 'none'`, but caching is required by `fetchMore`. Disabling caching for specific fields fixes the issue and is probably the better approach anyway.

### Screenshots

image

### Testing

- Add unit tests
- I attempted to write a test to verify behavior when clicking "Load more events" but it seems like there's [an issue](https://github.com/apollographql/apollo-client/issues/9761#issuecomment-1292369756) related to `fetchMore` that prevents reliably updating the data passed to the component :(


src/pages/projectSettings/tabs/EventLogTab/useEvents.ts
```diff
@@ -0,0 +1,69 @@
...
+  const [showLoadButton, setShowLoadButton] = useState(true);
+
+  // Hide Load More button when event count < limit is returned,
+  // or when the same event count is returned twice in a row.
+  const onCompleted = (eventData) => {
+    if (eventData.length === prevCount) {
+      setShowLoadButton(false);
+    } else if (eventData.length - prevCount < limit) {
+      setShowLoadButton(false);
+    } else {
```
minnakt, line 29: this may be a silly, silly question... but I am wondering what scenario requires the first if-statement... because I feel like it could be written equivalently as `eventData.length - prevCount === 0` which seems like it would fall under `eventData.length - prevCount < limit` as long as `limit` is positive 🤔

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.test.tsx
```diff
@@ -0,0 +1,235 @@
...
+      periodicBuilds: [],
+      tracksPushEvents: true,
+      hidden: false,
+      prTestingEnabled: true,
+      manualPrTestingEnabled: false,
+      githubChecksEnabled: false,
+      gitTagVersionsEnabled: true,
+      gitTagAuthorizedUsers: ["mohamed.khelif"],
+      gitTagAuthorizedTeams: ["evergreen"],
+      commitQueue: {
```
minnakt, line 215: I think you might be missing this field for the test mocks!

Screen Shot 2023-01-23 at 5 06 06 PM

minnakt (changes requested):
this is very cool, I didn't know about `fetchMore` ^^ it works really seamlessly too!

src/gql/GQLWrapper.tsx
```diff
@@ -40,6 +40,39 @@ interface ClientLinkParams {
 
 const cache = new InMemoryCache({
   typePolicies: {
```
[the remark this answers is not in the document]
khelif96: This is caused because the keyArgs are omitted. Apollo needs them to determine which keys belong in the cache.  https://www.apollographql.com/docs/react/pagination/key-args#setting-keyargs

src/gql/GQLWrapper.tsx
```diff
@@ -40,6 +40,39 @@ interface ClientLinkParams {
...
+      },
+    },
+    ProjectAlias: {
+      keyFields: false,
+    },
+    Project: {
+      keyFields: false,
+    },
+    ProjectSubscription: {
+      keyFields: false,
```
khelif96, line 74: You should be able to use some combination of projectIdentifier and some other field for this.

src/pages/projectSettings/tabs/EventLogTab/useEvents.ts
```diff
@@ -0,0 +1,69 @@
...
+      setPrevCount(eventData.length);
+    }
+  };
+
+  const { data: projectEventData, fetchMore: projectFetchMore } = useQuery<
+    ProjectEventLogsQuery,
+    ProjectEventLogsQueryVariables
+  >(GET_PROJECT_EVENT_LOGS, {
+    variables: { identifier, limit },
+    errorPolicy: "all",
```
khelif96, line 39: I see that this was here before but could you remind me why we used this errorPolicy here?

sophstad: Because of type changes over time and the fact that previous events don't necessarily adhere to our current types, we found that we were encountering errors somewhat often when trying to load event logs. But type safety isn't super important for a log like this; it's better to show a log then omit it, so we don't want to block on errors. (Added in https://github.com/evergreen-ci/spruce/pull/1283#issue-1243343462).

src/gql/GQLWrapper.tsx
sophstad: The problem with these few fields occurs when Apollo tries to merge the `before` and `after` fields in a `ProjectEventLogEntry`. These structs have no distinguishing field that could be used as a key; in fact they are usually almost exactly the same % whatever modification was made on save. The only field that could serve to distinguish them is whether they are a descendant of `before` or `after`, and Apollo doesn't provide a way to use the key name in in `keyFields`. Thus we have no choice but to pretty much disable caching altogether (which is fine—it really would not be useful in this context).

src/pages/projectSettings/tabs/EventLogTab/useEvents.ts
```diff
@@ -0,0 +1,77 @@
...
+    : projectPreviousData?.projectEvents;
+
+  // Hide Load More button when event count < limit is returned,
+  // or when an additional fetch fails to load more events.
+  const onCompleted = (count: number) => {
+    const { count: prevCount = 0 } = previousData ?? {};
+    if (count - prevCount < limit) {
+      setAllEventsFetched(true);
+    }
+  };
```
minnakt, line 74: This is a weird error, but when I click through about 4 to 5 projects and go to the event log each time, eventually I run into this problem:

Screen Shot 2023-01-24 at 12 35 22 PM

I'm not really sure how to fix it though because of the dependency on `previousData` 😔

src/pages/projectSettings/Tabs.tsx
```diff
@@ -204,7 +204,7 @@ export const ProjectSettingsTabs: React.VFC<Props> = ({
         />
         <Route
           path={ProjectSettingsTabRoutes.EventLog}
-          element={<EventLogTab projectType={projectType} />}
+          element={<EventLogTab projectType={projectType} key={identifier} />}
```
minnakt, line 207: do we need the `key` here?

sophstad: It's a pretty straightforward way to make sure all of the state gets reset when navigating between projects as you flagged in your first comment. Updating the cache policy helped somewhat but things like the `allEventsFetched` state in our hook would not necessarily update.

src/pages/projectSettings/tabs/EventLogTab/useEvents.ts
sophstad: Great catch Minna!

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.test.tsx
```diff
@@ -0,0 +1,241 @@
...
+import { EventLogTab } from "./EventLogTab";
+
+const Wrapper = ({ children, mocks = [] }) => (
+  <MockedProvider mocks={mocks} addTypename={false}>
+    {children}
+  </MockedProvider>
+);
+
+describe("loading events", () => {
+  afterEach(() => {});
```
minnakt, line 15:
```suggestion

```

src/pages/projectSettings/tabs/EventLogTab/EventLogTab.test.tsx
```diff
@@ -0,0 +1,241 @@
...
+      taskSync: {
+        configEnabled: false,
+        patchEnabled: false,
+        __typename: "TaskSyncOptions",
+      },
+      disabledStatsCache: false,
+      __typename: "Project",
+      private: false,
+      restricted: false,
+      admins: ["arjun.patel", "mohamed.khelif", "sophie.stadler"],
```
minnakt, line 103: oh, I see how it is 😠

minnakt (approved):
works great queen!

PR #1623, opened 2023-01-24
EVG-18749 Specify YAML defining option for alias fields and provide a link

hadjri:
EVG-18749

### Description
The UI forms for patch, PR, commit queue, git tag, and github checks aliases don't mention the nuance that they may be overridden by YAML definitions in the case that the UI aliases are removed (with the exception of patch aliases which are merged with YAML definitions).

Updated the help text for each alias to mention its behavior with respect to YAML definitions, as well as a link to the wiki on how to define them.

### Testing
Checked formatting locally.


src/constants/externalResources.ts
```diff
@@ -8,6 +8,16 @@ export const wikiUrl = "https://github.com/evergreen-ci/evergreen/wiki";
...
 
+export const patchAliasesDocumentationUrl = `${versionControlDocumentationUrl}#patch-aliases`;
+
+export const pullRequestAliasesDocumentationUrl = `${versionControlDocumentationUrl}#pr-aliases`;
+
+export const commitQueueAliasesDocumentationUrl = `${versionControlDocumentationUrl}#commit-queue-aliases`;
+
+export const gitTagAliasesDocumentationUrl = `${versionControlDocumentationUrl}#git-tag-aliases`;
+
+export const githubChecksAliasesDocumentationUrl = `${versionControlDocumentationUrl}#github-checks-aliases`;
```
khelif96, line 19: Looks like these urls don't work I get linked to the following URL when I click the link.
```
https://github.com/evergreen-ci/evergreen/wiki/Project-and-Distro-Settings#version-control%23github-checks-aliases
```
It looks like `versionControlDocumentationUrl` already has an anchor so when you generate these new URLs the second anchor gets auto encoded.

khelif96 (changes requested):

khelif96 (approved):

PR #1624, opened 2023-01-24
EVG-16822 Add the option to disable regex search for project health view

khelif96:
EVG-16822

### Description
Create new `TupleSelectWithRegex` component which wraps `TupleSelect` and conditionally escapes regular expressions.

### Screenshots
image

### Testing
Unit tests


src/pages/commits/TupleSelectWithRegexConditional/TupleSelectWithRegexConditional.stories.tsx
```diff
@@ -0,0 +1,34 @@
+import { Disclaimer } from "@leafygreen-ui/typography";
+import { action } from "@storybook/addon-actions";
+import { ProjectFilterOptions } from "types/commits";
+import TupleSelectWithRegexConditional from ".";
+
+export default {
+  title: "Pages/Commits/TupleSelectWithRegexConditional",
```
minnakt, line 7: rather than under the `Commits` folder, I feel like this might belong better under the `TupleSelect` folder as another variant 🤔

src/pages/commits/TupleSelectWithRegexConditional/TupleSelectWithRegexConditional.test.tsx
```diff
@@ -0,0 +1,126 @@
...
+    expect(input).toHaveValue("some-*");
+    userEvent.type(input, "{enter}");
+    expect(onSubmit).toHaveBeenCalledWith({
+      category: "build_variant",
+      value: "some\\-\\*",
+    });
+
+    expect(input).toHaveValue("");
+  });
+  it("should not attempt to validate input if using the `exact` input type", async () => {
```
minnakt, line 107:
```suggestion
  it("should not attempt to validate input if using the `exact` input type", () => {
```

src/pages/commits/TupleSelectWithRegexConditional/TupleSelectWithRegexConditional.test.tsx
```diff
@@ -0,0 +1,126 @@
...
+    expect(input).toHaveValue("bad");
+    userEvent.type(input, "{enter}");
+    expect(input).toHaveValue("bad");
+    expect(onSubmit).not.toHaveBeenCalled();
+    expect(validator).toHaveBeenLastCalledWith("bad");
+    expect(screen.getByDataCy("tuple-select-warning")).toBeInTheDocument();
+    userEvent.hover(screen.queryByDataCy("tuple-select-warning"));
+    await screen.findByText(validatorErrorMessage);
+  });
+  it("toggling the input type selector to `exact` should escape any regex characters", async () => {
```
minnakt, line 82:
```suggestion
  it("toggling the input type selector to `exact` should escape any regex characters", () => {
```

src/pages/commits/TupleSelectWithRegexConditional/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+      }
+    />
+  );
+};
+
+const PaddedSegmentedControl = styled(SegmentedControl)`
+  margin-left: ${size.xs};
+`;
+
+const escapeRegex = (str: string) =>
```
minnakt, line 66: this might be a separate ticket, but I noticed that the params don't get properly escaped in the URL

khelif96: How so? they seem to be escaped to me.
image

minnakt: Oh sorry, this was a quirk in Firefox I never noticed where the URL isn't escaped but when you copy it it is

src/components/TupleSelectWithRegexConditional/TupleSelectWithRegexConditional.stories.tsx
```diff
@@ -0,0 +1,33 @@
...
+    placeHolderText: "Search Build Variant names",
+  },
+  {
+    value: "task",
+    displayName: "Task",
+    placeHolderText: "Search Task names",
+  },
+];
+
+export const Default = () => (
```
minnakt, line 23: I think we don't want the nested folder:

Screen Shot 2023-01-27 at 12 15 56 PM

Changing the path/name will get rid of the nested folder:

```suggestion
export default {
  title: "Components/TupleSelect",
  component: TupleSelectWithRegexConditional,
};

const options = [
  {
    value: "build_variant",
    displayName: "Build Variant",
    placeHolderText: "Search Build Variant names",
  },
  {
    value: "task",
    displayName: "Task",
    placeHolderText: "Search Task names",
  },
];

export const TupleSelectWithConditional = () => (
```

Screen Shot 2023-01-27 at 12 21 51 PM

minnakt (approved):
lgtm! One thing is that I think there needs to be a better mechanism to easily determine which badges are regex vs exact, but that's something to bring up with design in the future.

Screen Shot 2023-01-27 at 12 11 49 PM

PR #1627, opened 2023-01-25
EVG-18789 upgrade ubuntu to 20.04

ybrill:
[EVG-18789](https://jira.mongodb.org/browse/EVG-18789)

### Description
We're moving evergreen's self-tests to Ubuntu 20.04, so let's move Spruce too. We're also updating the DB version for evergreen, but that's configured from Spruce's project variables.

### Screenshots
N/A

### Testing
Let's see how it lands 😄 

### Evergreen PR
https://github.com/evergreen-ci/evergreen/pull/6204

### DIFF
--- a/.evergreen.yml
+++ b/.evergreen.yml
@@ -438,10 +438,10 @@ tasks:
 
 
 buildvariants:
-  - name: ubuntu1604
-    display_name: Ubuntu 16.04
+  - name: ubuntu2004
+    display_name: Ubuntu 20.04
     run_on:
-    - ubuntu1604-large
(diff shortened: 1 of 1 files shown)


sophstad (commented):
A modern version of Node.js and Yarn come installed on the 16.04 images, but not on newer versions as far as I know (EVG-18682 has some details). We could probably do this ourselves using [nvm](https://github.com/nvm-sh/nvm) like other teams do, or could ask Build for some updates. The former may be better to have full control over the versions we use. (Also if you'd rather pass this off to a UI engineer that's no problem!)

PR #1632, opened 2023-01-26
EVG-12490 Fix the commit queue position being zero index 

malikchaya2:
[EVG-12490](https://jira.mongodb.org/browse/EVG-12490)
### Description
The commit queue position was displayed as 0 index, which didn't match how it was displayed in the commit queue itself. 

I added a function to convert zero to one index and used it in other parts of the codebase where it seems appropriate.

### DIFF
--- a/src/utils/numbers/index.ts
+++ b/src/utils/numbers/index.ts
@@ -29,6 +29,9 @@ const toPercent = (value: string | number): number => {
   return parseFloat(value) * 100;
 };
 
-const convertZeroToOneIndex = (value: number): number => value + 1;
+/**
+ * `formatForDisplay` - Formats a zero-indexed number for display in the UI.
+ */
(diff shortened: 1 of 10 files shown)


src/utils/numbers/index.ts
```diff
@@ -29,4 +29,6 @@ const toPercent = (value: string | number): number => {
   return parseFloat(value) * 100;
 };
 
-export { toDecimal, toPercent };
+const convertZeroToOneIndex = (value: number): number => value + 1;
```
khelif96, line 32: `formatForDisplay`? and add a jsdoc comment t
```suggestion
/**
 * `formatForDisplay` - Formats a zero-indexed number for display in the UI.
 */
const formatForDisplay = (value: number): number => value + 1;
```

khelif96 (commented):
LGTM just had a suggestion for the function name you chose

khelif96 (approved):

PR #1635, opened 2023-01-27
EVG-17549: Use dropdown for project org selection

sophstad:
EVG-17549

### Description

- Update Create Project and Move Project modals to use dropdown for selecting which GitHub org should be used.

### Screenshots

image

image

### DIFF
--- a/src/pages/projectSettings/CreateProjectModal.test.tsx
+++ b/src/pages/projectSettings/CreateProjectModal.test.tsx
@@ -11,13 +11,6 @@ import {
 import { selectLGOption } from "test_utils/utils";
 import { CreateProjectModal } from "./CreateProjectModal";
 
-// Mock out useNavigate to prevent a warning when the page is redirected to the new project on success.
-const mockedUseNavigate = jest.fn();
-jest.mock("react-router-dom", () => ({
-  ...jest.requireActual("react-router-dom"),
-  useNavigate: () => mockedUseNavigate,
-}));
-
 const defaultOwner = "evergreen-ci";
 const defaultRepo = "spruce";
 
@@ -140,7 +133,7 @@ describe("createProjectField", () => {
     const { Component, dispatchToast } = RenderFakeToastContext(
       <NewProjectModal />
     );
-    render(<Component />);
+    const { history } = render(<Component />);
 
     await waitFor(() =>
       expect(screen.queryByDataCy("create-project-modal")).toBeVisible()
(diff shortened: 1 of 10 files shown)


src/pages/projectSettings/CreateProjectModal.test.tsx
```diff
@@ -1,25 +1,37 @@
...
+import { selectLGOption } from "test_utils/utils";
 import { CreateProjectModal } from "./CreateProjectModal";
 
-const defaultOwner = "existing_owner";
-const defaultRepo = "existing_repo";
+const mockedUseNavigate = jest.fn();
+jest.mock("react-router-dom", () => ({
+  ...jest.requireActual("react-router-dom"),
+  useNavigate: () => mockedUseNavigate,
+}));
```
khelif96, line 18: If the goal of this mock was to assert that the modal redirected the user you can test this without mocking useNavigate. 
`renderWithRouterMatch` exposes a history object that you can use to assert on the current location. 


https://github.com/evergreen-ci/spruce/blob/620562260a9e06b7996671a45e9ab33f9f2399cf/src/components/FilterBadges/useFilterBadgeQueryParams.test.tsx#L62-L75

src/pages/projectSettings/CreateProjectModal.tsx
```diff
@@ -38,6 +41,11 @@ export const CreateProjectModal: React.VFC<Props> = ({
   });
   const [hasError, setHasError] = useState(true);
 
+  const { data } = useQuery<GetGithubOrgsQuery>(GET_GITHUB_ORGS);
```
khelif96, line 44: adding a `{skip: !open }` to the params would prevent this query from being kicked off if the user never opens this modal.

khelif96 (changes requested):
Can we add/augment some cypress tests for this?

src/pages/projectSettings/CreateProjectModal.test.tsx
sophstad: My goal here was really to remove a warning:
```
console.warn
    No routes matched location "/project/new-project-name/settings"
```
We've been kinda lax w.r.t. unit test warnings thus far but I think removing them would be preferable. Using the `renderWithRouterMatch` approach doesn't have a way to add additional routes at the moment, so I went with the mock instead.

khelif96: Oh great catch you should be able to augment `renderWithRouterMatch` to support that adding the following line to the method eliminates that warning and is reusable for all tasks.
image

khelif96 (approved):
LGTM

PR #1638, opened 2023-01-27
EVG-16434: Use projectIdentifier in place of projectID where appropriate

minnakt:
EVG-16434

### Description
This PR makes it so that we use the `projectIdentifier` param rather than `projectID` param for certain GraphQL queries.

### Testing
- Patch with module ([link](https://spruce.mongodb.com/version/63d423fa32f4173896f6a32a/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC))
- On staging

### Evergreen PR
* evergreen-ci/evergreen/pull/6218

### DIFF
--- a/src/pages/projectSettings/tabs/GithubCommitQueueTab/GithubCommitQueueTab.tsx
+++ b/src/pages/projectSettings/tabs/GithubCommitQueueTab/GithubCommitQueueTab.tsx
@@ -1,7 +1,6 @@
 import { useMemo } from "react";
 import { useQuery } from "@apollo/client";
 import Banner from "@leafygreen-ui/banner";
-import { useParams } from "react-router-dom";
 import { SpruceForm } from "components/SpruceForm";
 import { ValidateProps } from "components/SpruceForm/types";
 import { ProjectSettingsTabRoutes } from "constants/routes";
@@ -35,14 +34,13 @@ const getInitialFormState = (
 
 export const GithubCommitQueueTab: React.VFC<TabProps> = ({
   githubWebhooksEnabled,
+  identifier,
   projectData,
+  projectId,
   projectType,
   repoData,
   versionControlEnabled,
 }) => {
-  const { projectIdentifier: identifier } = useParams<{
-    projectIdentifier: string;
-  }>();
   const { getTab, updateForm } = useProjectSettingsContext();
   const { formData } = getTab(tab);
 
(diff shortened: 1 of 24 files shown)


src/pages/projectSettings/tabs/GithubCommitQueueTab/GithubCommitQueueTab.tsx
```diff
@@ -51,7 +49,7 @@ export const GithubCommitQueueTab: React.VFC<TabProps> = ({
     GithubProjectConflictsQueryVariables
   >(GET_GITHUB_PROJECT_CONFLICTS, {
     skip: projectType === ProjectType.Repo,
-    variables: { projectId: identifier },
```
minnakt, line 54: All other project settings related queries were using the `projectID` so I changed this query to use `projectID` as well. (Although it really doesn't matter because of how [`byID`](https://github.com/evergreen-ci/evergreen/blob/d1f28a88613c3789e22ad4b1ddd5f2c3935b284c/model/project_ref.go#L1380-L1387) is defined)

minnakt:
evergreen retry

src/pages/projectSettings/tabs/GithubCommitQueueTab/GithubCommitQueueTab.tsx
khelif96: Sorry if I missed this but why does this use `projectId` instead of `identifier`?

minnakt: Project settings uses `projectId` for a lot of different mutations. I assume that part of the reason is that repos don't have a concept of an `identifier`.

Screen Shot 2023-01-31 at 7 06 23 PM

This is fine and correct because we actually do send the `ID` to these queries on Spruce.

https://github.com/evergreen-ci/spruce/blob/19d98d6775831ce1ec8d4f681e1f7440839b7e69/src/pages/projectSettings/Tabs.tsx#L42

So all I'm doing here is using the `projectID` as the query specifies. There could be a more thorough audit of project settings queries/mutations but it would need a larger effort.

khelif96 (approved):

minnakt:
I'll merge this EOD as long as the Evergreen deploy does not get reverted

minnakt:
Merging because Storybook is flaky, though I feel like it might be getting worse 🧍

PR #1639, opened 2023-01-27
EVG-16928: Dynamic mainline commit column count

SupaJoon:
[EVG-16928](https://jira.mongodb.com/browse/EVG-16928)
### Description
The number mainline commit columns on the Commits page will adjust to the chart container width. 
You can test it locally here: http://localhost:3000/commits/spruce

### DIFF
--- a/src/pages/commits/hooks/useCommitLimit.ts
+++ b/src/pages/commits/hooks/useCommitLimit.ts
@@ -3,6 +3,9 @@ import { useDimensions } from "hooks/useDimensions";
 import { useQueryParam } from "hooks/useQueryParam";
 import { MainlineCommitQueryParams } from "types/commits";
 
+/**
+ * `useCommitLimit` is a hook that calculates the number of commits to fetch based on the width of the commits container.
+ * */
 export const useCommitLimit = <T extends HTMLElement>(): [
   MutableRefObject<T>,
   number


src/pages/Commits.tsx
```diff
@@ -200,13 +201,15 @@ export const Commits = () => {
...
           />
         </PaginationWrapper>
-        <CommitsWrapper
-          versions={versions}
-          error={error}
-          isLoading={loading || !projectIdentifier}
-          hasTaskFilter={hasTasks}
-          hasFilters={hasFilters}
-        />
+        <span ref={ref}>
```
khelif96, line 204:
```suggestion
        <div ref={ref}>
```

I think a div is the more "semantically correct" wrapper since `CommitsWrapper` represents several block components.

src/hooks/useDimensions.ts
```diff
@@ -0,0 +1,29 @@
+import { MutableRefObject, useLayoutEffect, useMemo, useState } from "react";
+
+interface DimensionState {
+  width: number;
+  height: number;
+}
+
+export const useDimensions = (ref: MutableRefObject<HTMLElement>) => {
```
khelif96, line 8: So we already have an AutoSizer component how do you feel about this hook approach should we be using it instead? If so I think we should consolidate on one approach for calculating the size of elements/ the page.

https://github.com/evergreen-ci/spruce/blob/a02985f93b266377640380220b1d3142cb5e0ae8/src/components/HistoryTable/HistoryTable.tsx#L84

src/hooks/useDimensions.ts
```diff
@@ -0,0 +1,29 @@
+import { MutableRefObject, useLayoutEffect, useMemo, useState } from "react";
```
khelif96, line 1: I wanna advocate for following this pattern for hooks so we can colocate the tests and hooks in the same folder. 
```bash
hooks/useDimensions/index.ts
hooks/useDimensions/useDimensions.test.ts
```

src/hooks/useDimensions.ts
khelif96: Makes sense to me thanks for the explanation!

khelif96, line 8: Can we add some jsdoc comments for this.

src/hooks/useDimensions.ts
```diff
@@ -0,0 +1,29 @@
...
+  width: number;
+  height: number;
+}
+
+export const useDimensions = (ref: MutableRefObject<HTMLElement>) => {
+  const [state, setState] = useState<DimensionState>({ width: 0, height: 0 });
+
+  const observer = useMemo(
+    () =>
+      new (window as any).ResizeObserver((entries) => {
```
khelif96, line 13: I'm confused by this type casting why can't you just use `new ResizeObserver...`

src/hooks/useDimensions.ts
```diff
@@ -0,0 +1,29 @@
...
+
+interface DimensionState {
+  width: number;
+  height: number;
+}
+
+export const useDimensions = (ref: MutableRefObject<HTMLElement>) => {
+  const [state, setState] = useState<DimensionState>({ width: 0, height: 0 });
+
+  const observer = useMemo(
```
khelif96, line 11: This useMemo isn't necessary you can declare this in a singular useEffect and have it handle the disconnect state.  `useIntersectionObserver` should behave similarly.
https://github.com/evergreen-ci/spruce/blob/c4b2c563e1f49d71e3a509122c594568025aa65a/src/hooks/useIntersectionObserver/index.ts

src/hooks/useDimensions.ts
```diff
@@ -0,0 +1,29 @@
...
+    () =>
+      new (window as any).ResizeObserver((entries) => {
+        const { width, height } = entries[0]?.contentRect ?? {};
+        setState({ width, height });
```
khelif96, line 15: Should this be debounced?

src/hooks/useCommitsLimit.ts
```diff
@@ -0,0 +1,25 @@
+import { MutableRefObject, useEffect, useRef } from "react";
```
khelif96, line 1: Same comment about folder structure. 
Although i would argue this should be colocated in the commits folder since this only applies to that page.

src/hooks/useCommitsLimit.ts
```diff
@@ -0,0 +1,25 @@
...
+import { useSearchParams } from "react-router-dom";
+import { MainlineCommitQueryParams } from "types/commits";
+import { useDimensions } from "./useDimensions";
+
+export const useCommitsLimit = <T extends HTMLElement>(): [
+  MutableRefObject<T>,
+  number
+] => {
+  const [params, setParams] = useSearchParams();
```
khelif96, line 10: lets use `useQueryParam`
https://github.com/evergreen-ci/spruce/blob/7681161d500c3783de6ed15106c618b8dae507e0/src/hooks/useQueryParam/index.ts

src/hooks/tests/useDimensions.test.ts
```diff
@@ -0,0 +1,133 @@
+/* eslint-disable class-methods-use-this */
```
khelif96, line 1: Check out the tests for `useIntersectionObserver` they behave similarly 
https://github.com/evergreen-ci/spruce/blob/c4b2c563e1f49d71e3a509122c594568025aa65a/src/hooks/useIntersectionObserver/useIntersectionObserver.test.ts

khelif96 (changes requested):

src/pages/commits/hooks/useCommitsLimit.ts
```diff
@@ -0,0 +1,25 @@
...
+] => {
+  const [params, setParams] = useQueryParams();
+  const commitsContainerRef = useRef<T>();
+  const { width } = useDimensions(commitsContainerRef);
+  const nextLimit = Math.max(Math.round(width / COL_WIDTH), MIN_LIMIT);
+
+  useEffect(() => {
+    const nextParams = { ...params };
+    delete nextParams[MainlineCommitQueryParams.SkipOrderNumber];
+    setParams(nextParams);
```
khelif96, line 18: I would recommend `useQueryParam` instead of `useQueryParams`  the distinction is that we only need to update the SkipOrderNumber param

const [skipOrderNumber
```suggestion
  const [, setSkipOrderNumber] = useQueryParam(
    MainlineCommitQueryParams.SkipOrderNumber,
    0
  );
  const commitsContainerRef = useRef<T>();
  const { width } = useDimensions(commitsContainerRef);
  const nextLimit = Math.max(Math.round(width / COL_WIDTH), MIN_LIMIT);

  useEffect(() => {
    setSkipOrderNumber(undefined);
  }, [nextLimit]); // eslint-disable-line react-hooks/exhaustive-deps


```

src/pages/commits/hooks/useCommitsLimit.ts
```diff
@@ -0,0 +1,25 @@
+import { MutableRefObject, useEffect, useRef } from "react";
+import { useDimensions } from "hooks/useDimensions";
+import { useQueryParams } from "hooks/useQueryParam";
+import { MainlineCommitQueryParams } from "types/commits";
+
+export const useCommitsLimit = <T extends HTMLElement>(): [
```
khelif96, line 6: What do you think about `useCommitLimit`?
```suggestion
export const useCommitLimit = <T extends HTMLElement>(): [
```

src/pages/commits/hooks/useCommitsLimit.ts
```diff
@@ -0,0 +1,25 @@
+import { MutableRefObject, useEffect, useRef } from "react";
+import { useDimensions } from "hooks/useDimensions";
+import { useQueryParams } from "hooks/useQueryParam";
+import { MainlineCommitQueryParams } from "types/commits";
+
```
khelif96, line 5:
```suggestion
/**
 * `useCommitLimit` is a hook that calculates the number of commits to fetch based on the width of the commits container. 
 * */ 
```

khelif96 (approved):
Great work on this just had a few suggestions!

PR #1641, opened 2023-01-31
EVG-18835: Don't access property of undefined

SupaJoon:
[EVG-18835](https://jira.mongodb.com/browse/EVG-18835)

### DIFF
--- a/src/components/Redirects/UserPatchesRedirect.tsx
+++ b/src/components/Redirects/UserPatchesRedirect.tsx
@@ -3,5 +3,5 @@ import { getUserPatchesRoute } from "constants/routes";
 
 export const UserPatchesRedirect: React.VFC = () => {
   const { id } = useParams<{ id: string }>();
-  return <Navigate to={getUserPatchesRoute(id)} />;
+  return <Navigate replace to={getUserPatchesRoute(id)} />;
 };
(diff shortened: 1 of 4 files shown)


src/components/PatchesPage/PatchCard.tsx
```diff
@@ -73,7 +73,7 @@ export const PatchCard: React.VFC<Props> = ({
         <strong>{projectIdentifier}</strong>
       </StyledRouterLink>
     ) : (
-      `${projectMetadata.owner}/${projectMetadata.repo}`
+      `${projectMetadata?.owner}/${projectMetadata?.repo}`
```
khelif96, line 76: While this temporarily alleviates the issue it doesn't get at the real cause. This results in these values being undefined. 
image

Ideally, we should  try to understand why these values are undefined in the first place. 

Looks like we are missing both  `projectIdentifier` and `projectMetadata`.

khelif96 (changes requested):

minnakt:
More pointers that might help: I noticed that these `undefined` values will show when a repo patch has not been configured. Repo patches that have been configured will properly show the owner and repo. This sort of makes sense to me as an unconfigured patch does not have a corresponding version, so if you try to get the `projectMetadata` from it, it will be null.

Screen Shot 2023-02-01 at 9 51 25 AM

Also I have a feeling that the project identifier is supposed to be null, otherwise why would we have the fallback in the first place. 🤔

khelif96 (approved):

PR #1642, opened 2023-01-31
EVG-18797: Make Hosts table Owner control always visible

SupaJoon:
[EVG-18797](https://jira.mongodb.com/browse/EVG-18797)


src/pages/hosts/HostsTable.tsx
```diff
@@ -301,3 +303,6 @@ export const HostsTable: React.VFC<Props> = ({
 };
 
 const rowKey = ({ id }: { id: string }): string => id;
+const Owner = styled.span`
+  word-break: break-all;
```
khelif96, line 307: You can reuse the word break style we already have
https://github.com/evergreen-ci/spruce/blob/620562260a9e06b7996671a45e9ab33f9f2399cf/src/components/styles/Typography.tsx#L11

khelif96 (changes requested):

minnakt (approved):
Aside from Mohamed's comment, I think wrapping the Current Task column in `WordBreak` would also be helpful because it frequently overflows its assigned width.

Also if you wanted to redistribute the width percentages I wouldn't be opposed to it!

src/pages/hosts/HostsTable.tsx
```diff
@@ -117,7 +117,7 @@ export const HostsTable: React.VFC<Props> = ({
       sorter: true,
       className: "cy-hosts-table-col-ID",
       defaultSortOrder: getDefaultSortOrder(HostSortBy.Id),
-      width: "15%",
+      width: "17%",
```
khelif96, line 120: nit.  This comment is more so about my personal feelings toward "magic numbers" as opposed to an actual complaint.  How would this look if we used more rounded numbers? Like maybe 15% and 20% (on line 171)

khelif96 (approved):
LGTM

PR #1654, opened 2023-02-02
EVG-18849: Upgrade Vite

sophstad:
EVG-18849

### Description

- Upgrade Vite to v4
- I noticed that this Parsley PR https://github.com/evergreen-ci/parsley/pull/173 converted `.storybook/main.js` to `main.ts`. I think there's a limitation with Storybook 6.x and Vite 3/4 that breaks Typescript for this file. It can be converted back to TS when upgrading Storybook in EVG-18851.

### Testing

- I validated that `yarn storybook` works locally


.storybook/main.js
```diff
@@ -1,8 +1,4 @@
-import type { StorybookViteConfig } from "@storybook/builder-vite";
-import { mergeConfig, PluginOption } from "vite";
-import viteConfig from "../vite.config";
-
-const storybookConfig: StorybookViteConfig = {
+module.exports = {
```
khelif96, line 1: I'm curious about what the bug you encountered was. Did it have to do with storyshots?

minnakt (approved):
This lgtm considering the Storybook config will be updated further in [EVG-18851](https://jira.mongodb.org/browse/EVG-18851). great job Sophie!

.storybook/main.js
sophstad: Not exactly, there's a TypeScript issue that pops up when we try to remove Storybook's React plugin. It never seems to have been properly resolved in v6.x.

```sh
.storybook/main.ts:57:17 - error TS2339: Property 'name' does not exist on type 'Plugin_2 | Promise<false | Plugin_2 | PluginOption[]>'.
  Property 'name' does not exist on type 'Promise<false | Plugin_2 | PluginOption[]>'.

57   return plugin.name === name;
```

Minna even mentioned it back in the day here 😆 https://github.com/evergreen-ci/parsley/pull/6

PR #1660, opened 2023-02-07
EVG-18269 Add s3 credential requester to project create/copy modals

bynn:
EVG-18269

### Description
add checkbox to create and copy project that will create a jira ticket when checked

### Screenshots
image

image

### Testing
unit + staging 
tickets are created for both create and copy


src/pages/projectSettings/CreateDuplicateProjectButton.tsx
```diff
@@ -90,13 +90,15 @@ export const CreateDuplicateProjectButton: React.VFC<Props> = ({
...
           repo={repo}
+          requestS3Creds={false}
         />
       )}
       <CopyProjectModal
         handleClose={() => setCopyModalOpen(false)}
         id={id}
         label={label}
         open={copyModalOpen}
+        requestS3Creds={false}
```
sophstad, line 101: It doesn't look like either of these components actually use this prop so no need to add it!

sophstad (changes requested):

src/pages/projectSettings/sharedFormSchema.ts
```diff
@@ -23,3 +23,15 @@ export const projectId = {
     "ui:optional": true,
   },
 };
+
+export const requestS3Creds = {
+  schema: {
+    type: "boolean" as "boolean",
+    title: "Request S3 Credentials",
```
sophstad, line 30: We could probably provide some more details here. Maybe a description that says something like "A JIRA ticket will be opened for the Build team enabling this project to read and write to an S3 bucket." Also, is there any means of providing users with the ticket number that is created?

bynn: done

src/pages/projectSettings/CreateDuplicateProjectButton.tsx
bynn: removed i think (i dont know if i understand this completely sorry)

sophstad (changes requested):
There are still some failing unit tests related to these changes!

src/pages/projectSettings/sharedFormSchema.ts
bynn: oh also there is no point in providing users with the ticket because we currently do not do that 
and it is difficult to get the ticket because we create a notification that triggers a jira ticket creation

bynn:
evergreen retry

ablack12:
We may want to change this to "Request S3 Bucket" since that's the wording on https://wiki.corp.mongodb.com/display/BUILD/How+to+Create+a+New+Evergreen+Project but whichever works (i think they don't use prefixes anymore for this though)

sophstad (approved):
🥳 thank you Bynn!

PR #1663, opened 2023-02-07
EVG-18384: Use task.taskFiles instead of query.taskFiles

minnakt:
EVG-18384

### Description
This PR makes it so that we get `taskFiles` from the task resolver, not the query resolver.

### Testing
- Patch with module ([link](https://spruce.mongodb.com/version/63e2aa8dd6d80a7f58babe87/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC))

### Evergreen PR
* evergreen-ci/evergreen/pull/6246


src/gql/queries/get-task.graphql
```diff
@@ -2,112 +2,112 @@
 #import "../fragments/annotation.graphql"
 
 query GetTask($taskId: String!, $execution: Int) {
```
khelif96, line 4: Assuming this diff is just reorganizing fields and the fields are the same lgtm.

src/pages/task/TaskTabs.tsx
```diff
@@ -21,29 +21,25 @@ const { parseQueryString } = queryString;
...
     execution,
-    totalTestCount,
     executionTasksFull,
+    failedTestCount,
+    isPerfPluginEnabled,
+    logs: logLinks,
+    status,
+    taskFiles,
+    totalTestCount,
   } = task ?? {};
```
khelif96, line 43: I think we should look into introducing some linter rules to auto-sort these for us in the future.

khelif96 (approved):

src/pages/task/TaskTabs.tsx
minnakt: Filed [EVG-18913](https://jira.mongodb.org/browse/EVG-18913)

PR #1664, opened 2023-02-07
EVG-18290: Use build ID for job logs page

sophstad:
EVG-18290

### Description

- Use build ID in job logs page path
- Disable page on production while being developed (based on New Relic stats this page is not really visited at all, <5 during the month which may have been mostly me, but this is still a nice safeguard to add)

### Screenshots

image
image


minnakt (approved):
lgtm!

PR #1668, opened 2023-02-08
EVG-18647 Include executions in exec tasks table links

khelif96:
EVG-18647

### Description
Execution tasks links now include an execution 
### Screenshots
image


minnakt (approved):
well that makes sense to me

PR #1670, opened 2023-02-08
EVG-18781: Safely search for patches by number

sophstad:
EVG-18781

### Description

- Stringify query used to filter patch names so that numbers don't throw an error

### Testing

- Update e2e test


minnakt (approved):
works great!

khelif96 (approved):

PR #1671, opened 2023-02-08
EVG-17355: Prevent patches page from locking up browser history

SupaJoon:
[EVG-17355](jira.mongodb.com/browse/EVG-17355)

### Description
You can reproduce the issue by navigating to localhost:3000, navigating to any page in the app and then trying to use the browser back button to navigate to the initial page. You can confirm the issue by logging the result of `useLocation` someplace directly underneath `<Router />` like GQLWrapper.
The same issues occurs when navigating to Spruce, then `/user/<user_name>` then trying to use the back button to go to the initial page before Spruce loaded. 
I saw that the browser tries to load `/` but then immediately pushes redirect params to the browser history causing a loop. I fixed this by passing in `replace={true}` to Navigate.
I'm unable to recreate this issue for commits by going to localhost:3000/commits, navigating to another page in the app and then clicking the back button.


khelif96 (commented):
Should we make this change in the below pages as well? They all perform similar immediate redirects
https://github.com/evergreen-ci/spruce/blob/71b8a6253f06da15d08562866aae25c5854f9608/src/pages/ConfigurePatch.tsx
https://github.com/evergreen-ci/spruce/blob/b592b125bb211f828af0f8b5961cbc1c78929660/src/components/Redirects/WaterfallCommitsRedirect.tsx
https://github.com/evergreen-ci/spruce/blob/620562260a9e06b7996671a45e9ab33f9f2399cf/src/pages/Preferences.tsx#L19
https://github.com/evergreen-ci/spruce/blob/3550bc848cbbd625e64a3bd5e43125cd9744b603/src/pages/Version.tsx#L133

khelif96 (approved):

PR #1679, opened 2023-02-13
EVG-18288: Add job logs page

sophstad:
EVG-18288

### Description

- I didn't make any design updates for this iteration vs the [legacy page](https://logkeeper2.build.10gen.cc/build/9b07c7f9677e49ddae4c53076ca4f4ca). We can involve Design for any future changes.
- I made a change to our env variables approach: local, non-sensitive variables for `local`, `devLocal`, `devStaging`, and `devProd` environments can live in `env/.cmdrc-local.json` and be checked into git. The other environments will continue to use the current untracked file. This will reduce the amount of file updates we'll have to share, and I think it's helpful to have the dev environment configured by default when pulling down Spruce.

### Screenshots

image

### Testing

- Added a logkeeper server for e2e testing of the job logs page. Test log contents are fetched from the Parsley s3 bucket—let me know if there are any objections to this. It seems potentially beneficial to operate with the same test data.

### Evergreen PR

This PR is merged and deployed but introduced relevant changes: https://github.com/evergreen-ci/evergreen/pull/6249


khelif96:
https://github.com/evergreen-ci/evergreen/blob/fafd9d3e06716c355a7fbcb259bfad325bbdb9fb/testdata/local/admin.json

Is used to seed the admin collection we populate other admin variables such as uiv2_url and parsley_url you should be able to seed the logkeeper URL for your tests to use a local logkeeper

minnakt (approved):
lgtm!

README.md
```diff
@@ -44,7 +44,7 @@ Follow these directions to enable query linting during local development so your
 
 ### Environment Variables
 
-[env-cmd](https://github.com/toddbluhm/env-cmd#readme) is used to configure build environments for production, staging and development. This file is git ignored because it contains API keys that we do not want to publish. It should be named `.cmdrc.json` and placed in the `env/` folder at the root of the project. This file is required to deploy Spruce to production and to staging. Ask a team member to send you their copy of the file, which should look like the following:
+[env-cmd](https://github.com/toddbluhm/env-cmd#readme) is used to configure build environments for production, staging and development. We use two files to represent these various environments: `.cmdrc-local.json` for local builds with non-sensitive information, and `.cmdrc.json` for builds deployed to S3. This file is git ignored because it contains API keys that we do not want to publish. It should be named `.cmdrc.json` and placed in the `env/` folder at the root of the project. This file is required to deploy Spruce to production and to staging. Ask a team member to send you their copy of the file, which should look like the following:
```
khelif96, line 47: This change makes sense to me although I would advocate for splitting it out of this pr, to make reverts easier. (We can revert the feature without having to worry about reverting how we handle credentials) 
Also could we do this on parsley for consistency?

src/pages/JobLogs.tsx
```diff
@@ -1,18 +1,49 @@
...
+  >(GET_LOGKEEPER_BUILD_METADATA, {
+    variables: { buildId },
+    onError: (err) => {
+      dispatchToast.error(
+        `There was an error retrieving logs for this build: ${err.message}`
+      );
+    },
+  });
+
+  if (!data) return null;
```
khelif96, line 41: Should we show a loading state?

khelif96 (approved):
This LGTM with the exception of the cmdrc changes being in this same pr. They don't seem related and should be in a separate commit imo.

README.md
sophstad: https://github.com/evergreen-ci/spruce/pull/1686

src/pages/JobLogs.tsx
```diff
@@ -32,22 +63,44 @@ export const JobLogs = () => {
...
+              data-cy="complete-test-logs-link"
+              target="_blank"
+            >
               Complete logs for all tests
             </Button>
           </CompleteLogsLink>
+          {tests.length ? (
+            <JobLogsTable buildId={buildId} tests={tests} />
+          ) : (
+            <TablePlaceholder message="No test results found." />
```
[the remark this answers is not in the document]
sophstad: Sadly I do not think the task duration table works in this way 😢 LeafyGreen rejects non-`Row` children within its tables. But I moved the placeholder into `JobLogsTable`.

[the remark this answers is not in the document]
minnakt: The `TablePlaceholder` for the task duration table is actually outside of the Leafygreen `Table`! It just exists in a wrapper around the table as a workaround. So that's why you can continue seeing the headers:

Screen Shot 2023-02-15 at 5 24 27 PM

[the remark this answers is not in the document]
sophstad: ahhh I see, the difference is that there's no ternary to render the table _or_ the placeholder. Makes sense, ty!

src/pages/JobLogs.tsx
khelif96: I was imagining using a skeleton and the 404 graphic to represent loading and not found like we do for the task and version pages.

PR #1682, opened 2023-02-14
EVG-17915 Code split Spruce pages

khelif96:
EVG-17915

### Description
Separates out less frequently visited pages from the main bundle. This should improve initial load times for users with slower internet connections by not sending code for pages they may not visit

### Screenshots
Bundle Before and After 
image

Main Bundle size
 Before: 5.94 MB
 After: 3.66 MB


khelif96:
> Is each loadable statement a separate bundle? Could you mention how I could produce the same bundle visualization locally?

Yes loadable is just a wrapper around [React.lazy and suspense ](https://reactjs.org/docs/code-splitting.html) 

https://github.com/evergreen-ci/spruce/blob/71b8a6253f06da15d08562866aae25c5854f9608/src/components/SpruceLoader/index.tsx 

You can find it under the build directory after running `yarn build` locally. It's saved to a file called `source_map.html`

PR #1686, opened 2023-02-15
EVG-18972: Split env file into two

sophstad:
EVG-18972

### Description

- Split our `env/.cmdrc.json` file into two: `env/.cmdrc.json` which is for deployed environments and contains secrets, and `env/.cmdrc-local.json`, which is tracked in git and does not.
- This PR split from https://github.com/evergreen-ci/spruce/pull/1679/files


khelif96 (approved):
💯

PR #1688, opened 2023-02-16
EVG-16928: Dynamic mainline commit count

SupaJoon:
[EVG-16928](jira.mongodb.com/browse/EVG-16928)

### Description
ResizeObserver code changes are sending a benign error to bugsnag. The ResizeObserver was not able to deliver all observations within a single animation frame so I wrapped the side effects of the resize requestAnimationFrame to limit the executions to a single frame.
I logged the constructor of the Resize Observer and I didn't detect any loops. What's likely happening is that the resize callback triggers a network request resulting in a resize of the observed box.
Ideally Bugsnag could allow us to prevent reporting the error originating from specific usage but I wasn't able to identify a configuration that allows this.
We already handle this exception in our cypress suite [here](https://github.com/evergreen-ci/spruce/blob/5d875423f27af33d86d14d094cb81e11a423d9f7/cypress/support/e2e.ts#L26)
You can read more information on the error here:
[Stackoverflow](https://stackoverflow.com/questions/49384120/resizeobserver-loop-limit-exceeded)
[Webkit](https://webkit.org/blog/9997/resizeobserver-in-webkit/#:~:text=observation.-,How%20Observations%20are%20Delivered,ResizeObserver,-strictly)

### DIFF
--- a/src/gql/generated/types.ts
+++ b/src/gql/generated/types.ts
@@ -1349,6 +1349,7 @@ export type Query = {
   subnetAvailabilityZones: Array<Scalars["String"]>;
   task?: Maybe<Task>;
   taskAllExecutions: Array<Task>;
+  /** @deprecated Use task.taskLogs instead */
   taskLogs: TaskLogs;
   taskNamesForBuildVariant?: Maybe<Array<Scalars["String"]>>;
   taskQueueDistros: Array<TaskQueueDistro>;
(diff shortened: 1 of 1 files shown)


src/hooks/useDimensions/index.ts
```diff
@@ -0,0 +1,37 @@
...
+ */
+export const useDimensions = (ref: MutableRefObject<HTMLElement>) => {
+  const [state, setState] = useState<DimensionState>({ width: 0, height: 0 });
+
+  const observer = useMemo(
+    () =>
+      new ResizeObserver((entries) => {
+        requestAnimationFrame(() => {
+          const { width, height } = entries[0]?.contentRect ?? {};
+          setState({ width, height });
```
khelif96, line 22: I'm not entirely sure if this is preferable to a debounce. Especially considering setState calls are async so there is no guarantee that they will actually execute within the requestAnimationFrame.  We can monitor bugsnag if this comes up.

khelif96 (approved):
LGTM apart from making the test changes

PR #1690, opened 2023-02-21
EVG-18676: Upgrade html-react-parser to v3.0.9

minnakt:
EVG-18676

### Description
This PR upgrades `html-react-parser`. Major changes include:
- bump domhandler from 4 to 5, html-dom-parser from 2 to 3

### Testing
- Looked at the task logs, since this is the only place where the library is called

### DIFF
--- a/src/pages/task/taskTabs/logs/LogTypes.tsx
+++ b/src/pages/task/taskTabs/logs/LogTypes.tsx
@@ -212,6 +212,6 @@ const StyledPre = styled.pre`
   border: 1px solid ${gray.light2};
   border-radius: ${size.xxs};
   font-size: ${fontSize.m};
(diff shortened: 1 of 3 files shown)


src/pages/task/taskTabs/logs/LogTypes.tsx
```diff
@@ -212,6 +212,6 @@ const StyledPre = styled.pre`
   border: 1px solid ${gray.light2};
   border-radius: ${size.xxs};
   font-size: ${fontSize.m};
-  overflow-x: scroll;
+  overflow: scroll hidden;
```
minnakt, line 215: there is a double vertical scrollbar on Firefox (you can look at a task on staging environment to verify this), so this is just a change to get rid of that

PR #1692, opened 2023-02-22
EVG-18763: Add job logs analytics

sophstad:
EVG-18763

### Description

- Add analytics when user clicks "complete logs" button or clicks on a link to Parsley
- Remove "Work in progress" placeholder"


src/analytics/joblogs/useJobLogsAnalytics.ts
```diff
@@ -9,7 +9,8 @@ type Action =
   | { name: "Filter Job Logs"; filterBy: string }
   | { name: "Clicked lobster testlog url"; testId: string }
   | { name: "Clicked HTML testlog url"; testId: string }
-  | { name: "Clicked complete logs link"; taskId: string };
+  | { name: "Clicked complete logs link"; taskId: string }
+  | { name: "Clicked Parsley test log link"; testId: string };
```
khelif96, line 13: Maybe tracking buildId is more valuable? So we can get a sense of the amount of logs visited per build? Product would provide a better sense of the analytics they want.

khelif96 (approved):

PR #1694, opened 2023-02-22
EVG-19035 Remove enabled from repo settings page

bynn:
EVG-19035

### Description
we are going to be removing this from the backend so were removing it from the ui first 

### Screenshots
image

### Testing
no existing tests fail

### DIFF
--- a/src/pages/projectSettings/tabs/GeneralTab/types.ts
+++ b/src/pages/projectSettings/tabs/GeneralTab/types.ts
@@ -2,7 +2,7 @@ import { ProjectType } from "../utils";
 
 export interface FormState {
   generalConfiguration: {
-    enabled?: boolean | null;
+    enabled?: boolean;
     repositoryInfo: {
       owner: string;
       repo: string;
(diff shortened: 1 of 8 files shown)


sophstad (changes requested):
Perhaps we should also remove the [request for this field](https://github.com/evergreen-ci/spruce/blob/635bac7d3b16a035adfe401040fa7ef4085ff334/src/gql/fragments/projectSettings/general.graphql#L24) in GraphQL? This will make it easier to remove the back end logic since there will be no existing front end dependencies.

src/pages/projectSettings/tabs/GeneralTab/types.ts
```diff
@@ -2,7 +2,7 @@ import { ProjectType } from "../utils";
 
 export interface FormState {
   generalConfiguration: {
-    enabled: boolean | null;
+    enabled?: boolean | null;
```
sophstad, line 5: Since attached projects can no longer default to repo (the `null` value) this can just be
```suggestion
    enabled?: boolean;
```

sophstad (approved):
🤙 nice!

PR #1706, opened 2023-02-28
EVG-17384: Surface warnings and errors on mainline commits

minnakt:
EVG-17384

### Description
This PR makes it so that warnings and errors in configuration files are surfaced on mainline commits. This matches the behavior on the legacy UI, as warnings and errors are not shown on user patches.

### Screenshots
Screenshot 2023-02-28 at 1 31 07 PM

Screenshot 2023-02-28 at 1 31 18 PM

### Testing
- Cypress tests in `version/banners.ts`
- Jest tests in `WarningBanner.test.tsx` and `ErrorBanner.test.tsx`
- Patch with module: ([link](https://spruce.mongodb.com/version/63fe55be3066154bf11484a3/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC))

### Evergreen PR
- evergreen-ci/evergreen/pull/6300
- You can test these changes locally by pulling the corresponding Evergreen PR, resetting your local data, and visiting [this link](http://localhost:3000/version/logkeeper_e864cf934194c161aa044e4599c8c81cee7b6237/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC).


src/pages/Version.tsx
```diff
@@ -157,6 +160,11 @@ export const VersionPage: React.VFC = () => {
 
   return (
     <PageWrapper data-cy="version-page">
+      <BannerWrapper>
+        {errors.length > 0 && <ErrorBanner errors={errors} />}
+        {warnings.length > 0 && <WarningBanner warnings={warnings} />}
+      </BannerWrapper>
```
khelif96, line 166: Can we omit the BannerWrapper if there are no errors or warnings?

khelif96 (approved):

src/pages/version/Banners/ErrorBanner/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+  const [showModal, setShowModal] = useState(false);
+
+  const errorTitle =
+    errors.length === 1
+      ? "1 error in configuration file"
+      : `${errors.length} errors in configuration file`;
+
+  return (
+    <BannerContainer data-cy="configuration-errors-banner">
+      <Callout variant="warning" title={errorTitle.toUpperCase()}>
```
sophstad, line 27: Might we not capitalize the title text here? Since it's not capitalized for warnings it seems preferable to standardize

src/pages/version/Banners/ErrorBanner/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+        {errors[0]}
```
sophstad, line 28: I'm not sure I'm convinced that we should show a full error here. Since the error banner isn't dismissible, it results in the banner taking up a large portion of the screen, and could persist for users who aren't really able to take action on it. These banners are already a lot more prominent on the page than the legacy implementation, so I'm hesitant to claim even more screen real estate. But what do you think?

minnakt: My only hang-up is how the Callout looks oddly bare (like something's missing) if we take away the full error:

Screenshot 2023-03-01 at 6 39 13 PM

I'll ask more about this in design sync today :>

sophstad (approved):

minnakt:
evergreen refresh

PR #1708, opened 2023-02-28
EVG-18317: Update userEvent import statements and remove fireEvent usage

minnakt:
EVG-18317

### Description
This PR changes the `userEvent` import statements from
```
import userEvent from "@testing-library/user-event"
```
to the following
```
import { userEvent } from "test_utils"
```

I also removed all instances of `fireEvent` and replaced them with `userEvent`.

### Testing
- Tests should pass.


khelif96 (approved):
Great refactor

PR #1719, opened 2023-03-02
EVG-16343: Improve set priority usability

minnakt:
EVG-16343

### Description
This PR aims to improve the usability of the set priority button. Now the priority input will be auto-focused as well as auto-selected, and the Enter key can be used to submit the input.

I created a new `Popconfirm` component using LeafyGreen components. There was already an existing component named `Popconfirm` that was using antd components, so I just renamed that one. I also replaced the other popovers within the menu dropdown because I thought it would look pretty weird if they didn't match.

### Screenshots

https://user-images.githubusercontent.com/47064971/222487799-1703f533-9fed-486a-bece-d58b20353b2b.mov

### Testing
- Added `Popconfirm.test.tsx`
- Modified Cypress tests


src/components/Popconfirm/index.tsx
```diff
@@ -1,72 +1,92 @@
...
-  checkboxLabel, // truthiness determines if checkbox is rendered
-  children,
-  onConfirm,
-  title,
+const Popconfirm: React.VFC<PopconfirmProps> = ({
+  active,
+  align = "top",
+  justify = "middle",
+  "data-cy": dataCy,
+  content,
```
sophstad, line 28: Using `children` instead of a `content` prop may feel more idiomatic here

src/components/PatchActionButtons/SetPatchPriority.tsx
```diff
@@ -37,38 +43,53 @@ export const SetPatchPriority: React.VFC<SetPriorityProps> = ({
...
 
-  const { sendEvent } = useVersionAnalytics(patchId);
+  const onConfirm = () => {
+    setPatchPriority({ variables: { patchId, priority } });
+    sendEvent({ name: "Set Priority", priority });
+  };
+
+  useEffect(() => {
+    inputRef?.focus();
+    inputRef?.select();
```
sophstad, line 53: This makes for a beautiful UX 😻

sophstad (approved):

src/components/AntdPopconfirm/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+interface PopconfirmWithCheckboxProps {
+  onConfirm: (e: React.MouseEvent) => void;
+  title: string;
+  checkboxLabel?: string;
+  children: React.ReactNode;
+}
+
+export const PopconfirmWithCheckbox: React.VFC<PopconfirmWithCheckboxProps> = ({
+  checkboxLabel, // truthiness determines if checkbox is rendered
```
khelif96, line 30: My above comment aside assuming you don't want to make those changes in this pr. 

I'm not sure how I feel about this implementation. If a user omits a checkbox label then this component is effectively just a normal Popconfirm. Either we should bake this functionality into the original Popconfirm or we should make checkboxLabel a required prop.

src/components/PatchActionButtons/SetPatchPriority.tsx
khelif96: Makes all the difference 💯

khelif96 (changes requested):
Love the improvements just had some questions!

src/components/AntdPopconfirm/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+import styled from "@emotion/styled";
+import Checkbox from "@leafygreen-ui/checkbox";
+import { Popconfirm as AntPopconfirm } from "antd";
+import { size } from "constants/tokens";
+
+interface Props extends React.ComponentProps<typeof AntPopconfirm> {
+  children: React.ReactNode;
+}
+
+export const Popconfirm: React.VFC<Props> = ({ children, ...props }) => (
```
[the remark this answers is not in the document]
minnakt: I would be a little nervous to make all of those changes here. I think this could be handled by the existing ticket [EVG-18525](https://jira.mongodb.org/browse/EVG-18525).

src/components/AntdPopconfirm/index.tsx
minnakt: This is not actually my change (I just renamed the file), but sure I can make it a required prop!

khelif96 (approved):

PR #1724, opened 2023-03-08
EVG-18945: Upgrade Menu

sophstad:
EVG-18945

### Description

- Upgrade Menu, which fixes some of the weird properties we've seen around `as`
- The type of `MenuItem` has been restricted so it does not accept refs anymore. This is pretty easy to work around for the new Popconfirm by wrapping the menu items in a div containing the corresponding ref.


minnakt (approved):
🥳

minnakt:
evergreen refresh

PR #1725, opened 2023-03-09
EVG-18525: Replace Antd Popconfirm

sophstad:
EVG-18525

### Description

- Replace Antd's popconfirm with our own implementation
- I added an uncontrolled version of Popconfirm that accepts just a trigger and internally manages the open state. This was easier when implementing components where the popconfirm element is mapped, which would have required tracking a lot of states and refs in the parent component.
- ...should we just use [Tooltip](https://www.mongodb.design/component/tooltip/example/) instead of Popconfirm? It seems to have a lot of features we would benefit from.


src/components/Popconfirm/index.tsx
```diff
@@ -1,50 +1,78 @@
...
   confirmText = "Yes",
   refEl,
   onCancel = () => {},
   onConfirm = () => {},
   setActive,
+  trigger,
 }) => {
+  const [uncontrolledOpen, uncontrolledSetOpen] = useState<boolean>(false);
+  const setOpen =
+    (typeof active === "boolean" && setActive) || uncontrolledSetOpen;
```
khelif96, line 56: Can you explain this type check? 
couldn't you do this?
```suggestion
    (active === null" && setActive) || uncontrolledSetOpen;
```

src/components/Popconfirm/index.tsx
```diff
@@ -73,17 +102,45 @@ const Popconfirm: React.VFC<PopconfirmProps> = ({
...
+
+    return (
+      <>
+        {renderedTrigger}
+        {popoverContent}
+      </>
+    );
+  }
+
+  return popoverContent;
```
khelif96, line 129:
```suggestion
  const triggerClickHandler = (e: React.MouseEvent) => {
    if (open) {
      onCancel();
    }

    setOpen(!open);
    trigger?.props?.onClick?.(e);
  };

  const renderedTrigger =
    trigger !== undefined &&
    cloneElement(trigger, {
      ref: triggerRef,
      onClick: triggerClickHandler,
    });

  return (
    <>
      {renderedTrigger}
      {popoverContent}
    </>
  );
```

khelif96 (changes requested):
Great job taking this on!

While I do agree the tooltip serves our needs, making it interactable goes against leafy greens design guidelines. 

https://www.mongodb.design/component/tooltip/guidelines/#:~:text=Never%20nest%20call%20to%20actions%20within%20tooltips.
> Never nest call to actions within tooltips.

sophstad:
> While I do agree the tooltip serves our needs, making it interactable goes against leafy greens design guidelines.

@khelif96 I noticed this too, but I feel like what we're doing by using the Popover here is effectively going against the guideline anyway. We're not _technically_ using Tooltip for an interaction, but we're effectively doing the same thing. Those guidelines also refer to mouseover instead of click interactions, so I wonder if they're a little outdated. Interested in any other thoughts on this!

khelif96:
> > While I do agree the tooltip serves our needs, making it interactable goes against leafy greens design guidelines.
> 
> @khelif96 I noticed this too, but I feel like what we're doing by using the Popover here is kinda going against the guideline anyway. We're not _technically_ using Tooltip for an interaction, but we're effectively doing the same thing. I think it's a reasonable UX pattern regardless! Those guidelines also refer to mouseover instead of click interactions, so I wonder if they're a little outdated. Interested in any other thoughts on this!

That is a good point I would be curious to hear from the design folks if they have any opinions on it. But it does look like cloud does use popover to define similar interactions like this. 
https://github.com/search?q=repo:10gen/mms+@leafygreen-ui/popover&type=code
while the few tooltip usages I found did not define any interactivity. (Although I didn't look at all of them)
https://github.com/search?q=repo:10gen/mms+@leafygreen-ui/tooltip&type=code&p=2

sophstad:
> That is a good point I would be curious to hear from the design folks if they have any opinions on it. But it does look like cloud does use popover to define similar interactions like this. https://github.com/search?q=repo:10gen/mms+@leafygreen-ui/popover&type=code while the few tooltip usages I found did not define any interactivity. (Although I didn't look at all of them) https://github.com/search?q=repo:10gen/mms+@leafygreen-ui/tooltip&type=code&p=2

Sounds good, I added to design sync for this week since this would make sense as a followup ticket anyway. Thanks!

src/components/Popconfirm/index.tsx
```diff
@@ -1,50 +1,78 @@
...
   children: React.ReactNode;
+  confirmDisabled?: boolean;
   confirmText?: string;
-  refEl?: React.RefObject<HTMLElement>;
+  "data-cy"?: string;
+  justify?: Justify;
   onCancel?: () => void;
-  onConfirm?: () => void;
-  setActive: (open: boolean) => void;
+  onConfirm?: (e?: React.MouseEvent) => void;
```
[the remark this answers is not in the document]
sophstad: `onConfirm` is called when the "ok" button is clicked, what kind of handler were you thinking of?

khelif96 (approved):
SGTM

sophstad:
evergreen retry

PR #1726, opened 2023-03-10
EVG-18314, EVG-18528: Replace antd Radio and Divider

minnakt:
EVG-18314, EVG-18528

### Description
Got inspired by Sophie to replace some antd components! 😙

This PR removes usage of `antd`'s Radio and Divider components. Also bumps LeafyGreen's Radio component. 
### Screenshots
Screenshot 2023-03-10 at 11 36 09 AM

### Testing
- Existing tests should pass


src/components/styles/Divider.tsx
```diff
@@ -1,9 +1,14 @@
...
 import { size } from "constants/tokens";
 
-const StyledDivider = styled(Divider)`
+const { gray } = palette;
+
+const Divider = styled("hr")`
+  background-color: ${gray.light2};
+  border: 0;
+  height: 1px;
   margin: ${size.xxs} 0 ${size.xs} 0;
```
sophstad, line 11: My preference for this component would probably be to have the same margin on top and bottom

src/pages/preferences/preferencesTabs/NotificationsTab.tsx
```diff
@@ -135,20 +139,24 @@ const handleFieldUpdate = (stateUpdate) => (e) => {
...
-  row-gap: 0;
-  width: 50%;
+  margin-bottom: ${size.s};
+  width: 350px;
 `;
 
-const GridField = styled.div`
-  grid-area: ${(props: { gridArea: string }): string => props.gridArea};
-`;
+const NotificationMethod = styled.div`
```
sophstad, line 155: [nit] What do you think of including this in NotificationField, maybe by adding a `header` or `columns` prop to pass in the column labels? Colocating it with the content component feels like it may be better.

minnakt: Yes, that's a good point that it's confusing to maintain a grid across two files. I put all of the grid-related stuff into the same file!

sophstad (approved):
beautiful

PR #1730, opened 2023-03-14
EVG-18529: Remove Antd collapsible

sophstad:
EVG-18529

### Description

- Replace Collapsible with our own Accordion component

### Screenshots

image

### Testing

- [Here](https://spruce.mongodb.com/host/i-0428c42429169abda) is a log that contains accordions; view it with changes using `yarn prod`


minnakt (changes requested):

src/components/Accordion/index.tsx
```diff
@@ -38,42 +40,37 @@ export const Accordion: React.VFC<AccordionProps> = ({
     <TitleTag>{toggledTitle ? showToggledTitle : title}</TitleTag>
   );
   return (
-    <>
+    <div data-cy={dataCy}>
```
minnakt, line 43: im so sory one more thing 😔 I think the accordion for folded commits is experiencing some weird visual lag effects:

https://user-images.githubusercontent.com/47064971/225719283-4317bbca-909e-4dcf-8940-ab46ceab232f.mov

sophstad: That's interesting, I could not reproduce (but also can't find 278 folded commits). Do you think this is maybe an existing bug having to do with the quantity of folded commits? Happy to dig further if you lmk where you saw this 8)

minnakt: Sorry I should have given more steps to reproduce, the 278 folded commits comes from the [`test-spruce`](http://localhost:3000/task-history/test-spruce/first_task_in_task_group) project on staging! I think this is a new bug introduced on this branch, because when I remove the parent `div` wrapper on the `Accordion` and change it back to a fragment, the issue goes away for me.

The fragment may somehow be important in how we process the folded commit in [FoldedCommit.tsx](https://github.com/evergreen-ci/spruce/blob/main/src/components/HistoryTable/HistoryTableRow/FoldedCommit.tsx#L80-L94). One idea might be to add a prop to the `Accordion` to disable the animation (`disableAnimation`) and use it here? 🤔

minnakt (approved):
looks great! 🌺

PR #1733, opened 2023-03-17
EVG-19151: Generate unique webhook secrets

minnakt:
EVG-19151

### Description
This PR addresses an issue where secrets were not being regenerated in a single session (note that  across multiple sessions, they would be different). This is because the form schema would just evaluate `generateWebhookSecret` once and stick with that value.

I changed the UI slightly such that the secret is only shown after saving the subscription. The only reason I did this is because it seemed like a more straightforward solution than messing with the internal form widgets (note that it's possible, but the solution is kind of ugly). Feel free to let me know if you'd like me to try something else!

I also wasn't sure how much we wanted to change the `generateWebhookSecret` function. For the time being I created a replacement function for `Math.random()` using `crypto`.

### Screenshots
Screenshot 2023-03-17 at 11 51 43 AM

### Testing
- Moved some tests around in order to be able to mock `generateWebhookSecret`


src/pages/projectSettings/tabs/NotificationsTab/getGqlPayload.test.ts
```diff
@@ -68,6 +70,38 @@ describe("getGqlPayload", () => {
     });
   });
 
+  it("should correctly format and generate a secret for webhook subscription", () => {
+    jest
```
khelif96, line 74: nice

src/pages/projectSettings/tabs/NotificationsTab/utils.ts
```diff
@@ -153,7 +41,7 @@ export const generateWebhookSecret = () => {
     "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
 
   for (let i = 0; i < 64; i++) {
-    text += possible.charAt(Math.floor(Math.random() * possible.length));
+    text += possible.charAt(Math.floor(cryptoRandom() * possible.length));
   }
```
khelif96, line 45: Have you run into [crypto.getRandomValues()](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues)?  We can do the same thing without maintaining a list of possible characters. 
```ts
  const buffer = new Uint8Array(32);
  crypto.getRandomValues(buffer);
  return Array.from(buffer, byte => byte.toString(16).padStart(2, '0')).join('');
```

minnakt: I used `crypto.getRandomValues` in the `cryptoRandom` function, but this is cool! 😲

 My one reservation is that this will generate a token containing only `0-9` and `a-f`. So we'll be missing lowercase `g-z` and uppercase `A-Z`, and I'm not sure if we want to deviate from the current format so much. What do you think?

khelif96: I don't think the characters matter so much. This is more so about ensuring there is enough entropy used to generate the key. You can also increase the radix value used in `toString` to allow for more characters. 
image

khelif96 (approved):

PR #1738, opened 2023-03-21
EVG-19219 Fix broken distro link

khelif96:
EVG-19219

### Description
When we upgraded `@leafygreen-ui/menu` in https://github.com/evergreen-ci/spruce/commit/31fa07eabf2df449de1ee952efa55b76cfc5b294 the distros link in the menu stopped working. 
This PR fixes the menu item component so it can accept both router and non-router links. 

### Testing
Added tests to ensure the links don't break again.


src/components/Header/NavDropdown/index.tsx
```diff
@@ -30,17 +30,26 @@ const NavDropdownItem: React.VFC<NavDropdownItemType> = ({
...
-);
+}) => {
+  const isLink = to !== undefined;
+  const linkProps = isLink
+    ? {
+        to,
+      }
+    : {
+        href,
+      };
```
khelif96, line 41: This was necessary so that the fields had exclusive types. If we tried to inline this typescript would complain because it wouldn't have a guarantee that to and href aren't used together.

src/components/Header/NavDropdown/index.tsx
```diff
@@ -30,17 +30,26 @@ const NavDropdownItem: React.VFC<NavDropdownItemType> = ({
...
-    to={to}
-    href={href}
-    data-cy={itemDataCy}
-    onClick={closeMenu}
-  >
-    {text}
-  </MenuItem>
-);
+}) => {
+  const isLink = to !== undefined;
```
sophstad, line 34: [nit] This component functions as a link in both cases, so maybe naming this `isReactRouterLink` or `isInternalLink` would be more helpful.

sophstad (approved):

PR #1739, opened 2023-03-21
EVG-18851: Upgrade Storybook

sophstad:
EVG-18851

### Description

- Upgrade Storybook to latest 7.0 release candidates
- Update stories to use new component story format
- Remove dependency on `storybook-addon-apollo-client`, which appears to be [unmaintained](https://github.com/lifeiscontent/storybook-addon-apollo-client/issues/89). Instead, I've moved the Apollo `MockedProvider` out of the shared Storybook setup and into each component that requires it. This allows any relevant mocks to be passed directly to MockedProvider instead of via the plugin.
- Add resolution to prevent `has-symbols@1.0.1` from being installed by a nested dependency of `@storybook/builder-manager`. This version of `has-symbols` lacks a polyfill for `global` which causes `yarn dev` to fail.


.evergreen.yml
```diff
@@ -316,6 +316,18 @@ functions:
...
+      params:
+        aws_key: ${aws_key}
+        aws_secret: ${aws_secret}
+        local_files_include_filter:
+          ["spruce/storybook-static/**/*.json"]
+        remote_file: spruce/${task_id}/storybook/
+        bucket: mciuploads
+        content_type: application/json
+        permissions: public-read
+        preserve_path: true
```
khelif96, line 331: It looks like artifacts aren't being uploaded to the right place. We are missing images in the uploaded stories and the fonts don't seem to be picked up
image

.storybook/preview.tsx
```diff
@@ -17,5 +16,10 @@ export const parameters = {
   },
 };
 
-// @storybook/addon-storyshots isn't successfully injecting the decorators from storybook-addon-react-router-v6 and storybook-addon-apollo-client so we need to manually wrap the stories.
-export const decorators = !isTest ?  [withRouter] : [(Story) => <MockedProvider><MemoryRouter>{<Story />}</MemoryRouter></MockedProvider>];
+export const decorators = [
```
khelif96, line 19: Adding the GlobalStyles decorator should fix the fonts not rendering. We do this in parsley.
https://github.com/evergreen-ci/parsley/blob/5d2269e8245499a286c84edc041a6383eaf94ec3/src/components/styles/GlobalStyles.tsx#L1

src/components/HistoryTable/HistoryTable.stories.tsx
```diff
@@ -7,21 +8,24 @@ import { mainlineCommitData } from "./testData";
...
-  title: "Components/History Table",
   component: HistoryTable,
 };
 
-export const TaskHistoryTable = () => (
-  <HistoryTableProvider>
-    <HistoryTableWrapper type="task" />
-  </HistoryTableProvider>
-);
+export const TaskHistoryTable: StoryObj<typeof HistoryTable> = {
```
khelif96, line 14: This needs an apollo provider

src/components/SpruceForm/SpruceForm.stories.tsx
```diff
@@ -1,66 +1,63 @@
...
-    <SpruceFormContainer title="Distro Projects">
-      <SpruceForm
-        schema={example1Def.schema}
-        uiSchema={example1Def.uiSchema}
-        onChange={onChange}
-        formData={formState}
-      />
-    </SpruceFormContainer>
-  );
+export default {
```
khelif96, line 6: The stories here look pretty weird. Maybe it needs the global reset styles?
image

src/pages/task/metadata/Metadata.stories.tsx
```diff
@@ -1,98 +1,111 @@
...
-            taskId: "some_task_id_4",
-          },
-        ],
-      }}
-      taskId={taskQuery.task.id}
-      error={null}
-    />
-  </Container>
-);
+export const Default: StoryObj<MetadataProps> = {
```
khelif96, line 20: Can we use 
```suggestion
export const Default: StoryObj<typeof Metadata> = {
```
instead of exporting metadata props

src/pages/task/metadata/index.tsx
```diff
@@ -30,14 +30,14 @@ const { applyStrictRegex, msToDuration, shortenGithash } = string;
 const { getUiUrl } = environmentalVariables;
 const { red } = palette;
 
-interface Props {
+export interface MetadataProps {
```
khelif96, line 33: Comment from above

khelif96 (changes requested):

package.json
```diff
@@ -217,14 +216,9 @@
...
   },
-  "overrides": {
-    "storybook-addon-react-router-v6": {
-      "react": "17.0.1",
-      "react-dom": "17.0.1"
-    }
-  },
   "resolutions": {
-    "@graphql-codegen/cli/**/ts-node": "10.8.2"
+    "@graphql-codegen/cli/**/ts-node": "10.8.2",
```
minnakt, line 220: Might be worth checking if this resolution could be removed, as the original reason it was installed was due to Storybook! ([comment](https://github.com/evergreen-ci/spruce/pull/1368#issuecomment-1182025643))

src/components/ProjectSelect/index.tsx
```diff
@@ -168,6 +168,7 @@ const getProjects: GetProjectsType = (
 
   // For Project Settings pages, move disabled projects to the bottom of the dropdown
   const projectGroups = viewableProjectsData?.viewableProjectRefs ?? [];
+  console.log(projectGroups);
```
minnakt, line 171:
```suggestion

```

src/components/Breadcrumbs/BreadCrumbs.stories.tsx
```diff
@@ -1,20 +1,22 @@
 import { actions } from "@storybook/addon-actions";
+import { StoryObj } from "@storybook/react";
 import Breadcrumbs from ".";
 
 export default {
-  title: "Components/Breadcrumbs",
   component: Breadcrumbs,
```
minnakt, line 6: Might want to give this story a `title`, since the nesting seems unnecessary:

Screenshot 2023-03-22 at 7 33 58 PM

Also, could we rename this file as `Breadcrumbs.stories.tsx`? (no capital C) :0

src/components/WelcomeModal/WelcomeModal.stories.tsx
```diff
@@ -15,42 +15,50 @@ export default {
...
 };
 
-export const NewSpruceUser = () => (
-  <WelcomeModal
-    title="Welcome to the New Evergreen UI!"
-    param="hasUsedSpruceBefore"
-    carouselCards={newSpruceUser}
-  />
-);
+export const NewSpruceUser: StoryObj<typeof WelcomeModal> = {
```
minnakt, line 18: Think you need an `MockedProvider` in this file as well!

Screenshot 2023-03-22 at 7 41 25 PM

src/components/Table/TasksTable.stories.tsx
```diff
@@ -1,5 +1,18 @@
+import { StoryObj } from "@storybook/react";
 import { TasksTable } from "./TasksTable";
 
+export default {
+  component: TasksTable,
```
minnakt, line 5: wondering somewhat if the `TasksTable` belongs under the `pages/Task` folder with the other tables!

src/components/SpruceForm/SpruceForm.stories.tsx
sophstad: This is actually an [existing issue](https://mciuploads.s3.amazonaws.com/spruce/spruce_ubuntu2004_storybook_5a5205cf6f3129423c382861d59c91435e6d7b88_23_03_22_22_03_15/storybook/spruce/storybook-static/index.html?path=/story/components-spruce-form--example-1) (from [this build](https://spruce.mongodb.com/version/spruce_5a5205cf6f3129423c382861d59c91435e6d7b88/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)). I added it as part of EVG-19241.

.storybook/preview.tsx
sophstad: See above, this fixes the issue locally but not remotely.

src/components/ProjectSelect/ProjectSelect.stories.tsx
```diff
@@ -1,29 +1,49 @@
...
+  ),
+};
 
-export const Default = () => (
-  <ProjectSelect
-    selectedProjectIdentifier="evergreen"
-    getRoute={getCommitsRoute}
-  />
-);
 const addFavoriteMock = {
```
minnakt, line 47: Should we type these mocks with `ApolloMock` like we do in the Jest tests?

src/components/PatchActionButtons/ScheduleTasks.stories.tsx
```diff
@@ -1,153 +1,157 @@
...
-                    buildVariant: "windows",
-                    buildVariantDisplayName: "Windows",
-                    displayName: "type_check",
-                    execution: 1,
-                  },
-                ],
+const mocks = [
+  {
+    request: {
+      query: GET_UNSCHEDULED_TASKS,
```
minnakt, line 30: same question as above!

src/components/SearchableDropdown/SearchableDropdown.stories.tsx
```diff
@@ -1,34 +1,28 @@
 import { useState } from "react";
-import SearchableDropdown from ".";
+import { StoryObj } from "@storybook/react";
+import SearchableDropdown, { SearchableDropdownProps } from ".";
 
 export default {
-  title: "Components/Searchable Dropdown",
   component: SearchableDropdown,
 };
```
minnakt, line 8: I think the `CustomOption` component might have been missed in this file (to turn into a `StoryObj` thing)

src/pages/preferences/preferencesTabs/cliTab/NodeList.stories.tsx
```diff
@@ -6,13 +7,15 @@ export default {
...
-);
+export const NodeElement: StoryObj<typeof Node> = {
+  render: () => (
+    <Node
+      title="Download the Command-Line Client."
+      child={<Child />}
+      stepNumber={1}
+    />
+  ),
+};
```
minnakt, line 19: `NodeListFull` might have been missed in this file

src/components/WelcomeModal/WelcomeModal.stories.tsx
```diff
@@ -1,3 +1,5 @@
+import { MockedProvider } from "@apollo/client/testing";
+import { StoryObj } from "@storybook/react";
 import {
```
minnakt, line 3: I'm seeing that the modals persist on the docs page. I googled it and it's a problem that others have run into, but I couldn't get their solutions to work 🤔 May be worth opening another ticket if it's not easy to fix!

Screenshot 2023-03-23 at 7 56 24 PM

minnakt (approved):
great job Sophie 🥳

khelif96 (approved):

PR #1743, opened 2023-03-22
EVG-19018 Create pod page

khelif96:
EVG-19018

### Description
Creates some types used by pods for statuses
Create Pod page, fetch data and render metadata
Render boilerplate placeholder for pod events (Eternal loading skeleton)
### Screenshots
image

### Testing
localhost: 
localhost:3000/container/localhost

staging:
<stagingURL>/container/641ae3bd0f40259343a6ffec

### DIFF
--- a/src/types/pod.ts
+++ b/src/types/pod.ts
@@ -1,0 +1,14 @@
+export enum PodStatus {
+  // yellow: pod-starting
+  Initializing = "initializing",
+  Starting = "starting",
+
+  // green: pod-running
+  Running = "running",
+
+  // grey: pod-unreachable
+  Decommissioned = "decommissioned",
+
+  // red: pod-terminated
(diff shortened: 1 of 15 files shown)


khelif96:
Tagging @Kimchelly  just to double check that the page data makes sense

src/types/pod.ts
```diff
@@ -0,0 +1,14 @@
+export enum PodStatus {
+  // yellow: pod-starting
+  Initializing = "initializing",
```
Kimchelly, line 3: `Initializing` is approximately equivalent to the host's `Uninitialized` state, so I would color it light grey for consistency.

Kimchelly (commented):
Didn't look that deeply into the change, but the picture looks good! 👍

PR #1744, opened 2023-03-23
EVG-19237 Avoid unnecessary project health network requests

khelif96:
EVG-19237

### Description
The project health page would perform a query to fetch test results from failed tasks every time a user hovers over an icon. If there are no failed tests this would result in a loading animation that would briefly flash and a network call. We can avoid this flash and network request by first checking if the task has failed test results in the first place.


sophstad (approved):
neat!

PR #1750, opened 2023-03-30
EVG-18402 Refactor Spruce deploy script

khelif96:
EVG-18402

### Description
Refactors spruce deploys script to be more like Parsleys. Making it easier to understand and eliminating some external dependencies.
Moves bucket names into cmdrc file so they are no longer hard coded in package.json

I will upload this to 1 password once this is merged.


package.json
```diff
@@ -15,11 +15,10 @@
     "cy:open": "cypress open",
     "cy:run": "cypress run",
     "cy:test": "cypress run --spec",
-    "deploy-prod:do-not-use-directly": "yarn build:prod && BUCKET=evergreen.spruce yarn deploy; yarn upload-source-maps;",
-    "deploy:beta": ". ./scripts/app-version.sh && yarn build:beta && BUCKET=evergreen-beta.spruce yarn deploy",
-    "deploy:prod": "node scripts/deploy-prod.js",
-    "deploy:staging": ". ./scripts/app-version.sh && yarn build:staging && BUCKET=evergreen-staging.spruce yarn deploy",
-    "deploy": "aws s3 sync build/ s3://${BUCKET}/ --acl public-read --follow-symlinks --delete --exclude env/.cmdrc.json",
+    "deploy:beta": "yarn build:beta && env-cmd -e beta  env/.cmdrc.json yarn deploy:do-not-use",
+    "deploy:staging": "yarn build:staging && env-cmd -e staging -r env/.cmdrc.json yarn deploy:do-not-use",
```
minnakt, line 19: Potentially silly question but why does one have `-r` and the other doesn't?
```suggestion
    "deploy:beta": "yarn build:beta && env-cmd -e beta -r env/.cmdrc.json yarn deploy:do-not-use",
    "deploy:staging": "yarn build:staging && env-cmd -e staging -r env/.cmdrc.json yarn deploy:do-not-use",
```

scripts/deploy-production.js
```diff
@@ -0,0 +1,126 @@
+const prompts = require("prompts");
+const { exec } = require("child_process");
+
+const localDeployScript =
+  "yarn build:prod && env-cmd -e production yarn deploy:do-not-use && env-cmd -e production ./scripts/email.sh";
```
minnakt, line 5: Not sure, but do you need to specify `-r` here?

scripts/deploy.sh
```diff
@@ -0,0 +1,12 @@
+# This script runs the aws cli command to deploy the app to s3
+# It also uploads source maps to bugsnag
+
+# Try this step and throw an error if it fails
+echo "Deploying to S3"
+aws s3 sync build/ s3://${BUCKET}/ --acl public-read --follow-symlinks --delete --exclude .cmdrc.json 
+echo "Deployed to S3"
```
minnakt, line 7: Oops sorry, I think this actually might be
```suggestion
aws s3 sync build/ s3://${BUCKET}/ --acl public-read --follow-symlinks --delete --exclude env/.cmdrc.json 
echo "Deployed to S3"
```

minnakt (changes requested):
Not saying to do this in this PR, but have we considered switching Spruce to use the default `.env-cmdrc` file instead, like Parsley? It might help eliminate some confusion!

package.json
khelif96: Great catch!

scripts/deploy.sh
khelif96: Don't need it actually the file doesn't exist in the build directory anymore.

khelif96:
> Not saying to do this in this PR, but have we considered switching Spruce to use the default `.env-cmdrc` file instead, like Parsley? It might help eliminate some confusion!

Done! I will communicate it once this is merged.

package.json
```diff
@@ -5,37 +5,36 @@
...
+    "deploy:beta": "yarn build:beta && env-cmd -e beta yarn deploy:do-not-use",
+    "deploy:staging": "yarn build:staging && env-cmd -e staging yarn deploy:do-not-use",
+    "deploy:prod": "node scripts/deploy-production.js",
+    "deploy:do-not-use": ". ./scripts/deploy.sh",
+    "dev": "env-cmd -e devLocal -r .env-cmdrc.local.json yarn start",
     "eslint:fix": "yarn eslint:strict --fix",
     "eslint:staged": "STRICT=1 eslint",
     "eslint:strict": "STRICT=1 eslint '*.{js,ts,tsx}' 'src/**/*.ts?(x)' 'cypress/**/*.ts' 'src/gql/**/*.graphql'",
-    "notify-email": "env-cmd -e prod -r env/.cmdrc.json bash email.sh",
+    "notify-email": "env-cmd -e prod bash email.sh",
```
minnakt, line 26: Might be able to delete this command, I can't find any references to it and Parsley doesn't have one
```suggestion
```

scripts/deploy-production.js
minnakt, line 5: Given that `prod` is referenced in `package.json`, I'm guessing this is supposed to be:
```suggestion
  "yarn build:prod && env-cmd -e prod yarn deploy:do-not-use && env-cmd -e prod ./scripts/email.sh";
```

Parsley uses `production`/`devProduction` rather than `prod`/`devProd`, could be worth standardizing them

scripts/deploy.sh
```diff
@@ -0,0 +1,12 @@
+# This script runs the aws cli command to deploy the app to s3
+# It also uploads source maps to bugsnag
+
+# Try this step and throw an error if it fails
+echo "Deploying to S3"
+aws s3 sync build/ s3://${BUCKET}/ --acl public-read --follow-symlinks --delete
```
minnakt, line 6: I think you may have already addressed this in your response to me but I wasn't sure HAHA just wanted to flag that Parsley has this extra thing! It might not be necessary at all but just wanted to point it out in case it is.
```suggestion
aws s3 sync build/ s3://${BUCKET}/ --acl public-read --follow-symlinks --delete --exclude .env-cmdrc.json 

```

minnakt (approved):
thanks for making our configurations between Spruce and Parsley more consistent 🥰 Let me know if you want me to give this another look, but other than these comments I think it looks good!

scripts/deploy-production.js
khelif96: Done will update the 1password file soon

PR #1752, opened 2023-03-30
EVG-19222 Abort if it takes over 10 minutes to start evergreen

khelif96:
EVG-19222


ybrill (commented):
Could it make sense to use `exec_timeout_secs` instead? It seems like even while we're waiting there's still stuff on stdout from the db logs.

khelif96:
> Could it make sense to use `exec_timeout_secs` instead? It seems like even while we're waiting there's still stuff on stdout from the db logs.

Done!

PR #1754, opened 2023-03-31
EVG-19295: Enable ESLint and Prettier on GraphQL files

minnakt:
EVG-19295

### Description
While working on linting GraphQL files on Parsley, I realized that linting for GraphQL files doesn't work on Spruce. This PR updates our configuration so that we can leverage these features.

I consolidated the three ESLint files into one because it was honestly so difficult to see what was going on and this just made it easier. 😭 

### Screenshots
(From Parsley)

https://user-images.githubusercontent.com/47064971/229238665-ab741591-c15b-45f5-88e9-444132fab0c8.mov

### Testing
- Manually fiddling with the query files and seeing the linting in action.  Note that you need to have `"graphql"` included in `eslint.validate` in your VSCode settings for this to work.


.eslintrc.js
```diff
@@ -39,85 +19,223 @@ module.exports = {
...
+    // For GraphQL files.
     {
-      files: ["*.graphql"],
-      parser: "@graphql-eslint/eslint-plugin",
-      plugins: ["@graphql-eslint"],
+      files: ["src/**/*.graphql"],
+      extends: "plugin:@graphql-eslint/operations-recommended",
       rules: {
-        "prettier/prettier": "error",
+        "@graphql-eslint/naming-convention": OFF, // This should be turned on eventually.
```
minnakt, line 229: This should actually be turned on, but it will cause like ~200 file changes so I left it out.

sophstad: Should we make a ticket for this and reference it here?

minnakt: Created [EVG-19316](https://jira.mongodb.org/browse/EVG-19316)!

.eslintrc.js
```diff
@@ -39,85 +19,223 @@ module.exports = {
...
-    ],
+    // Rules for ESLint.
     "arrow-body-style": [
       errorIfStrict,
       "as-needed",
       {
         requireReturnForObjectLiteral: false,
       },
     ],
-    "consistent-return": "off",
```
khelif96, line 68: :)

.eslintrc.js
```diff
@@ -1,10 +1,3 @@
-const fs = require("fs");
```
khelif96, line 1: For context the way eslint worked in the project was....
 
This file was the base config and contained all of the common configs and rules for each directory. 

The one in the `/src` directory extended this config. And applied its directory-specific rules on top of it.

The one in the `/cypress` directory also extended this config and applied its directory-specific rules on top of it.

.prettierrc
```diff
@@ -1,19 +0,0 @@
-{
```
khelif96, line 1: Why are these prettier rules being removed? Are they superseded by something else?

.eslintrc.js
```diff
@@ -39,85 +19,223 @@ module.exports = {
...
+      parserOptions: {
+        project: "cypress/tsconfig.json",
+      },
+    },
+    // For GraphQL files.
     {
-      files: ["*.graphql"],
-      parser: "@graphql-eslint/eslint-plugin",
-      plugins: ["@graphql-eslint"],
+      files: ["src/**/*.graphql"],
```
khelif96, line 226: It looks like the typescript rules are being applied to the graphql files.
image

khelif96 (changes requested):
Just a few comments and clarifications and one bug I noticed.

.prettierrc
minnakt: the prettier rules were set to the default (check each of the defaults [here](https://prettier.io/docs/en/options.html)) so I don't think there's a point in keeping the file. Also it was kind of wrong because we were explicitly setting the parser to `"typescript"` when Prettier should just infer the parser based on the file type. Parsley doesn't have a `.prettierrc` file!

.eslintrc.js
minnakt: I couldn't quite replicate this but I added the sdlschema folder to the `.eslintignore` file, let me know if that worked!

khelif96 (approved):
SGTM Thank you so much for doing this! 🚀

PR #1778, opened 2023-04-10
EVG-18860: replace "word-break: break-word;" with "word-break: normal;"

SupaJoon:
[EVG-18860](jira.mongodb.com/browse/EVG-18860)

### Description
deprecation docs: https://developer.mozilla.org/en-US/docs/Web/CSS/word-break.#values


khelif96 (changes requested):
> While word-break: break-word is deprecated, it has the same effect, when specified, as word-break: normal and overflow-wrap: anywhere

From the docs it only exhibits the same behavior if used with `overflow-wrap:anywhere`

khelif96 (approved):

PR #1782, opened 2023-04-11
EVG-19272: Commit Queue settings page is error-free for new projects

sophstad:
EVG-19272

### Description

- Update page schema to use a dependency so that Commit Queue fields that are not currently active/visible (e.g. when CQ is disabled) are not validated
- Improve error message for when commit queue is enabled

### Screenshots

image

### DIFF
--- a/src/components/SpruceForm/errors.ts
+++ b/src/components/SpruceForm/errors.ts
@@ -34,6 +34,11 @@ export const transformErrors = (errors: AjvError[]) =>
             ...error,
             message: Errors.Invisible,
           };
+        case "oneOf":
+          return {
+            ...error,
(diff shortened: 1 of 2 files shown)


src/components/SpruceForm/errors.ts
```diff
@@ -34,6 +34,11 @@ export const transformErrors = (errors: AjvError[]) =>
             ...error,
             message: Errors.Invisible,
           };
+        case "oneOf":
+          return {
+            ...error,
+            message: "Please select one of the available options.",
```
minnakt, line 40: :D

minnakt (approved):

PR #1785, opened 2023-04-13
EVG-19252: Skip fetching commits while window is resizing

minnakt:
EVG-19252

### Description
Currently, when the Project Health page is being resized, the `limit` is updated simultaneously, kicking off new `MainlineCommits` queries. Some of these queries are immediately cancelled because the limit updates so quickly. This PR makes it so that we only fetch the `MainlineCommits` query when the window is not resizing.

### Testing
- Manually inspecting the network requests while resizing the Project Health page
- Added `useResize.test.ts`

### Note
In the console, you may see this warning as you resize the page. I checked on the `main` branch and saw the same error originating from the same components, so I chose to ignore it. This particular error has been flagged for having a lot of false positives and was tweaked in React 18, so it might go away when we upgrade, though I can't say with 100% certainty. Read more about it in this [StackOverflow answer](https://stackoverflow.com/questions/58038008/how-to-stop-memory-leak-in-useeffect-hook-react).

Screenshot 2023-04-13 at 10 59 17 AM


src/hooks/useResize/index.ts
```diff
@@ -0,0 +1,33 @@
+import { useEffect, useState } from "react";
+
+/**
+ * `useResize` determines if the window is currently being resized or not.
+ * @returns a boolean indicating whether the window is being resized or not
+ */
+export const useResize = () => {
```
sophstad, line 7: To me this hook has an obvious use case of calling a function on resize—in this case we just only happen to care about the `isResizing` status. If you wanted I think it'd be handy to flesh this out with an `onResize` prop that gets called in the hook's `onResize` if it's provided. But that's definitely overkill so you could totally just leave a comment saying something to that effect if you like!

src/hooks/useResize/useResize.test.ts
```diff
@@ -0,0 +1,43 @@
+import { renderHook, act } from "@testing-library/react-hooks";
+import { useResize } from ".";
+
+describe("useResize", () => {
```
sophstad, line 4: very elegant tests

sophstad (approved):

PR #1787, opened 2023-04-18
EVG-17262: Show timezone

sophstad:
EVG-17262

### Description

- Show timezone when reporting timestamps. I opted to report it in `GMT+N` format, which I feel is understandable but not too verbose. All the options are enumerated [here](https://date-fns.org/v2.29.3/docs/format).
- Omit timezone in a few particularly space-constrained places

### Screenshots

Version page:
image

Hosts table:
image

### Testing

- Updated unit tests

### DIFF
--- a/src/utils/string/index.ts
+++ b/src/utils/string/index.ts
@@ -122,7 +122,7 @@ export const getDateCopy = (
   const finalDateFormat = dateOnly
     ? dateFormat
     : `${dateFormat}, h:mm${omitSeconds ? "" : ":ss"} aa${
-        omitTimezone ? "" : " O"
+        omitTimezone ? "" : " z"
       }`;
   if (tz) {
     return format(utcToZonedTime(time, tz), finalDateFormat, {
--- a/src/utils/string/index.ts
+++ b/src/utils/string/index.ts
@@ -102,6 +102,7 @@ export type DateCopyOptions = {
   tz?: string;
   dateOnly?: boolean;
   omitSeconds?: boolean;
+  omitTimezone?: boolean;
   dateFormat?: string;
 };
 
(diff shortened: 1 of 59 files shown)


khelif96 (changes requested):
I don't think GMT + some offset is the best way to surface the timezone. It doesn't help me understand what time zone I'm looking at because  I don't have the timezone to offset mappings memorized in the same way seeing the 3-letter code for `est`. Users also set their timezone in the readable format in the settings so showing the GMT + Offset everywhere else is confusing.

src/utils/string/index.ts
```diff
@@ -113,16 +114,20 @@ export const getDateCopy = (
...
+  const { tz, dateOnly, omitSeconds, omitTimezone } = options || {};
   let { dateFormat } = options || {};
   if (!dateFormat) {
     dateFormat = "MMM d, yyyy";
   }
   const finalDateFormat = dateOnly
     ? dateFormat
-    : `${dateFormat}, h:mm${omitSeconds ? "" : ":ss"} aa`;
+    : `${dateFormat}, h:mm${omitSeconds ? "" : ":ss"} aa${
+        omitTimezone ? "" : " O"
```
khelif96, line 125:
```suggestion
        omitTimezone ? "" : " z"
```

This formats the timezones in the readable `EST`/ `UST` format.

khelif96 (changes requested):
> > I don't think GMT + some offset is the best way to surface the timezone. It doesn't help me understand what time zone I'm looking at because I don't have the timezone to offset mappings memorized in the same way seeing the 3-letter code for `est`. Users also set their timezone in the readable format in the settings so showing the GMT + Offset everywhere else is confusing.
> 
> @khelif96 Agree but the timezone name is not an option offered by `date-fns` 😢 you can see all options at the link in the description. We could create some kind of mapping but even that is inconsistent because not all time zones have alphabetical representations (some are only numbered according to GMT). So I suppose we could special-case a few common time zones to represent with abbreviations, but I tend to prefer standardization. Let me know what you think!

It actually is! https://date-fns.org/v2.0.0-alpha.27/docs/Time-Zones#synopsis

From the above example, you can see that using the character `z` in the formatting string will show a timezone in a readable format.
Passing tests after using `z` instead of `O`
image

src/utils/string/index.ts
sophstad: Wow seems like they need to update their docs! That's not what I expected from this...
image

khelif96: This function is using the format method from `date-fns-tz` not `date-fns` which includes support for readable timezones. https://github.com/marnusw/date-fns-tz#format

khelif96 (approved):

PR #1806, opened 2023-05-01
EVG-19746 Hide Host in side panel for container tasks

khelif96:
EVG-19746

### Description
Sometimes we might show the `Host` label in a task metadata for container tasks that have not been assigned a podId yet. This adds an additional guard to ensure that a hostId also exists.


minnakt (approved):
nice!

PR #1807, opened 2023-05-01
EVG-8291 Add pagination to bottom of tasks tables

khelif96:
EVG-8291

### Description
Adds pagination and table controls to the bottom of the tasks and task duration tables.
Only show the options if the table has > 10 entries to avoid the "small sandwich" look.
I also noticed that the floating buttons on the bottom of the page tend to overlap with the pagination buttons at the bottom so I added transparency to them so they are less obtrusive. 
### Screenshots
image

### DIFF
--- a/src/components/Content/index.tsx
+++ b/src/components/Content/index.tsx
@@ -154,4 +154,10 @@ const FloatingContent = styled.div`
   background-color: white;
   padding: ${size.xs};
   border-radius: ${size.s};
+  transition: opacity 0.2s ease-in-out;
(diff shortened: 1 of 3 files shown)


src/components/Content/index.tsx
sophstad, line 1: Adding more margins to the bottom of the page may also get around the issue Arjun pointed out

khelif96: Great call!
image

sophstad (approved):
A four-digit ticket — how vintage!

PR #1818, opened 2023-05-03
EVG-18941: Upgrade react-router-dom to v6.11.1

minnakt:
EVG-18941

### Description
This PR upgrades the `react-router-dom` package and includes the following changes:
- In `App.tsx`, use new router `createBrowserRouter` over `<BrowserRouter />`.
- In Jest tests, use new router `createMemoryRouter` over `unstable_HistoryRouter`.
- In Storybook, use new router `createMemoryRouter` over `<MemoryRouter />`.

See the [docs](https://reactrouter.com/en/main/routers/picking-a-router) for more info.

### Testing
- Tests should pass


yarn.lock
```diff
@@ -4649,18 +4627,6 @@
     "@leafygreen-ui/polymorphic" "^1.2.0"
     "@leafygreen-ui/tokens" "^2.0.1"
 
-"@leafygreen-ui/typography@^16.1.0":
```
minnakt, line 4652: not sure why some LeafyGreen dependencies were affected but when I upgraded a completely different library the same thing happened so maybe this is just out of date 🤷

src/components/Content/index.tsx
```diff
@@ -143,6 +78,67 @@ export const Content: React.VFC = () => {
   );
 };
 
+export const Content: React.VFC = () => (
```
sophstad, line 81: Should we separate these components into different files? They seem fairly distinct 😳

minnakt: done and I will be careful about including the changes from your PR in case that goes in first 🫡

sophstad (approved):

PR #1820, opened 2023-05-04
EVG-18727 Make filter icons black when enabled to improve visibility

khelif96:
EVG-18727

### Description
Makes the filter icons black when they are enabled to make them more noticeable.

### Screenshots
image

### DIFF
--- a/src/components/Table/Filters.tsx
+++ b/src/components/Table/Filters.tsx
@@ -13,7 +13,7 @@ import {
 } from "components/TreeSelect";
 import { fontSize } from "constants/tokens";
 
-const { blue, gray } = palette;
+const { black, gray } = palette;
 const defaultColor = gray.light1;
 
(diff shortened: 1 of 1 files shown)


sophstad (commented):
Ehh, not really sure how I feel about these changes. Black barely feels more noticeable, and the LeafyGreen tables use blue to indicate active filters. It may not be ideal to move to black and then back to blue.

minnakt (commented):
Agree with Sophie that the enabled color should be consistent with the LG table. Also, by the ticket description I was under the impression that this change was supposed to be made for **not enabled** filters, but I still think we shouldn't make the icon color black in that case.

Screenshot 2023-05-05 at 2 49 20 PM

khelif96:
I'm going to close this out in favor of migrating to LG tables in the future.

PR #1822, opened 2023-05-04
EVG-19796 Remove unused webpack dependencies

khelif96:
EVG-19796

### Description
Removed the webpack dependency as we no longer use it. I checked each of our tools to make sure I didn't miss a spot. Local and production storybook. The dev server, prod builds, source map generation, codegen, local cypress, and CI headless cypress all still work.


minnakt (approved):
I believe 🙏

PR #1826, opened 2023-05-05
EVG-19807: Fix form validation for container size definitions

minnakt:
EVG-19807

### Description
The validation for container size definitions seems to not be working properly, as I'm able to save the changes when memory and CPU are empty. This PR updates the form schema so that it works correctly.

### Screenshots
Before:
Screenshot 2023-05-05 at 12 29 46 PM

After:
Screenshot 2023-05-05 at 12 30 23 PM

### Testing
- Tried to fix the flaky tests, which I ran 5 times on CI and it passed all times so I am hopeful 🤞

### DIFF
--- a/cypress/integration/projectSettings/project_settings.ts
+++ b/cypress/integration/projectSettings/project_settings.ts
@@ -1103,6 +1103,8 @@ describe("Containers", () => {
   const destination = getContainersRoute("spruce");
   beforeEach(() => {
     cy.visit(destination);
+    // Wait for page content to load.
+    cy.contains("Container Configurations").should("exist");
   });
   it("shouldn't have any container configurations defined", () => {
     cy.dataCy("container-size-row").should("not.exist");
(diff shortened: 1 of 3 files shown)


cypress/integration/projectSettings/project_settings.ts
```diff
@@ -1112,8 +1114,7 @@ describe("Containers", () => {
   });
   it("should be able to add a container configuration and save it", () => {
     cy.dataCy("add-button").should("be.visible");
-    cy.dataCy("add-button").click();
-
+    cy.dataCy("add-button").trigger("mouseover").click();
```
khelif96, line 1117: I wish I knew why just clicking on it is not enough.

khelif96 (approved):
Queen 👑 thanks for fixing this

khelif96 (approved):
Queen 👑 thanks for fixing this

cypress/integration/projectSettings/project_settings.ts
minnakt: honestly it is a little hack... 😭 but I can't replicate the failure on local, only on CI rip

PR #1832, opened 2023-05-16
EVG-18757 Allow canceling configuration for configured patches

khelif96:
EVG-18757

### Description
Adds a cancel button for patches that have already been configured 
### Screenshots
image

### DIFF
--- a/src/pages/configurePatch/ConfigurePatchCore.tsx
+++ b/src/pages/configurePatch/ConfigurePatchCore.tsx
@@ -141,7 +141,14 @@ export const ConfigurePatchCore: React.VFC<Props> = ({ patch }) => {
           onChange={(e) => setDescription(e.target.value)}
         />
         {activated && (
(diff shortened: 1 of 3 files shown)


src/pages/configurePatch/ConfigurePatchCore.tsx
```diff
@@ -118,6 +120,8 @@ export const ConfigurePatchCore: React.VFC<Props> = ({ patch }) => {
     });
   };
 
+  const Link = useLGButtonRouterLink(getVersionRoute(id));
```
sophstad, line 123: Would it make sense for the destination here to be the previous page? We could access it with `useNavigate()`. That would preserve the filters/pagination of the version page.

khelif96: Wow great catch y'all!

sophstad (approved):
💯

PR #1833, opened 2023-05-16
EVG-19903 Bump vite to latest

khelif96:
EVG-19903

### Description
Opted not to make the changes for env variable injection since that requires all env variables to be prefixed with vite_ and will be too risky for a simple dependency update. 

https://vitejs.dev/guide/env-and-mode.html#html-env-replacement

### DIFF
--- a/package.json
+++ b/package.json
@@ -161,7 +161,7 @@
     "@types/react-window-infinite-loader": "^1.0.6",
     "@typescript-eslint/eslint-plugin": "5.57.1",
     "@typescript-eslint/parser": "5.57.1",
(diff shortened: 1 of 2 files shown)


minnakt (approved):
sounds good to me 😎 do we want to file another ticket to change our variables to start with `VITE_`, or is that something we don't want to do?

khelif96:
> sounds good to me 😎 do we want to file another ticket to change our variables to start with `VITE_`, or is that something we don't want to do?

Opened a ticket although it is probably something I would be inclined to backlog since it we have a workaround.
https://jira.mongodb.org/browse/EVG-19908

PR #1834, opened 2023-05-16
EVG-18330: Base user subscriptions table

sophstad:
EVG-18330

### Description

- Install updated LG tables using `@leafygreen-ui/table/new` alias
- Add view-only user subscriptions table. The remaining features will be added in EVG-19904:
  - Filter table
  - Pagination
  - Delete individual subscriptions
  - Update design for delete all button

### Screenshots

image

### Testing

- Add e2e test

### Evergreen PR
https://github.com/evergreen-ci/evergreen/pull/6534


src/gql/queries/user-subscriptions.graphql
```diff
@@ -0,0 +1,54 @@
...
+              key
+              value
+            }
+            minDelayMs
+            retries
+            secret
+            timeoutMs
+            url
+          }
+        }
```
minnakt, line 46: I wonder if we need to request all of these subscribers for user subscriptions; I had an impression that we only offered `emailSubscriber`, `jiraCommentSubscriber`, and `slackSubscriber` at a user level

src/pages/preferences/preferencesTabs/notificationTab/UserSubscriptions.tsx
```diff
@@ -0,0 +1,151 @@
...
+      return resourceTypeToCopy?.[resourceType] ?? resourceType;
+    },
+  },
+  {
+    header: "ID",
+    accessorKey: "selectors",
+    cell: ({
+      getValue,
+      row: {
+        original: { resourceType },
```
minnakt, line 110: ooo cool stuff

src/pages/preferences/preferencesTabs/notificationTab/utils.tsx
```diff
@@ -0,0 +1,98 @@
...
+      return getVersionRoute(id);
+    case ResourceType.Patch:
+      return getPatchRoute(id, { configure: false });
+    case ResourceType.Task:
+      return getTaskRoute(id);
+    case ResourceType.Host:
+      return getHostRoute(id);
+    case ResourceType.CommitQueue:
+      return getCommitQueueRoute(id);
+    default:
```
minnakt, line 85: do these apply if we are excluding global subscriptions?

sophstad: They don't; lots of subscription design choices reside in a gray area where something is technically possible but not offered at the moment, so I added this in case a user were at one point able to subscribe to a host status. But it would probably be fine to just return the default in those cases and show the ID with no link.

src/pages/preferences/preferencesTabs/notificationTab/UserSubscriptions.tsx
```diff
@@ -0,0 +1,151 @@
...
+import { resourceTypeToCopy, triggerToCopy } from "types/triggers";
+import { getResourceRoute, processSubscriptionData } from "./utils";
+
+export const UserSubscriptions: React.VFC<{}> = () => {
+  const { data } = useQuery<
+    UserSubscriptionsQuery,
+    UserSubscriptionsQueryVariables
+  >(USER_SUBSCRIPTIONS);
+
+  const subscriptions = useMemo(
```
minnakt, line 38: I noticed that personal project subscriptions don't appear to be showing up in the table. Is it possible they are missing?

example project subscription I have on legacy UI:

Screenshot 2023-05-19 at 12 47 36 PM

sophstad:
> When I add a Notification from the first card on the page and save, the table appears and renders empty. Is this expected from the local dev environment?

Yes this is expected! Global subscriptions will not appear in this table. In EVG-19904 we'll add the ability to delete subscriptions in this table and want to ensure that global subs can only be managed in one location.

src/pages/preferences/preferencesTabs/notificationTab/UserSubscriptions.tsx
sophstad: Great catch! Based on how the data is actually structured projects look like this instead of how they do in the designs, if that's okay. I think it's still quite clear what is being monitored!
image

src/pages/preferences/preferencesTabs/notificationTab/UserSubscriptions.tsx
```diff
@@ -0,0 +1,151 @@
...
+  );
+
+  const tableContainerRef = useRef<HTMLDivElement>(null);
+  const table = useLeafyGreenTable<GeneralSubscription | any>({
+    columns,
+    containerRef: tableContainerRef,
+    data: subscriptions ?? [],
+  });
+
+  if (!subscriptions) return null;
```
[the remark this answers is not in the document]
sophstad: Good point, I added an empty state for when no subscriptions are found. I added disabling the "clear subscriptions" button to EVG-19904 since that button will be moved in that ticket making it easier to conditionally disable.
image

sophstad:
evergreen retry

src/pages/preferences/preferencesTabs/notificationTab/utils.tsx
```diff
@@ -0,0 +1,84 @@
...
+          : subscription;
+      })
+  );
+};
+
+const ExpandedBlock = styled.pre`
+  padding: ${size.s} ${size.l};
+`;
+
+export const getResourceRoute = (resourceType: ResourceType, id: string) => {
```
minnakt, line 58: I think this function might need to be updated to link to the project (linking to either the project patches page or project health page seems fine to me)

Screenshot 2023-05-24 at 10 07 35 AM


In case it's needed as a reference, I think this is the part that distinguishes project subscriptions: [[link]](https://github.com/evergreen-ci/spruce/blob/24907af333c612312bfb303255940bd0e95f65b0/src/components/Notifications/utils.ts#L80-L89)

src/pages/preferences/preferencesTabs/notificationTab/UserSubscriptions.tsx
```diff
@@ -0,0 +1,176 @@
...
+    cell: ({
+      getValue,
+      row: {
+        original: { resourceType },
+      },
+    }) => {
+      const selectors = getValue();
+      const selectorId = selectors.find(
+        (s: Selector) => s.type !== "object"
+      )?.data;
```
minnakt, line 141: might also want to prevent matching on `s.type !== "requester"`

minnakt (approved):
thanks for taking on this big table Sophie! 🌷🫶

src/pages/preferences/preferencesTabs/notificationTab/UserSubscriptions.tsx
```diff
@@ -0,0 +1,177 @@
...
+  },
+];
+
+const IdLink = styled(StyledRouterLink)`
+  span {
+    max-width: 200px;
+    overflow: hidden;
+    text-overflow: ellipsis;
+  }
+`;
```
khelif96, line 177: cc @SupaJoon  for the link formatting you were doing would this work better?

src/types/subscription.ts
```diff
@@ -37,3 +37,12 @@ export const projectSubscriptionMethods = [
     label: "Create a JIRA Issue",
   },
 ];
+
+// Map a NotificationMethods enum to its label
+export const notificationMethodToCopy = projectSubscriptionMethods.reduce(
```
khelif96, line 42: nit. Could this use jsdoc syntax so we can get IntelliSense previews?

src/components/SpruceForm/FieldTemplates/ArrayFieldTemplates/index.tsx
```diff
@@ -6,7 +6,7 @@ import { Body } from "@leafygreen-ui/typography";
 import { ArrayFieldTemplateProps } from "@rjsf/core";
 import { PlusButton } from "components/Buttons";
 import Icon from "components/Icon";
-import { formComponentSpacingCSS } from "components/SpruceForm/Container";
+import { formComponentSpacingCSS } from "components/SettingsCard";
```
khelif96, line 9: This makes me think this style should live somewhere else. We do have `components/styles`

src/pages/preferences/preferencesTabs/notificationTab/UserSubscriptions.tsx
```diff
@@ -0,0 +1,177 @@
...
+  }, [data?.userSettings?.notifications]);
+
+  const subscriptions = useMemo(
+    () =>
+      processSubscriptionData(data?.user?.subscriptions, globalSubscriptionIds),
+    [data?.user?.subscriptions, globalSubscriptionIds]
+  );
+
+  const tableContainerRef = useRef<HTMLDivElement>(null);
+  const table = useLeafyGreenTable<GeneralSubscription | any>({
```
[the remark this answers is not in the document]
sophstad: Oops yes I intended to remove this! Good catch

package.json
```diff
@@ -88,6 +88,7 @@
     "@leafygreen-ui/select": "10.2.0",
     "@leafygreen-ui/side-nav": "13.0.2",
     "@leafygreen-ui/table": "10.0.1",
+    "@leafygreen-ui/table/new": "npm:@leafygreen-ui/table@12.0.0",
```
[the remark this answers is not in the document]
sophstad: I'm open to this but it seems like a separate ticket to go through and add the adapter to every existing table. That would bloat this PR significantly. I created EVG-19995 and can grab it next sprint.

src/components/SpruceForm/FieldTemplates/ArrayFieldTemplates/index.tsx
sophstad: I think it makes sense colocated with the card component. It's specifically defining the card spacing, we just want to match that exact spacing for a placeholder that is used in lieu of a card.

src/pages/preferences/preferencesTabs/notificationTab/UserSubscriptions.tsx
```diff
@@ -0,0 +1,177 @@
...
+  const { rows } = table.getRowModel();
+
+  return (
+    <>
+      <SettingsCardTitle>Manage subscriptions</SettingsCardTitle>
+      <SettingsCard>
+        {!subscriptions?.length ? (
+          "No subscriptions found."
+        ) : (
+          <Table table={table} ref={tableContainerRef} shouldAlternateRowColor>
```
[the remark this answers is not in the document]
sophstad: Good eye, it seems like this is a bug that pops up when using renderExpandedContent is used. I'm reporting it to LG!

src/types/subscription.ts
sophstad: Lmk if there's something else you were looking for besides using `/** */` syntax, I wasn't exactly sure since this isn't a function!

sophstad: I wonder if we should look into an ES Lint plugin to enforce jsdoc! There's this: https://www.npmjs.com/package/eslint-plugin-jsdoc

khelif96: I like this i'll add it to the UI round table!

khelif96 (approved):

PR #1836, opened 2023-05-17
EVG-19083 Allow clearing all test filters and table refactors

khelif96:
EVG-19083

### Description
* Refactors the table controls to make them more reusable
* Adds a clear all button to the tests table
* Refactors pagination component and removes antd variant. 
* * This seemed like a natural step to move away from antd and make this component simpler to use (The [leafygreen pagination component](https://www.mongodb.design/component/pagination/example/) was an option but it seems to not be independent enough and probably makes sense to introduce once we fully migrate off of the antd tables)
* Add table controls to bottom of all tables for consistency
* 
### Screenshots
image


cypress/integration/version/task_table.ts
```diff
@@ -80,7 +80,7 @@ describe("Task table", () => {
 
   ["NAME", "STATUS", "BASE_STATUS", "VARIANT"].forEach((sortBy) => {
     // TODO: This test doesn't work bc of issues with assertCorrectRequestVariables
```
minnakt, line 82: Could you open a ticket for this if we don't already have one? Seems like it's from three years ago back when Spruce was on Cypress 5, and I would guess we might be able to fix it now. Alternatively if it seems unimportant just delete it 🤷

src/pages/host/HostTable.tsx
```diff
@@ -60,9 +60,8 @@ export const HostTable: React.VFC<{
         <PaginationWrapper>
           <Pagination
             data-cy="host-event-table-pagination"
-            pageSize={limit}
-            value={page}
-            totalResults={eventsCount}
+            currentPage={page}
```
minnakt, line 63: seeing some weird styles on event log pages!

Screenshot 2023-05-22 at 1 14 49 PM

minnakt (changes requested):

cypress/integration/version/task_table.ts
khelif96: Better yet I will include this in the table migration project. https://jira.mongodb.org/browse/EVG-19972

src/components/Table/TableWrapper.tsx
```diff
@@ -0,0 +1,28 @@
+import styled from "@emotion/styled";
+import { size } from "constants/tokens";
+
+interface TableWrapperProps {
+  children: React.ReactNode;
+  controls?: React.ReactNode;
+  shouldShowBottomTableControl?: boolean;
+}
+const TableWrapper: React.FC<TableWrapperProps> = ({
```
minnakt, line 9: this is awesome!! 😳

minnakt (approved):
looks great! 🤩

PR #1842, opened 2023-05-23
EVG-19738: Show searchable, truncated taskId with title in container events table

SupaJoon:
[EVG-19738](https://jira.mongodb.com/browse/EVG-19738)
### Description
Update the container events log to show a searchable, truncated taskID instead of display name. Hovering over the truncated taskID shows the full string. This makes it easier cmd+f search for a task ID or view the entire task ID without taking up too much of the screen width.

### Screenshots
Screenshot 2023-05-24 at 10 11 15 AM


src/pages/container/EventsTable/util.tsx
```diff
@@ -41,3 +47,15 @@ export const getEventCopy = (
...
+  position: relative;
+  top: 6px;
+`;
+const StyledLink = styled(Link)`
+  max-width: 300px;
+  overflow: hidden;
+  white-space: nowrap;
+  text-overflow: ellipsis;
+  display: inline-block;
+`;
```
khelif96, line 61: I think the `LinkWrapper` styling is a little strange you can fix the alignment issues by providing a wrapper around the entire text block and giving it a flex property. 
Followed by replacing the spans that wrap each of the strings with `<DescriptionWrapper />`

```suggestion
const DescriptionWrapper = styled.div`
  display: flex;
`;
const StyledLink = styled(Link)`
  display: inline-block;
  max-width: 200px; 
  white-space: nowrap;
  overflow: hidden;
  text-overflow: ellipsis;
`;

```

This accomplishes the same effect without needing to add custom positioning padding.
image

khelif96 (changes requested):

src/pages/container/EventsTable/util.tsx
minnakt: I think it should look correct if you remove `LinkWrapper` and use the styles from Sophie's PR i.e.

```
const StyledLink = styled(StyledRouterLink)`
  span {
    max-width: 200px;
    overflow: hidden;
    text-overflow: ellipsis;
  }
`;
```

No need to add `DescriptionWrapper` in this method

minnakt (approved):

khelif96 (approved):

PR #1843, opened 2023-05-24
EVG-18631: Avoid updating state in Task component body

SupaJoon:
[EVG-18631](jira.mongodb.com/browse/EVG-18631)

### Description
`updateQueryParams` was updating search params and triggering a state update within the component body which resulted in an error from BrowserRouter. To resolve this I moved the state update into a useEffect so the update happens in a callback. You can replicate the issue by clicking "Go" on the previous commit selector in staging on this page: http://localhost:3000/task/mci_osx_test_graphql_1db43f73f30b1f3711667ee6af4af8040203206f_20_04_21_18_11_22/logs?execution=0


minnakt (approved):
nice! 😸

PR #1848, opened 2023-05-26
EVG-19883: Create outline for Views & Filters tab

minnakt:
EVG-19883

### Description
This PR creates an outline for the new Views & Filters tab. This is just a basic skeleton to unblock [EVG-19891](https://jira.mongodb.org/browse/EVG-19891) and [EVG-19892](https://jira.mongodb.org/browse/EVG-19892).

Since Views & Filters settings are not available at the repo level, I made it so that the tab does not show up on repos and the "Go to repo settings" link does not appear. (Also if you want the tabs in a different order let me know 🙏)

### Screenshots
Screenshot 2023-05-26 at 9 26 37 AM

### DIFF
--- a/src/pages/projectSettings/Header.tsx
+++ b/src/pages/projectSettings/Header.tsx
@@ -11,6 +11,10 @@ import { HeaderButtons } from "./HeaderButtons";
 import { readOnlyTabs, WritableTabRoutes } from "./tabs/types";
 import { ProjectType } from "./tabs/utils";
 
+const notAvailableAtRepoLevel: Set<ProjectSettingsTabRoutes> = new Set([
+  ProjectSettingsTabRoutes.ViewsAndFilters,
+]);
+
 interface Props {
   attachedRepoId?: string;
   id: string;
@@ -26,12 +30,13 @@ export const Header: React.VFC<Props> = ({
 }) => {
   const { title } = getTabTitle(tab);
   const saveable = !(readOnlyTabs as ReadonlyArray<string>).includes(tab);
+  const showRepoLink = !notAvailableAtRepoLevel.has(tab);
 
   return (
     <Container>
       <TitleContainer>
         <H2 data-cy="project-settings-tab-title">{title}</H2>
-        {projectType === ProjectType.AttachedProject && (
+        {projectType === ProjectType.AttachedProject && showRepoLink && (
           <StyledRouterLink
             to={getProjectSettingsRoute(attachedRepoId, tab)}
             data-cy="attached-repo-link"
--- a/src/pages/projectSettings/Header.tsx
+++ b/src/pages/projectSettings/Header.tsx
@@ -8,12 +8,8 @@ import {
 import { size } from "constants/tokens";
 import { getTabTitle } from "./getTabTitle";
 import { HeaderButtons } from "./HeaderButtons";
-import { readOnlyTabs, WritableTabRoutes } from "./tabs/types";
+import { projectOnlyTabs, readOnlyTabs, WritableTabRoutes } from "./tabs/types";
 import { ProjectType } from "./tabs/utils";
-
-const notAvailableAtRepoLevel: Set<ProjectSettingsTabRoutes> = new Set([
-  ProjectSettingsTabRoutes.ViewsAndFilters,
-]);
 
 interface Props {
   attachedRepoId?: string;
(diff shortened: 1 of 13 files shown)


src/pages/projectSettings/tabs/ViewsAndFiltersTab/transformers.ts
```diff
@@ -0,0 +1,23 @@
...
+};
+
+export const formToGql: FormToGqlFunction<Tab> = (
+  _formState: FormState,
+  id: string
+) => ({
+  projectRef: {
+    id,
+  },
+});
```
sophstad, line 23: These can be better typed using the new TS `satisfies` operator, which applies types to the parameters and return value. There's a good writeup [here](https://www.learningtypescript.com/articles/the-satisfies-operator#real-world-usage-nextjs) about how it can apply to functions. I've used it for SpruceForm `validate` functions: https://github.com/evergreen-ci/spruce/blob/d1f969d680216e2ca45a73d74db380732a4d6169/src/pages/projectSettings/tabs/VariablesTab/VariablesTab.tsx#L91-L104

src/pages/projectSettings/Header.tsx
```diff
@@ -11,6 +11,10 @@ import { HeaderButtons } from "./HeaderButtons";
 import { readOnlyTabs, WritableTabRoutes } from "./tabs/types";
 import { ProjectType } from "./tabs/utils";
 
+const notAvailableAtRepoLevel: Set<ProjectSettingsTabRoutes> = new Set([
```
sophstad, line 14: We have a similar utility for read-only tabs in `projectSettings/tabs/types.ts`; perhaps this should be located there too. Also maybe `projectOnlyTabs` is more clear? Your call!

src/pages/projectSettings/tabs/ViewsAndFiltersTab/ViewsAndFiltersTab.tsx
```diff
@@ -0,0 +1,35 @@
...
+  return (
+    <SpruceForm
+      fields={fields}
+      formData={formData}
+      onChange={onChange}
+      schema={schema}
+      uiSchema={uiSchema}
+    />
+  );
+};
```
sophstad, line 35: At some point maybe we should make a generic project settings component that handles all this stuff 😅 but looks good for now!

src/pages/projectSettings/tabs/ViewsAndFiltersTab/types.ts
```diff
@@ -0,0 +1,10 @@
+import { ProjectType } from "../utils";
+
+export interface FormState {}
+
+export type TabProps = {
+  identifier: string;
+  projectData?: FormState;
+  projectType: ProjectType;
+  repoData?: FormState;
```
sophstad, line 9: Probably don't need this since repo isn't supported!

sophstad (changes requested):

src/pages/projectSettings/tabs/ViewsAndFiltersTab/transformers.ts
minnakt: thanks Sophie 🫶 please correct me if I used it wrong!

src/pages/projectSettings/tabs/ViewsAndFiltersTab/ViewsAndFiltersTab.tsx
minnakt: I shall file a ticket :)

sophstad (approved):
👏

minnakt:
> 👏 (not sure what's up with unit tests but otherwise LGTM!)

Ooh, I think there's problem in the Jest/Babel config as it can't parse the `satisfies` operator. I tried adding `@babel/preset-typescript` and that seemed to do the trick. It seems a little odd to include in this PR so I think I'll take the `satisfies` stuff out and put it into a new ticket -- which could be nice since we can update all of the transformers at once to use `satisfies` 😁

minnakt:
I'm going to merge with `check_codegen` failing, as updating the GraphQL types breaks `compile` and `type_check`. Checked in [this patch](https://spruce.mongodb.com/version/6470f1b91e2d1795ffbb5abc/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

PR #1849, opened 2023-05-26
EVG-19805: Periodic build interval can be specified with cron expression

sophstad:
EVG-19805

### Description

- Use toggle to control whether periodic build interval is specified in number of hours or via cron expression

### Screenshots

image
image

### Testing

- Update unit tests to ensure that the correct field is saved based on the toggle's value

### Evergreen PR

https://github.com/evergreen-ci/evergreen/pull/6556


src/pages/projectSettings/tabs/utils/alias.ts
```diff
@@ -431,7 +431,7 @@ export const gitTagArray = {
         gitTag: gitTag.schema,
         specifier: {
           type: "string" as "string",
-          title: "Specify Via",
```
minnakt, line 434: was there a reason to remove this? did you want to remove this instance too? :0

Screenshot 2023-05-31 at 11 32 04 AM

minnakt (approved):
looks great, I just had one question

src/pages/projectSettings/tabs/utils/alias.ts
sophstad: Created EVG-20037 to make some improvements to these!

PR #1850, opened 2023-05-26
EVG-19110: Create and apply ShortenedRouterLink 

SupaJoon:
EVG-19110
### Description
The abbreviated router link truncates and adds ellipsis to links so they are capped at 200px width. The new component applied to the host & container events table and user subscriptions. The user subscription and container pages had a similar solution without the shared component abstraction. The tests changed to search for the entire truncated string. 

### Screenshots
Screenshot 2023-05-26 at 3 03 52 PM

### DIFF
--- a/cypress/integration/host/host_events.ts
+++ b/cypress/integration/host/host_events.ts
@@ -97,7 +97,11 @@ describe("Host events", () => {
       },
       {
         hostType: "host-running-task-cleared",
-        text: "Current running task cleared (was: evergreen_ubuntu1604_test_command_patch_5e823e1f28baeaa22ae00823d83e03082cd148ab_5e4ff3abe3c3317e352062e4_20_02_21_15_13_48)",
+        text: "Current running task cleared (was: ",
+      },
+      {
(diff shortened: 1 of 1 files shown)


src/components/styles/StyledLink.tsx
```diff
@@ -20,3 +21,11 @@ export const StyledLink = (props) => (
 );
 
 export const StyledRouterLink = (props) => <StyledLink as={Link} {...props} />;
+
+export const AbbreviatedRouterLink = styled(StyledRouterLink)`
```
minnakt, line 25: naming-wise I might prefer `TruncatedRouterLink` or `ShortenedRouterLink`, abbreviated to me sounds like extracting characters from the string (e.g. like how tablespoon gets abbreviated as tbsp)

minnakt (approved):
looks great!

IamXander:
1. Nice work @SupaJoon ! Thank for owning this ticket!
2. Is the screenshot you are showing in the description a before or after?

IamXander (approved):

PR #1851, opened 2023-05-30
EVG-16959: Reintroduce navigation warning modal

sophstad:
EVG-16959

### Description

- Support for navigation blocking was reintroduced to react-router in https://github.com/remix-run/react-router/pull/9709. We can once again warn when a user tries to navigate away from project settings with unsaved changes.

### Screenshots

image

### Testing

- Re-add integration test

### DIFF
--- a/src/pages/projectSettings/Context.tsx
+++ b/src/pages/projectSettings/Context.tsx
@@ -208,9 +208,13 @@ const useHasUnsavedTab = (): {
   unsavedTabs: ProjectSettingsTabRoutes[];
 } => {
   const { tabs } = useProjectSettingsContext();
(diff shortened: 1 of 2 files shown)


sophstad:
> Is it possible to render the modal when the user selects a project from the project dropdown (since the selections are cleared) or click on "Switch to Legacy UI" or "Documentation."

The two navbar links function outside of react-router so it's impossible to catch them here sadly. But Documentation opens in a separate window, and soon we are deprecating the legacy project settings page and the button will be removed when users are looking at project settings, so I think the impact is minimal here. Good catch on warning when a user navigates to a new project, added handling for that!

PR #1852, opened 2023-05-30
EVG-18922: Upgrade typography package

sophstad:
EVG-18922

### Description

- Use new `as` prop with InlineCode links
- Remove lots of `@ts-expect-error` 🤩
- Clean up typing. When using a native HTML prop on a styled LeafyGreen component, type assertion is required using `as StyledComponent<ComponentProps>` (details from LG [here](https://mongodb.slack.com/archives/CFFA5BS79/p1682437079758049?thread_ts=1682435618.456929&cid=CFFA5BS79)).

### Testing

- Update snapshots


src/components/ErrorBoundary/ErrorFallback.tsx
```diff
@@ -25,11 +24,11 @@ const ErrorFallback = () => (
 
 export default ErrorFallback;
 
-const StyledHeader = styled<H1Type>(H1)`
+const StyledHeader = styled(H1)`
```
khelif96, line 27: Should these use the StyledComponent pattern the other components use?

src/pages/task/metadata/index.tsx
```diff
@@ -183,11 +184,12 @@ export const Metadata: React.VFC<Props> = ({
         <MetadataItem>
           Base commit:{" "}
           <InlineCode
+            as={Link}
```
khelif96, line 187: Nice

src/components/CodeChanges/CodeChanges.tsx
```diff
@@ -1,7 +1,12 @@
 import { useQuery } from "@apollo/client";
-import styled from "@emotion/styled";
+import styled, { StyledComponent } from "@emotion/styled";
```
khelif96, line 2: Didn't know this was a thing!

src/components/CodeChanges/CodeChanges.tsx
```diff
@@ -114,12 +118,11 @@ const StyledButton = styled(Button)`
...
 `;
 
-// @ts-expect-error
-const Title: SubtitleType = styled(Subtitle)`
+const Title = styled(Subtitle)`
   font-weight: normal;
   margin-right: ${size.s};
   margin-bottom: ${size.s};
-`;
+` as StyledComponent<SubtitleProps>;
```
khelif96, line 125: I'd be curious to understand why you went with this approach as opposed to something like this. Does the type assertion offer any advantages?

```
const Title = styled(Subtitle)<SubtitleProps>`
  font-weight: normal;
  margin-right: ${size.s};
  margin-bottom: ${size.s};
`;
```

sophstad: I went with this approach because it's what LeafyGreen recommended for this but I actually do agree this is nicer looking. Updated!

src/components/ErrorBoundary/ErrorFallback.tsx
sophstad: I added that assertion specifically where type errors were thrown. This is what we were doing previously with `H1Type` etc. so I think it's sufficient.

khelif96 (approved):
LGTM

PR #1854, opened 2023-05-31
EVG-19230: Convert codegen yaml configuration to typescript

SupaJoon:
[EVG-19230](https://jira.mongodb.com/browse/EVG-19230)

### Description
These code changes convert codegen.yml to codegen.ts. There was an invalid trailing comma on 16 of codegen.yml. Removing the comma resulted in a diff to `generated/types.ts` where primitive values are no longer wrapped in Maybe. The only way I was able to get the original behavior was with the typo.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "spruce",
-  "version": "3.0.83",
+  "version": "3.0.84",
   "private": true,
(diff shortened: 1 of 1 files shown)


khelif96 (approved):
The change makes sense to me 💯

PR #1856, opened 2023-05-31
EVG-20039 Rename patch status to task status

khelif96:
EVG-20039 

### Description
@julianedwards  pointed out we use the labels `Patch/ Version Status` on the tasks table over a column that represents task statuses. This updates the verbiage to reflect that.
### Screenshots

image


julianedwards (approved):
LGTM! Thanks for the speedy turnaround time!

PR #1870, opened 2023-06-02
EVG-20065 migrate patch status to success

ablack12:
EVG-20065
### Description
Attempting to migrate "succeeded" to "success" (re https://github.com/evergreen-ci/evergreen/pull/6567/files).

This needs to be handled for patch statuses, and for patch trigger alias statuses. For the latter I'm converting "succeeded" to "success" in BuildFromService so theoretically if these were deployed together we wouldn't need to support the legacy status, but ideally we could write this so we know how to support both. I'm also certainly not sure if I have found all cases of this.

Lowkey would love to take on this work even though I don't know what I'm doing to get some ⭐ exposure ⭐ so if it seems like conversation here would benefit from a pair program I would definitely be down. 

Will test with staging but eyes before that would probably be helpful to verify how off I am.

### DIFF
--- a/src/components/PatchStatusBadge.tsx
+++ b/src/components/PatchStatusBadge.tsx
@@ -17,6 +17,7 @@ const statusToBadgeVariant = {
   [PatchStatus.Failed]: Variant.Red,
   [PatchStatus.Started]: Variant.Yellow,
   [PatchStatus.Success]: Variant.Green,
+  [PatchStatus.LegacySuccess]: Variant.Green,
   [PatchStatus.Aborted]: Variant.LightGray,
 };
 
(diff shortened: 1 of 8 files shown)


khelif96 (changes requested):
This also needs to be changed in a few other places. The changes need to be made in a backward-compatible way to support both the old and new ways of representing patches. I think we can make the verbiage consistent but the mappings need to support both the old and new success status. Please feel free to reach out to someone on UI to pair. 
https://github.com/evergreen-ci/spruce/blob/2ebfcc6ab8c9da395cc44b16dfc8495d0e00c344/src/components/PatchStatusBadge.tsx
https://github.com/evergreen-ci/spruce/blob/2ebfcc6ab8c9da395cc44b16dfc8495d0e00c344/src/types/patch.ts#L9
https://github.com/evergreen-ci/spruce/blob/2ebfcc6ab8c9da395cc44b16dfc8495d0e00c344/src/components/PatchesPage/StatusSelector.tsx#L39

sophstad:
Completed in #2000

PR #1877, opened 2023-06-06
EVG-18722: Navigation should re-fetch page queries

SupaJoon:
[EVG-18722](https://jira.mongodb.com/browse/EVG-18722)

### Description
The goal of this PR is to prevent showing users stale data when they navigate to pages in the NavBar. Currently, users don't experience a loading state or new data in MyPatches or Commits for pages that have been visited. This allows for a smooth experience because the data is presented immediately. If we load fresh data for the first page in paginated data feeds like Commits and MyPatches, it's important that subsequent pages are refetched as well to prevent data loss. 
The ideal solution is to clear the cache for a specific query when a page is visited so the cache can be repopulated as the user paginates through. However, since the ApolloCache is a graph and data is normalized, it's not easily possible to create a function that accepts an operation name and deletes all of the related data. Instead, [an object ID for a specific data must be supplied](https://www.apollographql.com/docs/react/caching/garbage-collection#cacheevict) to release it from the cache. 
The next simplest solution I could come up with is to show cached data while a fetch is dispatched with the `cache-and-network` fetch policy. This essentially resembles an expedited poll for pages that enable polling. Instead of waiting for the poll to dispatch, a network request is dispatched immediately and then polling continues. The user experience is nearly the same at the cost of additional network traffic. I omitted the change for ProjectPatches because the network request could take 7 seconds to resolve for mongodb-mongo-master. MyHosts already refetches data and ProjectSettings doesn't experience frequent updates so I omitted the change for those pages as well.

### DIFF
--- a/cypress/integration/preferences/notifications.ts
+++ b/cypress/integration/preferences/notifications.ts
@@ -24,17 +24,65 @@ describe("global subscription settings", () => {
 });
 
 describe("user subscriptions table", () => {
-  it("shows all of a user's subscriptions and expands with details", () => {
+  beforeEach(() => {
     cy.visit(pageRoute);
+  });
+
+  it("shows all of a user's subscriptions and expands with details", () => {
     cy.dataCy("subscription-row").should("have.length", 3);
 
     cy.dataCy("regex-selectors").should("not.be.visible");
     cy.dataCy("trigger-data").should("not.be.visible");
-    cy.get("tr button").first().click();
+    cy.dataCy("subscription-row")
+      .eq(0)
+      .within(() => {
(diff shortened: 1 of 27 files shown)


src/pages/UserPatches.tsx
```diff
@@ -37,6 +37,7 @@ export const UserPatches = () => {
         includeCommitQueue: isCommitQueueCheckboxChecked,
       },
     },
+    fetchPolicy: "cache-and-network",
```
sophstad, line 40: This works really well!

sophstad (approved):

PR #1878, opened 2023-06-06
EVG-20089 Add useEffectDebugger hook

khelif96:
EVG-20089

### Description
This is just copying over the useEffectDebugger util hook from parsley over to spruce. 
Decided to split this off of EVG-18780

### DIFF
--- a/src/test_utils/hooks.ts
+++ b/src/test_utils/hooks.ts
@@ -1,18 +1,5 @@
-import { DependencyList, EffectCallback, useEffect, useRef } from "react";
-
-/**
- * `usePrevious` is a custom hook that returns the previous value of a given value
- * @param value
- * @param initialValue
- * @returns - the previous value
- */
-const usePrevious = <T>(value: T, initialValue: T) => {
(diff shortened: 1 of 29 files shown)


src/test_utils/hooks.ts
```diff
@@ -0,0 +1,68 @@
...
+ * @param initialValue
+ * @returns - the previous value
+ */
+const usePrevious = <T>(value: T, initialValue: T) => {
+  const ref = useRef(initialValue);
+  useEffect(() => {
+    ref.current = value;
+  });
+  return ref.current;
+};
```
minnakt, line 15: Do we need to redeclare `usePrevious` if we have an existing one in the hooks folder? :0

minnakt (approved):
yayayayay

src/test_utils/hooks.ts
khelif96: Great catch!

khelif96: Actually, nvm they are slightly different. One of them declares an initialValue Param.

PR #1881, opened 2023-06-07
EVG-19732: Fix flaky useBreadcrumbRoot test

minnakt:
EVG-19732

### Description
This PR fixes the flaky `useBreadcrumbRoot` test. The test was having an issue dealing with the cache data, since the UserQuery's key field is `"userId"` rather than the default `"id"`. 

### Testing
- The original test failure can be replicated locally, so I just ran the changes 20 times and made sure it never failed

### DIFF
--- a/src/hooks/tests/useBreadcrumbRoot.test.tsx
+++ b/src/hooks/tests/useBreadcrumbRoot.test.tsx
@@ -1,3 +1,4 @@
+import { InMemoryCache } from "@apollo/client";
 import { MockedProvider } from "@apollo/client/testing";
 import { renderHook } from "@testing-library/react-hooks";
 import { OtherUserQuery, OtherUserQueryVariables } from "gql/generated/types";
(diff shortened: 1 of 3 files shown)


src/hooks/tests/useBreadcrumbRoot.test.tsx
```diff
@@ -41,12 +50,11 @@ describe("useBreadcrumbRoot", () => {
...
   });
 
-  it("returns the correct breadcrumb root when the version is a commit", async () => {
-    const { result, waitForNextUpdate } = renderHook(
+  it("returns the correct breadcrumb root when the version is a commit", () => {
+    const { result } = renderHook(
       () => useBreadcrumbRoot(false, "admin", "spruce"),
       { wrapper: SameUserProvider }
     );
-    await waitForNextUpdate();
```
minnakt, line 49: It should make sense that we don't have to wait for an update if the version is a commit, because we skip the query in this case.

PR #1882, opened 2023-06-08
EVG-19995: Use V11 Table Adapter

sophstad:
EVG-19995

### Description

- Use `V11TableAdapter` when possible to convert legacy tables
- Sadly our Task Duration Table seemed to be too complex for the adapter 😢 I've left it out of this PR because it's a significant chunk of work. That means we can't get rid of the legacy table dependency yet, but it's ready for work at any time at EVG-20104.

### Screenshots

Looks pretty much the same—
image


sophstad:
Cypress tests are failing due to [LG-3313](https://jira.mongodb.org/browse/LG-3313), I think we should wait until that's resolved.

khelif96 (approved):
Thanks for following up on this! Can't wait until we can fully migrate our tables and not have 3 different table deps :)

PR #1883, opened 2023-06-08
EVG-19734: Delay pushing tags for less flaky deploys

sophstad:
EVG-19734

### Description

- Updated our postversion hook that pushes the deploy version and tags to GitHub.
  - Wait 10 seconds between pushing commit and tag
  - Push directly to our remote so that we don't rely on users' git remote configuration
- Remove deploy utils that are no longer used

### Screenshots

Should look something like this...
image

### Testing

- You can test it by renaming `postversion` to something else, updating `GITHUB_REMOTE` to point at your own fork, and running `yarn <test_command_name>`.


scripts/push-version.sh
```diff
@@ -0,0 +1,16 @@
...
+GITHUB_REMOTE=https://github.com/evergreen-ci/spruce
+
+git push $GITHUB_REMOTE
+
+echo -n "Waiting ${WAIT_TIME}s for Evergreen to pick up the version"
+
+i=0
+while [ $i -lt $WAIT_TIME ]
+  do echo -n .; sleep 1; i=$((i+1))
+done
```
khelif96, line 13: You can also do some carriage return hacking if you want to animate the countdown :) 
```suggestion
i=$WAIT_TIME
while [ $i -gt 0 ]
  do echo -en "Waiting $i s for Evergreen to pick up the version\r"; sleep 1; i=$((i-1))
done
```


https://github.com/evergreen-ci/spruce/assets/4605522/ba2dafd9-6fc8-41fb-a0fb-d7f1491ef4bc

scripts/deploy-utils.js
khelif96, line 1: Great catch thanks for cleaning up.

khelif96 (approved):
Thanks for taking this on! Would love to see this ported to Parsley as well!

PR #1887, opened 2023-06-08
EVG-20117: Unlink GitHub PR user

sophstad:
EVG-20117

### Description

- Unlink github_pull_request as a user in the project patches page. Querying its patches takes a long time and they are not very useful. I also threw in parent_patch since it is similarly useless. The pages can still be accessed directly if so desired.

### Screenshots

image

### DIFF
--- a/src/components/PatchesPage/PatchCard.tsx
+++ b/src/components/PatchesPage/PatchCard.tsx
@@ -4,6 +4,7 @@ import { Analytics } from "analytics/addPageAction";
 import { GroupedTaskStatusBadge } from "components/GroupedTaskStatusBadge";
 import { PatchStatusBadge } from "components/PatchStatusBadge";
 import { StyledRouterLink } from "components/styles";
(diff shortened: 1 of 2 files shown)


src/components/PatchesPage/PatchCard.tsx
```diff
@@ -21,8 +21,11 @@ import { DropdownMenu } from "./patchCard/DropdownMenu";
 
 type P = Unpacked<PatchesPagePatchesFragment["patches"]>;
 type PatchProps = Omit<P, "commitQueuePosition">;
+
 const { gray } = palette;
 
+const unlinkedPRUsers = new Set(["github_pull_request", "parent_patch"]);
```
khelif96, line 27: I think `constants/patch.ts` might be a good place for this.

khelif96 (approved):

minnakt (commented):
Is the ticket number for this PR possibly incorrect? I can't open the issue

PR #1894, opened 2023-06-13
EVG-20133: Remove mainline commits icon flicker

sophstad:
EVG-20133

### Description

Previously, moving your cursor between two icons on mainline commits would cause all icons to flicker back to `opacity: 1` and then become dim again:

https://github.com/evergreen-ci/spruce/assets/9298431/277db306-9f44-4feb-bd54-f5ef5d8ecc6d

I mitigated this by applying the dimming effect to all icons whenever an IconContainer is moused over. An IconContainer holds all icons for a build variant and thus is bigger, so there are fewer mouseenter/mouseleave events. Then, a second global style is applied to highlight all icons of the same task identifier.

https://github.com/evergreen-ci/spruce/assets/9298431/de0056a6-da8d-4586-bfa7-fade9c3cb0b1

I also slightly increased the timeout for fetching data from 500 to 600, as it resulted in less jerkiness when mousing around the page and wasn't perceptibly slower.

### Testing

-  Add test to confirm dim function is being applied


src/pages/commits/ActiveCommits/utils.ts
```diff
@@ -162,22 +162,43 @@ export const getStatusesWithZeroCount = (colors: ColorCount[]) => {
...
+
+export const injectGlobalHighlightStyle = (taskIdentifier: string) => {
+  // Remove style here again because hovering over LG tooltips triggers two consecutive mouseenter events.
+  removeGlobalHighlightStyle();
+
+  const hoverStyle = document.createElement("style");
+  hoverStyle.id = taskIconStyle;
+  hoverStyle.innerHTML = `
+     [data-task-icon="${taskIdentifier}"] {
+         opacity: 1 !important;
```
khelif96, line 200: smart!

src/pages/commits/ActiveCommits/BuildVariantCard.test.tsx
```diff
@@ -0,0 +1,78 @@
...
+    id: "2",
+    status: "succeeded",
+    displayName: "Two",
+  },
+  {
+    id: "3",
+    status: "succeeded",
+    displayName: "Three",
+  },
+];
```
minnakt, line 78: I think you might have to update these to say `"success"`:

Screenshot 2023-06-16 at 1 26 58 PM

src/pages/commits/ActiveCommits/BuildVariantCard.test.tsx
```diff
@@ -0,0 +1,78 @@
...
+        },
+      ]);
+      return {
+        observe: jest.fn(),
+        disconnect: jest.fn(),
+      };
+    });
+
+    // @ts-expect-error
+    window.IntersectionObserver = mockIntersectionObserver;
```
minnakt, line 28: If you want to avoid `@ts-expect-error`, I think this is also an option:
```suggestion
     Object.defineProperty(window, "IntersectionObserver", {
         writable: true,
         value: mockIntersectionObserver,
    });
```

khelif96 (approved):
Works great!

src/pages/commits/ActiveCommits/utils.test.ts
```diff
@@ -12,8 +12,8 @@ import {
   getStatusesWithZeroCount,
   constructBuildVariantDict,
   roundMax,
```
minnakt, line 14: Should the tests for `injectGlobalHighlightStyle` and `removeGlobalHighlightStyle` have been removed from this file? I thought it might make sense to have tests for all four functions (unless that seems too redundant)

minnakt (approved):
the dim style idea is so galaxy brain 🙌 And I really do notice the decrease in flickering =)

PR #1906, opened 2023-06-21
EVG-18032 Disable task selection checkboxes if they have already been activated

khelif96:
EVG-18032

### Description
Previously activated tasks will no longer be selectable on the task configure page and will be disabled.
### Screenshots
https://github.com/evergreen-ci/spruce/assets/4605522/0bf18ec3-f8ea-4e38-aa74-8e51b3a23a4c

### Testing
Wrote plenty of unit tests to test this functionality


src/pages/configurePatch/configurePatchCore/ConfigureTasks/utils.ts
```diff
@@ -0,0 +1,192 @@
...
+        default:
+          visibleTasks[taskName] = value
+            ? CheckboxState.CHECKED
+            : CheckboxState.UNCHECKED;
+          break;
+      }
+      if (previousBuildVariantTasks[taskName]) {
+        visibleTasks[taskName] = isCheckboxIndeterminate(visibleTasks[taskName])
+          ? CheckboxState.DISABLED_INDETERMINATE
+          : CheckboxState.DISABLED_CHECKED;
```
sophstad, line 64: Combining the checkboxes' disabled and checked states kinda feels like a derived state antipattern to me. Was there any particular benefit to this? It seems like it would straightforward to do this check in the component itself and apply the disabled prop that way.

sophstad (changes requested):

khelif96:
> There is unexepected behavior around select all checkbox state and the onChange handler. I think it's good that we did this code change first because the filter input will filter on top of the previously ran tasks. The video shows that the "select all" checkbox state responds differently to previously ran tasks and also shows that not all tasks are selected when they should have been.
> 
>  Screen.Recording.2023-06-22.at.11.19.55.AM.mov

Good catch!

src/pages/configurePatch/configurePatchCore/ConfigureTasks/utils.ts
sophstad: Got it, that's fair enough! Another lightweight idea may be to separate the states? I feel like this would be pretty flexible. But definitely respect if you want to ticket this as later work instead!

```js
interface DeduplicateTasksResult {
  [task: string]: {
    checkbox: CheckboxState;
    previouslySeen: boolean;
}
```

khelif96:
Closing in favor of #1912

PR #1907, opened 2023-06-22
EVG-20293 Add yarn clean command and update readme

khelif96:
EVG-20293

### Description
To help reduce some confusion around broken dependencies update the readme and introduce the `yarn clean` command


hadjri (approved):
Thanks for fixing this so fast!

PR #1910, opened 2023-06-23
EVG-20143: Update deploy script

sophstad:
EVG-20143

### Description

- Add support for redeploying the most recent commit tag: if no new commits are found and the user opts to redeploy, delete and re-push the latest tag.
- Deploying locally without the use of Evergreen is now done via `yarn deploy:prod --local`.

### Testing
- Added lots of tests for remote and local deploys


scripts/run-deploy.js
```diff
@@ -0,0 +1,12 @@
+const { evergreenDeploy, localDeploy } = require("./deploy-production.js");
+
+const main = async () => {
+  if (process.argv.includes("--local")) {
+    await localDeploy();
+    return;
+  }
+  await evergreenDeploy();
+  return;
```
khelif96, line 9: nit. I got a linter warning for an unnecessary return.

scripts/deploy-production.test.js
```diff
@@ -0,0 +1,302 @@
+import prompts from "prompts";
```
khelif96, line 1: Wow awesome job with these tests!

scripts/deploy-production.js
```diff
@@ -14,113 +22,89 @@ const main = async () => {
...
-        resolve(stdout);
-      }
-    );
+/* Deploy by generating a production build locally and pushing it directly to S3. */
+const localDeploy = async () => {
+  const response = await prompts({
+    type: "confirm",
+    name: "value",
+    message:
+      "Are you sure you'd like to build Spruce locally and push directly to S3?",
```
khelif96, line 81: Can we add some disclaimers here about how this is a dangerous operation? 

I want to ensure whoever is running this knows about the risks since many things can go wrong.
Such as deploying outdated deps,
outdated env variables,
deploying untested code,

scripts/deploy-production.js
```diff
@@ -14,113 +22,89 @@ const main = async () => {
...
+
+  // If there are no commit messages, ask the user if they want to delete and re-push the latest tag, thereby forcing a deploy with no new commits.
   if (commitMessages.length === 0) {
-    // If there are no commit messages, ask the user if they want to deploy anyway
     const response = await prompts({
       type: "confirm",
       name: "value",
-      message: "No new commits. Do you want to deploy anyway?",
+      message:
+        "No new commits. Do you want to deploy the most recent existing tag?",
```
khelif96, line 33: Out of curiosity what happens if we push a duplicate tag to evergreen? @ablack12

scripts/deploy-production.js
```diff
@@ -1,10 +1,18 @@
...
+  isWorkingDirectoryClean,
+  pushTags,
+  runLocalDeploy,
+} = require("./deploy-utils");
 
-const localDeployScript =
-  "yarn build:prod && env-cmd -e production yarn deploy:do-not-use && env-cmd -e production ./scripts/email.sh";
-
-const main = async () => {
+/* Deploy by pushing a git tag, to be picked up and pushed to S3 by Evergreen. */
```
khelif96, line 14: nit. since this reads as the git tag is pushed to S3
```suggestion
/* Deploy by pushing a git tag, to be picked up and built by Evergreen, and deployed to S3.
```
or even something more verbose

```suggestion
/* Deployment occurs by pushing a git tag, which is then picked up and built by Evergreen, followed by the deployment of the resulting artifacts to S3. */
```

scripts/deploy-production.js
```diff
@@ -14,113 +22,89 @@ const main = async () => {
...
-        if (err) {
-          reject(err);
-          return;
-        }
-        resolve(stdout);
-      }
-    );
+/* Deploy by generating a production build locally and pushing it directly to S3. */
+const localDeploy = async () => {
+  const response = await prompts({
```
khelif96, line 77: This doesn't have the same `isOnMainBranch` protections as the normal deploy.

khelif96 (changes requested):

scripts/run-deploy.js
sophstad: Do you know what rules you have applied locally that are not covered by Spruce's lint config? It'd be nice to have the two standardized.

scripts/deploy-production.js
```diff
@@ -14,113 +22,98 @@ const main = async () => {
...
-    console.log(commitMessages);
+
+    console.log(
+      "Deploy canceled. If systems are experiencing an outage and you'd like to push the deploy directly to S3, run yarn deploy:prod --local."
+    );
+    return;
   }
 
+  // Print all commits between the last tag and the current commit
+  console.log("Commit messages:\n" + commitMessages);
```
khelif96, line 50: My linter throws the following error 
image

scripts/deploy-production.js
```diff
@@ -14,113 +22,98 @@ const main = async () => {
...
-        reject(err);
-        return;
-      }
-      resolve(stdout.trim() === "");
-    });
-  });
+const deleteAndPushLatestTag = async () => {
+  try {
+    const latestTag = await getLatestTag();
+    console.log("Deleting and re-pushing latest tag (" + latestTag + ")");
```
khelif96, line 109: Same here.

scripts/deploy-production.js
khelif96: This is one of the default rules on the eslint Airbnb rules which is what we have applied project-wide. I'm not sure why it didn't pick up for you. Maybe this folder isn't checked by our lint scripts? https://airbnb.io/javascript/#es6-template-literals

scripts/run-deploy.js
khelif96: Same comment about this being a default rule in the airbnb config we are using. https://github.com/airbnb/javascript/blob/fd77bbebb77362ddecfef7aba3bf6abf7bdd81f2/packages/eslint-config-airbnb-base/rules/best-practices.js#L363
I'm unsure why it isn't flagging in CI or for you locally.

scripts/deploy-production.js
sophstad: Yep you were right we weren't looking in `scripts/`!

khelif96 (approved):
Great job!

PR #1914, opened 2023-06-27
EVG-20041: Default Patch list commit queue checkbox to cookie value when commitQueue query param is absent

SupaJoon:
[EVG-20041](https://jira.mongodb.com/browse/EVG-20041)

### Description
These code changes default the patch list commit queue checkbox to cookie value when commitQueue query param is absent.


khelif96 (approved):
🎉

PR #1915, opened 2023-06-27
EVG-8179: Introduce task filter input on configure page

SupaJoon:
[EVG-8179](https://jira.mongdb.com/browse/EVG-8179)

### Description
The task filter input filters tasks checkboxes based on task name. The "Select all" button should only select tasks that are in view after the filter is applied.
Screenshot 2023-06-27 at 1 00 50 PM

### DIFF
--- a/src/pages/configurePatch/configurePatchCore/ConfigureTasks/ConfigureTasks.test.tsx
+++ b/src/pages/configurePatch/configurePatchCore/ConfigureTasks/ConfigureTasks.test.tsx
@@ -225,7 +225,6 @@ describe("configureTasks", () => {
     });
     it("applying a search should filter the tasks", () => {
       const selectedBuildVariants = ["ubuntu2004", "ubuntu1804"];
-      const setSelectedBuildVariantTasks = jest.fn();
       render(
         <ConfigureTasks
           selectedBuildVariants={selectedBuildVariants}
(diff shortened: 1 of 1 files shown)


src/pages/configurePatch/configurePatchCore/ConfigureTasks/ConfigureTasks.test.tsx
```diff
@@ -223,6 +223,32 @@ describe("configureTasks", () => {
         ubuntu1804: { compile: true, lint: true },
       });
     });
+    it("applying a search should filter the tasks", () => {
+      const selectedBuildVariants = ["ubuntu2004", "ubuntu1804"];
+      const setSelectedBuildVariantTasks = jest.fn();
```
khelif96, line 228: Could also be a no-op since we don't assert on jest.fn

src/pages/configurePatch/configurePatchCore/ConfigureTasks/index.tsx
```diff
@@ -139,6 +148,7 @@ const ConfigureTasks: React.VFC<Props> = ({
     selectedBuildVariants.forEach((v) => {
       if (selectedBuildVariantsCopy?.[v] !== undefined) {
         Object.keys(selectedBuildVariantsCopy[v]).forEach((task) => {
+          if (search !== "" && !task.includes(search)) return;
```
khelif96, line 151: I wonder if this should be a regex match or a normal match. But we can address that in a future ticket.

khelif96 (approved):

PR #1942, opened 2023-07-14
EVG-20050: Update lint-staged with GQL codegen step

SupaJoon:
EVG-20050
### Description
These code changes introduce a script/lint staged step that downloads graphql files from main, generates a gql types file and compares it to the local types file to help ensure the codegen task won't fail.

### Testing
Try making an edit to your local generated/types.ts file and try committing to see to test that it works


scripts/download-generate-and-diff-latest-schema-types.ts
```diff
@@ -0,0 +1,132 @@
...
+
+const downloadAndGenerate = async () => {
+  const tempDir = os.tmpdir();
+  fs.mkdirSync(tempDir, { recursive: true });
+  await fetchAndDownloadFiles(
+    path.join(REPO, INITIAL_PATH),
+    path.join(tempDir, INITIAL_PATH)
+  );
+  const latestGeneratedTypesFileName = `${tempDir}/types.ts`;
+  await generate(
```
sophstad, line 89: Would invoking `yarn codegen` via `exec` be more complex than this?

scripts/download-generate-and-diff-latest-schema-types.ts
```diff
@@ -0,0 +1,132 @@
...
+          writeStream.close();
+          resolve(true);
+        });
+
+        writeStream.on("error", reject);
+      })
+      .on("error", reject);
+  });
+
+const fetchAndDownloadFiles = (repoPath: string, localPath: string) =>
```
sophstad, line 33: [nit] This feels slightly redundant, `fetchFiles` or `downloadFiles` could probably work.

khelif96 (changes requested):
After giving it more thought, I believe this change will effectively block workflows where developers are simultaneously making changes to both the Evergreen repo and the Spruce repo. The problem lies in Spruce constantly trying to validate against the production GraphQL schema, making it impossible to make changes to both repositories concurrently.

During our round table discussion, I initially suggested approaching this as a process change rather than a code change. I apologize for not foreseeing this edge case when we agreed to try an automated method.

khelif96:
> > After giving it more thought, I believe this change will effectively block workflows where developers are simultaneously making changes to both the Evergreen repo and the Spruce repo. The problem lies in Spruce constantly trying to validate against the production GraphQL schema, making it impossible to make changes to both repositories concurrently.
> > During our round table discussion, I initially suggested approaching this as a process change rather than a code change. I apologize for not foreseeing this edge case when we agreed to try an automated method.
> 
> I actually thought about this too and still think a version of these changes are valuable given the number of times codegen has failed in CI. What do you think about passing this check if: a local `yarn codegen` doesn't produce changes AND the latest local commit to the local graphql/schema directory as greater than or equal to that in remote main. This is all possible with the GitHub API without having to download the repository. These changes really encourage the user to develop against the latest schema and keep their codegen files up to date. Let me know what you think!

How do you feel about just validating the checksum is newer than what is in on evergreen prod? think it would have the same effect of ensuring main is kept up to date.
  https://stackoverflow.com/questions/3005392/how-can-i-tell-if-one-commit-is-a-descendant-of-another-commit

khelif96 (changes requested):

khelif96 (approved):
Great job I'm impressed by how quick this is!

khelif96:
One thing I noticed is that this does not work if I'm offline. Can we update this script to exit 0 if a user is offline? Just to not completely prevent offline development.

scripts/download-generate-and-diff-latest-schema-types.ts
sophstad, line 1: Could you add some unit tests for this? I recently added a bunch for deploys that could be a useful reference!

sophstad (approved):

PR #1945, opened 2023-07-19
EVG-20005: Reusable settings form components

sophstad:
EVG-20005 - followup to #1941

### Description

- Introduce reusable components for adding context-dependent Spruce Forms. Eliminated a lot of repetition of populating forms, setting onChange handlers, etc. I added a base form component and a project settings-specific one. Distro settings will get its own.
- Remove duplicate type definition for `GetFormSchema`


src/components/Settings/Form.tsx
khelif96, line 1: Would love to see some tests for this and context

src/pages/projectSettings/tabs/PatchAliasesTab/PatchAliasesTab.tsx
```diff
@@ -32,31 +28,18 @@ export const PatchAliasesTab: React.VFC<TabProps> = ({
   projectType,
   repoData,
 }) => {
-  const { getTab, updateForm } = useProjectSettingsContext();
-  const { formData } = getTab(tab);
-
   const initialFormState = useMemo(
     () => getInitialFormState(projectData, repoData),
```
khelif96, line 32: Could we type getInitialFormState above?

src/pages/projectSettings/tabs/VariablesTab/VariablesTab.tsx
```diff
@@ -36,9 +34,6 @@ export const VariablesTab: React.VFC<TabProps> = ({
     () => getInitialFormState(projectData, repoData),
```
khelif96, line 34: Same comment as above. Can we type this?

src/pages/projectSettings/tabs/VirtualWorkstationTab/VirtualWorkstationTab.tsx
```diff
@@ -31,18 +27,12 @@ export const VirtualWorkstationTab: React.VFC<TabProps> = ({
   projectType,
   repoData,
 }) => {
-  const { getTab, updateForm } = useProjectSettingsContext();
-  const { formData } = getTab(tab);
-
   const initialFormState = useMemo(
     () => getInitialFormState(projectData, repoData),
```
khelif96, line 31: Can we type this above as well.

src/pages/projectSettings/tabs/BaseTab.tsx
```diff
@@ -0,0 +1,29 @@
...
+export const BaseTab = <T extends WritableProjectSettingsType>({
+  initialFormState,
+  tab,
+  ...rest
+}: BaseTabProps<T>) => {
+  const state = useProjectSettingsContext();
+  usePopulateForm(initialFormState, tab);
+
+  return (
+    <Form<WritableProjectSettingsType, FormStateMap>
```
khelif96, line 23: I've never seen this method of using generics in a react component before 👀

src/components/Settings/Form.tsx
```diff
@@ -0,0 +1,47 @@
+import {
+  SpruceForm,
+  ValidateProps,
+  GetFormSchema,
+} from "components/SpruceForm";
+import { SettingsState } from "./Context";
+import { SettingsRoutes } from "./types";
+
+type FormProps<
+  T extends SettingsRoutes,
```
khelif96, line 10: Sorry, I didn't get a chance to review the last pr. But is the implication that you would need to add the type from the settings page to the types directory here?

Could we make it more generic/ use generics so that this component doesn't need to rely on page-specific logic? 

I ask because I would imagine in the future once we add another settings page the types would likely mix up.

khelif96 (changes requested):
Great job with this refactor! Just had a few comments about testing and how it would scale once we add more settings pages.

src/components/Settings/Form.tsx
sophstad: Well we definitely are using generics here :) Adding another settings section can be done by extending the `SettingsRoutes` type with e.g. `WritableDistroSettingsType`.

https://github.com/evergreen-ci/spruce/blob/e9fd310365ed7cdc2886d122b4c63492a0fef59b/src/components/Settings/types.ts#L3

I've already done most of the work to set up distro settings for EVG-19940 and since [context is set with a specific type](https://github.com/evergreen-ci/spruce/blob/e9fd310365ed7cdc2886d122b4c63492a0fef59b/src/pages/projectSettings/Context.tsx#L18) there hasn't been any type collision.

I think this is the best approach because it allows us to enforce that other parts of the context like the map of form state types include all pages associated with settings since the typing is more specific than something like `Record<string, string>`. Plus, there is some overhead required to set up a new settings context, so adding a type to `SettingsRoutes` would not be likely to be overlooked.

sophstad: Or we can just set `type SettingsRoutes = string` so that `T extends string`? Let me know if you have a preference, I think the specificity could be slightly helpful. Ultimately it doesn't matter too much though since a more specific type is always supplied.

khelif96: That makes sense thanks for the explanation!

src/components/Settings/test-utils.tsx
```diff
@@ -0,0 +1,67 @@
...
+};
+
+const TestContext = createSettingsContext<TestRoutes, FormStateMap>();
+
+const TestProvider = ({ children }) => {
+  const { getTab, saveTab, setInitialData, tabs, updateForm } =
+    useSettingsState(["foo", "bar"], formToGqlMap);
+
+  return (
+    <TestContext.Provider
```
khelif96, line 39: I know this is being used in a test but could we maybe use the workaround here instead of ignoring the rule? https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-constructed-context-values.md

khelif96 (approved):
Great job just left one comment!

PR #1947, opened 2023-07-20
EVG-18913 Add lint rule to sort destructured keys

khelif96:
EVG-18913

### Description 
Adds key sorting for destructured objects


src/constants/externalResources.ts
```diff
@@ -117,14 +117,14 @@ export const getHoneycombSystemMetricsUrl = (
...
+      { column: "system.cpu.utilization", op: "AVG" },
+      { column: "system.network.io.transmit", op: "RATE_AVG" },
+      { column: "system.network.io.receive", op: "RATE_AVG" },
     ],
-    filters: [{ op: "=", column: "evergreen.task.id", value: taskId }],
-    start_time: getUnixTime(new Date(startTs)),
     end_time: getUnixTime(new Date(endTs)),
+    filters: [{ column: "evergreen.task.id", op: "=", value: taskId }],
+    start_time: getUnixTime(new Date(startTs)),
   };
```
khelif96, line 128: @ybrill This changed in this pr and the keys are sorted now. I tried in staging and it appears the links still work I just wanted to call it out just in case.

khelif96:
I just realized this sort-keys-plus will break our json forms so I will not be including it here.

src/constants/externalResources.ts
ybrill: I don't see the change. But I'm with you that the order shouldn't matter.

khelif96: I had to revert it the change broke some other things.

PR #1948, opened 2023-07-20
EVG-20307: Add Sentry breadcrumbs

sophstad:
EVG-20307

### Description

- Send breadcrumbs to Sentry. Some conversion was needed to adhere to their [recommended field names and types](https://develop.sentry.dev/sdk/event-payloads/breadcrumbs/#breadcrumb-types).

### Testing

- Add unit test


src/utils/errorReporting.ts
```diff
@@ -35,16 +36,99 @@ const reportError = (
...
+ * @param breadcrumbType - the type of Bugsnag breadcrumb to be sent
+ * @returns a custom Sentry breadcrumb type
+ */
+const convertBreadcrumbType = (
+  breadcrumbType: BreadcrumbType
+): SentryBreadcrumb => {
+  switch (breadcrumbType) {
+    case "error":
+      return SentryBreadcrumb.Error;
+    case "navigation":
```
minnakt, line 99: I noticed that the navigation breadcrumb has a particular requirement for its data field. Do we just have to remember to include these fields in the metadata?

Screenshot 2023-07-24 at 11 10 28 AM

src/utils/errorReporting.ts
```diff
@@ -35,16 +36,99 @@ const reportError = (
...
   message: string,
-  metadata: { [key: string]: any },
+  metadata: Metadata,
   type: BreadcrumbType
 ) => {
   if (!isProductionBuild()) {
     console.debug({ message, metadata, type });
   } else {
     Bugsnag.leaveBreadcrumb(message, metadata, type);
+    addBreadcrumb(convertToSentryBreadcrumb(message, metadata, type));
```
minnakt, line 52: does the timestamp field get auto-populated? :0 (asking bc `Bugsnag.leaveBreadcrumb` doesn't have a timestamp field but Sentry does so I was wondering if that means anything)

sophstad: I think without a `timestamp` field the time is populated with the time the error is sent? It's not very clear... but I'll add the field!

minnakt: either one is fine with me as long as its there!! =)

minnakt (approved):

PR #1951, opened 2023-07-21
EVG-20078: Expose merge queue section to production

minnakt:
EVG-20078

### Description
This PR removes the `isProduction()` flag from the merge queue section. Also removes the "beta" badge since the GitHub merge queue is out of beta.

### Testing
For some reason, removing the beta badge caused the e2e tests to fail... the only way I could fix it was removing the invalid objects from the Radio widget. So the GitHub documentation link unfortunately had to be removed.

Also cleaned up some assertions.

### DIFF
--- a/src/constants/externalResources.ts
+++ b/src/constants/externalResources.ts
@@ -39,7 +39,7 @@ export const konamiSoundTrackUrl =
   "https://www.myinstants.com/media/sounds/mvssf-win.mp3";
 
 export const githubMergeQueueUrl =
-  "https://github.blog/changelog/2023-02-08-pull-request-merge-queue-public-beta";
+  "https://github.blog/2023-07-12-github-merge-queue-is-generally-available";
 
 export const legacyRoutes = {
   distros: "/distros",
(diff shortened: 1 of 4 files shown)


src/constants/externalResources.ts
```diff
@@ -38,9 +38,6 @@ export const getJiraImprovementUrl = (jiraHost: string) =>
 export const konamiSoundTrackUrl =
   "https://www.myinstants.com/media/sounds/mvssf-win.mp3";
 
-export const githubMergeQueueUrl =
```
khelif96, line 41: Should we include the documentation before we launch to reduce any potential confusion? @bsamek

khelif96 (approved):

src/constants/externalResources.ts
bsamek: Pushed to https://docs.devprod.prod.corp.mongodb.com/evergreen/Project-Configuration/Merge-Queue (will appear in a few minutes).

minnakt:
merging bc test is failing on main

PR #1956, opened 2023-07-25
EVG-19037: Upgrade SearchInput

sophstad:
EVG-19037

### Description

- Upgrade `SearchInput` to the latest version
- We don't have any uses where it would make sense to use the `SearchResult`/typeahead features of this release. The files table, for example, pretty much already implements this behavior in the table itself.
- The ticket suggested converting some `TextInput` to `SearchInput` to remove unit test warnings. However, all of the existing `type="search"` inputs are not convertible because they either use a prop not offered by `SearchInput`, like `label` or `description`, or use icons in a way that's better suited to an empty input. Too bad because those warnings are annoying!

### Testing
- Update Cypress tests because `SearchInput` now uses an `<input>` nested inside of a `<form>`, which is where the `data-cy` attribute is attached.

### DIFF
--- a/cypress/integration/myPatches/my_patches.ts
+++ b/cypress/integration/myPatches/my_patches.ts
@@ -34,7 +34,9 @@ describe("My Patches Page", () => {
   it("Typing in patch description input updates the url, requests patches and renders patches", () => {
     cy.visit(MY_PATCHES_ROUTE);
     const inputVal = "testtest";
-    cy.dataCy("patch-description-input").type(inputVal);
+    cy.dataCy("patch-description-input").within(() => {
+      cy.get("input").type(inputVal);
+    });
     urlSearchParamsAreUpdated({
       pathname: MY_PATCHES_ROUTE,
       paramName: "patchName",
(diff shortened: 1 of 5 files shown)


cypress/integration/myPatches/my_patches.ts
```diff
@@ -34,7 +34,9 @@ describe("My Patches Page", () => {
   it("Typing in patch description input updates the url, requests patches and renders patches", () => {
     cy.visit(MY_PATCHES_ROUTE);
     const inputVal = "testtest";
-    cy.dataCy("patch-description-input").type(inputVal);
+    cy.dataCy("patch-description-input").within(() => {
+      cy.get("input").type(inputVal);
```
khelif96, line 38: That's an annoying side effect. :/  Is this intentional on leafygreens end?

khelif96 (approved):

cypress/integration/myPatches/my_patches.ts
sophstad: I think so in order to group all the related elements of the form together, including any dropdown options being used. It's too bad it's included when none are used, though.

PR #1957, opened 2023-07-25
EVG-20512 Refactor deploy email to use same process as local and ensure email includes necessary commits

khelif96:
EVG-20512

### Description
Something broke with how we are sending deploy emails and it appeared  CI deploys were not getting all of the environmental variables necessary to generate the email. I updated the script and now things should perform the same actions for both local deploys and CI deploys.

### DIFF
--- a/scripts/deploy-production.js
+++ b/scripts/deploy-production.js
@@ -1,14 +1,13 @@
 const prompts = require("prompts");
 const {
   createNewTag,
-  deleteTag,
   getCommitMessages,
   getCurrentlyDeployedCommit,
-  getLatestTag,
   isOnMainBranch,
   isWorkingDirectoryClean,
-  pushTags,
-  runLocalDeploy,
+  deleteAndPushLatestTag,
+  isRunningOnCI,
+  runDeploy,
 } = require("./deploy-utils");
 
 /* Deploy by pushing a git tag, to be picked up and built by Evergreen, and deployed to S3. */
@@ -68,7 +67,7 @@ const evergreenDeploy = async () => {
       );
     } catch (err) {
       console.log(err);
-      console.log("Creating tag failed. Aborting.");
+      throw new Error("Creating tag failed. Aborting.");
     }
   }
 };
@@ -93,25 +92,29 @@ const localDeploy = async () => {
 
   if (response.value) {
     try {
-      const localDeployOutput = await runLocalDeploy();
+      const localDeployOutput = await runDeploy();
       console.log(localDeployOutput);
     } catch (err) {
-      console.error(err);
       console.error("Local deploy failed. Aborting.");
+      throw new Error(err);
     }
   }
 };
 
-const deleteAndPushLatestTag = async () => {
+const ciDeploy = async () => {
+  if (!(await isOnMainBranch())) {
+    throw new Error("You must be on the main branch to deploy!");
+  }
+  if (!isRunningOnCI()) {
+    throw new Error("Not running on CI");
+  }
   try {
-    const latestTag = await getLatestTag();
-    console.log(`Deleting and re-pushing latest tag (${latestTag})`);
-    console.log(await deleteTag(latestTag));
-    console.log(await pushTags());
+    const ciDeployOutput = await runDeploy();
+    console.log(ciDeployOutput);
   } catch (err) {
-    console.error(err);
-    console.error("Deleting and pushing tag failed. Aborting.");
+    console.error("CI deploy failed. Aborting.");
+    throw new Error(err);
   }
 };
 
-module.exports = { evergreenDeploy, localDeploy };
+module.exports = { evergreenDeploy, localDeploy, ciDeploy };
--- a/scripts/deploy-utils.js
+++ b/scripts/deploy-utils.js
@@ -1,13 +1,14 @@
 const { exec } = require("child_process");
 
 const githubRemote = "https://github.com/evergreen-ci/spruce";
-const localDeployScript =
+const deployScript =
   "yarn build:prod && env-cmd -e production yarn deploy:do-not-use && env-cmd -e production ./scripts/email.sh";
 
 const createNewTag = () =>
   new Promise((resolve, reject) => {
     exec("yarn version --new-version patch", (err, stdout) => {
       if (err) {
+        console.error(stdout);
         reject(err);
         return;
       }
@@ -21,6 +22,7 @@ const getCommitMessages = (currentlyDeployedCommit) =>
       `git log ${currentlyDeployedCommit}..HEAD --oneline`,
       (err, stdout) => {
         if (err) {
+          console.error(stdout);
           reject(err);
           return;
         }
@@ -33,7 +35,7 @@ const getCurrentlyDeployedCommit = () =>
   new Promise((resolve, reject) => {
     exec("bash scripts/get-current-deployed-commit.sh", (err, stdout) => {
       if (err) {
-        console.log(err);
+        console.error(stdout);
         reject(err);
         return;
       }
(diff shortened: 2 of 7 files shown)


scripts/deploy-production.js
```diff
@@ -68,7 +67,7 @@ const evergreenDeploy = async () => {
       );
     } catch (err) {
       console.log(err);
-      console.log("Creating tag failed. Aborting.");
+      throw new Error("Creating tag failed. Aborting.");
```
khelif96, line 70: Throw an error so we get proper exit codes.

scripts/deploy-utils.js
```diff
@@ -1,13 +1,14 @@
 const { exec } = require("child_process");
 
 const githubRemote = "https://github.com/evergreen-ci/spruce";
-const localDeployScript =
+const deployScript =
   "yarn build:prod && env-cmd -e production yarn deploy:do-not-use && env-cmd -e production ./scripts/email.sh";
```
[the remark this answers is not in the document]
khelif96: Refactored this so that these are all separate commands

scripts/deploy/deploy-utils.ts
```diff
@@ -0,0 +1,121 @@
...
+  console.log("BUILDING");
+  execSync("yarn build:prod", { stdio: "inherit" });
+  console.log("BUILD COMPLETE");
+  console.log("SAVING CURRENT COMMIT in build/commit.txt");
+  const currentCommit = getCurrentCommit();
+  execSync(`echo ${currentCommit} > build/commit.txt`, {
+    stdio: "inherit",
+  });
+  console.log("DEPLOYING");
+  if (!isDryRun) {
```
sophstad, line 93: This approach feels kinda brittle to me since it's dependent on including a conditional in every util function. It would be pretty easy to forget this check in a new function or change and the whole concept would break.

Since the intent is primarily to confirm that the deploy is wired up correctly without calling any "impactful" functions, here's one idea: we could introduce a file that mocks each of these functions (and the tag utils) for a dry run. Then, when exporting the utils, if `isDryRun`, the mock utils can be exported; if not, export the actual functions. This would hopefully reduce the number of failure points. Open to other ideas too!

scripts/deploy/deploy.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+# This script runs the aws cli command to deploy the app to s3
+# It also uploads source maps to bugsnag
+
+# Try this step and throw an error if it fails
```
sophstad, line 6: Does `aws` throw an error if the operation fails? I noticed the script doesn't implement that.

scripts/deploy/get-current-deployed-commit.sh
```diff
@@ -1,11 +1,7 @@
 #!/bin/bash
 
-BASE_URL=$REACT_APP_SPRUCE_URL
-if [ "$BASE_URL" == "" ]
-then
-    echo "BASE_URL not set"
-    exit 1
-fi
+BASE_URL=https://spruce.mongodb.com
```
sophstad, line 3: Ah, was an undefined `REACT_APP_SPRUCE_URL` the source of the problems on prod?

sophstad (changes requested):

.gitignore
```diff
@@ -35,4 +35,6 @@ sdlschema
 sdlschema.graphql
 .eslintcache
 
+body.txt
+body.txt-e
```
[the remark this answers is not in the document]
khelif96: These are files generated by the deploy script for sending an email. They will be uploaded as part of the deploy task so we can have a copy of the email. I added them to the gitignore so they don't get accidentally committed.

scripts/deploy/get-current-deployed-commit.sh
khelif96: yes the env variables wouldn't always be populated and it was random and difficult to reproduce. So I figured might as well remove the points of failure.

.gitignore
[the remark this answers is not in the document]
sophstad: Right I see we write to `body.txt`, but what's `body.txt-e`?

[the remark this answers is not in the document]
khelif96: Thats the big question. I'm not too sure what it is and I haven't been able to find anything related to it.

scripts/deploy/utils/deploy/index.ts
```diff
@@ -0,0 +1,43 @@
...
+ * It builds the production bundle, deploys it to the production server, and sends an email.
+ */
+const runDeploy = () => {
+  console.log(green("Getting currently deployed commit"));
+  const currentlyDeployedCommit = getCurrentlyDeployedCommit();
+  console.log(currentlyDeployedCommit);
+  console.log("BUILDING");
+  execSync("yarn build:prod", { stdio: "inherit" });
+  console.log("BUILD COMPLETE");
+  console.log("SAVING CURRENT COMMIT in build/commit.txt");
```
sophstad, line 17: Bump on caps lock

scripts/deploy/utils/git/index.ts
```diff
@@ -0,0 +1,77 @@
...
+    .trim();
+  return currentCommit;
+};
+
+/**
+ * `isOnMainBranch` is a helper function that checks if the current branch is the main branch.
+ * @returns true if the current branch is the main branch
+ */
+const isOnMainBranch = () => {
+  console.log("Checking if you are on the main branch");
```
sophstad, line 51: [nit] It seems like for all of the other util functions, we log this preemptive update outside of the function before it is called. Could we do the same here?

scripts/deploy/utils/tag/tag-utils.ts
sophstad, line 1: All of these utils are git-related too, might it be more straightforward to combine the `git` and `tag` util files?

scripts/deploy/run-deploy.ts
```diff
@@ -0,0 +1,31 @@
...
+    if (!isWorkingDirectoryClean()) {
+      console.log(red("You must have a clean working directory to deploy"));
+      process.exit(1);
+    }
+  }
+  if (process.argv.includes("--local")) {
+    await localDeploy();
+    return;
+  }
+  if (process.argv.includes("--ci")) {
```
sophstad, line 24: Is `isRunningOnCI()` insufficient here?

scripts/deploy/utils/deploy/index.ts
```diff
@@ -0,0 +1,43 @@
...
+import { green } from "../../../utils/colors";
+import { getCurrentCommit, getCurrentlyDeployedCommit } from "../git";
+import { isDryRun } from "../environment";
+
+/**
+ * `runDeploy` is a helper function that actually performs the deploy.
+ * It builds the production bundle, deploys it to the production server, and sends an email.
+ */
+const runDeploy = () => {
+  console.log(green("Getting currently deployed commit"));
```
sophstad, line 11: [nit] Green seems unnecessary here, the message isn't indicating any kind of success.

scripts/deploy/utils/tag/tag-utils.ts
khelif96: Done.

scripts/deploy/run-deploy.ts
khelif96: isRunningOnCI checks for the CI env variable that evergreen injects while this checks for the `--ci` flag. I figured it can be an extra safety check that it's both actually a CI build and we are intentionally triggering the CI path. (It also allows us to dry run the CI deploy this way)

sophstad: I see, my personal take is that defining whether we're on CI in 2 different ways is a bit confusing. We could also do a CI dry run by running `CI=true yarn deploy --dry-run`, no? But if this is strongly your preference that's okay with me too!

khelif96: Good point done!

sophstad (approved):
Excited to try this out!

PR #1958, opened 2023-07-26
EVG-20523 Bump caniuse-lite

khelif96:
EVG-20523

### Description
Noticed a warning when starting the dev server about this being out of date.


khelif96:
> package.json is missing

@SupaJoon there is no package.json associated with this change it is used by another dependency I am simply resolving the warning

PR #1963, opened 2023-07-27
EVG-20126: Use react-virtuoso in VersionRestartModal

minnakt:
EVG-20126

### Description
This PR:
* Replaces `FixedSizeList` with `Virtuoso`
* Removes `react-window` from dependencies
* Removes styles that were unnecessary + updates styles
* Alphabetizes props

### Screenshots
(Looks the same pretty much)

https://github.com/evergreen-ci/spruce/assets/47064971/9ad508e4-7807-4dbe-a256-d83c9d491851

### Testing
- Updated e2e tests

### DIFF
--- a/src/components/VersionRestartModal/TaskStatusCheckboxContainer.tsx
+++ b/src/components/VersionRestartModal/TaskStatusCheckboxContainer.tsx
@@ -1,17 +1,17 @@
-import { FixedSizeList } from "react-window";
+import { Virtuoso } from "react-virtuoso";
 import { selectedStrings } from "hooks/useVersionTaskStatusSelect";
 import { TaskStatusCheckbox } from "./TaskStatusCheckbox";
 
 interface TaskStatusCheckboxContainerProps {
-  versionId: string;
+  selectedTasks: selectedStrings;
   tasks: {
     id: string;
     status: string;
     baseStatus?: string;
     displayName: string;
   }[];
-  selectedTasks: selectedStrings;
   toggleSelectedTask: (taskIds: { [patchId: string]: string }) => void;
+  versionId: string;
 }
 export const TaskStatusCheckboxContainer: React.VFC<
   TaskStatusCheckboxContainerProps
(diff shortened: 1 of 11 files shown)


src/components/VersionRestartModal/VersionRestartModal.tsx
```diff
@@ -207,26 +201,25 @@ const VersionTasks: React.VFC<VersionTasksProps> = ({
...
-        />
-      ))}
+      <TaskStatusFilters
+        onChangeBaseStatusFilter={setBaseStatusFilterTerm}
+        onChangeStatusFilter={setVersionStatusFilterTerm}
+        versionId={version?.id}
+        selectedBaseStatuses={baseStatusFilterTerm || []}
+        selectedStatuses={versionStatusFilterTerm || []}
+      />
+      {[...buildVariants]
```
khelif96, line 211: Why are we spreading buildVariants into a new array?

src/components/VersionRestartModal/VersionRestartModal.tsx
```diff
@@ -207,26 +201,25 @@ const VersionTasks: React.VFC<VersionTasksProps> = ({
...
+        .sort((a, b) => a.displayName.localeCompare(b.displayName))
+        .map((patchBuildVariant) => (
+          <BuildVariantAccordion
+            versionId={version?.id}
```
khelif96, line 215: nit unnecessary optional chaining. If buildVariants exists so should id.

src/components/VersionRestartModal/TaskStatusCheckboxContainer.tsx
```diff
@@ -1,57 +1,50 @@
...
-    const { name } = e.target;
-    if (selectedTasks[name] !== undefined) {
-      toggleSelectedTask({ [versionId]: name });
-    }
-  };
+
   return (
-    // eslint-disable-next-line jsx-a11y/no-static-element-interactions,jsx-a11y/click-events-have-key-events
-    <div data-cy="patch-status-selector-container" onClick={toggleHandler}>
-      <FixedSizeList
```
khelif96, line 31: Great refactor

khelif96 (commented):
Great job with this refactor! Just left a few comments and a question

src/components/VersionRestartModal/VersionRestartModal.tsx
minnakt: the [`sort`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort) function sorts arrays in place so if the array is immutable an error will be thrown, and I was seeing the page crash because of it

It would be better to use [`toSorted`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSorted), but this is a really new function that doesn't have widespread support on browsers yet

khelif96 (approved):
Thanks for humoring me with the file changes

src/components/VersionRestartModal/VersionRestartModal.tsx
khelif96: discussed offline. This is supported by nearly all of our browsers but we might need to upgrade some of our tooling to use this given how new this feature is.

PR #1982, opened 2023-08-02
EVG-20577 Use correct file name for previous deploy file

khelif96:
EVG-20577

### Description
We look for a previous_deploy.txt but we were naming it previously-deployed-commit.txt


sophstad (approved):
Let's resolve the codegen failure here, otherwise lgtm!

khelif96:
> Let's resolve the codegen failure here, otherwise lgtm!

It will be resolved once https://github.com/evergreen-ci/spruce/pull/1967 is merged. (There are some type changes that needed to happen)

PR #1990, opened 2023-08-14
EVG-20617 Only upload to cypress if on a pr, mainline commit, or patch

khelif96:
EVG-20617

### Description
Following up on our conversation this should reduce unnecessary uploads to the Cypress cloud service.

### DIFF
--- a/.evergreen.yml
+++ b/.evergreen.yml
@@ -229,7 +229,7 @@ functions:
         ${PREPARE_SHELL}
         # Only record to cypress cloud if this is a pr or a mainline commit.
         echo "${requester}"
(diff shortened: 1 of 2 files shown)


.evergreen.yml
```diff
@@ -225,10 +225,16 @@ functions:
     command: shell.exec
     params:
       working_dir: spruce
+      shell: "bash"
```
minnakt, line 228: not sure if bash needs to be in quotes
```suggestion
      shell: bash
```

minnakt (approved):
remember to change the commit title to include patch!

PR #2000, opened 2023-08-18
EVG-20065: Read "success" patch status

sophstad:
EVG-20065

### Description

- Update value for `Success` status and rename `Succeeded` to `LegacySucceeded`. Use `LegacySucceeded` wherever the `Success` status is in use (namely in the patch filter dropdown).
- For patch trigger aliases, accept either "success" or "succeeded" status and save field as "succeeded"

### Testing

- Update unit test to ensure that "success" is correctly interpreted as a patch trigger alias status


src/pages/projectSettings/tabs/utils/types.ts
```diff
@@ -10,5 +10,6 @@ export enum ProjectType {
 export const PatchTriggerAliasStatus = {
```
minnakt, line 10: are the two "success" options intentional / temporary?
Screenshot 2023-08-20 at 12 11 46 PM

src/components/PatchesPage/StatusSelector.tsx
```diff
@@ -56,9 +56,9 @@ const treeData = [
     key: PatchStatus.Started,
   },
   {
-    title: statusValToCopy[PatchStatus.Success],
-    value: PatchStatus.Success,
-    key: PatchStatus.Success,
+    title: statusValToCopy[PatchStatus.LegacySucceeded],
+    value: PatchStatus.LegacySucceeded,
+    key: PatchStatus.LegacySucceeded,
   },
```
minnakt, line 62: Does this mean that the filter won't work for newer patches with the `"success"` status? I think one workaround could to include both the legacy and new success statuses when we [query for the patches](https://github.com/evergreen-ci/spruce/blob/07fc78fe676d76b86ebdddf648c7769ae0b3a39a/src/pages/UserPatches.tsx#L37-L40)

src/types/patch.ts
```diff
@@ -11,7 +11,8 @@ export enum PatchStatus {
   Created = "created",
   Failed = "failed",
   Started = "started",
-  Success = "succeeded",
+  LegacySucceeded = "succeeded",
+  Success = "success",
```
minnakt, line 15: this might be more of a question for annie rather than for you haha but just want to double check that `"success"` is expected to be the new status? EVG-20032 says that we are going to use `"succeeded"` for both patches and versions but based on this PR it sounds like it's the opposite? 🤔

src/components/PatchesPage/StatusSelector.tsx
sophstad: Handling both is going to be included on the backend! https://jira.mongodb.org/browse/EVG-19914?focusedCommentId=5626989&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-5626989

src/types/patch.ts
sophstad: Good point, @ablack12 `success` is the new status, correct? Should EVG-20032 be updated?

ablack12: Ah yeah that ticket description is old, I'll update it.

src/components/PatchesPage/StatusSelector.tsx
```diff
@@ -56,9 +56,9 @@ const treeData = [
     key: PatchStatus.Started,
   },
   {
-    title: statusValToCopy[PatchStatus.Success],
-    value: PatchStatus.Success,
-    key: PatchStatus.Success,
+    title: statusValToCopy[PatchStatus.LegacySucceeded],
```
minnakt, line 59: Just one last question: is this also a TODO thing that will need to be updated in a year? If so, would note in the ticket so it doesn't get lost :o

minnakt (approved):

src/components/PatchesPage/StatusSelector.tsx
sophstad: Left a comment on EVG-20032 that `PatchStatus.LegacySucceeded` should be removed! Doing so will ensure that this is updated.

PR #2003, opened 2023-08-23
EVG-20543: Hide merge method if GitHub merge queue is enabled

minnakt:
EVG-20543

### Description
This PR makes it so that the merge method input is hidden if the GitHub merge queue is selected. Setting the merge method here is not meaningful because it must be set in the GitHub UI.

Although the input is not visible, the merge method will be preserved. This is so that if a project switches back to the Evergreen commit queue, they won't have to configure this setting again.

Feel free to suggest any changes to the copy!

### Screenshots

https://github.com/evergreen-ci/spruce/assets/47064971/c57a2615-0355-43a9-96b7-ee796e4523b1

### Testing
- Cypress tests


src/pages/projectSettings/tabs/GithubCommitQueueTab/getFormSchema.tsx
```diff
@@ -212,60 +213,83 @@ export const getFormSchema = (
...
                           type: "string" as "string",
-                          title: "Merge",
-                          enum: ["merge"],
+                          oneOf: [
+                            {
+                              type: "string" as "string",
+                              title: "Evergreen",
+                              enum: [MergeQueue.Evergreen],
+                              description:
+                                "The standard commit queue owned and maintained by Evergreen.",
```
bsamek, line 232:
```suggestion
                                "Evergreen's commit queue.",
```
I'd suggest at least striking "standard", since the goal is to replace it with GitHub's, making that one the standard one. I suggested a simpler phrase to mirror the GitHub one.

minnakt: done! I wonder if I should say anything about the CLI only working with the evg one? :0

bsamek: I think that because there's a button in the GitHub UI, and also it's explained in the linked docs, that this is enough.

bsamek (approved):

PR #2005, opened 2023-08-23
EVG-20360: Remove redundant test from commit_queue.ts

minnakt:
EVG-20360

### Description
This test is failing on my Evergreen PR because it's trying to remove a patch from the Logkeeper commit queue, but the Logkeeper project does not have the commit queue enabled. There's already another test for removing from the commit queue in this same file, so I think it's okay to remove. It's not really meaningful to test it twice because the admin user is a superuser and can remove anything they want.

I'm planning to remove this patch data since it doesn't make sense that a patch is on a disabled commit queue.

### Evergreen PR
* evergreen-ci/evergreen/pull/6946


minnakt:
evergreen refresh

minnakt:
I don't know why the status checks aren't posting, but here's [my patch](https://spruce.mongodb.com/version/64e769d1c9ec443d0290b435/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) and everything passed, so I am going to merge this

PR #2017, opened 2023-09-01
EVG-19946: Support Docker on provider settings page

minnakt:
EVG-19946

### Description
This PR adds Docker to the provider settings page. The Docker configuration is kind of weird (container pool ID is not part of provider settings list) so I made some changes in the form schema and conversion functions.

### Screenshots
Screenshot 2023-09-01 at 11 00 13 AM

### Testing
- Cypress tests, Jest tests

### Evergreen PR
* evergreen-ci/evergreen/pull/6984


src/pages/distroSettings/tabs/ProviderTab/transformers.ts
```diff
@@ -1,22 +1,37 @@
...
+          ...staticProviderSettings(providerSettingsList[0]).form,
+        },
+      };
+    case Provider.Docker:
+      return {
+        provider: {
+          providerName: Provider.Docker,
+        },
+        providerSettings: {
+          ...dockerProviderSettings(providerSettingsList[0]).form,
```
sophstad, line 32: I switched the `localhost` distro to use Docker and then tried to switch it back to static and saw this error:
image

It seems like there's validation [here](https://github.com/evergreen-ci/evergreen/blob/00e0aa6f50e98304d251b3afa1108840b6538641/model/distro/distro.go#L550-L552) preventing container pool distros from using a different provider.

Maybe someone on app would have insight into the best approach here. Do we want to ban Docker distros from having their types changed? Or do we maybe want to update the save function to remove this distro from the list of Docker distros? 🤔

src/pages/distroSettings/tabs/ProviderTab/transformers.ts
```diff
@@ -29,13 +44,24 @@ export const formToGql = ((data, distro) => {
...
+    case Provider.Docker:
       return {
         ...distro,
-        provider: providerName,
-        ...staticProviderSettings(data.providerSettings).gql,
+        provider: Provider.Docker,
+        providerSettingsList: [
+          ...dockerProviderSettings(data.providerSettings).gql,
+        ],
+        // @ts-ignore-error - containerPoolId will exist in DockerFormState.
```
sophstad, line 62: We can narrow down the type here using the `in` operator to avoid this error ([docs](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#the-in-operator-narrowing)). Something like this:
```js
    case Provider.Docker: {
      if ("containerPoolId" in data.providerSettings) {
        return {
          ...distro,
          provider: Provider.Docker,
          providerSettingsList: [
            ...dockerProviderSettings(data.providerSettings).gql,
          ],
          containerPool: data.providerSettings.containerPoolId,
        };
      }
      break;
    }
```

sophstad (changes requested):

src/pages/distroSettings/tabs/ProviderTab/transformers.ts
minnakt: oh that's probably my fault. Have to clear the `containerPool` field in the transformers

minnakt:
hello hello, so after experimenting a bit with the EC2 providers, I ended up concluding that maybe we should just have unique fields for each provider (rather than trying to share the `providerSettings` field across all four providers). This solution will require us to redeclare some fields, but I think it might be simpler overall.

I first tried a more "elegant" solution where the `providerSettings` field would be an array in the schema, and we would populate the array with the appropriate amount of items (1 item for Static, 1 item for Docker, 4 items for EC2). This didn't work out so well due to a specific scenario: 
1. Start out with EC2 provider with 4 items in the array (each item corresponds to an AWS region)
2. Switch to the Static provider 
3. You'd still have 4 items in the array, which is wrong for Static providers

And this is why I ended up thinking that it would be simpler for each of the providers to have their own fields. Lmk what you think! 😲

src/pages/distroSettings/tabs/ProviderTab/types.ts
```diff
@@ -1,18 +1,31 @@
 import { Provider } from "gql/generated/types";
 
-interface StaticProviderFormState {
+export enum BuildType {
+  Import = "import",
+  Pull = "pull",
+}
+
+// TODO: Append type with additional provider options, e.g. type ProviderFormState = StaticProviderFormState | DockerProviderFormState
```
sophstad, line 8: [nit] I think this no longer applies in the way it's written here, could update or just remove.

src/pages/distroSettings/tabs/ProviderTab/ProviderTab.tsx
```diff
@@ -1,12 +1,34 @@
...
+import { useSpruceConfig } from "hooks";
+import { useDistroSettingsContext } from "pages/distroSettings/Context";
+import { omitTypename } from "utils/string";
 import { BaseTab } from "../BaseTab";
 import { getFormSchema } from "./getFormSchema";
-import { TabProps } from "./types";
+import { TabProps, ProviderFormState } from "./types";
 
 export const ProviderTab: React.FC<TabProps> = ({ distroData }) => {
   const initialFormState = distroData;
```
sophstad, line 11: [nit] No need to rename this var (even though I think we have been doing it unnecessarily in every tab 😂)

src/pages/distroSettings/tabs/ProviderTab/transformers.test.ts
```diff
@@ -1,39 +1,127 @@
...
-  provider: Provider.Static,
-  providerSettingsList: [
-    {
-      merge_user_data_parts: false,
-      security_group_ids: ["1"],
-      user_data: "",
-    },
-  ],
-};
+  describe("docker provider", () => {
```
sophstad, line 63: Would you mind adding a test that uses the `poolMappingInfo` field?

src/pages/distroSettings/tabs/ProviderTab/types.ts
```diff
@@ -1,18 +1,31 @@
...
-}
-
-// TODO: Append type with additional provider options, e.g. type ProviderFormState = StaticProviderFormState | DockerProviderFormState
-export type ProviderFormState = StaticProviderFormState;
+  dockerProviderSettings: {
+    imageUrl: string;
+    buildType: BuildType;
+    registryUsername: string;
+    registryPassword: string;
+    containerPoolId: string;
```
sophstad, line 23: It seems like `poolMappingInfo` could also be a field here no?

src/pages/distroSettings/tabs/ProviderTab/transformerUtils.ts
```diff
@@ -1,119 +1,52 @@
...
-})) satisfies FieldGetter;
+});
 
-const getRegistryPassword = ((providerSettings) => ({
-  form: {
-    registryPassword: providerSettings.docker_registry_pw ?? "",
+type ProviderSettings = ProviderFormState["staticProviderSettings"] &
+  ProviderFormState["dockerProviderSettings"];
+
+export const gqlProviderSettings = (
```
sophstad, line 35: This structure makes a lot more sense 😅 gorgeous!

sophstad (approved):

PR #2025, opened 2023-09-07
EVG-20173: Allow users to log out via user dropdown

minnakt:
EVG-20173

### Description
Adds a "Log out" button under the user menu.

I changed the `axios` calls to `fetch` since that seems to be what we prefer. (I didn't remove the `axios` dependency since there's still one other part of the codebase that uses it.)

### Screenshots

Screenshot 2023-09-07 at 4 20 29 PM

### Testing
* Added Cypress and Jest tests
* Manual testing locally and on staging
* How to test: open an incognito window, log in, then log out


khelif96 (approved):
💯

Is there a ticket to get rid of axios?

minnakt:
> 💯
> 
> Is there a ticket to get rid of axios?

there is now :shipit: [EVG-20849](https://jira.mongodb.org/browse/EVG-20849)

PR #2035, opened 2023-09-12
EVG-20177: Show Admin button to EVG admins

SupaJoon:
EVG-20177

### Description
Show the admin button when the `canUpdateAdminSettings` field is true
Relies on https://github.com/evergreen-ci/evergreen/pull/7025

### Screenshots

Screenshot 2023-09-14 at 3 42 31 PM

### DIFF
--- a/cypress/integration/nav_bar.ts
+++ b/cypress/integration/nav_bar.ts
@@ -1,4 +1,4 @@
-import { EVG_BASE_URL } from "../utils";
+import { EVG_BASE_URL } from "../constants";
 
 const PATCH_ID = "5e4ff3abe3c3317e352062e4";
 const USER_ID = "admin";
--- a/cypress/utils/index.ts
+++ b/cypress/utils/index.ts
@@ -72,5 +72,3 @@ export const clickSave = () => {
     .should("not.have.attr", "aria-disabled", "true")
     .click();
 };
-
-export const EVG_BASE_URL = "http://localhost:9090";
--- a/cypress/constants/index.ts
+++ b/cypress/constants/index.ts
@@ -1,0 +1,2 @@
+export const EVG_BASE_URL = "http://localhost:9090";
+export const GQL_URL = `${EVG_BASE_URL}/graphql/query`;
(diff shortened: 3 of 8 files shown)


cypress/integration/nav_bar.ts
```diff
@@ -100,4 +104,31 @@ describe("Nav Bar", () => {
...
+            permissions: {
+              canEditAdminSettings: false,
+            },
+          },
+        },
+      };
+      cy.overwriteGQL("User", userData);
+      cy.visit(SPRUCE_URLS.version);
+      cy.dataCy("admin-link").should("not.exist");
+    });
```
khelif96, line 125: As much as I'm not a fan of mocking requests in e2e tests, I understand why. I think this is a good motivation for us to begin thinking about having multiple users for e2e tests in the future. It will allow us to more accurately test permission things.

src/components/Header/Navbar.tsx
```diff
@@ -82,6 +82,11 @@ export const Navbar: React.FC = () => {
         >
           My Hosts
         </PrimaryLink>
+        {permissions?.canEditAdminSettings && (
+          <PrimaryLink data-cy="admin-link" to={adminSettingsURL}>
```
khelif96, line 86: This should use PrimaryA instead of PrimaryLink since this is an external link.

src/components/Header/Navbar.tsx
```diff
@@ -82,6 +82,11 @@ export const Navbar: React.FC = () => {
         >
           My Hosts
         </PrimaryLink>
+        {permissions?.canEditAdminSettings && (
```
khelif96, line 85: Also I don't think this should be a navbar item. I think it should be hidden in the dropdown menu like it is in the legacy UI.

src/components/Header/UserDropdown.tsx
```diff
@@ -1,20 +1,21 @@
...
   const { data } = useQuery<UserQuery>(GET_USER);
   const { user } = data || {};
-  const { displayName } = user || {};
+  const { displayName, permissions } = user || {};
 
   const { logoutAndRedirect } = useAuthDispatchContext();
   const { sendEvent } = useNavbarAnalytics();
 
-  const menuItems = [
+  const menuItems: MenuItemType[] = [
```
khelif96, line 18: nice.

cypress/utils/index.ts
```diff
@@ -72,3 +72,5 @@ export const clickSave = () => {
     .should("not.have.attr", "aria-disabled", "true")
     .click();
 };
+
+export const EVG_BASE_URL = "http://localhost:9090";
```
khelif96, line 76: Can we create a constants file for this and other cypress constants?

khelif96 (approved):
LGTM mod comments

PR #2038, opened 2023-09-13
EVG-20677 & EVG-20868: Make type.ts validation more robust

SupaJoon:
EVG-20677
EVG-20868

### Description
EVG-20677 was solved by not erroring on code 128 from `git merge-base --is-ancestor`. All it means is that the commit has never been fetched by the developers git program. 
EVG-20868 contains an overview of the rest of the code changes.
I made sure to fail the script only if `types.ts` is verified as stale. I pass the script for all other errors for the sake of developer productivity and also because this validation overlaps with `check_codegen`.
I used `git mv` for `diff-local-schema-with-remote` -> `check-schema-and-codegen` but GitHub didn't do a good job at lining up the diff 😬 .

### DIFF
--- a/scripts/check-schema-and-codegen/utils.ts
+++ b/scripts/check-schema-and-codegen/utils.ts
@@ -48,6 +48,7 @@ export const checkIsAncestor = async (commit: string): Promise<boolean> => {
   } catch (error) {
     process.chdir(originalDir);
     // Error status 1 and 128 means that the commit is not an anecestor and the user must fetch.
+    // Error code docs: https://www.git-scm.com/docs/api-error-handling/
     if (error.status === 1 || error.status === 128) {
       return false;
     }
--- a/src/gql/generated/types.ts
+++ b/src/gql/generated/types.ts
@@ -1531,6 +1531,7 @@ export type Permissions = {
   __typename?: "Permissions";
   canCreateDistro: Scalars["Boolean"]["output"];
   canCreateProject: Scalars["Boolean"]["output"];
+  canEditAdminSettings: Scalars["Boolean"]["output"];
   distroPermissions: DistroPermissions;
   userId: Scalars["String"]["output"];
 };


codegen.ts
```diff
@@ -3,35 +3,35 @@ import path from "path";
 
 export const getConfig = ({
```
khelif96, line 4: Sorry I know this was here but do you know why this is exported twice?

scripts/check-schema-and-codegen/utils.ts
```diff
@@ -0,0 +1,88 @@
...
+export const canResolveDNS = (domain: string) =>
+  new Promise((resolve) => {
+    dns.lookup(domain, (err) => {
+      if (err) {
+        resolve(false);
+      } else {
+        resolve(true);
+      }
+    });
+  });
```
khelif96, line 29: Is this necessary? We can avoid this request by just checking if the fetch request returned the same error.
image

scripts/check-schema-and-codegen/utils.ts
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * Get the latest commit that was made to the GQL folder of the remote Evergreen repository.
+ * @returns A Promise that resolves to the SHA of the latest commit.
+ * @throws {Error} When failed to fetch commits.
+ */
+export async function getLatestCommitFromRemote(): Promise<string> {
```
khelif96, line 36: Can we use arrow syntax for consistency?

scripts/check-schema-and-codegen/utils.ts
```diff
@@ -0,0 +1,88 @@
...
+  const localSchemaSymlink = fs.readlinkSync(LOCAL_SCHEMA);
+  const originalDir = process.cwd();
+  try {
+    process.chdir(localSchemaSymlink);
+    execSync(`git merge-base --is-ancestor ${commit} HEAD`);
+    process.chdir(originalDir);
+    return true;
+  } catch (error) {
+    process.chdir(originalDir);
+    if (error.status === 1 || error.status === 128) {
```
khelif96, line 67: It would be helpful to comment on what these statuses represent or extract them into named constants.

scripts/check-schema-and-codegen/utils.ts
```diff
@@ -64,6 +47,7 @@ export const checkIsAncestor = async (commit: string): Promise<boolean> => {
     return true;
   } catch (error) {
     process.chdir(originalDir);
+    // Error status 1 and 128 means that the commit is not an anecestor and the user must fetch.
```
khelif96, line 50: I discussed this offline but would love to see a reference to this here. Just so its easy to differentiate between what 1 and 128 means.

khelif96 (approved):

PR #2046, opened 2023-09-19
EVG-19815: Remove GET_ prefix from GraphQL queries

sophstad:
EVG-19815

### Description

- Remove `GET_` from query names and rename files to remove `get-`


src/gql/queries/index.ts
```diff
@@ -1,159 +1,159 @@
...
+import VERSION_TASK_DURATIONS from "./version-task-durations.graphql";
+import VERSION_TASKS from "./version-tasks.graphql";
+import VERSION from "./version.graphql";
+import VIEWABLE_PROJECTS from "./viewable-projects.graphql";
 
 export {
   DISTRO,
   DISTRO_EVENTS,
   DISTRO_TASK_QUEUE,
-  GET_AGENT_LOGS,
```
khelif96, line 85: Love that these are sortable now as well!

khelif96 (approved):
Thank you for taking this on!

PR #2049, opened 2023-09-19
EVG-20899: Remove Konami code

sophstad:
EVG-20899

### Description

- The Apollo cache work in `useKonamiCode` introduced errors in #2048. Rather than fix it I think we may as well remove this code 😢 but we'll always have the git history!


src/analytics/aprilFools/useAprilFoolsAnalytics.ts
```diff
@@ -1,6 +1,6 @@
 import { useAnalyticsRoot } from "analytics/useAnalyticsRoot";
 
-type Action = { name: "Triggered Konami Code" };
+type Action = { name: "2024 Boilerplate!" };
```
khelif96, line 3: 🎉

khelif96 (approved):
Gonna come back stronger in 2024!

PR #2077, opened 2023-10-03
EVG-20201: Use new LeafyGreen props in LoadingButton component

minnakt:
EVG-20201

### Description
Updates the `LoadingButton` component to use the native props on the LG button.

### DIFF
--- a/src/components/Buttons/LoadingButton.tsx
+++ b/src/components/Buttons/LoadingButton.tsx
@@ -10,7 +10,7 @@ type Props = Omit<ButtonProps, "isLoading"> & {
 export const LoadingButton: ExtendableBox<
   Props & { ref?: React.Ref<any> },
   "button"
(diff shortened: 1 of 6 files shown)


src/components/Buttons/LoadingButton.tsx
```diff
@@ -13,7 +13,8 @@ export const LoadingButton: ExtendableBox<
 > = forwardRef(({ leftGlyph, loading = false, ...rest }: Props, ref) => (
```
sophstad, line 13: The type definition above could be tightened up a little bit, I think; it currently includes `loading` and `isLoading`. We should probably choose one or the other for clarity.

minnakt: oops thanks sophie! I kind of like `loading` more so I kept that one

sophstad (approved):

PR #2080, opened 2023-10-03
EVG-20918: Update search input focus shortcut on configure page

SupaJoon:
EVG-20918

### Description
I updated the shortcut to `/ ` which is a common search hotkey in various tools like the GitHub web app and vim.


sophstad (approved):
This seems like a nice lightweight improvement to me! Hopefully it'll reduce user confusion on the page 🙂

PR #2082, opened 2023-10-04
EVG-21018 Catch logout errors

khelif96:
EVG-21018

### Description
This has been pretty noisy on the ops-alerts channel. Looks like we aren't catching any errors when the logout request happens.


khelif96:
> The spam will continue until we stop calling the logout endpoint [here](https://github.com/evergreen-ci/spruce/blob/7c087559e4aff2f97f2429e5d162b4921d1ba00f/src/gql/GQLWrapper.tsx#L132) since the logout endpoint lives behind auth.

The logout endpoint is called whenever we get an unauthed graphql request. So adding some error handling should prevent the unhandled error from being sent to sentry.

PR #2086, opened 2023-10-04
EVG-21009: Make distro settings page readonly if user lacks edit permissions

sophstad:
EVG-21009

### Description

- Make all forms disabled if user lacks edit permissions for a given distro. The save button would currently fail under these conditions, but making the page clearly view-only is more helpful than only notifying a user once their save fails.
- Limiting users who lack view permissions already works as designed
- Create, copy, and delete functionality is also appropriately gated

### Screenshots

Disabled general settings page. You can see that the "New distro" button is omitted, and the delete button is disabled as well.
Screenshot 2023-10-04 at 1 57 34 PM

Existing functionality for a user who lacks view permissions:
image

### DIFF
--- a/src/pages/distroSettings/tabs/BaseTab.tsx
+++ b/src/pages/distroSettings/tabs/BaseTab.tsx
@@ -1,6 +1,12 @@
+import { useQuery } from "@apollo/client";
 import { useParams } from "react-router-dom";
 import { Form } from "components/Settings/Form";
 import { GetFormSchema, ValidateProps } from "components/SpruceForm";
+import {
+  UserDistroSettingsPermissionsQuery,
+  UserDistroSettingsPermissionsQueryVariables,
+} from "gql/generated/types";
+import { USER_DISTRO_SETTINGS_PERMISSIONS } from "gql/queries";
 import { usePopulateForm, useDistroSettingsContext } from "../Context";
 import { FormStateMap, WritableDistroSettingsType } from "./types";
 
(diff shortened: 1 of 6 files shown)


src/pages/distroSettings/tabs/BaseTab.tsx
minnakt, line 1: Is it intentional that the disabled form auto-collapses the expandable cards? It may give a mistaken impression that they are empty 🤔

Screenshot 2023-10-04 at 8 06 17 PM

src/pages/distroSettings/tabs/BaseTab.tsx
```diff
@@ -15,12 +21,22 @@ export const BaseTab = <T extends WritableDistroSettingsType>({
   initialFormState,
   ...rest
 }: BaseTabProps<T>) => {
```
minnakt, line 23: If you'd like, I think it would be possible to overwrite the user's permissions in Cypress like what is done [here](https://github.com/evergreen-ci/spruce/blob/9b09e0609384d0a039165966bf933c92df5c8a60/cypress/integration/nav_bar.ts#L123), so that you can test the view-only UI in Cypress

P.S. I created [EVG-21024](https://jira.mongodb.org/browse/EVG-21024)

minnakt (approved):

src/pages/distroSettings/tabs/BaseTab.tsx
sophstad: Yeah I think this was set up as such for project settings so that read-only repo settings wouldn't be as prominent on the page. Frankly I don't think it's a huge problem as the expandable card is decently obvious but we can revisit if we receive feedback on this!

PR #2087, opened 2023-10-04
EVG-19973: Improve patch parameters modal

minnakt:
EVG-19973

### Description
Initially I was going to change this to use `Chip`, but the more I looked at it the less it made sense (like why do we list `key:value` in a badge in the first place)? And also, the `Chip` didn't look as good as I thought it would. 🤨

If it's a list of key-value pairs, I think it would make sense and be more readable if we put them into a table.

### Screenshots
Screenshot 2023-10-04 at 10 19 38 AM

### Testing
- Jest tests & existing Cypress tests should pass

### DIFF
--- a/src/pages/version/ParametersModal/index.tsx
+++ b/src/pages/version/ParametersModal/index.tsx
@@ -1,4 +1,5 @@
 import { useRef, useState } from "react";
+import styled from "@emotion/styled";
 import { useLeafyGreenTable, LGColumnDef } from "@leafygreen-ui/table/new";
 import { DisplayModal } from "components/DisplayModal";
 import { MetadataItem } from "components/MetadataCard";
@@ -40,11 +41,18 @@ export const ParametersModal: React.FC<ParametersProps> = ({ parameters }) => {
         size="large"
         title="Patch Parameters"
       >
-        <BaseTable table={table} />
+        <OverflowContainer>
+          <BaseTable table={table} />
+        </OverflowContainer>
       </DisplayModal>
     </>
   );
 };
+
+const OverflowContainer = styled.div`
+  max-height: 600px;
+  overflow-y: scroll;
+`;
 
 const columns: LGColumnDef<Parameter>[] = [
   {
--- a/src/pages/version/ParametersModal/index.tsx
+++ b/src/pages/version/ParametersModal/index.tsx
@@ -42,7 +42,7 @@ export const ParametersModal: React.FC<ParametersProps> = ({ parameters }) => {
         title="Patch Parameters"
       >
         <OverflowContainer>
-          <BaseTable table={table} />
+          <BaseTable table={table} shouldAlternateRowColor />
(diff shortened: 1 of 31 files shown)


src/pages/version/ParametersModal/index.tsx
```diff
@@ -26,23 +35,34 @@ export const ParametersModal: React.FC<ParametersProps> = ({ parameters }) => {
...
         title="Patch Parameters"
-        data-cy="parameters-modal"
       >
-        {parameters?.map((param) => (
-          <StyledBadge key={`param_${param.key}`}>
-            {param.key}: {param.value}
-          </StyledBadge>
-        ))}
+        <OverflowContainer>
+          <BaseTable table={table} />
```
sophstad, line 45: Imo the striping looks good here
```suggestion
          <BaseTable table={table} shouldAlternateRowColor />
```
image

src/pages/version/ParametersModal/index.tsx
```diff
@@ -26,23 +35,34 @@ export const ParametersModal: React.FC<ParametersProps> = ({ parameters }) => {
...
+const OverflowContainer = styled.div`
+  max-height: 600px;
+  overflow-y: scroll;
 `;
+
+const columns: LGColumnDef<Parameter>[] = [
+  {
+    accessorKey: "key",
+    header: "Key",
+    cell: (value) => value.getValue(),
```
sophstad, line 61: You can omit the cell field entirely because the accessor key exactly matches the field name and we don't need to render it in any special way
```suggestion
```

src/pages/version/ParametersModal/index.tsx
```diff
@@ -26,23 +35,34 @@ export const ParametersModal: React.FC<ParametersProps> = ({ parameters }) => {
...
+  },
+  {
+    accessorKey: "value",
+    header: "Value",
+    cell: (value) => value.getValue(),
```
sophstad, line 66: Same as above!
```suggestion
```

src/pages/version/ParametersModal/index.tsx
```diff
@@ -26,23 +35,34 @@ export const ParametersModal: React.FC<ParametersProps> = ({ parameters }) => {
         </MetadataItem>
       )}
       <DisplayModal
+        data-cy="parameters-modal"
         open={showModal}
         setOpen={setShowModal}
+        size="large"
```
sophstad, line 41: I kinda feel like this modal is super wide considering ime these values are often fairly short. However if you have seen examples otherwise I think large is fine!

sophstad (approved):
I think using a table was a great move here! Nice work

src/pages/version/ParametersModal/index.tsx
minnakt: ah yes they can be really long! I've seen a patch parameter that was a really long URL

PR #2092, opened 2023-10-06
EVG-19921: Allow unhiding patch card

SupaJoon:
EVG-19921

https://github.com/evergreen-ci/spruce/assets/10734386/2a1baa3a-e15d-4afa-8cf0-3c90265e8410

### Evergreen PR
https://github.com/evergreen-ci/evergreen/pull/7116

### DIFF
--- a/src/components/PatchActionButtons/SetPatchVisibility.tsx
+++ b/src/components/PatchActionButtons/SetPatchVisibility.tsx
@@ -26,7 +26,7 @@ export const SetPatchVisibility: React.FC<Props> = ({
   >(SET_PATCH_VISIBILITY, {
     onCompleted: (d) => {
       const copy = d.setPatchVisibility?.[0].hidden ? "hidden" : "unhidden";
-      dispatchToast.success(`This patch was successfully ${copy}`);
+      dispatchToast.success(`This patch was successfully ${copy}.`);
     },
     onError: (err) => {
       dispatchToast.error(`Unable to update patch visibility: ${err.message}`);


src/components/PatchesPage/PatchCard.tsx
```diff
@@ -167,8 +171,12 @@ const Left = styled(Center)`
 `;
 
 const Right = styled.div`
+  width: 110px;
   display: flex;
   justify-content: flex-end;
+  > button {
+    margin-left: ${size.xs};
+  }
```
sophstad, line 179: Since this is already using flexbox, you can use the `gap` property instead
```suggestion
  gap: ${size.xs};
```

src/components/PatchesPage/index.tsx
```diff
@@ -167,3 +193,8 @@ const CheckboxContainer = styled.div`
   display: flex;
   justify-content: end;
 `;
+
+/* @ts-expect-error */
+const StyledCheckbox = styled(Checkbox)`
+  padding-left: ${size.xs};
+`;
```
sophstad, line 200: Similar note as above, you can add a `gap` to `CheckboxContainer` instead of creating a specialized checkbox.

src/components/PatchesPage/index.tsx
```diff
@@ -70,13 +77,24 @@ export const PatchesPage: React.FC<Props> = ({
...
     Cookies.set(cookie, e.target.checked ? "true" : "false");
     // eslint-disable-next-line no-unused-expressions
     analyticsObject.sendEvent({ name: "Filter Commit Queue" });
   };
 
+  const includeHiddenCheckboxOnChange = (
+    e: React.ChangeEvent<HTMLInputElement>
+  ): void => {
+    setIsIncludeHiddenCheckboxChecked(e.target.checked);
+    Cookies.set(INCLUDE_HIDDEN_PATCHES, e.target.checked ? "true" : "false");
```
sophstad, line 93: When I load this page with this cookie set to true, the checkbox was checked but hidden patches are not visible in the list (i.e. the state is set but the query does not include the parameter).

Tbh I don't know how helpful a cookie is for this setting. Unlike the commit queue, it's probably something that would be looked at in special circumstances, not every time a user visits this page. But if you think it's worth including then the bug should be fixed!

sophstad (changes requested):

sophstad (approved):

PR #2095, opened 2023-10-09
EVG-20692: Indicate if a version has been ignored

minnakt:
EVG-20692

### Description
This PR makes it possible to tell if a version has been ignored on Spruce. Previously, there was no information about ignored versions at all.

### Screenshots
`Banner` on Version page (I used a `Banner` because the LeafyGreen design docs say `Callout` should be reserved for [docs only](https://www.mongodb.design/component/callout/guidelines/#basic-usage)):

Screenshot 2023-10-09 at 3 25 06 PM

`Icon` in the inactive commits tooltip (this is the same icon on the legacy UI, but if it doesn't seem clear enough, I could also just write out `IGNORED` explicitly):
Screenshot 2023-10-09 at 4 26 55 PM

### Testing
- Cypress tests

### DIFF
--- a/cypress/integration/version/banners.ts
+++ b/cypress/integration/version/banners.ts
@@ -2,11 +2,11 @@ describe("banners", () => {
   const versionWithBanners =
     "/version/logkeeper_e864cf934194c161aa044e4599c8c81cee7b6237/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC";
 
-  beforeEach(() => {
-    cy.visit(versionWithBanners);
-  });
-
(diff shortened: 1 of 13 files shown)


minnakt:
> LGTM, another possibility for the eyeball is IGNORED in a badge

I really like the badge idea, but when I tried it it seemed a bit too bulky for the tooltip — so I decided to make a custom icon that reads more like "ignored". We can update it if it ends up confusing people, but I just wanted to try it out. 🫣

Screenshot 2023-10-10 at 12 35 08 PM

PR #2099, opened 2023-10-12
EVG-20817: Restrict special characters in project name & ID input

SupaJoon:
EVG-20817

### Description
The validation matches that in [EVG](https://github.com/evergreen-ci/evergreen/blob/7c3cf47f2934eec0f9b6052db5b093076660b2e3/rest/data/project.go#L205) so none of the characters require escaping in the URL.

### Screenshots
Screenshot 2023-10-12 at 2 35 23 PM

### DIFF
--- a/src/constants/routes.ts
+++ b/src/constants/routes.ts
@@ -227,7 +227,9 @@ export const getProjectSettingsRoute = (
   projectId: string,
   tab?: ProjectSettingsTabRoutes
 ) => {
-  // Encode projectId for backwards compatibilty
+  // Encode projectId for backwards compatibilty.
+  // Encoding can be removed when all projectIDs
+  // are URL friendly withou encoding
   const encodedProjectId = encodeURIComponent(projectId);
   const root = `${paths.project}/${encodedProjectId}/${PageNames.Settings}`;
   return tab ? `${root}/${tab}` : root;


src/utils/validators/index.ts
```diff
@@ -132,16 +132,27 @@ const validateRegexp = (regexp: string): boolean => {
   }
 };
 
+/**
+ * `validateNoSpecialCharacters` tests if a provided string contains no special characters
+ * @param str - The string to test.
+ * @returns - true if the string has no special characters and false otherwise
+ */
+const validateNoSpecialCharacters = (str: string): boolean => {
+  const noSpecialCharacters = /^[0-9a-zA-Z-._~()]*$/;
```
sophstad, line 141: It could be cool to define these special chars as a var that can be used in the error message so that any updates are reflected. I think this would work
```suggestion
const specialChars = "-._~()";
const validateNoSpecialCharacters = (str: string): boolean => {
    const noSpecialCharacters = new RegExp(`^[0-9a-zA-Z${specialChars}]*$`);
```

src/constants/routes.ts
```diff
@@ -227,11 +227,9 @@ export const getProjectSettingsRoute = (
   projectId: string,
   tab?: ProjectSettingsTabRoutes
 ) => {
-  if (!tab) {
-    return `${paths.project}/${projectId}/${PageNames.Settings}`;
-  }
-
-  return `${paths.project}/${projectId}/${PageNames.Settings}/${tab}`;
+  const encodedProjectId = encodeURIComponent(projectId);
```
sophstad, line 230: The description says we don't need to encode the URL, should this be included

sophstad: Nice -- since the ticket mentions that we'll work with project admins to change their project identifiers, can we update the comment to note that we'll delete this when that is done?

sophstad (approved):

PR #2104, opened 2023-10-16
EVG-20817: Don't invalidate General project settings form based on initial project identifier

SupaJoon:
EVG-20817

### Description
This fixes an issue where the General project settings form is disabled when it loads with a project identifier that doesn't pass validation. The form will do normal validation if any changes are made to the project identifier.


sophstad (approved):
Good catch!

PR #2108, opened 2023-10-17
EVG-20983: Improve deploy script error handling and remove duplicate tag push

SupaJoon:
EVG-20983
Adds error handling and exit codes to `evergreenDeploy`, `ciDeploy`, `localDeploy` and the postversion hook (push-version.sh) and removes a duplicate tag push when force deploying.

### DIFF
--- a/scripts/deploy/utils/git/tag/mock-tag-utils.ts
+++ b/scripts/deploy/utils/git/tag/mock-tag-utils.ts
@@ -26,11 +26,9 @@ const deleteTag = (tag: string) => {
 
 /**
  * `pushTags` is a helper function that pushes tags to the remote.
- * @returns true
  */
 const pushTags = () => {
   console.log(yellow("Dry run mode enabled. Pushing tags."));
-  return true;
 };
 
 export { createTagAndPush, getLatestTag, deleteTag, pushTags };


scripts/deploy/ deploy-production.test.ts
khelif96, line 1: I think there might be a space in the name of this file? Either that or the github UI is playing tricks.
image

scripts/deploy/utils/git/tag/tag-utils.ts
```diff
@@ -3,12 +3,21 @@ import { githubRemote } from "./constants";
...
+    execSync("yarn version --new-version patch", {
+      encoding: "utf-8",
+      stdio: "inherit",
+    });
+  } catch (err) {
+    console.log("Creating new tag failed.");
+    console.log("output", err);
+    return false;
+  }
+  return true;
```
khelif96, line 20: I think this composition is pretty confusing. Some util functions return these booleans indicating the status while others throw errors. What criteria is used to determine what should do what? Could we be a little more consistent in how we error? 
How do you feel about treating errors the way we do in go and wrapping them? 
https://stackoverflow.com/questions/42754270/re-throwing-exception-in-nodejs-and-not-losing-stack-trace

khelif96 (changes requested):

scripts/deploy/deploy-production.ts
```diff
@@ -1,105 +1,87 @@
...
-    });
+    // If there are no commit messages, ask the user if they want to delete and re-push the latest tag, thereby forcing a deploy with no new commits.
+    if (commitMessages.length === 0) {
+      const latestTag = getLatestTag();
+      const { value: shouldForceDeploy } = await prompts({
+        type: "confirm",
+        name: "value",
+        message: `No new commits. Do you want to trigger a deploy on the most recent existing tag? (${latestTag})`,
+        initial: false,
+      });
```
sophstad, line 24: We ran into an issue last week when the deployer just hit enter and breezed through this message. Could we maybe have it default to false so that a user needs to be a little more careful to go through with this action?

khelif96 (approved):
LGTM

scripts/deploy/deploy-production.test.ts
sophstad, line 1: 😍

sophstad (approved):

PR #2111, opened 2023-10-18
EVG-20807 Expose Parsley link for task files.

khelif96:
EVG-20807

### Description
Adds a section that contains a link to the task uploaded file in Parsley. The button will be disabled if it is not available. 

Depends on #2110 to wire up the analytics events.
### Screenshots
image


src/analytics/task/useTaskAnalytics.ts
```diff
@@ -64,6 +64,10 @@ type Action =
       parsleyAvailable: boolean;
       fileName: string;
     }
+  | {
+      name: "Click Task File Parsley Link";
+      fileName: string;
```
sophstad, line 69: I suppose it can be gleaned from the URL, but do you think project name would be useful to track here? Perhaps for all of these events.

src/pages/task/taskTabs/FileTable/GroupedFileTable/index.tsx
```diff
@@ -17,23 +19,49 @@ const columns = (
...
+      <CellContainer>
+        <StyledLink
+          href={value.row.original.link}
+          data-cy="file-link"
+          target="_blank"
+          onClick={() => {
+            taskAnalytics.sendEvent({
+              name: "Click Task File Link",
+              parsleyAvailable: value.row.original.urlParsley !== null,
+              fileName: value.getValue() as GroupedFilesFile["name"],
```
sophstad, line 34: [nit] Could pull `value.getValue()` into a variable since it's used 3 times

sophstad (approved):

src/analytics/task/useTaskAnalytics.ts
khelif96: Great idea! Done

PR #2120, opened 2023-10-27
DEVPROD-1101: Update previous commits selector

sophstad:
DEVPROD-1101 - This is what I worked on over skunkworks because it's always irked me 😁 

### Description

- Replace select pattern with dropdown menu to be more in line with LeafyGreen design guidelines.
- Replace reducer with a memoized function that returns the relevant URLs. A state is not really necessary for this use case.

### Screenshots

image

### Testing

- Update tests to pass with new UI


src/pages/task/actionButtons/previousCommits/index.tsx
```diff
@@ -51,204 +36,138 @@ export const PreviousCommits: React.FC<PreviousCommitsProps> = ({ taskId }) => {
...
-            justify="middle"
-            triggerEvent="hover"
-            trigger={children}
-          >
-            {loading
-              ? `Fetching...`
-              : `There is no version that satisfies this criteria.`}
-          </Tooltip>
-        )}
+        {versionMetadata?.isPatch ? "Base" : "Previous"} commit
```
minnakt, line 143: I think that it's not obvious that these options send you to a different page, which I think could be fixed by just tweaking the menu options to start with "Go to..." or "View..."

Screenshot 2023-11-01 at 1 04 00 PM

src/pages/task/actionButtons/previousCommits/index.tsx
```diff
@@ -51,204 +36,138 @@ export const PreviousCommits: React.FC<PreviousCommitsProps> = ({ taskId }) => {
...
+    taskData?.task ?? {};
+  const { order: skipOrderNumber, projectIdentifier } =
+    versionMetadata?.baseVersion ?? {};
+  const bvOptionsBase = {
+    tasks: [applyStrictRegex(displayName)],
+    variants: [applyStrictRegex(buildVariant)],
+  };
+
   // We don't error for this query because it is the default query that is run when the page loads.
   // If it errors it probably means there is no base version, which is fine.
```
minnakt, line 49: I think you can delete this comment since there's a `skip` on this now :0 (doesn't run by default)

src/pages/task/actionButtons/previousCommits/reducer.ts
minnakt, line 1: yayyyy bye 👋

minnakt (approved):
nice job! 🥰

src/pages/task/actionButtons/previousCommits/utils.ts
```diff
@@ -0,0 +1,56 @@
...
+    return {
+      [CommitType.Base]: nullLink,
+      [CommitType.LastPassing]: nullLink,
+      [CommitType.LastExecuted]: nullLink,
+    };
+  }
+
+  return {
+    [CommitType.Base]: getTaskRoute(parentTask.id),
+    [CommitType.LastPassing]: getTaskRoute(
```
khelif96, line 28: This link would always default to the parent task even if it does not pass

sophstad: Sorry if I'm misunderstanding, doesn't [this test case](https://github.com/evergreen-ci/spruce/blob/dee29eda9ca94e917eb2407514e818bb630fcd46/src/pages/task/actionButtons/previousCommits/PreviousCommits.test.tsx#L120) assert that the last passing task works correctly?

src/pages/task/actionButtons/previousCommits/index.tsx
```diff
@@ -51,204 +37,146 @@ export const PreviousCommits: React.FC<PreviousCommitsProps> = ({ taskId }) => {
...
-        >
-          Go
-        </LoadingButton>
-      </ConditionalWrapper>
-    </PreviousCommitsWrapper>
+        Go to last passing version
+      </MenuItem>
+      <MenuItem
+        as={Link}
+        disabled={executedLoading}
```
khelif96, line 169: These can also be disabled if its the first execution of the task. Or if there wasn't one found in the last 300 versions for any of the above tasks.

src/pages/task/actionButtons/previousCommits/utils.ts
khelif96: Thanks for looping back I missed that!

src/pages/task/actionButtons/previousCommits/utils.ts
```diff
@@ -0,0 +1,56 @@
...
+// not exist. The logic to extract the task from it is written in this function.
+export const getTaskFromMainlineCommitsQuery = (
+  data: LastMainlineCommitQuery
+): CommitTask => {
+  const buildVariants =
+    data?.mainlineCommits.versions.find(({ version }) => version)?.version
+      .buildVariants ?? [];
+  if (buildVariants.length > 1) {
+    reportError(
+      new Error("Multiple build variants matched previous commit search.")
```
khelif96, line 47: Is there more info here that we can add to the sentry search? Or will sentry give us the necessary data be default?

khelif96 (approved):
LGTM mod one comment

src/pages/task/actionButtons/previousCommits/utils.ts
sophstad: Sentry will include a breadcrumb with the GraphQL query sent before this error. I think that's enough information to start investigating this.

PR #2121, opened 2023-10-30
EVG-20704: Save patch trigger aliases with new success status

sophstad:
EVG-20704

### Description

Save patch trigger aliases with "success" status.

After deploying this change, the following operation will be run, first on staging, to convert all patch trigger aliases from "succeeded" to "success" status:
```
db.project_ref.updateMany(
  { "patch_trigger_aliases": { $exists: true } },
  { $set: { "patch_trigger_aliases.$[elem].status": "success" } },
  { arrayFilters: [ { "elem.status": "succeeded" } ] }
)
```

### Testing

Validated DB command on local database, ensured that patch trigger aliases with either string function as expected.


ablack12 (approved):
🚀

PR #2123, opened 2023-10-30
EVG-19921: Set includeHidden to false in Patches query

SupaJoon:
EVG-19921

### Description
This change ensures backwards compatibility after https://github.com/evergreen-ci/evergreen/pull/7116 is merged and before https://github.com/evergreen-ci/spruce/pull/2092 is merged. Once https://github.com/evergreen-ci/evergreen/pull/7116 is merged `includeHidden` is optional and will evaluate to true if not provided in the PatchesInput. That's why it must be set to false. Without this intermediate change, hidden patches will be visible.

### DIFF
--- a/src/components/PatchesPage/index.tsx
+++ b/src/components/PatchesPage/index.tsx
@@ -138,10 +138,11 @@ export const usePatchesInputFromSearch = (search: string): PatchesInput => {
   );
   const statuses = rawStatuses.filter((v) => v && v !== ALL_PATCH_STATUS);
   return {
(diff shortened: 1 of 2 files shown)


khelif96 (changes requested):
I think we can avoid this by just making https://github.com/evergreen-ci/evergreen/pull/7116 default to false. e.g. https://github.com/evergreen-ci/evergreen/pull/7116/files#r1378196168 and then running gqlgen.

This will allow the backend change to be fully backward's compatible.

PR #2128, opened 2023-10-31
DEVPROD-920: Conditionally render base status badge

sophstad:
DEVPROD-920

### Description

- In the tests table, omit the base status badge if undefined. We already do this in the tasks table.

### Screenshots

Before:

before

After:
image


khelif96 (approved):
Short and sweet :)

PR #2139, opened 2023-11-06
DEVPROD-1230: Migrate Task Duration table to V12 table

minnakt:
DEVPROD-1230

### Description
Updates the Task Duration table to use the newer version of the LG table.

### Testing
- Existing Jest & Cypress tests should pass
- Updated Storybook story


src/components/Table/LGFilters.tsx
```diff
@@ -1,46 +1,108 @@
...
+type InputFilterProps = {
+  "data-cy"?: string;
+  onConfirm?: ({ id, value }: { id: string; value: string }) => void;
+};
+
+export const getColumnInputFilterProps = ({
+  "data-cy": dataCy,
+  onConfirm = () => {},
+}: InputFilterProps) => ({
+  meta: {
```
sophstad, line 60: Use of `meta` is pretty neat! Good find, will be helpful with not duplicating code!

src/components/TablePopover/TableSearchPopover.tsx
```diff
@@ -12,36 +12,40 @@ import { useOnClickOutside } from "hooks";
...
 
   // Handle onClickOutside
   useOnClickOutside([buttonRef, popoverRef], () => setActive(false));
 
-  const closePopup = () => {
-    onConfirm();
+  // If the value from the URL has changed, update the input.
+  useEffect(() => {
+    setInput(value);
+  }, [value]);
```
sophstad, line 40: What case does this handle, navigating to a URL that already has filters applied?

src/components/Table/BaseTable.tsx
```diff
@@ -17,56 +17,78 @@ type SpruceTableProps = {
...
-        ))}
-      </TableBody>
-    </Table>
-    {table.getRowModel().rows.length === 0 &&
+        </Row>
+      ))}
+    </TableBody>
+    {!loading &&
+      table.getRowModel().rows.length === 0 &&
       (emptyComponent || "No data to display")}
```
sophstad, line 92: Would it be possible to center the placeholder 😳 
image

src/components/TablePopover/TableSortIcon.tsx
```diff
@@ -0,0 +1,63 @@
...
+  "data-cy": dataCy,
+  onToggle,
+  value,
+}) => {
+  const iconColor = value === SortState.OFF ? gray.dark2 : blue.base;
+
+  const onClick = () => {
+    let update: string | undefined;
+    if (value === SortState.OFF) {
+      update = "DESC";
```
sophstad, line 39: It looks like we may have a GraphQL enum `SortDirection` that represents these values

src/components/TablePopover/TableSortIcon.tsx
```diff
@@ -0,0 +1,63 @@
+import styled from "@emotion/styled";
+import Icon from "@leafygreen-ui/icon";
+import IconButton from "@leafygreen-ui/icon-button";
+import { palette } from "@leafygreen-ui/palette";
+import { size } from "constants/tokens";
+
+const { blue, gray } = palette;
+
+enum SortState {
```
sophstad, line 9: [nit] Perhaps we should call this something like `TableSortState` so it's clear it refers to react-table values rather than what Evergreen expects

src/pages/version/taskDuration/TaskDurationTable.tsx
```diff
@@ -28,140 +27,146 @@ interface Props {
...
-  const { currentStatuses } = useTaskStatuses({ versionId });
+  const { currentStatuses: statusOptions } = useTaskStatuses({ versionId });
+
+  const [queryParams] = useQueryParams();
+  const {
+    [PatchTasksQueryParams.TaskName]: taskName = "",
+    [PatchTasksQueryParams.Statuses]: statuses = [],
+    [PatchTasksQueryParams.Variant]: variant = "",
+    [PatchTasksQueryParams.Duration]: duration = "",
+  } = queryParams;
```
sophstad, line 38:
```suggestion
  const [{
    [PatchTasksQueryParams.TaskName]: taskName = "",
    [PatchTasksQueryParams.Statuses]: statuses = [],
    [PatchTasksQueryParams.Variant]: variant = "",
    [PatchTasksQueryParams.Duration]: duration = "",
  }] = useQueryParams();
```

src/pages/version/taskDuration/TaskDurationTable.tsx
```diff
@@ -28,140 +27,146 @@ interface Props {
...
-    resetPage: true,
-    sendAnalyticsEvent: (filterBy: string) =>
-      sendEvent({ name: "Filter Tasks", filterBy }),
-  };
+  const sorting = useMemo(
+    () => [
+      ...(duration && [
+        {
+          id: PatchTasksQueryParams.Duration,
+          desc: duration === "DESC",
```
sophstad, line 57: If `SortOrder` enum can be used, we should use it here too

sophstad (commented):
This is really impressive 🌟

src/pages/version/taskDuration/TaskDurationTable.stories.tsx
```diff
@@ -10,34 +10,184 @@ export const Default: CustomStoryObj<typeof TaskDurationTable> = {
   render: () => <TaskDurationTable tasks={props.tasks} loading={false} />,
 };
 
+export const LongContent: CustomStoryObj<typeof TaskDurationTable> = {
+  render: () => <TaskDurationTable tasks={props.tasksLong} loading={false} />,
+};
+
 const props = {
-  tasks: [
+  tasksLong: [
```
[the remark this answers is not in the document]
minnakt: added more stories to `BaseTable`!

src/components/TablePopover/TableSearchPopover.tsx
minnakt: It's because of when you use the "Clear all filters" button 😞 Since the component manages an internal state, if I don't clear it here, you'll still see the filter in the input even after clicking "Clear all filters"

src/components/Table/BaseTable.stories.tsx
```diff
@@ -22,34 +23,83 @@ export const EmptyState: CustomStoryObj<typeof BaseTable> = {
...
+  subRows: [
+    {
+      name: `nested name ${i}`,
+      type: `nested type ${i}`,
+      size: `nested size ${i}`,
+    },
+  ],
+}));
+
+const repeated = "long ".repeat(100);
```
sophstad, line 68: TIL `repeat()`!

src/components/TablePopover/TableSearchPopover.tsx
```diff
@@ -12,36 +12,40 @@ import { useOnClickOutside } from "hooks";
...
 
   const buttonRef = useRef(null);
   const popoverRef = useRef(null);
 
   // Handle onClickOutside
   useOnClickOutside([buttonRef, popoverRef], () => setActive(false));
 
-  const closePopup = () => {
-    onConfirm();
+  // If the value from the URL has changed, update the input.
```
sophstad, line 37: I noticed that when changing the page size, filters and sorts are removed. Is that fixable? That may not be related to this `useEffect` but wanted to call it out.

minnakt: oops yeah seems to be a pre-existing bug since the other pages in Spruce also show this behavior! i think i fixed it

src/pages/version/taskDuration/TaskDurationTable.tsx
```diff
@@ -1,24 +1,23 @@
-import styled from "@emotion/styled";
-import { palette } from "@leafygreen-ui/palette";
-import { Table, TableHeader } from "@leafygreen-ui/table";
+import { useCallback, useMemo, useRef } from "react";
+import { useLeafyGreenTable } from "@leafygreen-ui/table/new";
```
sophstad, line 2: I think we can get rid of the non-`/new` dependency in this PR!

sophstad (changes requested):

src/pages/version/taskDuration/TaskDurationTable.tsx
minnakt: soo true...

sophstad (approved):
Thanks for the additional fixes!

PR #2140, opened 2023-11-06
DEVPROD-2234: Use correct project for querying task history

sophstad:
DEVPROD-2234

### Description

We've been querying for task history using the version's project instead of the task's. This is often fine, but downstream tasks will use the wrong value for queries.

The bug was present in the previous version of the history selector, but was encountered a lot less often because a user would have to select "Go to last passing task" in the dropdown before the query ran.

Now, get the project identifier directly from the task.

### Testing
- Update unit tests


khelif96 (approved):
Great catch!

PR #2146, opened 2023-11-09
DEVPROD-360: Allow spec filtering for user submitted patches

SupaJoon:
DEVPROD-360

The `cypress_spec` parameter maps to the Cypress [spec](https://docs.cypress.io/guides/guides/command-line#cypress-run-spec-lt-spec-gt) flag but only for user submitted patches


minnakt (approved):
works great!

PR #2147, opened 2023-11-09
DEVPROD-941, take 2: linkify user name on version metadata panel

bthood:
DEVPROD-941

### Description
Use the new authorEmail field to compute the username and linkify based on that. Also include a fallback based on author name in case the email is improperly specified.

### Screenshots
https://github.com/evergreen-ci/spruce/assets/1911841/243c190c-f450-4f32-be6a-be88369fb01a

### Testing
See screen video; will also verify on staging.

### Evergreen PR
https://github.com/evergreen-ci/evergreen/pull/7228

### DIFF
--- a/src/pages/version/Metadata.tsx
+++ b/src/pages/version/Metadata.tsx
@@ -122,7 +122,7 @@ export const Metadata: React.FC<Props> = ({ loading, version }) => {
       <MetadataItem>
         Submitted by:{" "}
         <StyledRouterLink
-          to={getUserPatchesRoute(getAuthorUsername(authorEmail, author))}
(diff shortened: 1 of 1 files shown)


src/pages/version/Metadata.tsx
```diff
@@ -204,3 +214,8 @@ export const Metadata: React.FC<Props> = ({ loading, version }) => {
     </MetadataCard>
   );
 };
+
+const getAuthorUsername = (email: string) => {
```
khelif96, line 218: I think this is an okay home for it for now since it isn't used elsewhere.

khelif96 (approved):
Great job on this!

PR #2151, opened 2023-11-13
DEVPROD-2415: Enable test isolation for distro settings e2e tests

SupaJoon:
DEVPROD-2415


minnakt (approved):
this pr didn't have any review requests but i just guessed that it was ready — let me know if it wasn't and you need another review

PR #2155, opened 2023-11-16
DEVPROD-981 Support notifications when SUCCESSFUL task exceeds some duration

malikchaya2:
DEVPROD-981

### Description
This allows projects to set up notifications for a task succeeding a duration only if the task is successful. This will be merged after the backend change is merged. 

### Screenshots
image

### Testing
Tested on staging

### Evergreen PR
[7251](https://github.com/evergreen-ci/evergreen/pull/7251)


src/components/Notifications/form/event.ts
```diff
@@ -396,6 +396,25 @@ export const getEventSchema = (
...
+                properties: {
+                  [ExtraFieldKey.TASK_DURATION_SECS]: taskDurationInput,
+                },
+              },
+              regexSelector: regexSelector(
+                regexEnumsToDisable,
+                taskRegexSelectors
+              ),
+            },
+          },
```
minnakt, line 417: Since it seems like this trigger shares the same fields as                   `ProjectTriggers.TASK_EXCEEDS_DURATION`, you can actually just delete this block and combine it with the above block so that it looks like:
```
       {
            properties: {
              eventSelect: {
                enum: [
                  ProjectTriggers.TASK_EXCEEDS_DURATION,
                  ProjectTriggers.SUCCESSFUL_TASK_EXCEEDS_DURATION,
                ],
              },
              ... blah blah blah
          },
```

src/constants/triggers.ts
```diff
@@ -306,10 +306,25 @@ export const projectTriggers: Trigger = {
...
+        format: "number",
+        default: "10",
+      },
+    ],
+  },
   [ProjectTriggers.TASK_EXCEEDS_DURATION]: {
     trigger: TriggerType.EXCEEDS_DURATION,
     resourceType: ResourceType.Task,
-    label: "The Runtime For a Task Exceeds Some Duration",
+    label: "The Runtime For any Task Exceeds Some Duration",
```
minnakt, line 327: Let's capitalize this word!

```suggestion
    label: "The Runtime For Any Task Exceeds Some Duration",
```

minnakt (approved):
nice job! I realize we probably shouldn't use weird capitalization in the select so I'll make another ticket to fix that

PR #2166, opened 2023-11-27
DEVPROD-1972: Update Hosts Table

sophstad:
DEVPROD-1972

### Description

- Convert hosts table to use LeafyGreen table
- Add support for sort using [`onSortingChange`](https://tanstack.com/table/v8/docs/api/features/sorting#onsortingchange) and filtering via [`onColumnFiltersChange`](https://tanstack.com/table/v8/docs/api/features/filters#oncolumnfilterschange). This means we no longer need to use the `getColumnTreeSelectFilterProps`, `getColumnInputFilterProps`, and `getColumnSortProps` utils introduced with the task duration table.

### Screenshots

image

### Testing

- Update hosts table page tests
- Add tests for new sort hook


src/types/task.ts
```diff
@@ -10,6 +10,7 @@ export enum RequiredQueryParams {
   Execution = "execution",
 }
 
+// TODO: Remove SortBy-Limit in favor of generic TableQueryParams
```
khelif96, line 13: Is this ticketed or planned in a follow-up?

src/pages/hosts/HostsTable.tsx
```diff
@@ -1,304 +1,254 @@
...
-      urlSearchParam: "startedBy",
-      sendAnalyticsEvent: sendHostsTableFilterEvent,
-    });
+  return (
+    <BaseTable
+      data-cy="hosts-table"
+      data-loading={loading}
+      loading={loading}
+      shouldAlternateRowColor
+      table={table}
```
khelif96, line 116: nit. Can we reflect the expected page size when loading? 
You can use the `loadingRows` prop.

src/types/host.ts
```diff
@@ -74,3 +77,30 @@ export enum HostMonitorOp {
...
+
+export const mapIdToFilterParam: PartialRecord<
+  HostSortBy,
+  keyof HostsQueryVariables
+> = Object.entries(mapQueryParamToId).reduce(
+  (obj, [id, param]) => ({
+    ...obj,
+    [param]: id,
+  }),
+  {}
```
khelif96, line 105: I'm confused by this what are we trying to do here?
Is this trying to establish a mapping between the query param and the react-table id field?

src/pages/hosts/index.tsx
```diff
@@ -203,6 +232,22 @@ const getQueryVariables = (search: string): HostsQueryVariables => {
   };
 };
 
+// Convert query param values into react-table's column filters state
+const getFilters = (queryParams: HostsQueryVariables): ColumnFiltersState => {
```
khelif96, line 236: nit. can you use jsdoc syntax so we get intellisense suggestions? :) 
image

khelif96 (approved):
Great job!

src/types/host.ts
sophstad: Yep that's it. More specifically, a given field has a value that is used with the `sortBy` param, as well as a field that is used as the key when filtering (i.e. `distroId` in `distroId=ubuntu`). The goal here is to associate them with each other and their column.

src/types/task.ts
sophstad: It can be done as part of the task table work.

src/components/Table/LGFilters.tsx
```diff
@@ -53,10 +57,14 @@ type InputFilterProps = {
   onConfirm?: ({ id, value }: { id: string; value: string }) => void;
 };
 
+/*
+ * @deprecated Use react-table's onColumnFiltersChange prop.
+ */
 export const getColumnInputFilterProps = ({
```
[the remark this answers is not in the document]
sophstad: Ah yes! Opened DEVPROD-3049 since this is a chunky PR.

PR #2170, opened 2023-11-28
DEVPROD-789: Fix spacing bug in ArrayFieldTemplate

minnakt:
DEVPROD-789

### Description
There was an odd spacing around array buttons because we always rendered the `<Placeholder>` component for an empty array, even if `placeholder` was not defined.

### Screenshots
(These are just a few examples, you would be able to see it on any array form element.)

Before       |  After
Screenshot 2023-11-28 at 12 12 20 PM  | Screenshot 2023-11-28 at 12 16 19 PM

Before     |  After
Screenshot 2023-11-28 at 12 14 00 PM |  Screenshot 2023-11-28 at 12 16 11 PM


sophstad (approved):
Yay 💆‍♀️

PR #2172, opened 2023-11-28
DEVPROD-2920: Mute flaky test

SupaJoon:
DEVPROD-2920


khelif96 (approved):
LGTM so long as we fix it soon

PR #2175, opened 2023-11-29
DEVPROD-2225 Replace history on project settings redirect

khelif96:
DEVPROD-2225

### Description
Visiting the project settings page with no tab provided in the URL would redirect you to a tab page but the non tab route would always exist in the URL and lead to an infinite redirect when using the back button.

https://github.com/evergreen-ci/spruce/assets/4605522/35a9072c-1149-441c-b283-4e393e61e99e


sophstad (commented):
Thanks, the distro settings page is missing this too if you wouldn't mind fixing my copypasta!

PR #2181, opened 2023-12-07
DEVPROD-3049: Remove LeafyGreen column utils

sophstad:
DEVPROD-3049

### Description

- Remove getColumnTreeSelectFilterProps, getColumnInputFilterProps, and getColumnSortProps in favor of native react-table filter/sort props

### Testing

- Update Cypress tests


cypress/integration/version/task_duration.ts
```diff
@@ -31,7 +31,7 @@ describe("Task Duration Tab", () => {
       cy.dataCy("leafygreen-table-row").should("have.length", 3);
       cy.location("search").should(
         "include",
-        "duration=DESC&page=0&statuses=running-umbrella,started,dispatched"
+        "duration=DESC&page=0&statuses=running-umbrella%2Cstarted%2Cdispatched"
```
khelif96, line 34: This indicates to me that the query params aren't being properly uriencoded or that we are double encoding them.

src/pages/version/taskDuration/TaskDurationTable.tsx
```diff
@@ -34,46 +33,44 @@ export const TaskDurationTable: React.FC<Props> = ({
...
+  const [sorting, setSorting] = useState<SortingState>(initialSort);
 
-  const updateQueryParams = useUpdateURLQueryParams();
-  const updateUrl = useCallback(
-    ({ id, value }) => {
-      updateQueryParams({ [id]: value || undefined, page: "0" });
-    },
-    [updateQueryParams]
-  );
+  const updateFilters = (filterState: ColumnFiltersState) => {
```
khelif96, line 46: I think at some point here we aren't properly transforming array values for filters which causes them to not be uriencoded correctly.

src/pages/version/taskDuration/TaskDurationTable.tsx
```diff
@@ -34,46 +33,44 @@ export const TaskDurationTable: React.FC<Props> = ({
...
-        {
-          id: PatchTasksQueryParams.Duration,
-          desc: duration === SortDirection.Desc,
-        },
-      ]),
-    ],
-    [duration]
-  );
+  const [filters, setFilters] = useState<ColumnFiltersState>(initialFilters);
+  const [sorting, setSorting] = useState<SortingState>(initialSort);
```
khelif96, line 44: This seems like we are just duplicating the state that's already represented in the URL.

khelif96: We should be able to just rely on the urlstate.

khelif96 (changes requested):

cypress/integration/version/task_duration.ts
sophstad: Well isn't the URL now being encoded, and it wasn't before? query-string's [`.stringify()`](https://www.npmjs.com/package/query-string#stringifyobject-options) has encode = true as default, which is what ultimately applies the encoding.

src/pages/version/taskDuration/TaskDurationTable.tsx
sophstad: Yeah, react-table's handling of sort/filter is kinda weird... I went with this approach because it allows us to use the `onColumnFilterChange`/`onSortingChange` callbacks; using non "manual" filtering (to use react-table terminology) means that those functions are never called. But I changed approaches to have a `useEffect` hook update the query params instead.

Let me know what you think! If you like it I'll make a similar change to HostsTable.

khelif96 (approved):
Looks great!

khelif96 (approved):

khelif96 (approved):
💯

PR #2199, opened 2023-12-18
DEVPROD-1975 Migrate TaskQueueTable to leafygreen table

khelif96:
DEVPROD-1975

### Description
Migrates the TaskQueueTable to use the LeafygreenTableComponent. Light refactoring to make the Table more self contained.

### Screenshots
image

### DIFF
--- a/src/pages/taskQueue/TaskQueueTable/index.tsx
+++ b/src/pages/taskQueue/TaskQueueTable/index.tsx
@@ -51,11 +51,11 @@ const TaskQueueTable: React.FC<TaskQueueTableProps> = ({
     if (taskId !== undefined && !performedInitialScroll.current) {
       const i = taskQueue.findIndex((t) => t.id === taskId);
       setSelectedRowIndexes([i]);
-      table.scrollToIndex(i, { align: "start" });
+      table.scrollToIndex(i, { align: "center" });
 
       setTimeout(() => {
         performedInitialScroll.current = true;
-        table.scrollToIndex(i, { align: "start" });
+        table.scrollToIndex(i, { align: "center" });
       }, 200);
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps


src/pages/taskQueue/TaskQueueTable/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+    />
+  );
+};
+
+const TaskCell = styled.div`
+  display: flex;
+  flex-direction: column;
+`;
+
+type TaskQueueColumnData = Omit<TaskQueueItem, "revision">;
```
minnakt, line 57: I think this type can be used on L19 and L35 as well

src/pages/taskQueue/TaskQueueTable/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+          <TaskCell>
+            <Body>
+              <StyledRouterLink
+                data-cy="current-task-link"
+                to={getTaskRoute(id)}
+                onClick={() => sendEvent({ name: "Click Task Link" })}
+              >
+                <WordBreak>{displayName}</WordBreak>
+              </StyledRouterLink>
+            </Body>
```
minnakt, line 86: Any reason for the `<Body>` tag here?
```suggestion
              <StyledRouterLink
                data-cy="current-task-link"
                to={getTaskRoute(id)}
                onClick={() => sendEvent({ name: "Click Task Link" })}
              >
                <WordBreak>{displayName}</WordBreak>
              </StyledRouterLink>
```

src/pages/taskQueue/TaskQueueTable/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+import { TaskQueueItem, TaskQueueItemType } from "gql/generated/types";
+import { formatZeroIndexForDisplay } from "utils/numbers";
+import { msToDuration } from "utils/string";
+
+interface TaskQueueTableProps {
+  taskQueue: Omit<TaskQueueItem, "revision">[];
+  loading: boolean;
+}
+
+const TaskQueueTable: React.FC<TaskQueueTableProps> = ({
```
minnakt, line 23: Does this support scrolling to the task if the task ID is defined in the URL?

src/pages/taskQueue/index.tsx
```diff
@@ -15,14 +15,16 @@ import {
 import { getTaskQueueRoute, getAllHostsRoute } from "constants/routes";
```
minnakt, line 15: I think the page is crashing for me when I visit the `/task-queue` route, which is linked via the nav bar

src/pages/taskQueue/TaskQueueTable/index.tsx
minnakt, line 23: After deploying on beta, I see that the performance is noticeably slow when changing the distro compared to production. It's noticeable because the dropdown freezes in place and then there's a lag before the distro name gets updated. Is there any way we can improve it?

src/pages/taskQueue/TaskQueueTable/index.tsx
khelif96: No reason at all apart from some copy pasta :)

src/pages/taskQueue/index.tsx
khelif96: It looks like the lg table is not performant on larger tables. I am adding virtualization support in #2205

minnakt: I'd be curious to know if the original tanstack table has the same problem 🤔 When I was working on the build visibility demo, I actually had to use the original tanstack table because the LG table was too slow...

khelif96: It does I tried updating the example table to render 10k rows and it would crash repeatedly https://codesandbox.io/p/devbox/optimistic-gianmarco-vpj64c

src/pages/taskQueue/TaskQueueTable/index.tsx
khelif96: I made some improvements and it should be better now.

khelif96: Fixed this.

src/pages/taskQueue/TaskQueueTable/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+  useEffect(() => {
+    if (taskId !== undefined && !performedInitialScroll.current) {
+      const i = taskQueue.findIndex((t) => t.id === taskId);
+      setSelectedRowIds([i.toString()]);
+      table.scrollToIndex(i, { align: "center" });
+
+      setTimeout(() => {
+        performedInitialScroll.current = true;
+        table.scrollToIndex(i, { align: "center" });
+      }, 200);
```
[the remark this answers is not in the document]
khelif96: It seems to be needed to produce a consistent scroll It seems similar to the bug we had in Parsley with our old virtualization library.

src/components/Table/BaseTable.tsx
```diff
@@ -140,11 +151,21 @@ export const BaseTable = forwardRef(
               ? virtualRows.map((vr) => {
                   const row = rows[vr.index];
                   return (
-                    <RenderableRow row={row} key={row.id} virtualRow={vr} />
+                    <RenderableRow
+                      row={row}
+                      key={row.id}
+                      virtualRow={vr}
+                      selectedRowIds={selectedRowIds}
```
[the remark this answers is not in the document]
khelif96: Great idea!

src/pages/taskQueue/TaskQueueTable/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+}
+
+const TaskQueueTable: React.FC<TaskQueueTableProps> = ({
+  loading,
+  taskId,
+  taskQueue = [],
+}) => {
+  const taskQueueAnalytics = useTaskQueueAnalytics();
+  const tableContainerRef = useRef<HTMLDivElement>(null);
+  const [selectedRowIds, setSelectedRowIds] = useState<string[]>([]);
```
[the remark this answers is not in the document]
khelif96: Good catch Idk why i did this. I was transforming the number to a string to make it fit this type 🤦

src/pages/taskQueue/TaskQueueTable/index.tsx
```diff
@@ -0,0 +1,168 @@
...
+const StyledBaseTable = styled(BaseTable)`
+  flex-grow: 1;
+`;
+
+const TaskCell = styled.div`
+  display: flex;
+  flex-direction: column;
+`;
+
+const taskQueueTableColumns = (
```
minnakt, line 84: it looks like the search icons in these column headers don't have any actual functionality, so they could be removed

src/pages/taskQueue/index.tsx
```diff
@@ -1,120 +1,137 @@
...
+    onError: (err) => {
+      dispatchToast.error(
+        `There was an error loading task queue: ${err.message}`,
+      );
+    },
+  });
+
+  const onChangeDistroSelection = (val: TaskQueueDistro) => {
     taskQueueAnalytics.sendEvent({ name: "Select Distro", distro: val.id });
+    navigate(getTaskQueueRoute(val.id, taskId));
```
minnakt, line 76: I think maybe you don't want to carry over the taskID, since it'll be a different distro

src/pages/taskQueue/index.tsx
```diff
@@ -1,120 +1,137 @@
...
+      const { taskQueueDistros } = data;
+      const firstDistroInList = taskQueueDistros[0]?.id;
       const defaultDistro = distro ?? firstDistroInList;
-      setSelectedDistro(distros.find((d) => d.id === defaultDistro));
-      navigate(getTaskQueueRoute(defaultDistro, taskId), { replace: true });
-    }
-  }, [firstDistroInList, distro, navigate, taskId, distros]);
+      setSelectedDistro(taskQueueDistros.find((d) => d.id === defaultDistro));
+      if (distro === undefined) {
+        navigate(getTaskQueueRoute(defaultDistro, taskId));
```
minnakt, line 52: I think you can remove taskID from here as well since it's navigating to `firstDistroInList`

src/pages/taskQueue/index.tsx
```diff
@@ -1,120 +1,137 @@
...
     taskQueueAnalytics.sendEvent({ name: "Select Distro", distro: val.id });
+    navigate(getTaskQueueRoute(val.id, taskId));
+    setSelectedDistro(val);
   };
 
-  const handleSearch = (options: { id: string }[], match: string) =>
+  const handleSearch = (options: TaskQueueDistro[], match: string) =>
     options.filter((d) => d.id.toLowerCase().includes(match.toLowerCase()));
 
+  console.log({ selectedDistro, loadingDistrosData });
```
minnakt, line 83:
```suggestion
```

src/pages/taskQueue/index.tsx
```diff
@@ -1,120 +1,137 @@
...
+        navigate(getTaskQueueRoute(defaultDistro, taskId));
+      }
+    },
+    onError: (err) => {
+      dispatchToast.error(`There was an error loading distros: ${err.message}`);
+    },
+  });
 
-  const onChangeDistroSelection = (val: { id: string }) => {
+  const { data: taskQueueItemsData, loading } = useQuery<
```
minnakt, line 60: nit: could we name the loading variable here as well? (will be easier to tell which query it belongs to)

minnakt (changes requested):
ready to approve after this one! great stuff

src/components/Table/BaseTable.tsx
```diff
@@ -72,11 +78,15 @@ export const BaseTable = forwardRef(
...
-
     return (
       <>
         <StyledTable data-cy={dataCyTable} table={table} ref={ref} {...args}>
-          <TableHead>
+          <TableHead
+            isSticky={hasVirtualRows}
+            className={css`
+              z-index: 1;
+            `}
```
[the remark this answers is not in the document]
khelif96: Good catch I think LG fixed it when they were fixing the virtualization bugs

src/pages/taskQueue/TaskQueueTable/index.tsx
```diff
@@ -0,0 +1,168 @@
...
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [taskId, taskQueue]);
+
+  return (
+    <StyledBaseTable
+      data-cy="task-queue-table"
+      table={table}
+      shouldAlternateRowColor
+      emptyComponent={<TablePlaceholder message="No tasks found in queue." />}
```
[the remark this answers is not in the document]
khelif96: I tried I think the table component creates an empty container for the virtual rows. I was unable to find a way around this. I'll reach out to the lg team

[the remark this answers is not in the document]
minnakt: we can just file a ticket to check this out later 🙌

src/pages/taskQueue/TaskQueueTable/index.tsx
```diff
@@ -0,0 +1,169 @@
...
+  const performedInitialScroll = useRef(false);
+  useEffect(() => {
+    if (taskId !== undefined && !performedInitialScroll.current) {
+      const i = taskQueue.findIndex((t) => t.id === taskId);
+      setSelectedRowIndexes([i]);
+      table.scrollToIndex(i, { align: "start" });
+
+      setTimeout(() => {
+        performedInitialScroll.current = true;
+        table.scrollToIndex(i, { align: "start" });
```
minnakt, line 58: nit. not a big deal but `"center"` could be preferable to prevent the problem where the `start` doesn't seem to properly take into account the sticky header height (but also I don't really mind either way)
Screenshot 2024-03-04 at 2 12 16 PM

src/pages/taskQueue/TaskQueueTable/index.tsx
```diff
@@ -0,0 +1,169 @@
...
+  const [selectedRowIndexes, setSelectedRowIndexes] = useState<number[]>([]);
+  const columns = useMemo(
+    () => taskQueueTableColumns(taskQueueAnalytics.sendEvent),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [],
+  );
+
+  const table = useLeafyGreenTable<TaskQueueColumnData>({
+    data: taskQueue,
+    columns,
```
minnakt, line 41: nit. do you know if there's a good way to make the column sizes fixed so they don't keep resizing if you scroll? No worries if not, it's a constant problem I have with the table component in general. 😓

minnakt (approved):
yay! 💯

src/pages/taskQueue/TaskQueueTable/index.tsx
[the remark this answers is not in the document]
khelif96: https://jira.mongodb.org/browse/DEVPROD-5337

PR #2205, opened 2023-12-27
DEVPROD-3283 Support virtualization for BaseTable component

khelif96:
DEVPROD-3283

### Description
It appears the LG table does not properly scale to handle tables that are over 1k rows. This PR implements virtualization so that the table can now render virtualized rows.

### Screenshots
image


src/components/Table/BaseTable.tsx
```diff
@@ -51,127 +55,171 @@ type SpruceTableProps = {
...
+      className,
+      "data-cy-row": dataCyRow,
+      "data-cy-table": dataCyTable,
+      emptyComponent,
+      loading,
+      loadingRows = 5,
+      table,
+      ...args
+    }: SpruceTableProps & TableProps<any>,
+    ref: ForwardedRef<HTMLDivElement>
```
sophstad, line 73: What exactly is the need for passing the ref as a component on the prop? As best I can tell, LG [supports](https://github.com/mongodb/leafygreen-ui/tree/main/packages/table#usevirtualscrolling) virtualized scrolling, and already requires a ref via [the `containerRef` prop](https://github.com/mongodb/leafygreen-ui/tree/main/packages/table#useleafygreentable), and so we may not need to deal with the ref ourselves.

src/components/Table/BaseTable.tsx
```diff
@@ -51,127 +55,171 @@ type SpruceTableProps = {
...
-              &[aria-hidden="false"] td > div {
-                max-height: unset;
-              }
-            `}
-          >
-            {row.getVisibleCells().map((cell) => (
-              <Cell key={cell.id} style={{ padding: `${size.xxs} 2px` }}>
-                {flexRender(cell.column.columnDef.cell, cell.getContext())}
-              </Cell>
+export const BaseTable = forwardRef(
```
sophstad, line 61: I believe the typing issues we discussed came from an inability to use generics with `forwardRef`. I think removing the `forwardRef` is the cleanest option, but if you see a strong need for it, I have some more ideas from [this article](https://fettblog.eu/typescript-react-generic-forward-refs/) that we could discuss!

src/components/Table/BaseTable.tsx
khelif96: The ref wasn't actually being passed to the underlying table component. [forwardRef](https://react.dev/reference/react/forwardRef) is required in order for the ref to be passed to the component. Without it the `containerRef` prop wouldn't have any effect and it is required for the leafygreen tables virtualization functionality.

src/components/Table/BaseTable.tsx
khelif96: I responded above on why forwardRef is required. Do you have any preference on which of those implementations we should use?

sophstad: I think Option 2 with a wrapper component seems pretty straightforward. But whatever you like works for me!

khelif96: Discussed this offline. It doesn't really make sense to type this with generics since this component is meant to be generic and reusable and is agnostic of any data type passed into it.

src/components/Table/BaseTable.tsx
```diff
@@ -51,127 +55,171 @@ type SpruceTableProps = {
   loading?: boolean;
   /** estimated number of rows the table will have */
   loadingRows?: number;
+  className?: string;
```
sophstad, line 58: [nit] Maintain alphabetization here

src/components/Table/BaseTable.tsx
```diff
@@ -51,127 +55,171 @@ type SpruceTableProps = {
...
+        </StyledTable>
+        {!loading &&
+          rows.length === 0 &&
+          (emptyComponent || "No data to display")}
+      </>
+    );
+  }
+);
+
+const RenderRows = <T extends LGRowData>({
```
sophstad, line 157: [nit] Slight preference to include this as a ternary in BaseTable instead of its own component, but your call.

src/components/Table/BaseTable.tsx
```diff
@@ -51,127 +55,171 @@ type SpruceTableProps = {
...
 
-    {!loading &&
-      table.getRowModel().rows.length === 0 &&
-      (emptyComponent || "No data to display")}
-  </>
+const RenderableRow = <T extends LGRowData>({
+  row,
+  virtualRow,
+}: {
+  row: LeafyGreenTableRow<LGTableDataType<T>>;
```
sophstad, line 184: I think the typing here should just be
```suggestion
  row: LeafyGreenTableRow<T>;
```
(see [LeafyGreen](https://github.com/mongodb/leafygreen-ui/blob/9a35ec367bb88f471082d58c535937e2b2be9d7f/packages/table/src/ExpandedContent/ExpandedContent.spec.tsx#L39))

sophstad (changes requested):

src/components/Table/BaseTable.tsx
```diff
@@ -45,6 +48,7 @@ declare module "@tanstack/table-core" {
 }
 
 type SpruceTableProps = {
+  className?: string;
```
sophstad, line 51: Would `className` maybe captured by [this declaration](https://github.com/mongodb/leafygreen-ui/blob/9a35ec367bb88f471082d58c535937e2b2be9d7f/packages/table/src/Table/Table.types.ts#L7) in `TableProps` and subsequently `args` below? If removing this errors though then LGTM!

sophstad (approved):

src/components/Table/BaseTable.tsx
khelif96: good catch!

PR #2206, opened 2023-12-27
DEVPROD-1122 Add stepback information to task metadata

ZackarySantana:
DEVPROD-1122

### Description
This adds and propogates the "stepbackInfo" field in the Task struct to the task page's metadata card on the left hand side.

### Screenshots
I explain in the video:

https://github.com/evergreen-ci/spruce/assets/64446617/52ce85dd-a9c9-4439-8ac5-9a34b89f1eee

### Testing
Tested via staging. I considered some tests for it but I'm not sure if it is worth testing, since it is only links from the task data. I thought it might be more baggage to test it via cypress than the returns- but if the UI team feels otherwise, I won't hesitate to add them! I don't hold this opinion strongly, it's more so I don't want to add roi tests

### Evergreen PR
[Evergreen PR](https://github.com/evergreen-ci/evergreen/pull/7341)

### DIFF
--- a/src/analytics/task/useTaskAnalytics.ts
+++ b/src/analytics/task/useTaskAnalytics.ts
@@ -70,6 +70,10 @@ type Action =
     }
   | { name: "Click Trace Link" }
   | { name: "Click Trace Metrics Link" }
+  | { name: "Click Last Passing Stepback Task Link" }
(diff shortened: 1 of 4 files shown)


ZackarySantana:
I'm not sure what to do about the cypress/flake. I don't think my code caused it

PR #2212, opened 2024-01-03
DEVPROD-802 DEVPROD-2811 Upgrade storybook and fix snapshot tests

khelif96:
DEVPROD-802
DEVPROD-2811
### Description
Upgrades storybook to latest stable release.
Snapshot testing is no longer officially supported and we instead needed to make our own testing fixture.


sophstad (commented):
Haven't looked at the code yet but wanted to say... I've never gotten any useful signal from a snapshot test 🤷‍♀️ most of the time, they just indicate a styling update via a semantically meaningless className change.

I'm curious about situations where you've found them useful, as well as input from the rest of the team!

khelif96:
> Haven't looked at the code yet but wanted to say... I've never gotten any useful signal from a snapshot test 🤷‍♀️ most of the time, they just indicate a styling update via a semantically meaningless className change.
> 
> I'm curious about situations where you've found them useful, as well as input from the rest of the team!

In the past I've found them useful for determining if a change has any downstream effects on other components that consume them. For example, we make a change in the padding of an icon and some other component that consumes it no longer looks the same way. It was easy to identify that based off on the diff. I.e. I changed the icon but the button component now also changed its snapshot so maybe I should look at it.

 I do agree that they are less useful now that `classNames` are randomly generated and that they seem to all update each other despite no visual changes. I don't feel to strongly if the team does not find them valuable. Or possibly we can find some other way of determining the impact of changes without relying on classnames

khelif96:
> @khelif96 @sophstad We could give [visual testing](https://storybook.js.org/docs/writing-tests/visual-testing) a try for the next round table. I think that could be a lot more useful than markup diffing.

That sounds great. I'll add it to the agenda.

sophstad:
I like that idea too. Would we want to consider holding off on this until we've discussed visual diffs?

khelif96:
> I like that idea too. Would we want to consider holding off on this until we've discussed visual diffs?

I'd like to merge this in and also I think the outcome of that would be a separate set of work either way. This pr includes storybook upgrades and just gets the snapshot tests back to a working state.

src/storybook.test.ts
sophstad, line 1: [nit] Maybe `snapshots.test.ts` would be a better name for this file

sophstad (approved):
Works for me. Maybe we can revisit removing these if we have success down the line with visual diffs as I still find the frustration:signal ratio to be quite high. But nice work!

khelif96:
> Works for me. Maybe we can revisit removing these if we have success down the line with visual diffs as I still find the frustration:signal ratio to be quite high. But nice work!

I totally agree. Yeah I think if we like using visual diffs we can eliminate the snapshot tests.

PR #2218, opened 2024-01-09
DEVPROD-823 Remove terminated status from hosts table filters

khelif96:
DEVPROD-823

### Description
Removes the terminated status so it's no longer a filterable option on the hosts table since it is omitted by the underlying query

### Screenshots
image


minnakt (approved):
woo!

PR #2228, opened 2024-01-17
DEVPROD-2920 Enable access settings page tests

khelif96:
DEVPROD-2920

### Description
I tried running these 10 times and could not get it to flake so it possibly went away?


sophstad (approved):
🤞

PR #2234, opened 2024-01-22
Dependabot: Add team as reviewer

sophstad:
### Description

- I sometimes forget to review dependabot PRs on rotation because they don't show up in my inbox without a request from `evg-ui`. I figured adding the team would be helpful to flag these PRs, and then the rotator can change the reviewer to their own user.
- [Config file docs](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#reviewers)


khelif96 (approved):
Great idea! Can we do this on Parsley too?

PR #2238, opened 2024-01-24
DEVPROD-801 Remove axios 

khelif96:
DEVPROD-801

### Description
Remove axios as a dependency and rewrite the function it depends on using native fetch.

### Testing
Tested that the one place this is used still functions and is able to successfully reset API keys


src/utils/request.ts
```diff
@@ -1,44 +1,36 @@
-import axios from "axios";
 import { getUiUrl } from "./environmentVariables";
 import { reportError } from "./errorReporting";
 
 type optionsType = {
-  onFailure?: (e) => void;
+  onFailure?: (e: Error) => void;
```
minnakt, line 5: Do we need to support this option? The one place that calls this function doesn't use it, and I don't imagine we will use this function in other places

```suggestion
```

minnakt (approved):
this lgtm for getting rid of axios! I was just wondering, should we make a ticket to turn this into a GraphQL mutation or was there some reason we needed to hit the legacy API directly?

PR #2239, opened 2024-01-24
DEVPROD-826 Query hasCedarResults instead of failedTestCount

khelif96:
DEVPROD-826

### Description
Previously we would query cedar to determine if a task has test results in order to determine if we should attempt to load them on the UI. This is a slow operation since it involved thousands of requests to cedar before we could return anything on the UI. This pr instead relies on a field that is already cached in the db.

### Testing
Tested locally and existing tests test that this still works.

### Evergreen PR
https://github.com/evergreen-ci/evergreen/pull/7444


khelif96:
evergreen retry

minnakt (approved):

PR #2241, opened 2024-01-24
DEVPROD-3533: Add admin-only distro setting

sophstad:
DEVPROD-3533

### Description

- Add field on general distro settings tab to mark distro as admin-only. This has no effect at present, but DEVPROD-3532 will hide these distros from user-facing menus, like distro selection when spawning hosts.

### Screenshots

image

### Testing

- Update unit and Cypress tests

### Evergreen PR

https://github.com/evergreen-ci/evergreen/pull/7376

### DIFF
--- a/src/pages/distroSettings/tabs/GeneralTab/getFormSchema.ts
+++ b/src/pages/distroSettings/tabs/GeneralTab/getFormSchema.ts
@@ -39,6 +39,11 @@ export const getFormSchema = (
         type: "object" as "object",
         title: "Distro Options",
         properties: {
+          adminOnly: {
+            type: "boolean" as "boolean",
+            title: "Admin only",
+            default: false,
+          },
           isCluster: {
             type: "boolean" as "boolean",
             title: "Mark distro as cluster",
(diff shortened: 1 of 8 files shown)


src/pages/distroSettings/tabs/GeneralTab/getFormSchema.ts
```diff
@@ -88,6 +93,10 @@ export const getFormSchema = (
     },
     distroOptions: {
       "ui:ObjectFieldTemplate": CardFieldTemplate,
+      adminOnly: {
+        "ui:description":
+          "Admin-only distros are not selectable by general users (e.g. when spawning a host). They do not have their access controlled beyond being hidden.",
```
minnakt, line 98: the example `"when spawning a host"` is somewhat confusing to me because I think we already hide distros in the spawn host modal through the `userSpawnAllowed` attribute. Unless maybe `adminOnly` is going to take precedence over `userSpawnAllowed` in the next PR? 👀 otherwise LGTM!

minnakt (approved):

src/pages/distroSettings/tabs/GeneralTab/getFormSchema.ts
sophstad: Thomas clarified that there are lots of these temporary distros where `userSpawnEnabled` is true. The next PR will have admin-only take precedence over that setting, but will still allow these distros to be visible to admins in the dropdown.

PR #2242, opened 2024-01-24
DEVPROD-3935 Use correct query for project health view setting

khelif96:
DEVPROD-3935

### Description
We have logged nearly 7k events to sentry due to a failing permission check caused by the `projectHealthView` query consistently failing. It was trying to use the `ProjectSettings` query to fetch the `projectHealthView` state
Since most users don't have project settings access this query always fails and will consistently report an error.
This pr updates it so that we use a query that does not require the same permissions.

### Testing
Since I am an admin it is difficult to test but since the query does not have a permissions check directive this should work.

### DIFF
--- a/src/pages/commits/ViewToggle.tsx
+++ b/src/pages/commits/ViewToggle.tsx
@@ -42,7 +42,7 @@ export const ViewToggle: React.FC<Props> = ({ identifier }) => {
 
   useEffect(() => {
     if (!view) {
-      setView(data?.project.projectHealthView);
+      setView(data?.project?.projectHealthView);
     }
   }, [data, setView, view]);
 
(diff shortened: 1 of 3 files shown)


src/pages/commits/ViewToggle.tsx
```diff
@@ -42,7 +42,7 @@ export const ViewToggle: React.FC<Props> = ({ identifier }) => {
 
   useEffect(() => {
     if (!view) {
-      setView(data?.projectSettings?.projectRef?.projectHealthView);
+      setView(data?.project.projectHealthView);
```
sophstad, line 45:
```suggestion
      setView(data?.project?.projectHealthView);
```

sophstad (approved):
Thank you!

PR #2243, opened 2024-01-25
DEVPROD-788: Hide Slack banner more reliably

sophstad:
DEVPROD-788

### Description

- Don't show Slack banner if user has already added their Slack username. I think if they've done so it's safe to assume they have used notifications and don't need prompting.


khelif96 (approved):

sophstad:
evergreen retry

PR #2252, opened 2024-02-06
DEVPROD-3295 update volume size default

ablack12:
DEVPROD-3295
### Description
500 GB is supposed to be the max volume size that users can request, not the default that we suggest. Infra suggested that we start recommending 256 as a power of two.


src/components/Spawn/spawnHostModal/getFormSchema.tsx
```diff
@@ -408,7 +408,7 @@ export const getFormSchema = ({
                       volumeSize: {
                         title: "Volume size (GB)",
                         type: "number" as "number",
-                        default: 500,
```
sophstad, line 411: Would prefer to update this constant and then use that variable in place of this magic number! https://github.com/evergreen-ci/spruce/blob/de2fac705d2400ea62c68708a3ee8ebbc4b18a98/src/components/Spawn/spawnHostModal/constants.ts#L1

sophstad (changes requested):
I think a test in `src/pages/spawn/spawnVolume/SpawnVolumeModal.test.tsx` may need updating as well

src/components/Spawn/spawnHostModal/getFormSchema.tsx
ablack12: perf thanks!

ablack12:
I'm not sure what to look at for the cypress/flake, can you let me know if there's something to be done here?

sophstad (approved):
LGTM! No worries about the flake

PR #2253, opened 2024-02-08
DEVPROD-3919 Don't show eta timer if the eta has been exceeded.

khelif96:
DEVPROD-3919

### Description
Fixes a bug where the eta timer would turn negative if the eta was exceeded.


SupaJoon (approved):
Yee haw!! 🤠

PR #2262, opened 2024-02-15
DEVPROD-356: Add Sentry grouping

sophstad:
DEVPROD-356

### Description

Added two new fields to GraphQL errors that should help with grouping in the app:
- Adding a [fingerprint](https://docs.sentry.io/platforms/javascript/usage/sdk-fingerprinting/) based on the operation name should force these errors to be grouped together. I've also appended the path to the fingerprint, which means that errors will only be grouped if the same field was the cause of the error. This is a little bit more aggressive than just grouping on operation name, but I think it will be helpful
- Adding operation name as a [tag](https://docs.sentry.io/platforms/javascript/guides/react/enriching-events/tags/) allows it to be searchable.

### Testing

- The [event grouping information](https://mongodb-org.sentry.io/issues/4981024769/?environment=staging&project=4505211042856960&referrer=issue-stream&statsPeriod=7d&stream_index=2#grouping-info) for this event shows the custom fingerprint in use.
- Clicking "All Events" at the top of that page allows you to see all errors that were thrown by `UserProjectSettingsPermissions` being thrown with an error in the project identifier.
- `operationName` is now a tag appearing at the top of the page:
  image

I think this is a pretty logical way of grouping things. You can also take a look on staging and see how incoming errors are grouped. Sentry's "recommended error" view is still confusing to me, and the error titles are still somewhat useless, but I think the grouping is improved.

### DIFF
--- a/src/components/ErrorHandling/Sentry.tsx
+++ b/src/components/ErrorHandling/Sentry.tsx
@@ -1,14 +1,14 @@
-import { ErrorResponse } from "@apollo/client/link/error";
 import {
   captureException,
   ErrorBoundary as SentryErrorBoundary,
   getCurrentHub,
   init,
   Replay,
-  setTag,
+  setTags,
   withScope,
 } from "@sentry/react";
 import type { Scope, SeverityLevel } from "@sentry/react";
+import type { Context, Primitive } from "@sentry/types";
 import { environmentVariables } from "utils";
 import ErrorFallback from "./ErrorFallback";
 
@@ -40,33 +40,32 @@ const initializeSentry = () => {
 
 const isInitialized = () => !!getCurrentHub().getClient();
 
-export type ErrorMetadata = {
-  gqlErr?: ErrorResponse["graphQLErrors"][0];
-  operationName?: ErrorResponse["operation"]["operationName"];
-  variables?: ErrorResponse["operation"]["variables"];
+export type ErrorInput = {
+  err: Error;
+  fingerprint?: string[];
+  context?: Context;
+  severity: SeverityLevel;
+  tags?: { [key: string]: Primitive };
 };
 
-const sendError = (
-  err: Error,
-  severity: SeverityLevel,
-  metadata?: ErrorMetadata,
-) => {
+const sendError = ({
+  context,
+  err,
+  fingerprint,
+  severity,
+  tags,
+}: ErrorInput) => {
   withScope((scope) => {
-    setScope(scope, { level: severity, context: metadata });
+    setScope(scope, { level: severity, context });
 
-    const { gqlErr, operationName } = metadata ?? {};
-
-    // Add additional sorting for GraphQL errors
-    if (operationName) {
+    if (fingerprint) {
       // A custom fingerprint allows for more intelligent grouping
-      const fingerprint = [operationName];
-      if (gqlErr?.path && Array.isArray(gqlErr.path)) {
-        fingerprint.push(...gqlErr.path);
-      }
       scope.setFingerprint(fingerprint);
+    }
 
-      // Apply tag, which is a searchable/filterable property
-      setTag("operationName", operationName);
+    if (tags) {
+      // Apply tags, which are a searchable/filterable property
+      setTags(tags);
     }
 
     captureException(err);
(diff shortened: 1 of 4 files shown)


src/components/ErrorHandling/Sentry.tsx
```diff
@@ -46,6 +47,21 @@ const sendError = (
   withScope((scope) => {
     setScope(scope, { level: severity, context: metadata });
 
+    const { gqlErr, operationName } = metadata ?? {};
```
khelif96, line 50: Can we improve the type safety on these by adding `gqlErr` and `operationName` as option fields in the metadata struct on line 46.

src/components/ErrorHandling/Sentry.tsx
```diff
@@ -46,6 +47,21 @@ const sendError = (
...
+    if (operationName) {
+      // A custom fingerprint allows for more intelligent grouping
+      const fingerprint = [operationName];
+      if (gqlErr?.path && Array.isArray(gqlErr.path)) {
+        fingerprint.push(...gqlErr.path);
+      }
+      scope.setFingerprint(fingerprint);
+
+      // Apply tag, which is a searchable/filterable property
+      setTag("operationName", metadata.operationName);
```
khelif96, line 62:
```suggestion
      setTag("operationName", operationName);
```

khelif96 (approved):
Great job investigating this! I just had a few comments but it looks great to me and I'm excited to see how it looks in production

khelif96:
Can we also make these changes in parsley

src/components/ErrorHandling/Sentry.tsx
```diff
@@ -39,10 +40,16 @@ const initializeSentry = () => {
...
+  gqlErr?: ErrorResponse["graphQLErrors"][0];
+  operationName?: ErrorResponse["operation"]["operationName"];
+  variables?: ErrorResponse["operation"]["variables"];
+};
+
 const sendError = (
   err: Error,
   severity: SeverityLevel,
-  metadata?: { [key: string]: any },
+  metadata?: ErrorMetadata,
```
khelif96, line 52: This typing is technically broken since https://github.com/evergreen-ci/spruce/blob/45550f1a6d19b1825ec5beeed7f8ada29c857f99/src/utils/errorReporting.ts#L10 passes a generic `{ [key:string]: any }` so this type is never actually used.

khelif96: It is also not flexible enough to allow for any metadata to be passed in once the typing in reportError is fixed.

src/components/ErrorHandling/Sentry.tsx
```diff
@@ -39,10 +40,16 @@ const initializeSentry = () => {
 
 const isInitialized = () => !!getCurrentHub().getClient();
 
+type ErrorMetadata = {
+  gqlErr?: ErrorResponse["graphQLErrors"][0];
+  operationName?: ErrorResponse["operation"]["operationName"];
+  variables?: ErrorResponse["operation"]["variables"];
```
khelif96, line 46: These are all actually typed from
`import { GraphQLError } from "graphql";` since they are properties of the GraphQL request and not specific to Apollo.

khelif96 (changes requested):

src/components/ErrorHandling/Sentry.tsx
sophstad: This is the typing of Apollo's onError handlers if I'm not mistaken: https://github.com/apollographql/apollo-client/blob/17ccc42825d0b0355212f3ac823780edb467ea77/src/link/error/index.ts#L28

src/components/ErrorHandling/Sentry.tsx
sophstad: Updated the typing in `errorReporting.ts`. If you want strict typing I think we should just go all the way and update this type as more metadata fields are added rather than allow the type to be flexible.

khelif96 (commented):
We discussed offline and we will make this more generic to support general grouping in the future. Feel free to re-request a review when you are ready @sophstad

khelif96 (approved):
LGTM 🚀

khelif96:
Can we do this on Parsley as well please! @sophstad

PR #2267, opened 2024-02-20
DEVPROD-933 Trim trailing and ending whitespace for project settings fields

khelif96:
DEVPROD-933

### Description
Creates a new validator which validates that white space was trimmed from inputs on the project settings page. I applied this to most inputs that I felt made sense. If there are any that I missed that you think should be included lmk please.

### Screenshots
image


khelif96:
> Why not apply white space formatting during request payload generation so we don't block the user?

I opted for this approach since it is more implicit to the user that these characteristics are required for certain fields (and not all).  I think transforming a field a user modifies without their knowledge is an anti-pattern especially if it doesn't happen to all fields.

khelif96:
> > > Why not apply white space formatting during request payload generation so we don't block the user?
> > 
> > 
> > I opted for this approach since it is more implicit to the user that these characteristics are required for certain fields (and not all). I think transforming a field a user modifies without their knowledge is an anti-pattern especially if it doesn't happen to all fields.
> 
> Sounds good. A UX improvement could be to be explicit about whether the input has trailing or leading whitespace in the error message.

I couldn't find a way to do this since we don't get access to the contents when we generate the error message but it should be pretty simple to determine which side the whitespace is on since there are only 2 options.

PR #2282, opened 2024-03-01
DEVPROD-5132: Make tags unique

sophstad:
DEVPROD-5132

### Description

- Ahead of the monorepo merge, make Spruce's tags unique by using `spruce/vX.Y.Z`. The commit message has also been changed from "vX.Y.Z" to "spruce/vX.Y.Z"
- Remove references to remote URL. We can assume users have their upstream properly configured (and if they don't commands will fail, which is fine)
- Added a new git tag regex to Spruce's project settings. When this is merged I will delete the old one:
  image

### Testing

- Tested with my wifi off 🤣 the correct tag names were generated and git attempted to push to the correct place as part of the postversion script!

### DIFF
--- a/.yarnrc
+++ b/.yarnrc
@@ -1,0 +1,2 @@
+version-tag-prefix spruce/v
+version-git-message "spruce/v%s"
--- a/scripts/deploy/utils/git/tag/constants.ts
+++ b/scripts/deploy/utils/git/tag/constants.ts
@@ -1,1 +1,0 @@
-export const githubRemote = "https://github.com/evergreen-ci/spruce";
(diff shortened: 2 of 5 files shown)


.yarnrc
```diff
@@ -0,0 +1,2 @@
+version-tag-prefix spruce/v
```
minnakt, line 1: (comment placement is random) I think you'll want to use the Variant Regex + Task Regex combination instead of the Variant Tags + Task Tags combination! I don't think we use tags in our yaml

minnakt (approved):
I tested with my wifi off too 🫣

.yarnrc
sophstad: Thank you, great catch! I made this update in Spruce project settings.

PR #2284, opened 2024-03-01
DEVPROD-4509: Exclude "secret fields" in Sentry Breadcrumbs

SupaJoon:
DEVPROD-4509

### Description
These code changes query SecretFields and use them to clean up GQL variables that are sent to Sentry. Since Sentry breadcrumbs are sent within an Apollo Link which is set during Apollo Client initialization, these code changes block Apollo Client initialization until SecretFields are fetched. 
These code changes also update `Login.tsx` to utilize `fetch` instead of the Apollo client. If Apollo cannot be initialized because of failure to fetch Secret Fields, then the user is redirected to the Login page where Apollo cannot be initialized. The login page is used in the dev environment only.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "spruce",
-  "version": "3.0.227",
+  "version": "3.0.228",
   "private": true,
   "scripts": {
     "bootstrap-logkeeper": "./scripts/bootstrap-logkeeper.sh",
--- a/src/components/Header/Navbar.tsx
+++ b/src/components/Header/Navbar.tsx
@@ -185,7 +185,6 @@ const SecondaryLink = styled.a`
 `;
 
 const HybridTreeIcon = styled.img`
-  height: 70px;
-  width: 70px;
+  height: 72px;
   position: relative;
 `;


src/utils/request.ts
```diff
@@ -9,19 +13,50 @@ export const post = async (url: string, body: unknown) => {
...
 const handleError = (error: string) => {
   reportError(new Error(error)).warning();
 };
+
+export const fetchWithRetry = <T = any>(
+  url: string,
+  options: RequestInit,
+  retries: number = 3,
+  backoff: number = 150,
+): Promise<T> =>
```
khelif96, line 38: I think the typing for this should actually be 
```suggestion
): Promise<{ data: T }> =>
```

src/hooks/useCreateGQLClient/index.ts
```diff
@@ -0,0 +1,65 @@
...
+
+const { getGQLUrl } = environmentVariables;
+
+export const useCreateGQLCLient = (): ApolloClient<NormalizedCacheObject> => {
+  const { dispatchAuthenticated, logoutAndRedirect } = useAuthDispatchContext();
+  const [secretFields, setSecretFields] = useState<string[]>();
+  const [gqlClient, setGQLClient] = useState<any>();
+
+  useEffect(() => {
+    fetchWithRetry(getGQLUrl(), secretFieldsReq)
```
khelif96, line 25: With the above/below? suggestion you can add some type safety here by using the generic 
```suggestion
    fetchWithRetry<SecretFieldsQuery>(getGQLUrl(), secretFieldsReq)
```

src/utils/request.test.ts
khelif96, line 1: Great tests!

src/gql/fetch/index.ts
```diff
@@ -0,0 +1,17 @@
+import { OperationDefinitionNode } from "graphql";
+import { SECRET_FIELDS } from "gql/queries";
+
+export const secretFieldsReq = {
```
khelif96, line 4: nit for a little bit more type safety
```suggestion
export const secretFieldsReq: RequestInit = {
```

khelif96 (commented):
Great work here it works really well!  I just had a few suggestions

src/utils/object/omit.ts
```diff
@@ -1,8 +1,10 @@
+import { deleteNestedKey } from "./deleteNestedKey";
+
 export const omit = <T extends object, K extends [...(keyof T)[]]>(
   obj: T,
   params: K,
 ) => {
   const newObj = { ...obj };
-  params.forEach((param) => delete newObj[param]);
+  deleteNestedKey(newObj, params as string[]);
   return newObj as Omit<T, K[number]>;
```
khelif96, line 9: The typing here is no longer valid it does not handle nested keys.

src/utils/object/deleteNestedKey.ts
```diff
@@ -0,0 +1,40 @@
...
+ * @param keyToUpdate The name of key(s) to be removed or updated.
+ * @param redactedString The string to replace the key with if provided.
+ * @returns The object with the key removed.
+ */
+export function deleteNestedKey<T extends object>(
+  obj: T,
+  keyToUpdate: string | string[],
+  redactedString?: string,
+): Partial<T> {
+  const deleteKey = (currentObject: any) => {
```
khelif96, line 20:
```suggestion
  const deleteKey = (currentObject: { [key:string]: any}) => {
```

src/utils/object/omit.ts
khelif96: I think it might makes sense to keep these utility functions separate since one handles simple key value maps and the other handles deeply nested objects. What do you think?

khelif96 (commented):
Looks great! sorry I realized theres one more spot where we need to redact variables.
https://github.com/evergreen-ci/spruce/blob/21ff1bb3dcdca2cf56570e6d07435fbbf0a41a20/src/gql/client/link/index.ts#L39

image

khelif96 (approved):
Great work on this! 🚀

PR #2290, opened 2024-03-06
DEVPROD-4193: Redirect to project identifier on Project Health page

minnakt:
DEVPROD-4193

### Description
This PR creates a new hook called `useProjectRedirect` which handles the redirect. I made it a bit general since I think it might be reused for other pages.

I also slightly tweaked the logic of how the `CURRENT_PROJECT` cookie is set so that it is never set to an Object ID.

### Testing
- Jest tests


src/components/Header/Navbar.tsx
```diff
@@ -33,16 +36,20 @@ export const Navbar: React.FC = () => {
...
 
-  // Update current project cookie if the project in the URL does not equal the cookie value.
+  // Update current project cookie if the project in the URL is not an objectId and is not equal
+  // to the current project.
   // This will inform future navigations to the /commits page.
   useEffect(() => {
-    if (projectFromUrl && projectFromUrl !== Cookies.get(CURRENT_PROJECT)) {
+    if (
+      projectFromUrl &&
+      !validateObjectId(projectFromUrl) &&
```
sophstad, line 47: Great improvement! I think it's possible this has a large impact on the issue altogether.

src/hooks/useProjectRedirect/index.ts
```diff
@@ -0,0 +1,36 @@
...
+  const { loading } = useQuery<ProjectQuery, ProjectQueryVariables>(PROJECT, {
+    skip: !needsRedirect,
+    variables: {
+      idOrIdentifier: project,
+    },
+    onCompleted: (projectData) => {
+      const { identifier } = projectData.project;
+      const currentUrl = location.pathname.concat(location.search);
+      const redirectPathname = currentUrl.replace(project, identifier);
+      navigate(redirectPathname);
```
sophstad, line 31: This could be overkill, but maybe adding an analytics event would be helpful to track whether we're still seeing this behavior frequently after the cookie change.

src/hooks/useProjectRedirect/index.ts
```diff
@@ -0,0 +1,36 @@
...
+    },
+  });
+
+  return { isRedirecting: needsRedirect && loading };
```
sophstad, line 35: This works great but would you mind explaining what `&& loading` does? Just curious

src/pages/commits/index.tsx
```diff
@@ -233,7 +235,10 @@ const Commits = () => {
             versions={versions}
             revision={revision}
             isLoading={
-              (loading && !versions) || !projectIdentifier || isResizing
+              (loading && !versions) ||
+              !projectIdentifier ||
+              isRedirecting ||
```
sophstad, line 240: [nit] Is this maybe overkill since we check `loading` for a function that is skipped if we're redirecting?

src/hooks/useProjectRedirect/index.ts
minnakt: for sure! I think the only linkable way to get to an ID link now is from the legacy UI 🫠

src/hooks/useProjectRedirect/index.ts
minnakt: in theory it should never really happen, but it handles the case where someone visits `/commits/{repoId}`! Without `&& loading`, `isRedirecting` will keep being true even though the repo could never redirect to an identifier

src/pages/commits/index.tsx
minnakt: good question, I checked it manually, and it looks like the screen will flicker without `isRedirecting` 😔

sophstad (approved):
wahoo!

minnakt:
failing tests on Hosts page seems unrelated

PR #2295, opened 2024-03-08
DEVPROD-750 remove unused dispatcher option

ablack12:
DEVPROD-750
### Description
I'm thinking it would be good to leave the dropdown on the UI page, in case we end up adding more things (we're likely to keep scaffolding in the backend as well), but if we'd prefer to remove it I can do that, and then maybe just verify that the resolvers still save correctly?

### Testing
Updated an existing test.

### Evergreen PR
https://github.com/evergreen-ci/evergreen/pull/7628


src/gql/generated/types.ts
```diff
@@ -347,7 +347,6 @@ export type DispatcherSettingsInput = {
 };
 
 export enum DispatcherVersion {
-  Revised = "REVISED",
```
minnakt, line 350: i think you may want to leave this file unchanged, I think we typically don't regenerate this file until a backend change is fully merged

minnakt (approved):
im okay with leaving the dropdown! most users cannot edit the page so they won't even be able to tell there's only 1 option 🫣

PR #2299, opened 2024-03-12
DEVPROD-4608 Add stepback to the metadata of tasks

ZackarySantana:
DEVPROD-4608

### Description
This adds the metadata item for stepback to show "In progress" or "completed" and an info icon accompanying it. I implemented according to the design except I added a "loading" state while the graphql query completes.

This also isolates and extracts the different relevant commit queries in to each their own hook: 'useParentTask', 'useLastPassingTask', 'useBreakingCommit', and 'useLastExecutedTask'. The first three are re-used (within each other, in the relevant commits dropdown, and the metadata card) while the last one is only used in the relevant commits dropdown- but it figured it would be better to extract and test that logic separately to be consistent (it removes bloat from the relevant commits dropdown component).

I separated my commits to be distinct so anyone who reviews can look over the commits for an easier time!

#### In progress
Screenshot 2024-03-12 at 12 01 31 PM

#### Completed
Screenshot 2024-03-12 at 12 02 29 PM

#### Hovering dropdown
Screenshot 2024-03-12 at 12 02 59 PM

### Testing
I tested the buttons on multiple different tasks on staging- before and after this change. All the buttons respond the exact same- and as for the metadata item I also tested on staging making sure it shows the correct state.

I plan to add similar testing for all the new hooks if the reviewer thinks the tests for the hook I added is good! Please let me know- it shouldn't be too long to add them!


ZackarySantana:
Could anyone on UI help me with the hooks tests? I defined the mocks and I'm using the `renderHook` with a wrapper that is a provider with the mocks- but it does not get any of the mocks.

Here's a quick view of it but it's available in the `useBreakingTask` hook and test file.

Test usage part:
```
   const { result } = renderHook(() => useBreakingTask("t1"), {
      wrapper: ({ children }) =>
        ProviderWrapper({
          children,
          mocks: [
            getPatchTaskWithFailingBaseTask,
            getLastPassingVersion,
            getBreakingCommit,
          ],
        }),
    });
 ```
 
 The mock I'm providing:
```
const getPatchTaskWithFailingBaseTask: ApolloMock<
  BaseVersionAndTaskQuery,
  BaseVersionAndTaskQueryVariables
> = {
  request: {
    query: BASE_VERSION_AND_TASK,
    variables: {
      taskId: "t1",
    },
  },
  result: {
    data: {
      task: {
      ...
  ```
  
  The query I'm expecting the mock to fill (but it prints undefined)
  ```
    const { data: taskData } = useQuery<
    BaseVersionAndTaskQuery,
    BaseVersionAndTaskQueryVariables
  >(BASE_VERSION_AND_TASK, {
    variables: { taskId },
  });

  const { buildVariant, displayName, projectIdentifier, status } =
    taskData?.task ?? {};
  console.log(taskId);
  console.log(taskData?.task);
  ```
  
  taskId properly puts out `t1` but the taskData?.task is undefined.

src/pages/task/metadata/Stepback.tsx
```diff
@@ -0,0 +1,75 @@
...
+ * @returns Whether the task is in stepback.
+ */
+export function inStepback(task: TaskQuery["task"]) {
+  // lastFailingStepbackTaskId is set only in the middle/last of stepback (not the first task).
+  const stepback =
+    task?.stepbackInfo?.lastFailingStepbackTaskId !== undefined &&
+    task?.stepbackInfo?.lastFailingStepbackTaskId !== "";
+
+  // nextStepbackTaskId is set when the next task in stepback in known, in the beginning
+  // of stepback, it is known right away. In the rest of stepback, it is not.
```
minnakt, line 28: nit. there might be a few typos in this comment e.g. `in stepback in known`. `in the beginning of...` seems like it should be its own sentence as well

src/pages/task/metadata/Stepback.tsx
```diff
@@ -0,0 +1,75 @@
...
+            <Icon glyph="InfoWithCircle" size="small" />
+          </IconContainer>
+        }
+        triggerEvent="hover"
+      >
+        <Body>
+          When Stepback is completed you can access the breaking commit via the
+          relevant commits dropdown.
+        </Body>
+      </Tooltip>{" "}
```
minnakt, line 64: You might be able to replace this block with the `InfoSprinkle` component

src/pages/task/metadata/Stepback.tsx
```diff
@@ -0,0 +1,75 @@
...
+      {loading && <Badge variant="lightgray">Loading</Badge>}
```
minnakt, line 65: Having `Loading` in a badge makes it seem like it's a state of stepback, rather than an indication that we're still querying for a response. Perhaps a [Skeleton Loader](https://www.mongodb.design/component/skeleton-loader/example/) or [Spinner](https://www.mongodb.design/component/loading-indicator/example/) could be a good replacement? (a skeleton loader is good for longer-term loading operations > 300ms, a spinner is good for shorter-term loading operations < 300ms)

minnakt (changes requested):
test structure lgtm 😎 feel free to add the rest! and great reorganization with the hooks!

ZackarySantana:
New loading state!:
Screenshot 2024-03-20 at 12 26 53 PM

Using new sprinkle, looks near identical:
Screenshot 2024-03-20 at 12 27 33 PM

src/pages/task/metadata/Stepback.tsx
ZackarySantana: It's hard to guage how fast this might take. I based it on relevant commits since it uses the same queries, and guessed on average it would be a longer operation so I did the skeleton loader for the badge!

src/pages/task/metadata/Stepback.tsx
```diff
@@ -0,0 +1,66 @@
...
+    <StepbackWrapper>
+      Stepback:
+      <InfoSprinkle baseFontSize={BaseFontSize.Body1}>
+        When Stepback is completed you can access the breaking commit via the
+        relevant commits dropdown.
+      </InfoSprinkle>
+      {loading && <Skeleton size="small" />}
+      {!loading && !finished && <Badge variant="lightgray">In progress</Badge>}
+      {!loading && finished && <Badge variant="green">Complete</Badge>}
+    </StepbackWrapper>
```
minnakt, line 57: My only reservation about this is that now this doesn't have the correct font styles from `MetadataItem` (if you look closely the font size is wrong). Perhaps we can change the underlying HTML of `MetadataItem` into a `<div>` and then rewrite it as the following:

```suggestion
   <MetadataItem>
      <StepbackWrapper>
        // content goes here
      </StepbackWrapper>
    </MetadataItem>
```

src/pages/task/actionButtons/RelevantCommits/index.tsx
```diff
@@ -37,101 +32,18 @@ export const RelevantCommits: React.FC<RelevantCommitsProps> = ({ taskId }) => {
...
-  const {
-    baseTask,
-    buildVariant,
-    displayName,
-    projectIdentifier,
-    status,
-    versionMetadata,
-  } = taskData?.task ?? {};
-  const { order: skipOrderNumber } = versionMetadata?.baseVersion ?? {};
+  const { baseTask, versionMetadata } = taskData?.task ?? {};
```
minnakt, line 35: I just realized that there doesn't seem to be a reason to perform the `BaseVersionAndTaskQuery` at all; it looks like the parent component of this component (`ActionButtons.tsx`) can pass these values down as props 🤔

minnakt (changes requested):
added tests look great!

src/hooks/useBreakingTask/useBreakingTask.test.tsx
```diff
@@ -0,0 +1,240 @@
...
+interface ProviderProps {
+  mocks?: MockedProviderProps["mocks"];
+  children: React.ReactNode;
+}
+const ProviderWrapper: React.FC<ProviderProps> = ({ children, mocks = [] }) => (
+  <MockedProvider mocks={mocks}>{children}</MockedProvider>
+);
+
+describe("useBreakingTask", () => {
+  it("no breaking task is found when task is not found", async () => {
```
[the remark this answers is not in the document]
ZackarySantana: Nice catch! I wonder why the linter didn't catch it...

src/pages/task/metadata/Stepback.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+export const Stepback: React.FC<Props> = ({ taskId }) => {
+  const { loading, task: breakingTask } = useBreakingTask(taskId);
+  const { task: lastPassingTask } = useLastPassingTask(taskId);
+
+  // The last stepback task has an undefined last passing task (it is passing itself).
+  const isLastStepbackTask = lastPassingTask === undefined;
+
+  // The stepback is finished if there is a breaking task or we are on the last stepback task.
+  const finished = breakingTask !== undefined || isLastStepbackTask;
```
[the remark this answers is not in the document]
ZackarySantana: If you are on the breaking task, stepback should still be counted as finished- the breaking task is when `lastPassingTask === undefined` is true. I'll redo the comment!

src/pages/task/ActionButtons.tsx
```diff
@@ -254,7 +254,7 @@ export const ActionButtons: React.FC<Props> = ({
       <PageButtonRow>
         {!isExecutionTask && (
           <>
-            <RelevantCommits taskId={taskId} />
+            <RelevantCommits taskId={taskId} task={task} />
```
minnakt, line 257: You can remove the `taskId` prop if it can be accessed from the `task`!

src/components/MetadataCard.tsx
```diff
@@ -39,14 +40,20 @@ export const MetadataTitle: React.FC<{ children: React.ReactNode }> = ({
...
   "data-cy"?: string;
 }
 
 export const MetadataItem: React.FC<ItemProps> = ({
+  as = "p",
   children,
   "data-cy": dataCy,
-}) => <Item data-cy={dataCy}>{children}</Item>;
+}) => (
+  <Item data-cy={dataCy} as={as}>
```
minnakt, line 53: This is cool! The only issue with this is that the margin from `:not(:last-of-type)` selector doesn't quite work as intended anymore due to the mix of `<p>` and `<div>` tags.

Maybe we can change `:not(:last-of-type)` -> `:not(:last-child)`

src/pages/task/metadata/Stepback.tsx
```diff
@@ -0,0 +1,70 @@
...
+          <Badge variant="lightgray">In progress</Badge>
+        )}
+        {!loading && finished && <Badge variant="green">Complete</Badge>}
+      </StepbackWrapper>
+    </MetadataItem>
+  );
+};
+
+const StepbackWrapper = styled.div`
+  margin-top: ${size.xxs};
```
minnakt, line 66: you can remove this `margin-top` style if you get the margin from the style change mentioned in the previous comment

src/pages/task/metadata/Stepback.tsx
```diff
@@ -0,0 +1,70 @@
...
+  // The 'lastFailingStepbackTaskId' is set for all stepback tasks except the first one.
+  const hasLastStepback =
+    task?.stepbackInfo?.lastFailingStepbackTaskId?.length > 0;
+
+  // The 'nextStepbackTaskId' is only set when the next task in stepback is running/finished.
+  // This happens in the beginning of stepback or them middle of stepback. This condition is
+  // covering for the beginning of stepback.
+  const isBeginningStepback =
+    task?.stepbackInfo?.nextStepbackTaskId !== undefined &&
+    task?.stepbackInfo?.nextStepbackTaskId !== "";
```
minnakt, line 31: Can this be rewritten with length as well, or did you run into a problem? 👀

minnakt (changes requested):

src/pages/task/metadata/Stepback.tsx
ZackarySantana: Nope! Just got confused but I updated it to use the length trick too

src/pages/task/metadata/index.tsx
```diff
@@ -387,7 +388,9 @@ export const Metadata: React.FC<Props> = ({ error, loading, task, taskId }) => {
             Honeycomb Trace
           </StyledLink>
           <StyledLink
-            data-cy="task-metrics-link"
+            data-cy="t
+      task: patchTaskWithSuccessfulBaseTask,
+    ask-metrics-link"
```
minnakt, line 393: looks like this was an unintentional change

src/pages/task/metadata/Stepback.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  return (
+    <MetadataItem as="div">
+      <StepbackWrapper>
+        Stepback:
+        <InfoSprinkle baseFontSize={BaseFontSize.Body1}>
+          When Stepback is completed you can access the breaking commit via the
+          relevant commits dropdown.
+        </InfoSprinkle>
+        {loading && <Skeleton size="small" />}
```
minnakt, line 54: nit. loading state looks like it might need a small style tweak

Screenshot 2024-03-21 at 12 52 29 PM

minnakt (approved):
great work!

src/pages/task/metadata/Stepback.tsx
ZackarySantana: Screenshot 2024-03-21 at 2 01 00 PM

Added no wrap and it looks like this now!

PR #2302, opened 2024-03-15
DEVPROD-1976: Migrate execution tasks table

minnakt:
DEVPROD-1976

### Description
This PR updates `ExecutionTasksTable` to use LeafyGreen table. I will be opening two more PRs for the remaining tables.

### Background Info
Currently Downstream Projects, Execution Tasks, and Patch Tasks all share the same table instance. I've decided to split them up into separate tables because it is difficult to reconcile their differences. Namely:
* Execution task table does not have filters, and sorting state is maintained internally by the table instance
* Downstream project table filter and sorting state is maintained by a reducer
* Patch task table filter and sorting state is maintained by URL query params

I think splitting them up into different tables will make it easier to debug them in the future. And it will also make it easier to review these PRs. 😃 

### Screenshots
Screenshot 2024-03-15 at 2 08 01 PM

### Testing
- Updated Cypress tests

### DIFF
--- a/src/components/TasksTable/Columns.tsx
+++ b/src/components/TasksTable/Columns.tsx
@@ -1,0 +1,155 @@
+import { LGColumnDef } from "@leafygreen-ui/table";
+import Tooltip from "@leafygreen-ui/tooltip";
+import pluralize from "pluralize";
+import { ConditionalWrapper } from "components/ConditionalWrapper";
+import { StyledRouterLink } from "components/styles";
+import TaskStatusBadge from "components/TaskStatusBadge";
+import { TreeDataEntry } from "components/TreeSelect";
+import { getVariantHistoryRoute } from "constants/routes";
+import { mergeTaskVariant } from "constants/task";
+import { zIndex } from "constants/tokens";
+import { TaskSortCategory } from "gql/generated/types";
+import { TaskStatus } from "types/task";
+import { TaskLink } from "./TaskLink";
+import { TaskTableInfo } from "./types";
+
+export const getColumnsTemplate = ({
+  baseStatusOptions = [],
+  isPatch = false,
+  onClickTaskLink = () => {},
+  showTaskExecutionLabel = false,
+  statusOptions = [],
+}: {
+  baseStatusOptions?: TreeDataEntry[];
+  isPatch?: boolean;
+  onClickTaskLink?: (taskId: string) => void;
+  showTaskExecutionLabel?: boolean;
+  statusOptions?: TreeDataEntry[];
+}): LGColumnDef<TaskTableInfo>[] => [
+  {
+    header: "Name",
+    accessorKey: "displayName",
+    id: TaskSortCategory.Name,
+    cell: ({
+      getValue,
+      row: {
+        original: { execution, id },
+      },
+    }): JSX.Element => (
+      <TaskLink
+        execution={execution}
+        onClick={onClickTaskLink}
+        showTaskExecutionLabel={showTaskExecutionLabel}
+        taskId={id}
+        taskName={getValue() as string}
+      />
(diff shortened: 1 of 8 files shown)


src/components/TasksTable/Columns.tsx
```diff
@@ -0,0 +1,155 @@
...
+        taskId={id}
+        taskName={getValue() as string}
+      />
+    ),
+    meta: {
+      search: {
+        "data-cy": "task-name-filter-popover",
+        placeholder: "Task name regex",
+      },
+    },
```
sophstad, line 52: Should this be removed since task name isn't a searchable column for the execution tasks table? Or will it be used later?

src/components/TasksTable/Columns.tsx
```diff
@@ -0,0 +1,155 @@
...
+      >
+        {getValue() as string}
+      </ConditionalWrapper>
+    ),
+    meta: {
+      search: {
+        "data-cy": "variant-filter-popover",
+        placeholder: "Variant name regex",
+      },
+    },
```
sophstad, line 151: Same here

src/pages/task/taskTabs/ExecutionTasksTable.tsx
```diff
@@ -1,69 +1,106 @@
...
+  });
+
+  const tableContainerRef = useRef<HTMLDivElement>(null);
+  const table: LeafyGreenTable<TaskTableInfo> =
+    useLeafyGreenTable<TaskTableInfo>({
+      columns,
+      containerRef: tableContainerRef,
+      data: executionTasksFull ?? [],
+      defaultColumn: {
+        enableColumnFilter: false,
```
sophstad, line 78: The current implementation allows multisort but it's missing here. You can check out the tests table for an example!

src/components/TasksTable/Columns.tsx
```diff
@@ -0,0 +1,155 @@
...
+          <TaskStatusBadge status={status} id={id} execution={execution} />
+        )
+      );
+    },
+    meta: {
+      treeSelect: {
+        "data-cy": "status-filter-popover",
+        options: statusOptions,
+      },
+    },
```
sophstad, line 93: Similarly, the current table doesn't allow filtering so this may be superfluous

src/components/TasksTable/Columns.tsx
```diff
@@ -0,0 +1,155 @@
...
+          id={baseTask?.id}
+          execution={baseTask?.execution}
+        />
+      ),
+    meta: {
+      treeSelect: {
+        "data-cy": "base-status-filter-popover",
+        options: baseStatusOptions,
+      },
+    },
```
sophstad, line 119: Same as above

src/pages/task/taskTabs/ExecutionTasksTable.tsx
sophstad: Ok I just noticed that I don't think this works currently 🤣 do you think we should just omit multisort altogether?

src/components/TasksTable/Columns.tsx
minnakt: this column definition will be reused for the upcoming tables, and they'll be shown in those tables! (initially i wrote these changes in one big PR which got split up now into 3 PRs, so that's why they are here 😅)

src/pages/task/taskTabs/ExecutionTasksTable.tsx
minnakt: added it back! i think it works now...

sophstad (approved):
💅

minnakt:
failing test just seems like a flaky project settings test, will merge

PR #2303, opened 2024-03-18
DEVPROD-4196: Display persistent DNS name

sophstad:
DEVPROD-4196 cc @Kimchelly 

### Description

When a persistent DNS name exists for a workstation, use it in place of the host URL. This applies to the spawn host table (including the "Copy SSH" button) and the host's event log page.

### Screenshots

Spawn hosts table:
image

Host page:
image

### Testing

Validated in staging, updated Storybook

### Evergreen PR

https://github.com/evergreen-ci/evergreen/pull/7650


Kimchelly (commented):
I was actually envisioning that the persistent DNS name would be a separate field from the current DNS name to give users time to try it out (and tell us if they have issues with it). The ephemeral EC2 DNS name is still valid, so it would be nice to give people either option while it's still a new feature.

khelif96 (approved):
🚀

PR #2310, opened 2024-03-20
DEVPROD-1970: Migrate CodeChanges Table to LG

SupaJoon:
DEVPROD-1970

### Description
These code changes swap the antd table for the LG table. Originally the code changes table on the commit queue card didn't include table column headers but I added them back in. I think the table looks better with the headers and it's also kind of complicated styling to not show the table headers and have the table cells respect width props.

### Screenshots
New render:
Screenshot 2024-03-20 at 5 56 41 PM
Screenshot 2024-03-20 at 5 57 10 PM

Old render:
Screenshot 2024-03-20 at 6 04 13 PM
Screenshot 2024-03-20 at 6 03 14 PM

### DIFF
--- a/src/components/CodeChangesTable/index.tsx
+++ b/src/components/CodeChangesTable/index.tsx
@@ -3,7 +3,6 @@ import { useLeafyGreenTable } from "@leafygreen-ui/table";
 import { FileDiffText } from "components/CodeChangesBadge";
 import { StyledLink, WordBreak } from "components/styles";
 import { BaseTable } from "components/Table/BaseTable";
-import { TablePlaceholder } from "components/Table/TablePlaceholder";
 import { FileDiffsFragment } from "gql/generated/types";
 
 interface CodeChangesTableProps {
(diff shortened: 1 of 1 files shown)


src/components/CodeChangesTable/index.tsx
```diff
@@ -1,65 +1,69 @@
...
+    enableSorting: false,
+  });
 
-const rowKey = (record: FileDiffsFragment): string =>
-  `${record.diffLink}_code_table`;
+  return (
+    <BaseTable
+      data-cy="code-changes-table"
+      data-cy-row="code-changes-table-row"
+      emptyComponent={<TablePlaceholder message="No diffs." />}
```
sophstad, line 28: [nit] I don't think you need this because we don't show the table if there are no diffs
image

src/components/CodeChangesTable/index.tsx
```diff
@@ -1,65 +1,69 @@
...
-  showHeader: boolean,
-) => Array<ColumnProps<FileDiffsFragment>> = (showHeader: boolean) => [
+const columns = [
   {
-    title: <span data-cy="file-column">File</span>,
-    width: "60%",
-    dataIndex: "fileName",
-    key: "fileName",
-    render: (text: string, record: FileDiffsFragment): JSX.Element => (
+    id: "fileName",
```
sophstad, line 37: [nit] You can delete the `id` fields since they're not used for anything. Other tables use them for sorting, but here they just duplicate the `accessorKey` field, which is what's used by TanStack table when no `id` field is provided.

sophstad (approved):

PR #2311, opened 2024-03-21
DEVPROD-5595: Downgrade LG text input

sophstad:
DEVPROD-5595

### Description

- Downgrade package to the same version on Parsley. This version does not use the `@leafygreen-ui/form-field` package which introduced the styling bugs.

### Screenshots

The input on this page is styled correctly again:
image


minnakt (approved):
perf

PR #2315, opened 2024-03-22
DEVPROD-5788: Allow upgrading to minor or major version

sophstad:
DEVPROD-5788

### Description

Noticed that our deploy script no longer allows upgrading anything besides patch version. I'm going to upgrade the major version when we switch to a monorepo so we need this functionality. Will add to Parsley after!

### Screenshots

image

### Testing

Ran `createTagAndPush` command locally/wifi off to verify prompt works as designed


khelif96 (approved):
LGTM lets update the deploy guide so people aren't confused on what to do.

